PR #5, opened 2023-08-16
Reload rules threadsafely

willmostly:
Refactor rule selector to standalone class and add synchronization logic to prevent multiple threads from attempting to parse the rules yaml simultaneously. The rules file is only parsed once, when it is [re]loaded.


gateway-ha/src/main/java/com/lyft/data/gateway/ha/router/RuleReloadingRoutingGroupSelector.java
```diff
@@ -0,0 +1,76 @@
...
+
+@Slf4j
+public class RuleReloadingRoutingGroupSelector
+    implements RoutingGroupSelector  {
+
+  RulesEngine rulesEngine = new DefaultRulesEngine();
+  MVELRuleFactory ruleFactory = new MVELRuleFactory(new YamlRuleDefinitionReader());
+  String rulesConfigPath;
+  Rules rules = new Rules();
+  long lastUpdatedTime;
```
xkrogen, line 28: These need to be marked [volatile](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.3.1.4) to allow for consistent updates to be seen across threads.

But also I would recommend using a bit stronger synchronization here? This should not be a hot enough path to be worth optimizing for lock-free read access. You can use a r-w lock or atomics + CAS to get more consistent behavior

xkrogen, line 28: should these all be `private`?

mosabua:
As agreed we will wait with this until the other PR from @vishalya  are in .. right?

willmostly:
I threw this up since @vishalya had an alternative approach of removing the reloading capability. If we want to retain the ability to update the rules I wanted to save him the trouble of submitting that change. Let me know your thoughts - I will apply @xkrogen 's suggestions (thanks!) if we want to use this.

vishalya:
Our other PR will be in soon, meanwhile we can get this reviewed and not merged.

vishalya:
My solution has limitation that the rules file would get read at the load time. On the other hand, I found that the reading before every request is much slower (as it was before, apart from the synchronization issue).

willmostly:
I applied Erik's suggestions on using stronger synchronization, and will be doing some performance testing with Starburst Insights. It generates a lot of concurrent requests for css, icon images, etc so should be pretty good at finding any performance issues with the read/write locking strategy.

willmostly:
> My solution has limitation that the rules file would get read at the load time. On the other hand, I found that the reading before every request is much slower (as it was before, apart from the synchronization issue).

I'm using the file timestamp to check if it has changed, so this should avoid the expense of parsing for every request

willmostly:
@xkrogen could you give this another pass when you get a chance? Thanks!

xkrogen:
Hey @willmostly sorry for the delay, current version LGTM 👍🏻

mosabua:
@willmostly @vishalya @xkrogen  ... we you okay with this as it stands now or do we need to update something before I review and merge?

xkrogen (approved):

willmostly:
I just noticed that the test for this functionality in `TestRoutingGroupSelector.testByRoutingRulesEngineFileChange()` was disabled at some point, my commit this morning reenbles it. _Now_ this is ready :)

mosabua (approved):
This looks good now. Ultimately I agree that we should not have to reload rules so often - seems like a bad idea and must have a bad impact on performance. Fixing that however is a bigger task and will need to be done separately.

PR #7, opened 2023-08-16
Update to Java 17

willmostly:
This updates the source and target version to java 17. This requires adding VM options for the surefire plugin and `org.javalite:activejdbc-instrumentation`. A `mvnw` is added to facilitate passing the required option to `activejdbc-instrumentation`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,8 +12,8 @@
     <version>1.9.5</version>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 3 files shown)


mosabua:
Also .. @vishalya .. do we need to get your PRs in first?

pom.xml
```diff
@@ -12,8 +12,8 @@
     <version>1.9.5</version>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>17</maven.compiler.source>
```
mosabua, line 15: Not sure we want to do this now already .. we could just make it work with 17 to build .. and then switch language level later

willmostly: Sounds good. I closed this and will add the build options in once the maven wrapper PR https://github.com/trinodb/trino-gateway/pull/8 is merged. I don't think this should conflict with Vishal's package renaming.

vishalya: Yes, that makes sense. I will have 2 more PR to bring our code in sync with this repo and then we can go on enhancing the rest.

PR #8, opened 2023-08-17
Add maven wrapper

willmostly:
This adds a generated maven wrapper.


mosabua:
Did you confirm that the build works with Maven 3.9.4 @willmostly ?

willmostly:
> Did you confirm that the build works with Maven 3.9.4 @willmostly ?

Yes - I am using 3.9.4 locally

PR #9, opened 2023-08-17
Enhancements - Authentication, Roles, docker-compose

vishalya:
Feature enhancements:
- Authentication & Authorization - LDAP, OIDC and user list from config file
- Support Roles - User, Admin and API
- Healthcheck for Trino backends using JDBC
- Read the rules file only once during the load time to avoid concurrency issue
- Use docker-compose to support multiple trino backends during development
- Filter logs for sensitive information
- TCP check for routing

The readme has the information on how to use the features, but we might need to enhance it further.


mosabua:
Can you expand the content of the commit message again and confirm that this builds?

PR #11, opened 2023-08-21
Enhancements: Postgres, use /ui/api to fetch backend status, refactor to io.trino.gateway, add timeout to prepareStatement operation

andythsu:
- Refactored from Presto to Trino
- Added postgres support
- Updated readme
- Added useApi in config to determine whether to check backend status using api
- Removed X-Presto-* headers
- Refactored directories to io.trino.gateway
- Added timeout to prepareStatement operation


mosabua (changes requested):
Can you change commit messages to 

* Adding routing for requests to healthy backends
* Deactivate clusters with zero workers
* Rename and repackage from Presto to Trino
* Add PostgreSQL support for backend database

And then the rest of them probably should be merged into these as applicable, unless you think some stand alone ..

Also .. could we add PostgreSQL support before the renaming or at least not between two renaming commits..

andythsu:
> Can you change commit messages to
> 
> * Add routing for requests to healthy backends
> * Deactivate clusters with zero workers
> * Rename and repackage from Presto to Trino
> * Add PostgreSQL support for backend database
> 
> And then the rest of them probably should be merged into these as applicable, unless you think some stand alone ..
> 
> Also .. could we add PostgreSQL support before the renaming or at least not between two renaming commits..

Fixed. Please see https://github.com/trinodb/trino-gateway/pull/13

PR #13, opened 2023-08-22
Enhancements: Postgres, use /ui/api to fetch backend status, refactor to io.trino.gateway, add timeout to prepareStatement operation

andythsu:
- Add routing for requests to healthy backends
- Deactivate clusters with zero workers
- Add PostgreSQL support for backend database
- Rename and repackage from Presto to Trino


LICENSE
```diff
@@ -1,201 +0,0 @@
-                                 Apache License
```
mosabua, line 1: Do not delete this file please

baseapp/pom.xml
```diff
@@ -15,8 +15,8 @@
     <packaging>jar</packaging>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>1.9</maven.compiler.source>
```
mosabua, line 18: Why are we changing to Java 1.9  .. ultimately will change to 17 but is this change necessary?

gateway-ha/gateway-ha-config.yml
```diff
@@ -4,42 +4,22 @@ routingRules:
...
-  name: prestoRouter
+  name: trinoRouter
   historySize: 1000
 
 dataStore:
-  jdbcUrl: jdbc:mysql://localhost:3306/prestogateway
-  user: root
-  password: root123
-  driver: com.mysql.cj.jdbc.Driver
+  jdbcUrl: jdbc:postgresql://localhost:5432/trino_gateway_db
```
mosabua, line 11: Will either of these work now?

gateway-ha/pom.xml
```diff
@@ -17,8 +17,8 @@
     <name>gateway-ha</name>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>1.9</maven.compiler.source>
```
mosabua, line 20: Do we need 1.9?

gateway-ha/pom.xml
```diff
@@ -56,14 +56,14 @@
           <version>2.10.1</version>
         </dependency>
         <dependency>
-            <groupId>com.lyft.data</groupId>
+            <groupId>io.trino.gateway</groupId>
             <artifactId>baseapp</artifactId>
-            <version>${project.parent.version}</version>
+            <version>1.9.5</version>
```
mosabua, line 61: Why hardcode the version here .. this seems wrong .. if anything you can set the version in dependencyManagement in the parent.

mosabua (changes requested):

gateway-ha/pom.xml
andythsu: yes, we can just take the parent's version

gateway-ha/gateway-ha-config.yml
andythsu: yes, these should work now

gateway-ha/pom.xml
andythsu: ok!

baseapp/pom.xml
andythsu: this can stay 1.8, but we will need 1.9 for gateway-ha because last time it failed when I built it with 1.8.

andythsu:
I'd like to put everything in one commit. Let me know if there are anything else to fix!

andythsu:
> Can you please just do fixup commits and stick with the four we have
> 
> so you just do
> 
> git commmit --fixup=hashofcommittofix
> 
> and then once you have them all in
> 
> git rebase -i HEAD~5 --autosquash

I see. Np.

PR #18, opened 2023-08-25
Move VM options to jvm.config

willmostly:
This allows the build to succeed using `mvn install` instead of requiring `./mvnw install`

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,1 @@
+--add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED
--- a/mvnw
+++ b/mvnw
@@ -280,9 +280,7 @@ if [ -n "$wrapperSha256Sum" ]; then
   fi
 fi
 
-#This option is required by the javalite db-migrator-maven-plugin, which does not support adding VM options in its config
-REQUIRED_BUILD_OPTIONS='--add-opens=java.base/java.net=ALL-UNNAMED'
-MAVEN_OPTS="$(concat_lines "$MAVEN_PROJECTBASEDIR/.mvn/jvm.config") $REQUIRED_BUILD_OPTIONS $MAVEN_OPTS"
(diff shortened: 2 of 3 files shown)


.mvn/jvm.config
```diff
@@ -0,0 +1 @@
+--add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED
```
vishalya, line 1: if you could get rid of that eye sore end of line.

vishalya (approved):
LGTM, minor comment.

andythsu (approved):

.mvn/jvm.config
willmostly: haha, done 🖖

vishalya (approved):
LGTM

mvnw
```diff
@@ -281,8 +281,7 @@ if [ -n "$wrapperSha256Sum" ]; then
 fi
 
 #This option is required by the javalite db-migrator-maven-plugin, which does not support adding VM options in its config
-REQUIRED_BUILD_OPTIONS='--add-opens=java.base/java.net=ALL-UNNAMED'
-MAVEN_OPTS="$(concat_lines "$MAVEN_PROJECTBASEDIR/.mvn/jvm.config") $REQUIRED_BUILD_OPTIONS $MAVEN_OPTS"
+MAVEN_OPTS="$(concat_lines "$MAVEN_PROJECTBASEDIR/.mvn/jvm.config") $MAVEN_OPTS"
```
mosabua, line 284: do NOT modify the generated mvnw wrapper script .. this will be wiped whenever we upgrade the wrapper

mosabua (changes requested):
We should take the related setting in the pom out as part of this PR as well

mvnw
willmostly: this changes it back to the generated version

willmostly: except for the comment - let me regenerate the wrapper

mosabua: oh .. good

mosabua:
This fails to build locally for me .. you want output @willmostly ?

mosabua:
Is this ready now @willmostly ... I dont think you pushed the regenerated wrapper..

PR #19, opened 2023-08-28
Add minimal GHA/Maven build

mosabua:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
```
willmostly, line 27: I thought Trino was using Zulu - is that not the case?

willmostly (approved):
LGTM!

.github/workflows/ci.yml
mosabua: But let me switch to Zulu.

willmostly: yeah i agree it shouldn't matter - just trying to keep things uniform

.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'zulu'
```
regadas, line 27:
```suggestion
          distribution: 'temurin'
```
:art: Wondering if it should be consistent to the JDK used in Trino.

mosabua: we are .. Trino uses Zulu and suggests that in the docs

regadas: ah! yes you are completely right 👍 I guess I saw a `temurin` reference somewhere and misjudged things. ❤️

regadas: Actually, I think it was a recent change https://github.com/trinodb/trino/pull/17685 and docs didn't follow 🤷 I'm happy to help out let me know if this is smth that needs cleaning.

mosabua:
Had a discussion about Java with other Trino maintainers and got confirmation that we should switch back to use Temurin. The Trino docs will be updated by myself as well. 

fyi @regadas

.github/workflows/ci.yml
mosabua: Changed back to Temurin now and will update the Trino docs.

dain (approved):

PR #20, opened 2023-08-29
Improve documentation setup

mosabua:
- Break up readme content
- Move content around to multiple pages
- No content modifications
- Format paragraphs in source to 80 column width
- Add simplistic navigation

Follow up PRs will address the actual content and maybe in the long run even use a proper GH pages site or whatever.


dain (approved):

docs/development.md
```diff
@@ -0,0 +1,102 @@
...
+
+run `mvn clean install` to build `trino-gateway`
+
+Edit the [config file](/gateway-ha/gateway-ha-config.yml) and update the mysql
+db information.
+
+```
+cd gateway-ha/target/
+java -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar server ../gateway-ha-config.yml
+```
```
Chaho12, line 71: Hi, since update to java 17 causes the following error
> Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @49c9930c

How about adding java opts in the doc for now until there is a different shell script or sth that runs the server? I noticed that this option was added when building mvn.
`java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar server ../gateway-ha-config.yml`

mosabua: It would be great if you could send a PR to update the docs for that. I think @willmostly is also looking at working on more the necessary updates towards Java 17 usage.

PR #23, opened 2023-09-01
Add java opts guide

Chaho12:
Added guide to use java opts when running with java cmd because of java version issue causing the following error

> Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @49c9930c

- https://github.com/trinodb/trino-gateway/pull/20#discussion_r1312440769

I linked `[jvm.config](../.mvn/jvm.config)` as this will be added soon in the following PR
- https://github.com/trinodb/trino-gateway/pull/18/files

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -67,8 +67,12 @@ db information.
 
 ```
 cd gateway-ha/target/
-java -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar server ../gateway-ha-config.yml
+java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar server ../gateway-ha-config.yml
 ```
+
+If you encounter `module java.base does not "opens java.lang" to unnamed module` error, 
+this may be due to newer version of java.
+Do check opts at [jvm.config](../.mvn/jvm.config) when running java command.
 
 If you encounter a `Failed to connect to JDBC URL` error, this may be due to
 newer versions of java disabling certain algorithms when using SSL/TLS, in
--- a/docs/development.md
+++ b/docs/development.md
@@ -57,8 +57,7 @@ testing
 
 ### Build and run
 
-Please note these steps have been verified with JDK 8 and 11. Higher versions of
-Java might run into unexpected issues.
+Please note these steps have been verified with JDK 17.
 
 run `mvn clean install` to build `trino-gateway`
 


mosabua:
Can you also update the section above and remove Java 8 and 11 and change it to 17. We will only support 17 .. just have to clean it all up.

Chaho12:
> Can you also update the section above and remove Java 8 and 11 and change it to 17. We will only support 17 .. just have to clean it all up.

Changed :)

docs/development.md
```diff
@@ -67,9 +66,13 @@ db information.
 
 ```
 cd gateway-ha/target/
-java -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar server ../gateway-ha-config.yml
+java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar server ../gateway-ha-config.yml
 ```
 
+If you encounter `module java.base does not "opens java.lang" to unnamed module` error, 
+this may be due to newer version of java.
```
mosabua, line 73: Newer than 17? I think I will merge #25 first. And we just document what must be done with 17 ...

docs/development.md
```diff
@@ -67,9 +66,13 @@ db information.
...
+Do check opts at [jvm.config](../.mvn/jvm.config) when running java command.
```
mosabua, line 74: Not sure what this means ..

mosabua (changes requested):
Wait until #25 is merged and then we can add this info

docs/development.md
Chaho12: Oh what i meant was that the guide might be out-of-date so if any error happens with `module java.base does not "opens java.lang" to unnamed module` error, user can check .mvn/jvm.config file for other jvm opts. I noticed that [trinodb/trino](https://github.com/trinodb/trino/blob/master/.mvn/jvm.config) has more opts.

docs/development.md
Chaho12: I tried to say that this error is caused by the jdk as it has higher restrictions on internal JDK modules(since jdk 9 i assume), so user has to add additional parameters.
https://stackoverflow.com/questions/42538750/unable-to-export-a-package-from-java-base-module/42541096#42541096

andythsu:
looks like I needed to run ` java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha-3-SNAPSHOT-jar-with-dependencies.jar server ../gateway-ha-config.yml` to successfully launch the gateway. Otherwise I get `Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @2ba867a6`. Are you re-opening this PR or am I missing anything?

Chaho12:
> Are you re-opening this PR or am I missing anything?

Mistakenly deleted branch xD. 
I can't reopen so I will add PR with just the command changed (as there is other [PR for development instructions](https://github.com/trinodb/trino-gateway/pull/25))

PR #25, opened 2023-09-01
Update development instructions for Java 17

willmostly:
This updates the documentation to align with the Java update.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -57,10 +57,11 @@ testing
 
 ### Build and run
 
-Please note these steps have been verified with JDK 8 and 11. Higher versions of
-Java might run into unexpected issues.
+This project requires Java 17. Note that higher version of Java have not been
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -57,10 +57,10 @@ testing
 
 ### Build and run
 
-Please note these steps have been verified with JDK 8 and 11. Higher versions of
-Java might run into unexpected issues.
+This project requires Java 17, and is tested with Azul Zulu. Please note that higher version of Java have not been
```
mosabua, line 60: Actually Eclipse Temurin .. but we can leave that out.

mosabua: Drop "Please"

mosabua (changes requested):

mosabua (changes requested):
Can you fix the wrapping in the source to be 80 char column width.

PR #26, opened 2023-09-04
Improve security documentation

Chaho12:
- odic -> OAuth/OpenIDConnect
  - `odic` is typo for `oidc` which means OpenIDConnect
  - Conf is set using OAuth so added example to set defaultType as OAuth to use OIDC/OAuth
- Separated `Form` into `basic auth` and `LDAP auth`
  - I haven't tested LDAP conf yet, but [FormConfiguration code](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/config/FormAuthConfiguration.java) has LDAP as string type so I just added simple guide
  - `privateKeyRsa` and `publicKeyRsa` had should have R`sa` as lowercase
  - [keypair is read as file, not string](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbKeyProvider.java#L35-L53)

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -22,11 +22,13 @@ The authentication would happen on https protocol only. Add the
 set using `defaultType: "form"` Following types of the authentications are
 supported.
 
-### odic
+### OAuth/OpenIDConnect
 
 It can be configured as below
 
 ```
+authentication:
+  defaultType: "oauth"
   oauth:
     issuer:
     clientId:
@@ -42,7 +44,7 @@ It can be configured as below
       - s3
 ```
 
-### form
+### Form/Basic Auth
 
 The authentication happens with the pre-defined users from the configuration
 file. To define the preset user use the following section.
@@ -57,16 +59,26 @@ presetUsers:
     privileges: "lb_api"
 ```
 
-Provide a key pair ins RSA format
+Also provide a random key pair in RSA format.
 
 ```
+authentication:
+  defaultType: "form"
   form:
     selfSignKeyPair:
-      privateKeyRSA: |
-      <PEM PRIVATE KEY>
-      publicKeyRSA: |
-      <PEM PUBLIC key>
+      privateKeyRsa: <private_key_path>
+      publicKeyRsa: <public_key_path>
 ```
+
+### Form/LDAP
+
+```
+authentication:
+  defaultType: "form"
+  form:
+    ldapConfigPath: <ldap_config_path>
+```
+
 
 ## Authorization
 


mosabua (commented):
Looks good to me. Can you confirm @vishalya @andythsu @willmostly and any others?

docs/security.md
```diff
@@ -57,17 +59,27 @@ presetUsers:
...
     selfSignKeyPair:
-      privateKeyRSA: |
-      <PEM PRIVATE KEY>
-      publicKeyRSA: |
-      <PEM PUBLIC key>
+      privateKeyRsa: <private_key_path>
+      publicKeyRsa: <public_key_path>
 ```
 
+### Form/LDAP
```
andythsu, line 73: I would merge this section with the above so it becomes 

```
  authentication:
    defaultType: "oauth"
    form:
      ldapConfigPath: <ldap_config_path>
      selfSignKeyPair:
        publicKeyRsa: "<public_key_path>"
        privateKeyRsa: "<private_key_path>"
```
Since `selfSignKeyPair` and `ldapConfigPath` are both needed by `FormAuthConfiguration`

Chaho12: Is both `ldapConfigPath` and `selfSignKeyPair` needed for ldap? because ldap config can be left empty if we do use FormAuthConfiguration but not use ldap

mosabua: I think its fine to have them separate and explain each section .. and then also maybe have examples that combine each. We can iterate on that over time.

andythsu: @Chaho12 I think both `ldapConfigPath` and `selfSignKeyPair` are needed if we have `defaultType = form` for authentication.

Chaho12: > if we have defaultType = form for authentication.

For both ldap and basic auth? 
With current main version, we are using the conf as below and it works for admin/user1 to login and blocks any other users. 
```yaml
authentication:
      defaultType: "form"
      form:
        selfSignKeyPair:
          privateKeyRsa: /app/config/private.key
          publicKeyRsa: /app/config/public.key
presetUsers:
      admin:
        password: ...
        privileges: "lb_admin, lb_user, lb_api"
      user1:
        password: ...
        privileges: "lb_user"
    authorization:
      admin: "lb_admin"
      user: "lb_user"
      api: "lb_api"
```

andythsu: oh I see. In this case we don't need to have both then.

mosabua: So I think this PR is ready to go then.

mosabua (approved):

PR #27, opened 2023-09-06
Allow /v1/node to be routed by gateway

Chaho12:
Allow gateway to route `v1/node` that returns node info. 
- We use this to API to find all workers address, then request `{worker_addr}/v1/status` to find cpu/memory stats.


mosabua (approved):
Looks good to me. Any concerns from others?

andythsu (approved):
LGTM

mosabua:
Can you rebase from main .. that should fix CI

mosabua:
I rebased and pushed now.

PR #31, opened 2023-09-11
Handle NPE for reading user records

Chaho12:
1. Handle NPE error caused while getting entry from UserRecord object.
    - If the attribute key for `memberOf` conf, which is set by user using `ldapGroupMemberAttribute` in LDAP config, does not exist, then NPE happens because `engtry.get(memberOf)` returns null in `entry.get(memberOf).toString()` code.
    - Although error is caught by throwing LdapException, I suggest returning empty record instead.

2. Error above is caught by LbFilter, but the error message is not logged, so added logging code.

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFilter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFilter.java
@@ -13,6 +13,9 @@ import javax.ws.rs.core.Cookie;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.SecurityContext;
 
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
 @Priority(Priorities.AUTHENTICATION)
 public class LbFilter<P extends Principal> extends AuthFilter<String, P> {
 
@@ -41,6 +44,7 @@ public class LbFilter<P extends Principal> extends AuthFilter<String, P> {
       }
 
     } catch (Exception e) {
+      log.debug("Error while filtering request for authentication", e);
       throw new WebApplicationException(unauthorizedHandler.buildResponse(prefix, realm));
     }
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
@@ -3,9 +3,12 @@ package io.trino.gateway.ha.security;
 import io.trino.gateway.ha.config.LdapConfiguration;
 
 import java.util.List;
+import java.util.Optional;
 
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import org.apache.directory.api.ldap.model.entry.Attribute;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.exception.LdapException;
 import org.apache.directory.api.ldap.model.message.SearchScope;
@@ -33,7 +36,7 @@ public class LbLdapClient {
     }
   }
 
-  public static class UserEntryMapper implements EntryMapper<UserRecord> {
+  public static class UserEntryMapper implements EntryMapper<Optional<UserRecord>> {
     String memberOf;
 
     public UserEntryMapper(String memberOfAttr) {
@@ -41,8 +44,11 @@ public class LbLdapClient {
     }
 
     @Override
-    public UserRecord map(Entry entry) throws LdapException {
-      return new UserRecord(entry.get(memberOf).toString());
+    public Optional<UserRecord> map(Entry entry) throws LdapException {
+      Attribute attribute = entry.get(memberOf);
+      return ObjectUtils.isEmpty(attribute)
+              ? Optional.empty()
+              : Optional.of(new UserRecord(attribute.getString()));
     }
   }
 
@@ -109,15 +115,18 @@ public class LbLdapClient {
     String filter = config.getLdapUserSearch().replace("${USER}", user);
 
     String[] attributes = new String[]{config.getLdapGroupMemberAttribute()};
-    List<UserRecord> list = ldapConnectionTemplate.search(config.getLdapUserBaseDn(),
+    List<Optional<UserRecord>> userRecords = ldapConnectionTemplate.search(config.getLdapUserBaseDn(),
         filter,
         SearchScope.SUBTREE,
         attributes,
         userRecordEntryMapper);
 
     String memberOf = "";
-    if (list != null && !list.isEmpty()) {
-      memberOf = list.listIterator().next().getMemberOf();
+    if (userRecords != null && !userRecords.isEmpty()) {
+      Optional<UserRecord> userRecord = userRecords.listIterator().next();
+      if (userRecord.isPresent()) {
+        memberOf = userRecord.get().getMemberOf();
+      }
       log.debug("Member of {}", memberOf);
     }
     return memberOf;


mosabua (commented):
The same behavior should be implemented for other supported authentication methods. Please check if they already work that way or changes are needed, or this is not applicable.

Chaho12:
I checked once again and since all authentication methods use `EntryMapper<UserRecord>` (except presetUser) to search, this will fix all the NPE error that could happen while finding user records.

Here is some flow
- When server loads, [getAuthenticationFilter](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java#L103) is called and uses LbFilter.Builder to set authenticators (this is why exception is caught at LbFilter)
- There are 3 types of authenticators called : [LbAuthenticator](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthenticator.java), [FormAuthenticator](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/FormAuthenticator.java), [ApiAuthenticator](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/ApiAuthenticator.java)
- All the authenticators call [getPrivileges function](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java#L39C27-L39C40) which gets privileges.
- It calls [lbLdapClient.getMemberOf](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java#L112) if user does not exist in presetUsers
- getMemberOf function calls [ldapConnectionTemplate.search](https://github.com/apache/directory-ldap-api/blob/2.1.2/ldap/client/api/src/main/java/org/apache/directory/ldap/client/template/LdapConnectionTemplate.java#L831) function to find user and store record by using EntryMapper#map function, which map function is [overrided in LBLdapClient](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java#L44-L47). 
- There is potential NPE error if there is no match find for the user (can be tested by setting [ldapGroupMemberAttribute value](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/test/resources/auth/ldapTestConfig.yml#L9C1-L9C25) to a non-existing attribute key so that users found in ldapConnectionTemplate.search function does not match any attribute conf)

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -42,7 +47,9 @@ public UserEntryMapper(String memberOfAttr) {
 
     @Override
     public UserRecord map(Entry entry) throws LdapException {
-      return new UserRecord(entry.get(memberOf).toString());
+      Attribute attribute = entry.get(memberOf);
+      return ObjectUtils.isEmpty(attribute) ? emptyRecord : new UserRecord(attribute.getString());
```
willmostly, line 51: Is it possible to change the signature to `Optional<UserRecord>` and return `Optional.empty()` if there is no `memberOf` attribute? `Optional.empty()` is the usual way this type of situation is handled in Trino.

vishalya (commented):
LGTM

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFilter.java
```diff
@@ -41,6 +44,7 @@ public void filter(final ContainerRequestContext requestContext) throws WebAppli
       }
 
     } catch (Exception e) {
+      log.error("Error while filtering request for authentication", e);
```
willmostly, line 47: I think this should probably be `log.debug` to avoid filling up the logs when someone supplies incorrect credentials. Alternatively, maybe a more specific exception can be thrown if `authenticate` returns false, so that bad credentials can be segregated from other exceptions.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
willmostly, line 51: Is it possible to change the signature to `Optional<UserRecord>` and return `Optional.empty()` if there is no `memberOf` attribute? `Optional.empty()` is the usual way this type of situation is handled in Trino.

willmostly (commented):
Looks good, just a couple questions

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
Chaho12: Nope. 
image

Map function is used in two places below and cannot be empty :( 
- [return entry == null ? null : entryMapper.map( entry );](https://github.com/apache/directory-ldap-api/blob/2.1.2/ldap/client/api/src/main/java/org/apache/directory/ldap/client/template/LdapConnectionTemplate.java#L831)
- [entries.add( entryMapper.map( entry ) );](https://github.com/apache/directory-ldap-api/blob/2.1.2/ldap/client/api/src/main/java/org/apache/directory/ldap/client/template/LdapConnectionTemplate.java#L354)

Well it is possible to simply return a `new UserRecord(null)` instead though. What do you think?

willmostly: Sorry I should have been more clear. If you change 
```
public static class UserEntryMapper implements EntryMapper<UserRecord>
```
to
```
public static class UserEntryMapper implements EntryMapper<Optional<UserRecord>>
```
then make the necessary changes to other parts of the code to use `Optional`, then I _think_ this should all work.

Chaho12: Oh my bad. I understood it wrong xD. Changed as you said and rebased it too!

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -109,15 +115,18 @@ public String getMemberOf(String user) {
...
+    List<Optional<UserRecord>> list = ldapConnectionTemplate.search(config.getLdapUserBaseDn(),
         filter,
         SearchScope.SUBTREE,
         attributes,
         userRecordEntryMapper);
 
     String memberOf = "";
     if (list != null && !list.isEmpty()) {
-      memberOf = list.listIterator().next().getMemberOf();
+      Optional<UserRecord> userRecord = list.listIterator().next();
```
willmostly, line 126: Is it possible that some items in the list might have the `memberOf` attribute defined an others won't? We just take the first item in the list here, so it looks like we assume that that everything returned by the search will be the same with respect to the memberOf attribute. If this isn't the case, you could filter the list using sth like `stream().filter(e -> e.isPresent())`.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -109,15 +115,18 @@ public String getMemberOf(String user) {
     String filter = config.getLdapUserSearch().replace("${USER}", user);
 
     String[] attributes = new String[]{config.getLdapGroupMemberAttribute()};
-    List<UserRecord> list = ldapConnectionTemplate.search(config.getLdapUserBaseDn(),
+    List<Optional<UserRecord>> list = ldapConnectionTemplate.search(config.getLdapUserBaseDn(),
```
willmostly, line 118: nit: we should use more meaningful name than "list"

Chaho12: How about `userRecords`?

willmostly: sounds good to me

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
willmostly: good point. I don't see anywhere where we validate that the configured `filter` returns a unique result though. I assume it is meant to be something like `sAMAccountName=${USER}`. Maybe I have missed that though, @vishalya could you chime in here?

Chaho12: @vishalya any thoughts?

vishalya: memberOf should be a single record of string type, at least that's what we have here. The string is `,` delimited list.
You change looks generic, but we will have to use config to hint at what field to use in order to find the membership.

vishalya: `ldapGroupMemberAttribute` should help the configuration and might void this issue.

Chaho12: Can you elaborate when you mean by "use config to hint at what field to use in order to find the membership."

Also @willmostly , if I run the test codes and it fails on [line 126](https://github.com/trinodb/trino-gateway/pull/31/files#diff-d59eac35750dcc974b98c0c5387a7ae147e3a3fbb339b4c1c5d2b4a1f39bb8a1R126). It seems that the [map function](https://github.com/trinodb/trino-gateway/pull/31/files#diff-d59eac35750dcc974b98c0c5387a7ae147e3a3fbb339b4c1c5d2b4a1f39bb8a1R47-R51) that is overriding has issues when changed to Optional. Do you happen to know how to fix this? 
```
java.lang.ClassCastException: class io.trino.gateway.ha.security.LbLdapClient$UserRecord cannot be cast to class java.util.Optional (io.trino.gateway.ha.security.LbLdapClient$UserRecord is in unnamed module of loader 'app'; java.util.Optional is in module java.base of loader 'bootstrap')
```

mosabua:
@Chaho12 can you please rebase and ensure that the build passes. Currently it fails for me with a checkstyle message but rebasing might remove that

Chaho12:
> can you please rebase and ensure that the build passes. Currently it fails for me with a checkstyle message but rebasing might remove that

I rebased, and tried on my local and it fails stills on https://github.com/trinodb/trino-gateway/pull/31/files#diff-d59eac35750dcc974b98c0c5387a7ae147e3a3fbb339b4c1c5d2b4a1f39bb8a1R126 part :(

mosabua:
You might want to redo this on a new branch .. codebase changed quite a bit

Chaho12:
Will make a new branch :)

PR #35, opened 2023-09-12
Add doc for contributing

### LINKED ISSUE #34: CONTRIBUTING.md is missing
The link in https://github.com/trinodb/trino-gateway/blob/main/docs/development.md#contributing return 404. 

Screenshot 2023-09-12 at 14 18 14

mosabua:
Fixes #34

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,14 @@
+# Contributing
+
+## Contributor License Agreement ("CLA")
+
+In order to accept your pull request, we need you to [submit a CLA](https://github.com/trinodb/cla).
+
+## License
+
+By contributing, you agree that your contributions will be licensed under the
+[Apache License Version 2.0 (APLv2)](../LICENSE).
+
+## Contribution process
+
+See the suggested [process for contributions](https://trino.io/development/process.html).
--- a/README.md
+++ b/README.md
@@ -13,3 +13,4 @@ Find out more details from our documentation:
 * [Resource groups API](./docs/resource-groups-api.md)
 * [Routing rules](./docs/routing-rules.md)
 * [References](./docs/references.md)
+* [Contributing](./.github/CONTRIBUTING.md)
--- a/docs/development.md
+++ b/docs/development.md
@@ -98,5 +98,5 @@ for more details.
 
 ## Contributing
 
-Want to help build trino Gateway? Check out our [contributing
-documentation](CONTRIBUTING.md)
+Want to help build Trino Gateway? Check out our [contributing
+documentation](../.github/CONTRIBUTING.md)


CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
+# Contributing to Trino
```
ebyhr, line 1: Trino → Trino Gateway

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Contributor License Agreement ("CLA")
+
+In order to accept your pull request, we need you to [submit a CLA](https://github.com/trinodb/cla).
+
+## License
+
+By contributing to Trino, you agree that your contributions will be licensed under the [Apache License Version 2.0 (APLv2)](LICENSE).
```
ebyhr, line 9: Trino → Trino Gateway

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Contribution process
+
+See the suggested [process for contributions](https://trino.io/development/process.html).
```
ebyhr, line 13: trino and trino-python-client repositories put this file under `.github` repository. I would move to the same path.

ebyhr (approved):

CONTRIBUTING.md
mosabua: Well .. I think it makes sense to keep it Trino.. since they contribute to the Trino Software Foundation and the trinodb org on GitHub .. but I am happy to change as well.. wdyt @dain

CONTRIBUTING.md
mosabua: See above comment.

CONTRIBUTING.md
mosabua: That seems like a bad idea .. a LOT less visible and against the common approach. I did see that on the other repos though .. 

Maybe move it but add mention in the README?

CONTRIBUTING.md
ebyhr: This project isn't "Trino". See https://github.com/trinodb/trino-python-client/blob/master/.github/CONTRIBUTING.md

mosabua: Also the CLA is for the TSF

CONTRIBUTING.md
ebyhr: Relates to https://github.com/trinodb/trino/pull/9168

CONTRIBUTING.md
mosabua: I can see both sides ... hence asking @dain to chime in.

CONTRIBUTING.md
mosabua: Cool .. thanks for pointing to that discussion.. will move

mosabua: Done

CONTRIBUTING.md
mosabua: I got an even better fix now .. the name can just be removed (and is therefore implied).

mosabua:
I will wait for another day or so for feedback.

PR #32, opened 2023-09-11
Use regex expression for authorization

Chaho12:
Use regex instead of contains to support more cases where more than 1 type of user members can exist and not all but some of the groups/members are allowed for User role.

PR #36, opened 2023-09-12
Improve LDAP and general security documentation

vishalya:


docs/security.md
```diff
@@ -100,12 +118,23 @@ authorization:
...
+  ldapHost: '<ldap sever>'
+  ldapPort: <port>
+  useTls: <true/false>
+  useSsl: <true/false>
+  ldapAdminBindDn: <>
+  ldapUserBaseDn: <>
+  ldapUserSearch: <>
+  ldapGroupMemberAttribute: <>
+  ldapAdminPassword: <>
+```
```
Chaho12, line 136: kinda duplicate of what i suggested in other [PR](https://github.com/trinodb/trino-gateway/pull/32/commits/a11a30d7ef9e8dfcbd3e34648f66d40cf092d781#diff-2c506cc756a5333f8fb5acb850d443980706602048f9ddb3dc515abacbfbce6a) but anyway, 
aren't these configs `ldapTrustStorePath` and `ldapTrustStorePassword` conf necessary if using either TLS/SSL? 
I think it's rare that user doesn't use either one so how about adding this confs?

docs/security.md
```diff
@@ -15,6 +15,24 @@
 
 # Security
 
+## Overview
+Trino-gateway has it's own security with it's own authetication and authorization.
+It's security is used only to authenticate/authorize it's own ui and the APIs.
```
willmostly, line 20: Both of the It's should have the apostrophe removed, it's is a contraction whereas its is possessive.

docs/security.md
```diff
@@ -15,6 +15,24 @@
...
+
+### Note for oauth2 flow:
+For oauth2 gateway uses `oidc/callback` where as trino uses `oauth2` path
+gateway should have it's own client id and all the trino clusters should have a single client id.
+gateway needs to pass thorugh the trino oauth2 requests to only one of the cluster. One way to handle that is to set a spectial rule like below:
```
willmostly, line 25: Should this be a bulleted list? Per our meeting Wednesday we should replace all instances of "gateway" with Trino Gateway.

docs/security.md
```diff
@@ -100,12 +118,23 @@ authorization:
...
 ```
+  ldapHost: '<ldap sever>'
+  ldapPort: <port>
+  useTls: <true/false>
+  useSsl: <true/false>
+  ldapAdminBindDn: <>
+  ldapUserBaseDn: <>
+  ldapUserSearch: <>
+  ldapGroupMemberAttribute: <>
+  ldapAdminPassword: <>
```
willmostly, line 135: It would be good to have some description and examples of these properties for those less familiar with ldap. E.g. the code appears to assume `ldapUserSearch` is sth like `'(&(objectclass=user)(sAMAccountName=${USER}))`. 

`ldapAdminPassword` should be grouped with `ldapAdminBindDn`, and the doc should note they are the credentials of the service user that performs the ldap search. Many users may be familiar with Trino which doesn't require a service user.

willmostly (changes requested):
Looks good - a few grammar nits, but mainly I think we need a little more explanation of the config properties.

mosabua:
Are you fixing this PR @vishalya ?

docs/security.md
vishalya: Did your PR get merged?
Yes, if you are using internal issued certs then you will need to use those specific paths.

vishalya: Is the other PR merged?

vishalya:
yes, I was focusing on the library upgrade. I will get to this one now.

docs/security.md
vishalya: Add those extra properties. somehow I missed those in the first update.

docs/security.md
vishalya: I agree that we need good documentation ldap and all the other security features like oauth2 and documentation in general. Let's do it in another PR.
My motivation behind this PR is to have a better solution than answering those repeated questions on the slack.

docs/security.md
```diff
@@ -16,6 +16,24 @@
...
+- Trino Gateway needs to pass thorugh the trino oauth2 requests to only one of the cluster. One way to handle that is to set a spectial rule like below:
+```
+  ---
+  name: "Oauth requests"
+  description: "Oauth requests need to go to a single backed"
+  condition: "request.getRequestURI.startsWith(\"/oauth2\")"
+  actions:
+    - "result.put(\"routingGroup\", \"oauth2-handler\")"
+```
+That also means you need to have a cluster with that routing group. It's ok to replicate an existing cluster backend record with a different name for that purpose.
```
mosabua, line 35: wrapping

mosabua (changes requested):

mosabua:
Let us know when feedback is addressed @vishalya

docs/security.md
```diff
@@ -16,6 +16,24 @@
 
 # Security
 
+## Overview
+Trino Gateway has its own security with its own authetication and authorization.
+These features are used only to authenticate/authorize its ui and the APIs.
+
+### Note for oauth2 flow:
```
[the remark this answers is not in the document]
vishalya: Yes, that's better!

mosabua (approved):
I fixed wrapping and grammar a bit, but left content as is. Looks good now.

Issue #39, opened 2023-09-15
Ensure consistent project name usage


mosabua:
Approved and agreed upon names are 

* "Trino Gateway" overall
* "trino-gateway" for repository and similar technical usage
* "io.trino.gateway.*" as Java package, groupId and similar namespace usage

```[tasklist]
### Tasks
- [x] Review and update documentation
- [x] Review and update readme  (maybe adding this info there..)
- [x] Review and update source code and inline javadoc
- [x] Change title in web user interface
```

mosabua:
I am gonna call this done. If we find more instances of wrong naming we can fix on the go.

PR #51, opened 2023-09-17
Ensure consistency in naming

andythsu:
https://github.com/trinodb/trino-gateway/issues/39

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# trino-gateway
+# Trino Gateway
 
 A load balancer, proxy server, and configurable routing gateway for multiple
 [Trino](https://trino.io) clusters.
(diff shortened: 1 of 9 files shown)


mosabua (changes requested):
Lets take the pom changes out of this PR .. they are in the airbase update already .. that way we avoid conflicts and this can get merged.

willmostly (approved):
Looks good, a 
```
grep -Ri -e 'trino.*gateway' | grep -v 'io.trino.gateway'
```
didn't turn up any additional nonconforming usages so I think this is good to go once @mosabua 's comment regarding titles for nav links is addressed.

mosabua (approved):
Looks good now. Thank you.

PR #53, opened 2023-09-22
Fix development environment configs

Chaho12:
Upgrade docker-compose file version
- The `start_period` option was added in file format 3.4
- https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

Specify `PGUSER=trino_gateway_db_admin` in postgresql so that it doesn't use current OS user as database user name. 
- Postgre keeps logging `FATAL:  role "root" does not exist` 
- https://stackoverflow.com/questions/60193781/postgres-with-docker-compose-gives-fatal-role-root-does-not-exist-error/60194261#60194261

Specify `PGDATABASE=trino_gateway_db` so that it doesn't use user name as DB name to access in default.
- `FATAL:  database "trino_gateway_db_admin" does not exist`

Update mysql healthcheck cmd (needs pwd)
- Mysql keeps logging `[Note] Access denied for user 'root'@'localhost' (using password: NO)`

Add support for development using m1
- Add `platform: linux/amd64` so that it can download image
- upgrade mysql version to 5.7 as there is bug with 5.5 (maybe use 8.x version as 5.x is too old?)
- https://github.com/docker/for-mac/issues/6137

### DIFF
--- a/localdev/docker-compose.yml
+++ b/localdev/docker-compose.yml
@@ -1,14 +1,15 @@
-version: "3.3"
+version: "3.4"
 services:
   mysqldb:
     environment:
       - MYSQL_ROOT_PASSWORD=root123
       - MYSQL_DATABASE=trinogateway
-    image: mysql:5.5
+    image: mysql:5.7
+    platform: linux/amd64
     ports:
       - "3306:3306"
     healthcheck:
-      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost", "-proot123"]
       start_period: 20s
       timeout: 5s
       retries: 5
@@ -26,7 +27,9 @@ services:
       - PGPORT=5432
       - POSTGRES_PASSWORD=P0stG&es
       - POSTGRES_DB=trino_gateway_db
+      - PGDATABASE=trino_gateway_db
       - POSTGRES_USER=trino_gateway_db_admin
+      - PGUSER=trino_gateway_db_admin
     ports:
       - "5432:5432"
     healthcheck:
--- a/localdev/docker-compose.yml
+++ b/localdev/docker-compose.yml
@@ -9,6 +9,7 @@ services:
     ports:
       - "3306:3306"
     healthcheck:
+      # Need root password to check MySQL server status
       test: ["CMD", "mysqladmin", "ping", "-h", "localhost", "-proot123"]
       start_period: 20s
       timeout: 5s
@@ -27,13 +28,12 @@ services:
       - PGPORT=5432
       - POSTGRES_PASSWORD=P0stG&es
       - POSTGRES_DB=trino_gateway_db
-      - PGDATABASE=trino_gateway_db
       - POSTGRES_USER=trino_gateway_db_admin
-      - PGUSER=trino_gateway_db_admin
     ports:
       - "5432:5432"
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready || exit 1"]
+      # Need user and database name to check PostgreSQL server status
+      test: ["CMD-SHELL", "pg_isready -U $$POSTGRES_USER -d $$POSTGRES_DB || exit 1"]
       interval: 10s
       timeout: 3s
       retries: 3


mosabua:
Awesome .. I will look more in detail tomorrow

localdev/docker-compose.yml
```diff
@@ -1,14 +1,15 @@
-version: "3.3"
+version: "3.4"
 services:
   mysqldb:
     environment:
       - MYSQL_ROOT_PASSWORD=root123
       - MYSQL_DATABASE=trinogateway
-    image: mysql:5.5
+    image: mysql:5.7
```
mosabua, line 7: If a newer mysql version works I would much rather upgrade more

Chaho12: I'll check to see with 8.x versions. 
5.7 was just mere patch for m1 version to work.

Chaho12: I just tried with 8.0 and there is issue with index size in [exact_match_source_selectors](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/resources/gateway-ha-persistence.sql#L67-L77) table's unique index, as it is innodb with utf8mb4 (4 bytes)

So, to resolve the issue, we need to either
1. decrease size of the variables: resource_group_id(256), source(512), environment(128), query_type(128) from 1024 -> 768 (-256).
2. Or use utf8(mb3) instead of mb4.

willmostly: It sounds like a more complex migration to MySQL 8 - maybe that should be a separate PR? If we change the schema then we should provide a migration as well.

localdev/docker-compose.yml
```diff
@@ -26,7 +27,9 @@ services:
       - PGPORT=5432
       - POSTGRES_PASSWORD=P0stG&es
       - POSTGRES_DB=trino_gateway_db
+      - PGDATABASE=trino_gateway_db
       - POSTGRES_USER=trino_gateway_db_admin
+      - PGUSER=trino_gateway_db_admin
```
willmostly, line 32: The postgres docker image uses POSTGRES_DB and POSTGRES_USER instead of PGDATABASE and PGUSER https://hub.docker.com/_/postgres. We shouldn't need to specify the same value in two different ways.

willmostly (changes requested):

localdev/docker-compose.yml
Chaho12: I gree separating PR would be better. I noticed upgrading to mysql 8 may cause other issues.

FYI. trino solves this issue by giving limits to each unique key index `UNIQUE (source(128), environment, query_type(128), resource_group_id)` but when I tried to do this in gateway
there is syntax issue when activejdbc executes sql. 
- https://github.com/trinodb/trino/blob/master/plugin/trino-resource-group-managers/src/main/resources/db/migration/mysql/V4__add_exact_match_source_selectors.sql#L8

mosabua: Yea .. lets stick to 5 for now

localdev/docker-compose.yml
mosabua: So if we remove this .. then we get the error message that @Chaho12 mentioned in the description.. so we must be doing something else wrong..

andythsu: We can remove this, and change https://github.com/trinodb/trino-gateway/pull/53/files#diff-51be12c366011b89d122169876c596658140cb407629a3eafac89827fcb3480cL33 to 
` ["CMD-SHELL","pg_isready -U $$POSTGRES_USER -d $$POSTGRES_DB || exit 1"]` (notice the `$$` instead of `$`)

andythsu (changes requested):

localdev/docker-compose.yml
mosabua: That would be better imho .. might need to add a comment though

localdev/docker-compose.yml
Chaho12: Sure i just created an issue to be handled in future
- https://github.com/trinodb/trino-gateway/issues/54

localdev/docker-compose.yml
Chaho12: Fixed and added comment :)

localdev/docker-compose.yml
```diff
@@ -1,14 +1,16 @@
...
       - MYSQL_ROOT_PASSWORD=root123
       - MYSQL_DATABASE=trinogateway
-    image: mysql:5.5
+    image: mysql:5.7
+    platform: linux/amd64
     ports:
       - "3306:3306"
     healthcheck:
-      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      # Need root password to check mysql server status
```
mosabua, line 12:
```suggestion
      # Need root password to check MySQL server status
```

localdev/docker-compose.yml
```diff
@@ -30,7 +32,8 @@ services:
     ports:
       - "5432:5432"
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready || exit 1"]
+      # Need User & DB to check postgres server status
```
mosabua, line 35:
```suggestion
      # Need user and database name to to check PostgreSQL server status
```

mosabua (approved):
Two minor updates.. then ready to go.

andythsu (approved):

Chaho12:
> Two minor updates.. then ready to go.

Updated comments!

PR #56, opened 2023-09-26
Disable javadoc linting

mosabua:
This disables javadoc linting and enables source jar creation. Therefore enables release process.

With this I can run

```
./mvnw clean install -P oss-release -Dgpg.skip=true
```

successfully and get source and javadoc jar files.


martint (approved):
Approved for now, but we should work towards being able to re-enable this.

PR #61, opened 2023-09-29
fix test name

andythsu:


gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
```diff
@@ -46,7 +46,7 @@ public void testRequestDelivery() throws Exception {
             .post(requestBody)
             .build();
     Response response = httpClient.newCall(request).execute();
-    Assert.assertEquals(EXPECTED_RESPONSE, response.body().string());
+    Assert.assertEquals(response.body().string(), EXPECTED_RESPONSE);
```
vishalya, line 49: Is this just for better understanding?

vishalya (approved):
LGTM

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
mosabua: The order is significant because it also affects the error message on a failing test.

"Presents assertion methods with a more natural parameter order. The order is always actualValue, expectedValue [, message]."

See https://www.javadoc.io/static/org.testng/testng/7.8.0/org/testng/Assert.html

mosabua: Is this messed up in other tests @andythsu ?

andythsu: Last time I checked it's just this one that's reversed

andythsu: @mosabua some files use junit4 assert and some files use testng assert

For junit4 assert, it's actually (expected, actual) https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html

For testng, it's (actual, expected)

Do we want to change everything to testng?

mosabua: Ideally we want testng removed and use junit 5 only. I am not sure how much work this is but ultimtely thats the direction we are taking on Trino and Trino Gateway. We could do this module by module if necessary. Maybe for now leave this change out of the PR and we do a separate run at this clean up.

andythsu: just out of curiosity, why do we prefer junit to testng? Is it because we want to be consistent with Trino or is junit the industrial standard? A quick research seems that testng is better than junit5 (https://www.guru99.com/junit-vs-testng.html)

mosabua: TestNG was a better a long time ago. At this stage it just is not really maintained anymore. Trino is moving from testng to junit 5 and we want to be inline with Trino itself.

andythsu: I see. I can do the migration from testng to junit5. I have removed the change from this PR.

PR #63, opened 2023-10-04
Add option for extra whitelisted paths

willmostly:
This submission allows configuring extra paths to proxy without a code change. This is useful for cases in which Trino has been extended to add extra APIs, and cases in which access to one of the paths not whitelisted by default is desired.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,45 @@
+**Trino Gateway documentation**
+
+<table>
+  <tr>
+    <td><a href="configuration.md">Configuration</a></td>
+    <td><a href="design.md">Design</a></td>
+    <td><b><a href="development.md">Development</a></b></td>
+    <td><a href="security.md">Security</a></td>
+    <td><a href="operation.md">Operation</a></td>
+    <td><a href="gateway-api.md">Gateway API</a></td>
+    <td><a href="resource-groups-api.md">Resource groups API</a></td>
+    <td><a href="routing-rules.md">Routing rules</a></td>
+    <td><a href="references.md">References</a></td>
+    <td><a href="release-notes.md">Release notes</a></td>
+  </tr>
+</table>
+
+# Configuration
+
+The Trino Gateway is configured by passing a yaml when running the start command.
+```shell
+java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha.jar server gateway-config.yml
+```
+Each component of the Trino Gateway will have a corresponding node in the configuration yaml. 
+
+## Proxying additional paths
+
+By default, Trino Gateway only proxies requests to paths starting with 
+`/v1/statement`, `/v1/query`, `/ui`, `/v1/info`, `/v1/node`, 
+`/ui/api/stats` and `/oauth`. 
+
+If you want to proxy additional paths, 
+you can add them by adding the `extraWhitelistPaths` node to your gateway 
+configuration yaml:
+
+```yaml
+extraWhitelistPaths:
+  - "/ui/insights"
+  - "/api/v1/biac"
+  - "/api/v1/dataProduct"
+  - "/api/v1/dataproduct"
+  - "/ext/faster"
+```
+
+This example enables additional proxying of any requests to path starting with the specified paths. 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
@@ -55,18 +55,20 @@ public class QueryIdCachingProxyHandler extends ProxyHandler {
 
   private final Meter requestMeter;
   private final int serverApplicationPort;
+  private final List<String> extraWhitelistPaths;
 
   public QueryIdCachingProxyHandler(
       QueryHistoryManager queryHistoryManager,
       RoutingManager routingManager,
       RoutingGroupSelector routingGroupSelector,
       int serverApplicationPort,
-      Meter requestMeter) {
+      Meter requestMeter, List<String> extraWhitelistPaths) {
     this.requestMeter = requestMeter;
     this.routingManager = routingManager;
     this.routingGroupSelector = routingGroupSelector;
     this.queryHistoryManager = queryHistoryManager;
     this.serverApplicationPort = serverApplicationPort;
+    this.extraWhitelistPaths = extraWhitelistPaths;
   }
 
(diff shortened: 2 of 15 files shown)


mosabua:
Needs docs on how to use this at least.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -218,7 +219,8 @@ private boolean isPathWhiteListed(String path) {
         || path.startsWith(V1_INFO_PATH)
         || path.startsWith(V1_NODE_PATH)
         || path.startsWith(UI_API_STATS_PATH)
-        || path.startsWith(OAUTH_PATH);
+        || path.startsWith(OAUTH_PATH)
+        || extraWhitelistPaths.stream().anyMatch(s -> path.startsWith(s));
```
vishalya, line 223: To make this code look cleaner, I am wondering if all the paths could be derived from a config list.
We can provide a default (hardcoded initialization) list for the existing paths and then append the `extraWhitelistPaths`
Also, how much overhead is `stream` going to add? This should be a really performance code.

vishalya (commented):
Are we expecting any paths to be whitelisted in near future?

mosabua:
> Are we expecting any paths to be whitelisted in near future?

Yes ... there are custom paths that need to be whitelisted for Starburst products and other custom deployments where Trino is embedded with more plugins and setups. It also potentially allows you to proxy metrics and whatever else you want to expose on additional end points on the Trino coordinator and workers. 

Furthermore we will work towards a new client protocol for Trino .. and can use this feature to support whatever endpoints it will surface on ..

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
mosabua: Have not looked at the code but... will this only load at start up time of the server.. then performance wont matter.. also wont matter if the reload is done on demand via a trigger in the UI or a REST endpoint .. I would prefer these two methods over reloading automatically all the time. That would only be okay if we know there is basically no performance impact.

vishalya: This code get's called for every request sent to the cluster and hence the comment.

willmostly: Java should evaluate chained `||` operators lazily, so as soon as any of these conditions are true it will skip the rest. Because of that, I think it is a good idea to retain the current structure and order them so that the most latency sensitive paths such as `v1/statement` are first in the list, even if it is a little ugly. 

> Are we expecting any paths to be whitelisted in near future?
This is useful for commercial extensions to Trino such as Starburst. Starburst has extra endpoints for its features such as data products and the access control API.

Lastly, we could clean up this list. At least `/ui/api/stats` should probably be removed. If someone is using it (it is an internal endpoint, not intended to be exposed but who knows...) they could add it back in through the new config.

docs/configuration.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Configuration
+
+The Trino Gateway is configured by passing a yaml when running the start command.
+```shell
+java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha.jar server gateway-config.yml
+```
+Each component of the Trino Gateway will have a corresponding node in the configuration yaml. 
+
+## Proxying additional paths
```
mosabua, line 26: add empty liner after title

docs/configuration.md
```diff
@@ -0,0 +1,43 @@
...
+By default, Trino Gateway will only proxy requests to paths starting with
```
mosabua, line 27:
```suggestion
By default, Trino Gateway only proxies requests to paths starting with 
```

docs/configuration.md
```diff
@@ -0,0 +1,43 @@
...
+The Trino Gateway is configured by passing a yaml when running the start command.
+```shell
+java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha.jar server gateway-config.yml
+```
+Each component of the Trino Gateway will have a corresponding node in the configuration yaml. 
+
+## Proxying additional paths
+By default, Trino Gateway will only proxy requests to paths starting with 
+`/v1/statement`, `/v1/query`, `/ui`, `/v1/info`, `/v1/node`, 
+`/ui/api/stats` and `/oauth`. If you would like to proxy additional paths,
```
mosabua, line 29:
```suggestion
`/ui/api/stats` and `/oauth`. 

If you want to proxy additional paths, 
```

docs/configuration.md
```diff
@@ -0,0 +1,43 @@
...
+java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar gateway-ha.jar server gateway-config.yml
+```
+Each component of the Trino Gateway will have a corresponding node in the configuration yaml. 
+
+## Proxying additional paths
+By default, Trino Gateway will only proxy requests to paths starting with 
+`/v1/statement`, `/v1/query`, `/ui`, `/v1/info`, `/v1/node`, 
+`/ui/api/stats` and `/oauth`. If you would like to proxy additional paths, 
+you can add them by adding the `extraWhitelistPaths` node to your gateway 
+configuration yaml. For example adding
```
mosabua, line 31:
```suggestion
configuration yaml:
```

docs/configuration.md
```diff
@@ -0,0 +1,43 @@
...
+```yaml
+extraWhitelistPaths:
+  - "/ui/insights"
+  - "/api/v1/biac"
+  - "/api/v1/dataProduct"
+  - "/api/v1/dataproduct"
+  - "/ext/faster"
+```
+
+would allow requests to paths starting with any of the above to be forwarded
```
mosabua, line 42: This example enables additional proxying of any requests to path starting with the specified paths.

mosabua (approved):
Just a few minor edits and resolving the conflict needed. Then this should be good.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
mosabua: We can do that cleanup later

mosabua:
Thank you .. looks great now.

PR #67, opened 2023-10-05
add dynamic header size config for http server and client

Chaho12:
Add dynamic header size config for http server and client
- Idk where to document it. Any ideas?
- https://trinodb.slack.com/archives/C059KUNPTSP/p1695708028907499

Values set in the conf are the default sizes set by Jetty
[HttpConfiguration](https://github.com/eclipse/jetty.project/blob/jetty-9.4.48.v20220622/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java#L61-L64) & [HttpClient](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java#L142-L143)
```java
  // Set size for HttpConfiguration
  private int outputBufferSize = 32 * 1024;
  private int requestHeaderSize = 8 * 1024;
  private int responseHeaderSize = 8 * 1024;

  // Set size for HttpClient
  private int requestBufferSize = 4 * 1024;
  private int responseBufferSize = 16 * 1024;
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,7 @@
         <dep.testng.version>6.10</dep.testng.version>
         <dep.trino.version>387</dep.trino.version>
         <dep.wiremock.version>3.0.1</dep.wiremock.version>
+        <dep.apache.commons.version>3.13.0</dep.apache.commons.version>
 
     </properties>
 
@@ -114,6 +115,12 @@
                 <artifactId>testng</artifactId>
                 <version>${dep.testng.version}</version>
                 <scope>test</scope>
+            </dependency>
+
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
+                <version>${dep.apache.commons.version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/proxyserver/src/main/java/io/trino/gateway/proxyserver/ProxyServerConfiguration.java
+++ b/proxyserver/src/main/java/io/trino/gateway/proxyserver/ProxyServerConfiguration.java
@@ -14,6 +14,11 @@ public class ProxyServerConfiguration {
   private String keystorePath;
   private String keystorePass;
   private boolean forwardKeystore;
+  private int outputBufferSize = 2 * 1024 * 1024;
+  private int requestHeaderSize = 2 * 1024 * 1024;
+  private int responseHeaderSize = 8 * 1024;
+  private int requestBufferSize = 4 * 1024;
+  private int responseBufferSize = 16 * 1024;
 
   protected String getPrefix() {
     return prefix;
@@ -45,5 +50,25 @@ public class ProxyServerConfiguration {
 
   protected int getLocalPort() {
     return localPort;
+  }
+
+  protected int getOutputBufferSize() {
+    return outputBufferSize;
+  }
+
+  protected int getRequestHeaderSize() {
+    return requestHeaderSize;
+  }
+
+  protected int getResponseHeaderSize() {
+    return responseHeaderSize;
+  }
+
+  protected int getRequestBufferSize() {
+    return requestBufferSize;
+  }
+
+  protected int getResponseBufferSize() {
+    return responseBufferSize;
   }
 }
--- a/proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
+++ b/proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
@@ -6,13 +6,10 @@ import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
 import com.squareup.okhttp.mockwebserver.RecordedRequest;
 
-import io.trino.gateway.proxyserver.ProxyHandler;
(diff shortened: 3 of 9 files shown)


proxyserver/src/main/java/io/trino/gateway/proxyserver/ProxyServerConfiguration.java
```diff
@@ -14,6 +14,11 @@ public class ProxyServerConfiguration {
   private String keystorePath;
   private String keystorePass;
   private boolean forwardKeystore;
+  private int outputBufferSize = 32 * 1024;
+  private int requestHeaderSize = 8 * 1024;
```
willmostly, line 18: Trino's default max header size is [2MB](https://github.com/trinodb/trino/blob/f769e9d3c35d319d16dfe8bb8d2f9f51b4223ca2/core/trino-main/src/main/java/io/trino/server/CoordinatorModule.java#L159). IMO the defaults should support that.

proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
```diff
@@ -79,13 +80,50 @@ public void testCustomHeader() throws Exception {
     }
   }
 
+  @Test
+  public void testLongHeader() throws Exception {
+    String mockResponseText = "CUSTOM LONG HEADER TEST";
+    String mockLongHeaderKey = "HEADER_LONG";
+    // Mockserver has max 8k for HTTP Header values so test with header value larger than default 4k
+    String mockLongHeaderValue = RandomStringUtils.random(5*1024, true, true);
```
willmostly, line 88: Nit: make the header size a parameter like `int headerLength = 1*1024*1024`

Similarly to above comment, lets use something close to the Trino default max header size.

proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
```diff
@@ -79,13 +80,50 @@ public void testCustomHeader() throws Exception {
...
+    int backendPort = 30000 + new Random().nextInt(1000);
+
+    MockWebServer backend = new MockWebServer();
+    backend.enqueue(new MockResponse().setBody(mockResponseText));
+    backend.play(backendPort);
+
+    int serverPort = backendPort + 1;
+    ProxyServerConfiguration config = buildConfig(backend.getUrl("/").toString(), serverPort);
+    ProxyServer proxyServer = new ProxyServer(config, new ProxyHandler());
```
willmostly, line 97: This code is used in several of the `TestProxyServer` tests - while you are in here could you extract it to a separate method?

willmostly (commented):
Looks about ready, a couple minor comments

pom.xml
```diff
@@ -66,6 +66,7 @@
         <reflections.version>0.9.10</reflections.version>
         <slf4j.version>1.7.25</slf4j.version>
         <testng.version>6.10</testng.version>
+        <apache.commons.version>3.13.0</apache.commons.version>
```
willmostly, line 69: change to `dep.apache.commons.version` to be consistent with airbase style.

willmostly (commented):
One more small change.

proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
Chaho12: Is there a way to increase jetty server's default maximum header size from 8K?

proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
Chaho12: extracted to setProxyServer method

proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
willmostly: As per David's request here https://github.com/jetty/jetty.project/issues/4541, Jetty allows large headers. Do you get an error?

Chaho12: yes, i must be missing sth. it works fine when i deploy in production so i guess it's issue with test code.
image

willmostly: Ok yeah, maybe its due to using MockWebServer, even though the error appears to be from jetty. This might be better to sort out once we start migrating away from mocks. If you can't get it working, maybe just open an issue to improve the test so we don't lose track of it.

willmostly (approved):
Please squash commits. Approved % investigation into why the test can't use a header > 8kb even though the real server can.

proxyserver/src/test/java/io/trino/gateway/proxyserver/TestProxyServer.java
Chaho12: I created an issue here :) 
- https://github.com/trinodb/trino-gateway/issues/72

Chaho12:
Squashed and rebased :)

mosabua (commented):
Looks good. I changed the commit message and merged.

PR #69, opened 2023-10-06
Ignore the unhandled ODIC properties gracefully

vishalya:
We currently have the bare minimum oauth properties supported. Any additional properties in the token result in parsing exception. The scope of the fix is to avoid the exception by ignoring unsupported properties.

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
@@ -11,7 +11,11 @@ import org.junit.jupiter.api.Test;
 @Slf4j
 public class TestOidcToken {
   @Test
-  public void testOidcTokenParams() {
+  public void testParseTokenParamsGracefully() {
+    // This test is to make sure that we simulate the condition where
+    // the OIDC providers send additional parameters in the token
+    // 'to_be_ignored' parameter should not cause any parsing exception
+    // All the other parameters should have the correct values
     OidcTokens oidcTokens = null;
     try {
       ObjectMapper objectMapper = new ObjectMapper();
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
@@ -1,0 +1,40 @@
+package io.trino.gateway.ha.security;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.trino.gateway.ha.security.LbOAuthManager.OidcTokens;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+public class TestOidcToken {
+  @Test
+  public void testOidcTokenParams() {
+    OidcTokens oidcTokens = null;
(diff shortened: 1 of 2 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
```diff
@@ -0,0 +1,39 @@
+package io.trino.gateway.ha.security;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.trino.gateway.ha.security.LbOAuthManager.OidcTokens;
+import lombok.extern.slf4j.Slf4j;
+import org.testng.Assert;
+import org.testng.annotations.Test;
```
willmostly, line 8: Should these be JUnit now?

willmostly (commented):
Looks good % the transition from testng to junit

mosabua:
Please rebase and update to use junit since that PR is now merged.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
mosabua: Yes

mosabua (changes requested):
Needs to be rebased and use junit .. then good to go imho

vishalya:
rebased and changed to junit, waiting on the check.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
```diff
@@ -0,0 +1,40 @@
...
+    OidcTokens oidcTokens = null;
+    try {
+      ObjectMapper objectMapper = new ObjectMapper();
+
+      String jsonStr = "{\"id_token\" : \"ABC235234\", "
+          + "\"access_token\" : \"AcessABCD123\", "
+          + "\"refresh_token\" : \"RefreshTKN\", "
+          + "\"token_type\" : \"TOKENType\", "
+          + "\"expires_in\" : \"123456\", "
+          + "\"to_be_ignored\" : \"XYX123456\", "
```
mosabua, line 24: Can we somehow make it more explicit what this test actually does .. e.g. it verifies that unknown parameters dont trip up parsing

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
```diff
@@ -0,0 +1,40 @@
...
+          + "\"scope\" : \"global\" "
+          + "}";
+
+      oidcTokens = objectMapper.readValue(jsonStr, OidcTokens.class);
+    } catch (JsonProcessingException ex) {
+      log.error(ex.getMessage());
+      assertTrue(false);
```
mosabua, line 31: Maybe with a message here?

mosabua (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
vishalya: It's a test, so we should be ok.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
vishalya: should be good with the last check in.

mosabua (approved):

PR #71, opened 2023-10-07
Check if url has https protocol scheme and add test

andythsu:


Chaho12 (approved):
LGTM 👍

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
```diff
@@ -62,7 +62,7 @@ public ClusterStats monitor(ProxyBackendConfiguration backend) {
       jdbcUrl = String
```
willmostly, line 62: Since we are getting smarter about setting the port based on the protocol, we should also use the backend URL to set the SSL parameter. Currently it is set in `BackendStateConfiguration`, so if you have a mixture of `http` and `https` backends it will necessarily be invalid for some of them.

gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsJdbcMonitor.java
```diff
@@ -0,0 +1,91 @@
...
+@Slf4j
+@TestInstance(Lifecycle.PER_CLASS)
+public class TestClusterStatsJdbcMonitor {
+    ClusterStatsJdbcMonitor clusterStatsJdbcMonitor;
+    @Mock
+    private Connection con;
+    @Mock
+    private PreparedStatement stmt;
+    @Mock
+    private ResultSet rs;
```
willmostly, line 37: nit - use full names like connection, preparedStatement and resultSet.

willmostly (changes requested):
Looks good, I think there is an opportunity for further improvement per my comment.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
andythsu: I'm not familiar with SSL parameters. So I'm assuming SSL can be set depending on http or https?

willmostly: yep  
https => SSL=true 
http => SSL=false

mosabua: Lets add that and then I think we should be good to merge this soon.

proxyserver/pom.xml
```diff
@@ -54,11 +54,6 @@
             <artifactId>commons-lang3</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <scope>test</scope>
-        </dependency>
```
andythsu, line 61: removing this as this is a dup of line 42

willmostly (approved):
LGTM!

PR #74, opened 2023-10-20
Fix LbAuthorizer test

mosabua:
This fixes import and signature usage so tests actually compile and run.


gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbAuthorizer.java
```diff
@@ -1,5 +1,8 @@
 package io.trino.gateway.ha.security;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
Chaho12, line 4: Oh it's jupiter API. got it. I'll use this next time :)

martint (approved):

PR #75, opened 2023-10-20
Update and clean up Trino JDBC driver dependency

mosabua:


wendigo (approved):

vishalya (commented):
We use trino jdbc connection here - 
https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java#L63

vishalya:
The `useApi` flag `false` would execute that route. so please test that route.

mosabua:
Okay .. well.. then we cant merge the PR as it stands .. need to think about how to proceed. Maybe we should refactor that code to use the JDBC driver methods for real .. 

On the other hand .. is that one query worth the overhead .. and if not .. maybe its fine to remove this feature.

Or do we like the flexibility to run whatever query in the future..  and therefore need to keep this all in place - I kinda think yes..

vishalya:
This is need to get the stats from the backend cluster. This flow is needed when the cluster is enabled with authentication which can't be handled with the basic auth. We need to think it through. We can revisit this issue in 2024, by that time the useApi route will become a solid one.

mosabua:
Updated now. Can I get an approval @vishalya @willmostly ?

PR #76, opened 2023-10-20
Fix bug on create backend to support external URL

Chaho12:
I was looking at some [PRs](https://github.com/lyft/presto-gateway/pull/122/files) from lyft/gateway to see any improvements that can be made and noticed that `externalUrl` was not added on create.
If the user wanted to set external url as [Gateway API guide](https://github.com/trinodb/trino-gateway/blob/main/docs/gateway-api.md#add-or-update-a-backend) saids, I need to create add backend, then edit external url.

There was no issue code-wise until now because proxyTo conf (a.k.a backend_url) was used instead of external_url
- https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/config/ProxyBackendConfiguration.java#L16-L21

Also updated docs/gateway-api docs as it has typos/curl syntax issue

FYI. ~Test fails until https://github.com/trinodb/trino-gateway/pull/74 is merged.~ PR is merged and test is success

### DIFF
--- a/docs/gateway-api.md
+++ b/docs/gateway-api.md
@@ -21,10 +21,10 @@
 
 ```$xslt
 curl -X POST http://localhost:8080/entity?entityType=GATEWAY_BACKEND \
- -d '{  "name": "trino-3", \
-        "proxyTo": "http://localhost:8083",\
-        "active": true, \
-        "routingGroup": "adhoc" \
+ -d '{  "name": "trino-3",
+        "proxyTo": "http://localhost:8083",
+        "active": true,
+        "routingGroup": "adhoc"
     }'
 ```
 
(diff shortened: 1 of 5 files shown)


mosabua (approved):

mosabua:
Can you also review @willmostly and @vishalya  .. from my perspective this looks good.

mosabua:
Can you change commit message to these maybe

* Correct code snippets in API docs
* Add support to use external URL for backend creation

Chaho12:
Rebased and changed commit message as you said :)

lambrospetrou:
Thanks for this fix. Can we also change the view to use the `externalUrl` instead of the `proxyTo` since when some are only resolved internally, we need to use the `externalUrl` to be able to inspect the queries.

See https://github.com/trinodb/trino-gateway/issues/81 for context.

mosabua:
@Chaho12 can you please resolve the conflict

@willmostly @vishalya @andythsu - can I get your review input

mosabua:
@lambrospetrou I think it would be good to show both in the UI. This can be done via a separate PR by @Chaho12 or yourself or whoever.

Chaho12:
> Thanks for this fix. Can we also change the view to use the externalUrl instead of the proxyTo since when some are only resolved internally, we need to use the externalUrl to be able to inspect the queries.

@lambrospetrou you can have a go to fix it! Or i can do it next week or so cause am off for holiday for the rest of the week.

andythsu (approved):
looks good to me!

PR #89, opened 2023-11-08
Add support for Oracle as backend database

posulliv:


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
```diff
@@ -29,14 +29,10 @@ public void submitQueryDetail(QueryDetail queryDetail) {
...
       connectionManager.open();
-      String sql = "select * from query_history";
       if (user.isPresent()) {
-        sql += " where user_name = '" + user.get() + "'";
+        return QueryHistory.upcast(QueryHistory.where("user_name = '" + user.get() + "'").limit(2000).orderBy("created desc"));
       }
-      return QueryHistory.upcast(QueryHistory.findBySQL(String.join(" ",
-          sql,
-          "order by created desc",
-          "limit 2000")));
```
posulliv, line 39: I removed this hard-coded SQL since `LIMIT` is not supported in Oracle.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ResourceGroups.java
```diff
@@ -77,7 +77,6 @@ public static List<ResourceGroupsDetail> upcast(List<ResourceGroups> resourceGro
    * @param resourceGroupDetail
    */
   public static void create(ResourceGroups model, ResourceGroupsDetail resourceGroupDetail) {
-    model.set(resourceGroupId, resourceGroupDetail.getResourceGroupId());
```
posulliv, line 80: I made this change to ensure auto-increment resource group ID column is used.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
```diff
@@ -29,14 +29,10 @@ public void submitQueryDetail(QueryDetail queryDetail) {
   public List<QueryDetail> fetchQueryHistory(Optional<String> user) {
     try {
       connectionManager.open();
-      String sql = "select * from query_history";
       if (user.isPresent()) {
-        sql += " where user_name = '" + user.get() + "'";
+        return QueryHistory.upcast(QueryHistory.where("user_name = '" + user.get() + "'").limit(2000).orderBy("created desc"));
```
willmostly, line 33: use `QueryHistory.userName` instead of `"user_name"` and `QueryHistory.created` instead of `"created"`

gateway-ha/src/main/resources/gateway-ha-persistence-oracle.sql
```diff
@@ -0,0 +1,70 @@
...
+    query_text VARCHAR (256),
+    created NUMBER,
+    backend_url VARCHAR (256),
+    user_name VARCHAR(256),
+    source VARCHAR(256)
+);
+CREATE INDEX query_history_created_idx ON query_history(created);
+
+CREATE TABLE resource_groups (
+    resource_group_id NUMBER GENERATED ALWAYS as IDENTITY(START with 1 INCREMENT by 1),
```
willmostly, line 20: I don't know Oracle very well, but would a `LONG` be a better choice than a `NUMBER`? Iiuc a `NUMBER` with unspecified scale supports real numbers as well as integers. Not that it probably matters much

gateway-ha/src/main/resources/gateway-ha-persistence.sql
```diff
@@ -18,7 +18,7 @@ CREATE INDEX query_history_created_idx ON query_history(created);
 
 CREATE TABLE IF NOT EXISTS resource_groups (
     resource_group_id BIGINT NOT NULL AUTO_INCREMENT,
-    name VARCHAR(250) NOT NULL UNIQUE,
+    name VARCHAR(250) NOT NULL,
```
willmostly, line 21: Why are we changing the MySQL sql for this PR?

willmostly (commented):
Just a few minor comments, thanks for the submission!

gateway-ha/src/main/resources/gateway-ha-persistence.sql
posulliv: I updated it to match the schema in trino for consistency - https://github.com/trinodb/trino/blob/master/plugin/trino-resource-group-managers/src/main/resources/db/migration/mysql/V2__add_resource_groups.sql

Happy to revert if unique names need to be enforced for gateway.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
posulliv: It appears all the attributes of the activejdbc models are private. Should we add getters for them? Not sure what the typical activejdbc pattern is for this.

gateway-ha/src/main/resources/gateway-ha-persistence-oracle.sql
posulliv: a `LONG` in oracle is not actually a numeric data type confusingly enough :) https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-F6309DF8-162F-48A4-9454-FEE59EC6644F

I would vote to use `NUMBER` as the data type for the identity column here.

gateway-ha/src/main/resources/gateway-ha-persistence.sql
posulliv: One use case for non-unique names I have had in the past with Trino is storing the same resource group definitions for different environments in the same database. This is possible since resource groups are determined by the `environment` column of the `resource_groups` table. So you could use a single database and have the same resource group definitions for multiple environments.

vishalya:
Also, a general comment here - we are supporting almost 3 dbs now, will it be a good idea to use migrations now than actual DDL like Flyway.

mosabua:
> Also, a general comment here - we are supporting almost 3 dbs now, will it be a good idea to use migrations now than actual DDL like Flyway.

Totally agree .. that could also help with installation/ bootstrap

gateway-ha/src/main/resources/gateway-ha-persistence-oracle.sql
willmostly: After reviewing, I'd agree

gateway-ha/src/main/resources/gateway-ha-persistence.sql
willmostly: I'm not sure. Trino clusters require an environment name set, while the gateway could front backends with different environment names. So environment could be less meaningful in the gateway context. OTOH, this is defined by the user, and they can choose not to complicate things. 

I guess being able to import a set of policies from Trino to the gateway probably wins out.

posulliv:
> Also, a general comment here - we are supporting almost 3 dbs now, will it be a good idea to use migrations now than actual DDL like Flyway.

We are using flyway in trino for resource group migrations so could likely reuse some of what we did there. Are we good with doing this in a follow up PR though?

willmostly:
> Are we good with doing this in a follow up PR though?
I agree that we should split these up. Either first Oracle, then Flyway, or the other way round.

willmostly (approved):

mosabua:
Fyi @ebyhr  ... any thought of adding an abstraction layer like Flyway or so? Also given you are ripping over from activejdbc to jdbi it would be good to have a coherent plan in place

willmostly (approved):
This looks ready to go. @posulliv and I have discussed offline and will begin work to introduce Flyway for schema management once this PR is merged

pom.xml
```diff
@@ -62,6 +62,7 @@
         <dep.mockito.version>5.8.0</dep.mockito.version>
         <dep.mysqlconnector.version>8.0.17</dep.mysqlconnector.version>
         <dep.okhttp.version>3.9.0</dep.okhttp.version>
+        <dep.oracle.version>21.9.0.0</dep.oracle.version>
```
ebyhr, line 65: It would be nice to use LTS version like https://github.com/trinodb/trino/pull/19851

Also, please inline this property since it's used from single place.

mosabua:
At minimum a rebase has to be done. I can review more then ..

docs/resource-groups-api.md
```diff
@@ -96,12 +96,12 @@ curl -X POST http://localhost:8080/trino/resourcegroup/delete/{INSERT_ID_HERE}
...
 ## Add a selector
 
-To add a single selector, specify all relevant fields in the body. Resource
-group id should not be specified since the database should autoincrement it.
+To add a single selector, specify all relevant fields in the body.
 
 ```$xslt
 curl -X POST http://localhost:8080/trino/selector/create \
  -d '{
+        "resourceGroupId": 1, \
```
mosabua, line 104: that seems like a bad idea .. how would you know what number to use? An id should be created automatically. Also .. this change should be independent and separate

pom.xml
mosabua: Agreed with @ebyhr ..

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistory.java
```diff
@@ -28,12 +28,12 @@
 public class QueryHistory
         extends Model
 {
-    private static final String queryId = "query_id";
-    private static final String queryText = "query_text";
-    private static final String backendUrl = "backend_url";
-    private static final String userName = "user_name";
-    private static final String source = "source";
-    private static final String created = "created";
+    public static final String queryId = "query_id";
```
mosabua, line 31: If they are public .. should they not be changed to QUERY_ID and such @ebyhr ?

gateway-ha/pom.xml
```diff
@@ -60,6 +60,12 @@
             <version>${dep.h2.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.oracle.database.jdbc</groupId>
+            <artifactId>ojdbc11</artifactId>
+            <version>${dep.oracle.version}</version>
```
mosabua, line 66: inline version number here

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistory.java
mosabua: Thought so .. thanks for confirming

posulliv: looks like these fields have been removed in latest updates.

posulliv:
@ebyhr the problem I'm facing now is that the SQL in `QueryHistoryDao` is not compatible with Oracle (`limit` is not supported in Oracle). Didn't have this issue with `activejdbc` when calling `QueryHistory.limit()`.

Any suggestions on the preferred approach to deal with this in JDBI?

ebyhr:
@posulliv We can call different dao methods based on the connection-url prefix `jdbc:oracle`.

mosabua:
Chatted with @willmostly .. I think this can be closed and restarted when it becomes relevant. Feel free to reopen after rebasing and adjusting

PR #90, opened 2023-11-08
Ensure file change is observable

willmostly:
The routing rules engine will reload the rules file if it is newer than the last set of rules, based on the timestamp from the file attributes. However, `java.nio.file.attribute.FileTime` offers a maximum precision of 1 ms. 

We see an intermittent test failure due to a rules file being changed but not reloaded. The theory is that the test writes a new rules file <1ms after the first one, so a 1ms sleep is added to prevent this.


vishalya (approved):
LGTM

mosabua (approved):

mosabua:
This is a bad hack but I think for now it can stabilize our CI setup. Over time we should refactor and clean up these and other tests to be more robust.

Issue #80, opened 2023-10-26
Feature request: Introduce different states to prevent clusters from being turned inactive programmatically


andythsu:
Currently, if the backends are not responding to healthcheck, trino-gateway will inactivate the backends. Since healthcheck only checks against healthy backends, inactive backends will never be turned to active again. Therefore, it is better to introduce different states. It is proposed to introduce healthy/pending/unhealthy states depending on the health of the backends. These states will be independent from the active/inactive states, so it ensures that unhealthy but active backends will always be checked and possibly turned active again.

`Healthy state` is defined as backends are healthy and ready to be served
`Pending state` is defined as when the backend is switched from inactive to active. It will wait until healthcheck returns success before turning backend to healthy
`Unhealthy state` is defined as backends are returning error or not responding to healthcheck. At this point, if the backends are still active, healthcheck will still check on these backends.

andythsu:
When backend is active and health check passes, health state should be HEALTHY
![image](https://github.com/trinodb/trino-gateway/assets/18670495/c7c23c93-cf98-4ab2-ba09-bf87cef6826a)

When backend is inactive, health state should be PENDING
![image](https://github.com/trinodb/trino-gateway/assets/18670495/696e0d96-206f-4ad0-b592-767b871ef421)

When backend is turned active from inactive, health state should be PENDING until health check passes
![image](https://github.com/trinodb/trino-gateway/assets/18670495/a2c6b57d-1497-4520-bcba-35557f5a27fd)

willmostly:
Instead of deactivating backends, we could switch to using the `TrinoQueueLengthRoutingTable` as default. This routing strategy will automatically ignore any backend that does not respond with the number of running and queued queries.

Chaho12:
At first I thought gateway only checks against active backends because when user disables backends, e.g for deployment purpose, it should not go back to enabled status automatically. A manual interaction is needed. 

> unhealthy but active backends can still be checked and possibly turned active again.

So, if i understood it correctly, healthy status checks only to those who has `active status as true` right? Those in active as false is not checked? Am just clarifying what you intended :)

andythsu:
> At first I thought gateway only checks against active backends because when user disables backends, e.g for deployment purpose, it should not go back to enabled status automatically. A manual interaction is needed.
> 
> > unhealthy but active backends can still be checked and possibly turned active again.
> 
[...]

PR #91, opened 2023-11-13
Introduced 'Pending' state for backend's health

andythsu:
This commit adds a 'Pending' state to keep track of backend's health.

Resolves: #80

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
@@ -1,5 +1,13 @@
 package io.trino.gateway.ha;
 
+import static io.trino.gateway.ha.HaGatewayTestUtils.WAIT_FOR_BACKEND_IN_SECONDS;
+import static io.trino.gateway.ha.HaGatewayTestUtils.prepareMockGetBackend;
+import static io.trino.gateway.ha.HaGatewayTestUtils.prepareMockPostBackend;
+import static io.trino.gateway.ha.HaGatewayTestUtils.setUpBackend;
+import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.UI_API_QUEUED_LIST_PATH;
+import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.UI_API_STATS_PATH;
+import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.UI_LOGIN_PATH;
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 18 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaRoutingManager.java
```diff
@@ -49,16 +50,16 @@ private void addMockBackends() {
...
+      haRoutingManager.upateBackEndHealth(backend, BackendHealthState.HEALTHY);
     }
 
     //Keep only 1st backend as healthy, mark all the others as unhealthy
     assert (!backendManager.getAllActiveBackends().isEmpty());
 
     for (int i = 1; i < numBackends; i++) {
       backend = groupName + i;
-      haRoutingManager.upateBackEndHealth(backend, false);
+      haRoutingManager.upateBackEndHealth(backend, BackendHealthState.UNHEALTHY);
```
rdsarvar, line 62: nitpick: perhaps here we want to hit all enum states in the test?

for ex: 50% pending and 50% unhealthy which both represent "not yet healthy"
-> `i % 2 == 0 ? BackendHealthState.UNHEALTHY : BackendHealthState.PENDING`

(note: adding this note to all blocks as i might not be able to get back to them; feel free to resolve with or without change)

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
```diff
@@ -18,8 +28,11 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.api.extension.ExtendWith;
 
 @TestInstance(Lifecycle.PER_CLASS)
+@ExtendWith(DropwizardExtensionsSupport.class)
+@Slf4j
 public class TestGatewayHaSingleBackend {
```
rdsarvar, line 36: question: is the goal of this test case to be a sub-system test?


my thinking is: if we just wanted to test that the gateway chooses backends properly based on health as a single unit you could replace:
```
    SECONDS.sleep(WAIT_FOR_BACKEND_IN_SECONDS);
```

with:

```
    HaGatewayLauncher haGatewayLauncher = new HaGatewayLauncher("io.trino");
    haGatewayLauncher.run(args);

    // Now populate the backend
    setUpBackend(
            "trino1", "http://localhost:" + backendPort, "externalUrl", true, "adhoc", routerPort);

    // When a backend becomes active for the first time OR is transitioned from inactive to active it will not yet be
    // treated as healthy. Without a real health check happening we will extract the RoutingManager and set them to healthy manually
    Field privateInjectorField = BaseApp.class.
            getDeclaredField("injector");
    privateInjectorField.setAccessible(true);

    Injector injector = (Injector) privateInjectorField.get(haGatewayLauncher);
    injector.getInstance(RoutingManager.class).updateBackendHealth("trino1", BackendHealth.HEALTHY);
```

This hack to grab the class with reflection is just to keep the test as close to the module as possible (and do without a sleep pause). Curious on your thoughts for it.

If the goal is to cover as much repository as possible in the test as sub-system instead of integration then fair enough 👍

andythsu: great idea! included this in my new PR

PR #92, opened 2023-11-14
fix basic config load issue, register log filtering

vishalya:


andythsu (approved):
LGTM

gateway-ha/gateway-ha-config.yml
```diff
@@ -6,7 +6,7 @@ requestRouter:
   port: 8080
   name: trinoRouter
   historySize: 1000
-  requestBufferSize: 8 * 1024
+  requestBufferSize: 8195
```
rdsarvar, line 9: question: how come we opted into 8195 instead of 8192?

vishalya: It's a typo, thanks for catching it.

mosabua (approved):
Looks good. I will squash and merge

PR #94, opened 2023-11-17
added support for oauth2 audience

andythsu:


gateway-ha/src/main/java/io/trino/gateway/ha/config/SelfSignKeyPairConfiguration.java
```diff
@@ -1,8 +1,12 @@
 package io.trino.gateway.ha.config;
 
+import lombok.AllArgsConstructor;
```
willmostly, line 3: We are in the process of removing Lombok, ptal at @mosabua 's recent PRs. This will need to be refactored.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbTokenUtil.java
```diff
@@ -28,11 +30,16 @@ public static boolean validateToken(String idToken, RSAPublicKey publicKey, Stri
...
-      JWT.require(algorithm)
+      Verification verification =
+          JWT.require(algorithm)
           .withIssuer(issuer)
-          .acceptLeeway(60 * 60) // Expired tokens are valid for an hour
-          .build()
-          .verify(idToken);
+          .acceptLeeway(60 * 60); // Expired tokens are valid for an hour
+
+      if(audiences != null){
```
willmostly, line 38: Trino codestyle prefers using `Optional` to handling nulls explicitly, so the signature should be changed to `Optional<List<String>> audiences` and the check here should be `audiences.isPresent()`.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -76,7 +76,7 @@ public Optional<Map<String, Claim>> getClaimsFromIdToken(String idToken) {
     try {
       DecodedJWT jwt = JWT.decode(idToken);
 
-      if (LbTokenUtil.validateToken(idToken, lbKeyProvider.getRsaPublicKey(), jwt.getIssuer())) {
+      if (LbTokenUtil.validateToken(idToken, lbKeyProvider.getRsaPublicKey(), jwt.getIssuer(), null)) {
```
willmostly, line 79: use `Optional.empty()` instead of `null`, see below comment.

willmostly (changes requested):
Looks good mod comments

gateway-ha/src/main/java/io/trino/gateway/ha/config/SelfSignKeyPairConfiguration.java
andythsu: I can go ahead and get rid of lombok here but since @mosabua already has a PR for it, I think it is best to keep this PR as is and merge this PR in before his PR to avoid merge conflicts

mosabua:
Delomboking is done .. you need to rebase this now .. sorry .. also might want to wait a bit since we are about to get the checkstyle refactor in as well hopefully

mosabua:
Should all be good to rebase now .. lots of refactoring got in

mosabua:
Is this replaced by #164 and we can close this PR @andythsu ?

andythsu:
yes I'll close it

PR #97, opened 2023-11-21
Add installation instructions

mosabua:
Could you review and confirm @willmostly @vishalya @Chaho12 and @ebyhr 

Also note that I am not happy with the navigation and the order but will address that in a separate follow up PR.


Chaho12 (approved):
LGTM~

ebyhr (approved):

Issue #77, opened 2023-10-23
Feature Request: Routing based on number of queued and running queries


alaturqua:
### Description
Trino Gateway routes queries randomly within a routing group and does not consider the number of already running and queued queries on a cluster within a routing group

### Issue
It can happen that some queries take longer and and queue of the routed cluster keeps growing

#### Example scenario:
Assumption 3 Clusters in a routing group.
- Each cluster receive 50 queries
- cluster 1 and 2 process all the queries and have 0 queries remaining in queued queries
- cluster 3 has received some complex queries and processing needs longer and blocks other queries
- Gateway routes randomly again additional 50 queries to each cluster
- all clusters receive 50 queries to process again
- cluster 3 has 100 queries now, because the one or couple complex queries are in running state and blocking others.

![image](https://github.com/trinodb/trino-gateway/assets/17232317/f55a8610-1555-40ac-a119-501cb6cdb020)
![image](https://github.com/trinodb/trino-gateway/assets/17232317/99949eb5-158f-4fcf-9f40-d318d03df2f9)

### Feature Request
Implement a solution which routes queries based on health status of the backend. It should consider the number of running and queued queries and route new queries to clusters with less load

### Findings on current code
There is already code for this, but it is not used as of now. [ClusterStateListenerModule](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/module/ClusterStateListenerModule.java#L42-L43) does not have code for it.
https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueueLengthRoutingTable.java
https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoQueueLengthChecker.java

vishalya:
I can send a PR to try it out, we will need to test out thoroughly.

alaturqua:
I could test it in our dev environment with realistic scenarios.

alaturqua:
I run some tests on the fork/PR in our dev environment.

Tests were done in a realistic scenario using 3x Clusters as trino backend with a routing group named **looker**. Routing rules make sure that every query with source = 'Looker' is routed to routing group **looker**.
```yaml
---
name: "Looker"
description: "Route Looker queries to Looker"
priority: 6
condition: "request.getHeader(\"X-Trino-Source\") contains \"Looker\""
actions:
[...]

Issue #84, opened 2023-10-30
Load balancing scheme in Trino gateway?


pulkitkalia1994:
What's the load balancing scheme used by Trino gateway? And is there a way to change it?

lambrospetrou:
By default the gateway picks a random cluster within the matched routing group. See https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java#L249 and https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java#L91

mosabua:
@vishalya can you potentially take this on to document as part of the update for the linked PR ?

oneonestar:
Please see https://trinodb.github.io/trino-gateway/routing-logic/#routing-based-on-cookies

PR #98, opened 2023-11-24
Configure the router policy via router provider module

vishalya:
- This PR providers a way to configure the que length based router. 
- This router and it's logic was there, now it can be configured via the config file.
- Please check the configuration.md, Router Policy section to understand on how to configure the current default router vs the queue length based router
- HaGatewayProviderModule needed to be refactored and the db dependencies are moved into another base module, because of that the router module needs to be configured explicitly.
- Hopefully this PR fixes https://github.com/trinodb/trino-gateway/issues/77 and https://github.com/trinodb/trino-gateway/issues/84


gateway-ha/gateway-ha-config.yml
```diff
@@ -36,8 +36,11 @@ monitor:
 
 modules:
   - io.trino.gateway.ha.module.HaGatewayProviderModule
-  - io.trino.gateway.ha.module.ClusterStateListenerModule
+#  - io.trino.gateway.ha.module.ClusterStateListenerModule
+#  - io.trino.gateway.ha.module.BasicRouterProvider
   - io.trino.gateway.ha.module.ClusterStatsMonitorModule
+  - io.trino.gateway.ha.module.QueueLengthListenerModule
+  - io.trino.gateway.ha.module.QueueLengthRouterProvider
```
willmostly, line 43: The `QueueLengthRouterProvider` requires that the `QueueLengthListenerModule` is configured (right?). If it isn't, what kind of error does the end user see? 

If its not an easy one to understand, we should add some checking in Baseapp to ensure that dependent modules are loaded. `AppConfiguration.getModules()` should give us the list of loaded modules.

willmostly (approved):
Looks good overall. My only real concern is that as config gets more complex, we should make sure we're giving understandable error messages to users and avoiding returning hundreds of lines of Guice injection errors.

mosabua:
This needs a rebase before we can look at it again.

vishalya:
Yes, let me rebase and then I can also add a new router which is much simpler. Currently, we are testing it internally.

vishalya:
- I have added a new router - `QueryCountBasedRouter` which is proved by `QueryCountBasedRouterProvider`. The clusterstats are supplied to the routers and any hence can be used by any new router in future. The new router can be enabled by adding the provider to config. The full modules section in the config would look like below 
```
modules:
  - io.trino.gateway.ha.module.HaGatewayProviderModule
  - io.trino.gateway.ha.module.ClusterStateListenerModule
  - io.trino.gateway.ha.module.ClusterStatsMonitorModule
  - io.trino.gateway.ha.module.QueryCountBasedRouterProvider
```
- The existing router could be used replace the last line with `BasicRouterProvider`. We can also set the existing router as default one, if none is provided explicitly in the config.
```
modules:
  - io.trino.gateway.ha.module.HaGatewayProviderModule
  - io.trino.gateway.ha.module.ClusterStateListenerModule
  - io.trino.gateway.ha.module.ClusterStatsMonitorModule
  - io.trino.gateway.ha.module.BasicRouterProvider
```
- `TrinoQueueLengthRoutingTable` didn't seem to work that well, so I am dropping it and eventually we can get rid of that class all together.

- The background task which collected stats can be configured to run with second as an unit, instead of the minute.  It is a part of the `monitor` section of the config. It's not a mandatory parameter, as the previous default of 1 minute is respected with 60 second default. The old parameter `taskDelayMin` is removed.
```
monitor:
  connectionTimeout: 15
  taskDelaySeconds: 10
```

The logic for the `QueryCountBasedRouter` is in the comments.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,147 @@
...
+    // else we assume it would be scheduled immediately and we increment the stats for the running
+    // queries
+
+    private Optional<ClusterStats> getClusterToRoute(String user, String routingGroup)
+    {
+        List<ClusterStats> filteredList = null;
+        Lock readLock = lock.readLock();
+        try {
+            readLock.lock();
+            filteredList = clusterStats.stream()
```
willmostly, line 67: This would be a good use case for a `Map<Array<ClusterStats>>` or potentially `com.google.common.collect.Table` if you need to look up on various attributes vs filtering a list

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    public String getBackendForRoutingGroup(String routingGroup, String user)
+    {
+        Optional<ClusterStats> cluster = getClusterToRoute(user, routingGroup);
+        if (cluster.isPresent()) {
+            cluster.orElseThrow().updateLocalStats(user);
+            return cluster.orElseThrow().getProxyTo();
+        }
+        return null;
```
willmostly, line 115: Use `Optional`

willmostly (changes requested):
A few things to address, mainly around the locking and potential performance impacts

alaturqua:
Worked today on resolving the issues alongside @vishalya. The solution now operates as intended, effectively distributing queries based on the number of queued queries evenly across a routing group. A merge would greatly enhance our current setup and would be great feature for the community.

docs/configuration.md
```diff
@@ -42,4 +42,22 @@ extraWhitelistPaths:
...
 ```
 
-This example enables additional proxying of any requests to path starting with the specified paths. 
\ No newline at end of file
+This example enables additional proxying of any requests to path starting with the specified paths.
+
+## Router Policy
+There are 2 routers you can choose from
+### Que length based
+- This distributed the queries based on the number of queries running
```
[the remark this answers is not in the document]
vishalya: Yes, it should route it to the cluster with the fewest running queries.

gateway-ha/gateway-ha-config.yml
vishalya: In my latest change, I am providing the BasicRouter by default and config can override it. Is that ok?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -177,4 +177,20 @@ public String toString()
                 .add("userQueuedCount", userQueuedCount)
                 .toString();
     }
+
+    public synchronized void updateLocalStats(String user)
+    {
+        // The live stats refresh every few seconds, so we update the stats immediately
+        // so that they can be used for next queries to route
+        // We assume that if a user has queued queries then newly arriving queries
+        // for that user would also be queued
```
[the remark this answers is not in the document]
vishalya: - I hear you on the assumption, but one could only assume in absense of the live stats. Currently, the cluster stats interval can be set with a granularity of seconds (default is 60 secs), but we are doing ok with 10 sec interval. This assumption phase is applicable during that 10 seconds and when the fresh stats arrive we start afresh with them.
- There is tremendous overall gain in the performance, the BasicRouter (the one on the main branch) check db for every request and it's literally 50 to 100 times slower. I have verified that.

gateway-ha/src/main/java/io/trino/gateway/ha/module/BasicRouterProvider.java
```diff
@@ -0,0 +1,46 @@
...
+package io.trino.gateway.ha.module;
+
+import com.google.inject.Provides;
+import io.dropwizard.core.setup.Environment;
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.router.HaQueryHistoryManager;
+import io.trino.gateway.ha.router.HaRoutingManager;
+import io.trino.gateway.ha.router.RoutingManager;
+
+public class BasicRouterProvider
```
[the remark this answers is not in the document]
vishalya: so `StochasticRoutingManager` and `StochasticRoutingManagerProvide`? I am ok with that.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: Yeah, If it becomes more complicated then yes, may be the next iteration when we get more stats for the cost based routing we can use it. At this time the filter and sort is just simple and clean IMO.

gateway-ha/gateway-ha-config.yml
willmostly: sgtm

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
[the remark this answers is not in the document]
willmostly: You mean this is faster compared to the `TrinoQueueLengthRoutingTable` I assume?

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+        finally {
+            readLock.unlock();
+        }
+
+        if (filteredList.isEmpty()) {
+            return Optional.empty();
+        }
+
+        Collections.sort(filteredList, new Comparator<ClusterStats>()
```
willmostly, line 84: replace with `Collections.max()` instead of sort and get 0

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,151 @@
...
+        {
+            public int compare(ClusterStats lhs, ClusterStats rhs)
```
willmostly, line 86: `ClusterStats` should extend `Comparable`, then you can implement `compareTo` using this logic. That way it can be reused in other routers.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,151 @@
...
+        });
+        // The first cluster is right one
+        return Optional.of(filteredList.get(0));
+    }
+
+    private Optional<String> getBackendForRoutingGroup(String routingGroup, String user)
+    {
+        Optional<ClusterStats> cluster = getClusterToRoute(user, routingGroup);
+        if (cluster.isPresent()) {
+            cluster.orElseThrow().updateLocalStats(user);
```
willmostly, line 116: Why is `orElseThrow` used? `isPresent()` was just checked. If this is needed then an exception supplier should be passed.

willmostly (commented):
Getting close, just spotted a few more code improvements to make. 

Can you additionally remove `TrinoQueueLengthRoutingTable` as part of this PR? It will be orphaned once this merged.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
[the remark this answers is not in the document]
vishalya: Even compared to the BasicRouter it's very fast, since BasicRouter keeps cheking db for the active/inactive state.

gateway-ha/gateway-ha-config.yml
vishalya: done.

gateway-ha/src/main/java/io/trino/gateway/ha/module/BasicRouterProvider.java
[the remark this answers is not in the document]
vishalya: done

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: That's true!

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: That's because the static check doesn't allow it, but I guess there should be a way around it.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsMonitor.java
```diff
@@ -15,7 +15,14 @@
 
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 
-public interface ClusterStatsMonitor
+public abstract class ClusterStatsMonitor
 {
-    ClusterStats monitor(ProxyBackendConfiguration backend);
+    abstract ClusterStats monitor(ProxyBackendConfiguration backend);
+
+    protected void populateClusterStats(ClusterStats.Builder stats, ProxyBackendConfiguration backend)
```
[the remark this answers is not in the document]
vishalya: I was thinking about it, but I was trying to avoid a dependency on the `ProxyBackendConfiguration`, but since you brought up the same point, makes me thing that that's a better approach.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: It would be `Collections.min()` but I got the idea.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
willmostly: looks good, can you remove the comment block of the previous code?

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    @Override
+    public String provideBackendForRoutingGroup(String routingGroup, String user)
+    {
+        Optional<String> proxyUrl = getBackendForRoutingGroup(routingGroup, user);
+        if (proxyUrl.isPresent()) {
+            return proxyUrl.orElseThrow();
+        }
+        return provideAdhocBackend(user);
```
willmostly, line 143:
```suggestion
return getBackendForRoutingGroup(routingGroup, user).OrElse(provideAdhocBackend(user));
```

willmostly (changes requested):
Just a few loose ends

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
```diff
@@ -0,0 +1,250 @@
...
+        queryCountBasedRouter = null;
+        clusters = null;
+    }
+
+    // Helper function to generate the ClusterStat list
+    private static ImmutableList getClusterStatsList(String routingGroup)
+    {
+        ImmutableList.Builder<ClusterStats> clustersBuilder = new ImmutableList.Builder();
+        // Set Cluster1 stats
+        {
```
ebyhr, line 52: Use `clustersBuilder.add()` instead.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
```diff
@@ -0,0 +1,250 @@
...
+    static final int LEAST_QUEUED_COUNT = 1;
+    static final int SAME_QUERY_COUNT = 5;
+    QueryCountBasedRouter queryCountBasedRouter;
+    ImmutableList<ClusterStats> clusters;
+
+    TestQueryCountBasedRouter()
+    {
+        queryCountBasedRouter = null;
+        clusters = null;
+    }
```
ebyhr, line 45: This constructor seems redundant.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
```diff
@@ -0,0 +1,250 @@
...
+
+    // Helper function to generate the ClusterStat list
+    private static ImmutableList getClusterStatsList(String routingGroup)
```
ebyhr, line 48: Please fix the warning about raw usage of parameterized list class in this class. 
```suggestion
    private static List<ClusterStats> getClusterStatsList(String routingGroup)
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
```diff
@@ -0,0 +1,250 @@
...
+import com.google.common.collect.ImmutableList;
+import io.trino.gateway.ha.clustermonitor.ClusterStats;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
```
ebyhr, line 25: Use AssertJ.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+
+public class QueryCountBasedRouter
```
ebyhr, line 29: Fix the indentation in this class.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,119 @@
...
+            extends StochasticRoutingManager
+{
+    private static final Logger log = LoggerFactory.getLogger(QueryCountBasedRouter.class);
+    private ArrayList<ClusterStats> clusterStats;
+    private static final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
+
+    public QueryCountBasedRouter(GatewayBackendManager gatewayBackendManager,
+                                 QueryHistoryManager queryHistoryManager)
```
ebyhr, line 37: Note that the convention of the method argument is single line or wrap per arguments:
```java
    public QueryCountBasedRouter(GatewayBackendManager gatewayBackendManager, QueryHistoryManager queryHistoryManager)
```
or
```java
    public QueryCountBasedRouter(
            GatewayBackendManager gatewayBackendManager,
            QueryHistoryManager queryHistoryManager)
```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -47,9 +187,10 @@ public static final class Builder
         private String routingGroup;
         private Map<String, Integer> userQueuedCount;
 
-        private Builder(String clusterId)
+        public Builder(String clusterId)
```
ebyhr, line 190: Revert. See another comment about getClusterStatsBuilder method.

ebyhr (commented):
Just skimmed.

gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
```diff
@@ -66,6 +69,34 @@ public abstract class BaseApp<T extends AppConfiguration>
...
+            logger.info("Trying to load module [{}]", clazz);
+            Object ob =
+                    Class.forName(clazz)
+                            .getConstructor(configuration.getClass(), Environment.class)
+                            .newInstance(configuration, environment);
+            return Optional.of((AppModule) ob);
+        }
+        catch (Exception e) {
+            logger.error("Could not instantiate module [" + clazz + "]", e);
+        }
```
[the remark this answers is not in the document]
vishalya: that makes sense, caller code would be clean.
I would rethrow the exception as they could be many different exceptions to handle after logging the exception.
How are we planning to continue using guice in future?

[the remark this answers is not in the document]
vishalya: How do you want us to fail? :-)

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: not sure, what's the issue here, can you elaborate? I rely on the check-style to guide me.

vishalya:
I have taken care of the review comments, the rebase is coming next.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
ebyhr: > I rely on the check-style to guide me.

Are you really using Airlift code style? https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#code-style

The indention at L30 & L72-74 are wrong.
```java
            extends StochasticRoutingManager
↓
        extends StochasticRoutingManager
```

```java
            filteredList = clusterStats.stream()
                .filter(stats -> stats.healthy())
                .filter(stats -> routingGroup.equals(stats.routingGroup()))
                .collect(Collectors.toList());
↓
            filteredList = clusterStats.stream()
                    .filter(stats -> stats.healthy())
                    .filter(stats -> routingGroup.equals(stats.routingGroup()))
                    .collect(Collectors.toList());
```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -96,22 +237,63 @@ public Builder routingGroup(String routingGroup)
...
+            userQueuedCount.put(user, count + 1);
+            return;
+        }
+        // Else the we assume that the query would be running
+        // so update the clusterstat with the +1 running queries
+        ++runningQueryCount;
+    }
+
+    // The method compared the stats from a user's perspective to find out the most suitable queue
+    public int compareStats(ClusterStats rhs, String user)
```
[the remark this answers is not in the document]
ebyhr: Reminder.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -96,22 +237,63 @@ public Builder routingGroup(String routingGroup)
...
+        // First check if the user has any queries queued
+        int compareUserQueue = Integer.compare(userQueuedCount().getOrDefault(user, 0),
+                rhs.userQueuedCount().getOrDefault(user, 0));
+
+        if (compareUserQueue != 0) {
+            return compareUserQueue;
         }
+
+        int compareClusterQueue = Integer.compare(queuedQueryCount(),
+                rhs.queuedQueryCount());
```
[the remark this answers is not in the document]
ebyhr: Reminder.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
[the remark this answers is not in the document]
oneonestar: This synchronized usage is incorrect.
Reads and writes on `userQueuedCount` and `runningQueryCount` must be synchronized.

Change from record class to normal class requires a lot of works on reimplementing getter/setters/equals.  I think we can continue to use record class by using AtomicInteger and ConcurrentHashMap.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+
+public class QueryCountBasedRouter
+            extends StochasticRoutingManager
+{
+    private static final Logger log = LoggerFactory.getLogger(QueryCountBasedRouter.class);
+    private ArrayList<ClusterStats> clusterStats;
+    private static final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
```
oneonestar, line 34: The operations on `clusterStats` are read all and replace all.
`ReentrantReadWriteLock` can be replaced by `AtomicReference<List<ClusterStats>>`.

oneonestar (changes requested):
Some comments on the implementation of concurrency.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: Yes, I am using the airbase check-style. The checks on the PR are also not catching these error.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
vishalya: This one line change is there with last commit, but not sure that is the complete fix. How do I see the details on the errors (or original error).  I don't see it on my local build with `./mvnw test-compile`; also didn't find any warning here -
https://github.com/trinodb/trino-gateway/actions/runs/8249053834/job/22560656476?pr=98

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: Good catch on the read, thanks!
Both suggestions for the synchronization are good.
I might need to modify the flow a bit to use the record, with record I won't be able to modify the member attributes.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
andythsu: The current check-style doesn't seem to throw exception https://github.com/airlift/airbase/blob/ac2bd5fe5a9cbfe351096906fe729dcb6aeab460/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml#L156
Not sure if there's a way to enforce the indentation to let check-style fail if errors like this occur again going forward

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
vishalya: - I am going back to the class for `ClusterStats` as the router needs to update it, it can be achieved by copying the records back and forth, but it won't be performant.
- I have moved the reading/sorting and updating of the cluster stats at the router level - inside `QueryCountBasedRouter`, for 2 reasons.

  1. The logic is specific to the router.

  2. The logic needs to read the stats, sort and the update the counters in the `ClusterStats` in a single transaction. This would avoid the race condition while processing simultaneous requests.

vishalya:
I have pushed the changes to copy the stats locally and the router now works with them as discussed.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,234 @@
...
+public class QueryCountBasedRouter
+        extends StochasticRoutingManager
+{
+    private static final Logger log = LoggerFactory.getLogger(QueryCountBasedRouter.class);
+    private List<LocalStats> clusterStats;
+
+    List<LocalStats> clusterStats()
+    {
+        return clusterStats;
+    }
```
oneonestar, line 38: Let's prevent someone misuse this function in the future.
```suggestion
    @VisibleForTesting
    synchronized List<LocalStats> clusterStats()
    {
        return new ArrayList<>(clusterStats);
    }
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -0,0 +1,234 @@
...
+                    .collect(Collectors.toList());
+
+        if (filteredList.isEmpty()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(Collections.min(filteredList, (lhs, rhs) -> compareStats(lhs, rhs, user)));
+    }
+
+    private void updateLocalStats(LocalStats stats, String user)
```
oneonestar, line 193: Add `synchronized`.

oneonestar (commented):
Just some minor comments. Others LGTM.

I think coding style can be improved overtime.
Hope we can get this merged soon to avoid conflicts on migration to JDBI & Airlift.

mosabua (approved):
Looks good. As agreed with @willmostly @Chaho12 and others in the dev sync, we are merging this now and will address any code style and other minor issues in follow up PRs.

PR #99, opened 2023-11-29
Add check before borrowing from pool

Chaho12:
Add checking connection before borrowing from the pool for stability.
- In our k8s, unless KA is set for sockets, it cleans up process that has been idle for 10 min.
- So, if the LDAP connection socket is not used and goes idle for more than 10 min, it gets deleted without gateway server knowing that is is invalid.
- Then next time ldapClient borrows the connection and makes request, it fails as log below.
image

- https://trinodb.slack.com/archives/C059KUNPTSP/p1700796849182119

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
@@ -62,9 +62,10 @@ public class LbLdapClient
 
         //A single connection and keep it alive
         GenericObjectPoolConfig poolConfig = new GenericObjectPoolConfig();
-        poolConfig.setMaxIdle(1);
-        poolConfig.setMaxTotal(1);
-        poolConfig.setMinIdle(0);
+        poolConfig.setMaxIdle(ldapConfig.getPoolMaxIdle());
+        poolConfig.setMaxTotal(ldapConfig.getPoolMaxTotal());
+        poolConfig.setMinIdle(ldapConfig.getPoolMinIdle());
+        poolConfig.setTestOnBorrow(ldapConfig.isPoolTestOnBorrow());
 
         ValidatingPoolableLdapConnectionFactory validatingFactory =
                 new ValidatingPoolableLdapConnectionFactory(defaultFactory);
--- a/docs/security.md
+++ b/docs/security.md
@@ -151,4 +151,8 @@ The LDAP config file should have the following contents:
   ldapAdminPassword: <>
   ldapTrustStorePath: <for a secure ldap connectivity>
   ldapTrustStorePassword: '<for a secure ldap connectivity>'
+  poolMaxIdle: 8
+  poolMaxTotal: 8
+  poolMinIdle: 0
+  poolTestOnBorrow: true
 ```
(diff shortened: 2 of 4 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -76,6 +76,7 @@ public LbLdapClient(LdapConfiguration ldapConfig) {
     poolConfig.setMaxIdle(1);
     poolConfig.setMaxTotal(1);
```
Chaho12, line 77: Btw, how about increasing number of connections to idk default value of 8? 
Any ideas on why it is set as 1?

willmostly (commented):
Looks like a good improvement, I think this will be ready once the connection validation is sorted out.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -76,6 +76,7 @@ public LbLdapClient(LdapConfiguration ldapConfig) {
...
     poolConfig.setMinIdle(0);
+    poolConfig.setTestOnBorrow(true);
```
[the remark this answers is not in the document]
Chaho12: I think your looking at the document for tomcat jdbc, which in this case it is rather apache commons pool2. Here is the [document](https://commons.apache.org/proper/commons-pool/apidocs/org/apache/commons/pool2/impl/BaseGenericObjectPool.html#setTestOnBorrow(boolean)) for it.

> Sets whether objects borrowed from the pool will be validated before being returned from the borrowObject() method. Validation is performed by the validateObject() method of the factory associated with the pool. If the object fails to validate, it will be removed from the pool and destroyed, and a new attempt will be made to borrow an object from the pool.

To elaborate more, `validateObject()` is depended on by the validating factory in line 80, `ValidatingPoolableLdapConnectionFactory validatingFactory = new ValidatingPoolableLdapConnectionFactory(defaultFactory);`. [ValidatingPoolableLdapConnectionFactory](https://nightlies.apache.org/directory/api/2.1.2/apidocs/org/apache/directory/ldap/client/api/ValidatingPoolableLdapConnectionFactory.html) extends [AbstractPoolableLdapConnectionFactory](https://nightlies.apache.org/directory/api/2.1.2/apidocs/org/apache/directory/ldap/client/api/AbstractPoolableLdapConnectionFactory.html#validateObject-org.apache.commons.pool2.PooledObject-), where validateObject is defined. `setValidator` explains the default validator (An implementation of [LdapConnectionValidator](https://nightlies.apache.org/directory/api/2.1.2/apidocs/org/apache/directory/ldap/client/api/LdapConnectionValidator.html) that attempts a simple lookup on the rootDSE.)

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
Chaho12: Sure, changed it so that it reads Authentication ldap conf. PTAL

willmostly (approved):
Looks good!

willmostly:
@mosabua I think this one is good to go

Chaho12:
rebased and fixed conflicts :)

mosabua (approved):

mosabua:
We should document all the security config in more detail btw @willmostly @Chaho12 and @vishalya

Issue #102, opened 2023-11-30
Remove usage of Lombok


mosabua:
Lombok uses private API in Javac and that blocks the upgrade to Java 21. We should not need Lombok in general anyway.

mosabua:
Probably can do this with IntelliJ and the Delombok feature - https://stackoverflow.com/questions/75209785/how-to-quickly-replace-all-lombok-annotations-with-generated-methods

mosabua:
All PRs merged

PR #104, opened 2023-11-30
Remove Lombok usage in baseapp

mosabua:
Part of fixing https://github.com/trinodb/trino-gateway/issues/102

Using the Delombok refactor tool from IntelliJ


baseapp/src/main/java/io/trino/gateway/baseapp/BaseApp.java
```diff
@@ -45,7 +45,6 @@
  * <p>GuiceApplication also makes {@link com.codahale.metrics.MetricRegistry} available for
  * injection.
  */
-@Slf4j
 public abstract class BaseApp<T extends AppConfiguration> extends Application<T> {
 
   private static final Logger logger = LoggerFactory.getLogger(BaseApp.class);
```
mosabua, line 50: Seems like there were two separate loggers (log and logger) before this cleanup

findepi (approved):

mosabua:
Thank you @willmostly and @findepi

PR #110, opened 2023-12-14
Add quickstart scripts

willmostly:
These pages add an example of how to stand up a Trino Gateway locally and add two backends. They are intended as a Quickstart option for someone interested in exploring the Gateway, and not to replace the development enviroment. They were presented in the Fall 2023 Trino Summit.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2,9 +2,10 @@
 
 <table>
   <tr>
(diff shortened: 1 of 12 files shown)


mosabua (approved):
Looks good. I did some minor navigation and grammar adjustments. Shipping.

Issue #86, opened 2023-11-03
Add container


mosabua:
Besides the executable JAR we should at least provide a container image to run Trino Gateway.

```[tasklist]
### Tasks
- [x] Add separate folder and Maven module to build docker container
- [x] Add release step to publish container
- [x] Update documentation to add how to run with Docker container and custom config
```

rdsarvar:
Hi @mosabua, are we looking to have the gateway follow the same release strategy as the core Trino repository? Do you happen to have a link to any docs/runbooks/etc which detail how the existing releases are done?

(I've seen https://github.com/trinodb/trino/tree/master/core/docker but was curious to where the building and releasing of the images actually occurs)

Or is that portion of it done manually by a maintainer?

mosabua:
The build for the release is done by @martint typically as part of the release - scripts are in the release-scripts repo .. 

For trino-gateway I think we want a similar maven module setup that runs as part of the build but the scripts for the release will also be in this same trino-gateway repo

mosabua:
Also see https://github.com/trinodb/release-scripts/blob/master/trino-docker.sh

mosabua:
Duplicate of #48

mosabua:
We could also consider reviving a docker compose setup that we removed with https://github.com/trinodb/trino-gateway/pull/146/files .. if users desire that.

mosabua:
All done with release 6.

mosabua:
Also includes work in 

* #198 
* #227 
* #209

PR #113, opened 2023-12-17
Added dockerization for the Trino Gateway

rdsarvar:
Adding documentation, image creation, and compose file modifications required for running the Trino Gateway in a docker container.

Docker setup based on what @wendigo did with Trino

Potential implementation for #86


docker/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+ARG TRINO_GATEWAY_VERSION
+COPY --chown=trino:trino gateway-ha-${TRINO_GATEWAY_VERSION}-jar-with-dependencies.jar /usr/lib/trino
+COPY --chown=trino:trino gateway-ha /usr/lib/trino
+
+ENV GATEWAY_JAR_PATH "/usr/lib/trino/gateway-ha-${TRINO_GATEWAY_VERSION}-jar-with-dependencies.jar"
+
+EXPOSE 8080
+USER trino:trino
+CMD java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar "${GATEWAY_JAR_PATH}" "server" "/opt/trino/gateway-ha-config.yml"
```
wendigo, line 57: https://github.com/airlift/airlift/tree/master/launcher could be used here

rdsarvar: Thanks for the quick review 😄 

I began looking at the Airlift launcher but would that be something we should be doing once we migrate the Gateway off of Dropwizard? Or is there a simple way to wrap the existing web server that I am missing?

Will try to find more time to dive into the launcher setup within the next week

mosabua: I think we might delay using launcher .. the problem is also that it pull in a requirement for Python .. ideally we could avoid that. For now I think we are okay with just using the command.

rdsarvar: @mosabua valid point on the Python requirements, was mostly looking for a non-hacky solution for pulling configurations from YAML in the health checker

I also considered installing jq+yq then doing it in bash, open to other solutions

would wrapping the installs in a venv be sufficient? or would we prefer to limit pip installs completely for Python in this image

mosabua:
So I have not really looked at the details of this PR at all, but here are a few points to consider

* we should use same maven and docker setup as in trino itself so that the docker build is part of the maven build
* we should use the same container image as base
* we should use Eclipse Temurin 17 
* we will try to get to Java 21 and Eclipse Temurin 21 over time 
* release script for publishing should be in this repo (I think I saw it in the PR already) and should support multi arch

I am currently going through a refactor PR and then we will try to get all other PRs reviewed and merged .. including this one. Thank you for all the work you have done already.

docker/Dockerfile
rdsarvar: To loop back here, the Trino health checking script is a bit simpler because extraction from a .properties file is pretty straight forward

In our case the Gateway is leveraging YAML which complicates things slightly, that's why I was hoping to use a qualified YAML parsing package in Python.

I can think of a couple of options if we'd like to skip out on Python:

1. Write something with awk, an example I've tested is:
```bash
function get_2nd_level_yaml_key() {
    local parent_key=$1
    local target_key=$2
    local yaml_file_path=$3

    # In order to get the nested key we will leverage awk to do some parsing
    # First we search for the parent_key, once we find it we set 'in_block' to true
    # Then we will attempt to find the 2nd level target key based on:
    #  1. There's only spaces before the target_key
    #  2. There are no '#'s before the target_key indicating it's in a comment
    # If we determine that we have left the 'parent_key' block we will set 'in_block' to false
    awk -v parent_key=${parent_key} \
        -v target_key=${target_key} \
        '
        match($0, "^" parent_key ":") { in_block=1; next } \
        in_block {
            if (match($0, "^[[:space:]]+" target_key ":") \
              && !/^[[:space:]]*#/ ) {
               sub(/:/, "", $2); print $2; exit
            } else if ($0 ~ /^[^\t ]/) {
                in_block=0
            }
        }
        ' \
        ${yaml_file_path}
}
```
2. Install jq + yq (note: jq you can microdnf install, yq we'd need to explicitly download)

The awk version works but would break in the future if we ever nested the keys we were attempting to match inside of requestRouter. Example:
```
requestRouter:
    newKey:
        port: 8081 # this gets matched
    port: 8080
```

I'm open to suggestions on best way to extract the proper ports and scheme

rdsarvar: If we go along the 'awk' path, to handle the case above I've modified the awk to attempt to find the indentation required to be a 2nd level key:
```bash
function get_2nd_level_yaml_key() {
    local parent_key=$1
    local target_key=$2
    local yaml_file_path=$3

    # In order to get the nested key we will leverage awk to do some parsing
    # First we search for the parent_key, once we find it we set 'in_block' to true
    # Next because we don't know the indentation levels being provided, we will attempt
    # to find the indentation level of the 2nd level keys.
    # Then we will attempt to find the 2nd level target key based on:
    #  1. There's only spaces before the target_key
    #  2. The indentation level is equal to the 'found' 2nd level indentation
    #  3. There are no '#'s before the target_key indicating it's in a comment
    # If we decide that the key matches our expectations, we will print the matched value
    # If we determine that we have left the 'parent_key' block we will set 'in_block' to false
    awk -v parent_key=${parent_key} \
        -v target_key=${target_key} \
        '
        function count_indentation(line) {
            match(line, /^[[:space:]]*/);
            return RLENGTH;
        }
        match($0, "^" parent_key ":") {
            in_block=1;
            parent_indent_level = count_indentation($0);
            next;
        } \
        in_block {
            current_indent_level = count_indentation($0);
            if ( !first_level_indentation \
              && match($0, "^[[:space:]]+.*:") \
              && current_indent_level > parent_indent_level \
              && !/^[[:space:]]*#/ ) {
                first_level_indentation = current_indent_level;
            }
            if (match($0, "^[[:space:]]+" target_key ":") \
              && current_indent_level == first_level_indentation \
              && !/^[[:space:]]*#/ ) {
               sub(/:/, "", $2);
               print $2;
            } else if ($0 ~ /^[^\t  #]/) {
                in_block=0
            }
        }
        ' \
        ${yaml_file_path}
}
```

Sort of unwieldy 😄

mosabua: Cant we just hit a HTTP endpoint for healthcheck .. also maybe I am missing the point .. but how are healthcheck and launcher related?

rdsarvar: They're unrelated we could probably spin up a new thread to talk about the health checker, I was replying in this thread as it answered the question of 'why Python'.

> Cant we just hit a HTTP endpoint for healthcheck
The overkill above was mostly just to pull out 2 fields from YAML: `requestRouter.port` and `requestRouter.ssl`. With those 2 parameters we can decide on the proper endpoint to hit for health checking, ex:
```
# Where scheme is http or https
endpoint="${scheme}://localhost:${port}/api/public/backends"
```

Without code changes if we were to roll this out for just HTTP, users would need to override the health checker when using `ssl: true` to use a test healthiness on HTTPS.

TBH there should be dedicated liveness/readiness routes being served by the Gateway itself that we can use, I propose we create a new Github issue for it to minimize the scope of this MR to just docker + release. Thoughts?

(If I'm making any assumptions anywhere please correct me, I'm quite new to this repo 😄 )

docker/build.sh
```diff
@@ -0,0 +1,138 @@
...
+
+usage() {
+    cat <<EOF 1>&2
+Usage: $0 [-h] [-a <ARCHITECTURES>] [-r <VERSION>]
+Builds the Trino Docker image
+
+-h       Display help
+-a       Build the specified comma-separated architectures, defaults to amd64,arm64,ppc64le
+-r       Build the specified Trino release version, downloads all required artifacts
+-j       Build the Trino release with specified Temurin JDK release
```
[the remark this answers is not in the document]
rdsarvar: I've updated all instances in this file of `Trino` to `Trino Gateway` 👍

docker/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    update-alternatives --install /usr/bin/python python /usr/bin/python3 1 && \
+    pip3 install requests pyyaml && \
+    groupadd trino --gid 1000 && \
+    useradd trino --uid 1000 --gid 1000 --create-home && \
+    mkdir -p /usr/lib/trino && \
+    chown -R "trino:trino" /usr/lib/trino /opt/trino
+
+ARG TRINO_GATEWAY_VERSION
+COPY --chown=trino:trino gateway-ha-${TRINO_GATEWAY_VERSION}-jar-with-dependencies.jar /usr/lib/trino
+COPY --chown=trino:trino gateway-ha /usr/lib/trino
```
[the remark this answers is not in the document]
rdsarvar: I've removed the COPY from line 50 by adding the jar to the 'gateway-ha' directory in the build.sh script, let me know if that matches what you meant

docker/bin/health-check
```diff
@@ -0,0 +1,75 @@
...
+if [ "$(get_2nd_level_yaml_key 'requestRouter' 'ssl' "$config")" == "true" ]; then
+    scheme=https
+fi
+
+potential_port=$(get_2nd_level_yaml_key 'requestRouter' 'port' "$config")
+if [ "${potential_port}" != "" ]; then
+    port=${potential_port}
+fi
+
+endpoint="${scheme}://localhost:${port}/api/public/backends"
```
rdsarvar, line 69: @mosabua is it sufficient to leave the health checking endpoint on 'backends' for now?

I was thinking it might be worth a separate Github issue for implementing liveness/readiness endpoints in preparation for K8s. That story could then have a task to modify the health-checker here to use one of them

mosabua: Thats fine.

mosabua:
Please rebase and resolve conflicts.

mosabua:
@cla-bot check

mosabua:
Let me know if you are ready for review again @rdsarvar

pom.xml
```diff
@@ -43,7 +43,8 @@
 
         <air.java.version>17</air.java.version>
         <air.check.skip-enforcer>true</air.check.skip-enforcer>
-        <air.check.skip-dependency>true</air.check.skip-dependency>
+        <air.check.skip-dependency>false</air.check.skip-dependency>
```
mosabua, line 46: that wont pass the build

pom.xml
```diff
@@ -43,7 +43,8 @@
...
+        <air.check.fail-dependency>false</air.check.fail-dependency>
```
mosabua, line 47: leave this out as well .. we are fixing all this in separate work

mosabua:
I think we should merge this setup to get started and improve from there. What do you think @willmostly @Chaho12 @vishalya @wendigo 

We just cut version 5 so we have some time to improve before we ship this .. and we could even still release 6 without pushing the container.

mosabua:
Please address any feedback that it still outstanding and squash commits @rdsarvar

pom.xml
rdsarvar: updated both by reverting the changes

Chaho12 (approved):
Sounds good :) 
Btw do you need gateway-ha-config-docker.yml file? I find it duplicate with [gateway-ha-config.yml file](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/gateway-ha-config.yml). Will there be any needs for config to be different?

rdsarvar:
> Sounds good :) Btw do you need gateway-ha-config-docker.yml file? I find it duplicate with [gateway-ha-config.yml file](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/gateway-ha-config.yml). Will there be any needs for config to be different?

The only real reason I did it is to be able to target the `postgres` container from inside of the gateway container:
```yaml
  jdbcUrl: jdbc:postgresql://postgres:5432/trino_gateway_db
```

Due to 'localhost' inside of the container referring to itself and not the underlying host running the container.

If we don't want to use the isolated bridged networking mode, we could use `host` network mode and that would bind the containers directly to the network running the daemon. I'm just not 100% sure how well that'd play if we eventually wanted to run this in CI, would need to see if using docker-in-docker if it thinks of the DinD host as the network to use, or if it tries to get to the physical host. 🤔 

Do you think there's an easy way to maybe at runtime convert instances of `127.0.0.1`/`localhost` to `host.docker.internal`?

mosabua (approved):
Okay .. I tested it all out now and got it all working.

The PR as it stands is functional and good enough to merge.

There were a bunch of things I ran into with regards to macos/arm64 and maven wrapper. I will send a follow up PR for improvements related to that.

PR #115, opened 2023-12-21
Apply checkstyle in baseapp and proxyserver

ebyhr:


mosabua:
Please confirm that this commit does not change any functionality and is a pure source reformat and maybe add this info in the commit message @ebyhr 

Thanks for sending the PR btw.. that was our next step ..

ebyhr:
@mosabua Added commit body.

baseapp/pom.xml
```diff
@@ -12,6 +12,10 @@
     <artifactId>baseapp</artifactId>
     <packaging>jar</packaging>
 
+    <properties>
```
mosabua, line 15: Great .. this is a good temporary approach until the whole project works.

mosabua (approved):
Awesome. Thank you so much for taking this on.

PR #119, opened 2023-12-29
Remove redundant code

ebyhr:
Removed AppConfiguration class as the setters methods are unsued.


mosabua:
There is a lot of refactoring and clean up like that we need to do. I would like to first get the source formatting fixed if thats ok..

PR #120, opened 2023-12-30
Remove `org.apache.commons:commons-lang3` and sort entries in pom.xml

ebyhr:


pom.xml
```diff
@@ -6,13 +6,13 @@
     <parent>
         <groupId>io.airlift</groupId>
         <artifactId>airbase</artifactId>
-        <version>146</version>
+        <version>147</version>
```
mosabua, line 9: Please mention this update in the a commit message or do it in a separate commit

mosabua (approved):
One minor nit .. but essentially good to go.

pom.xml
ebyhr: The commit body already mentions: 
```
Also, upgrade airbase to 147 that contains sortpom-maven-plugin.
```

PR #125, opened 2023-12-31
Remove redundant code

ebyhr:


mosabua:
This goes very far in removing code... and you please explain reasoning for the various removals in the commit message or split things up into separate commits that have more info? 

We need to be sure we are not removing stuff that is actually used or needed and we just dont know details yet. 

I would also love some input after these details are provided from @willmostly @vishalya and @Chaho12

mosabua (changes requested):
Needs more info added on reasoning and relevant details.

PR #126, opened 2023-12-31
Fix COMPACT_NO_ARRAY checkstyle failure

ebyhr:


mosabua (approved):
I am fine with this change.. but how would we allow multiple roles.. is the array syntax allowed then?

wendigo (approved):

wendigo:
@mosabua Yes

PR #131, opened 2023-12-31
Avoid defining redundant dependency versions in root pom.xml

ebyhr:
No need to define `dep.{name}.version` when it's called from one place. Also, the dependency version which is used in a module should be moved to the module's pom.xml.


mosabua (changes requested):
Considering we are inheriting dep mangement from airbase I think this approach is not ideal. Instead we should keep the existing approach to ensure correct override across all modules and all transitive dependencies as well. 

Ideally as a clean up we add all dependencies to the parent pom in dependency management and then remove all versions from the the module poms. I can do that in a separate PR if you want.

PR #132, opened 2024-01-02
Enforce dependency

ebyhr:


mosabua (changes requested):
See remarks in other PR. Ideally we have all dependencies managed in the parent pom and then we use them in the module poms without any version. 

If we have properities for all dependencies is a separate topic and I can go with either approach .. but as soon as there is more than one dependency with a shared number then it should be controlled by a property. And it should also be controlled by a property if we inherit it from airbase.

PR #135, opened 2024-01-02
Convert UserConfiguration to record

ebyhr:


mosabua (approved):
This is awesome .. gotta love that we managed to get it upgraded to Java 17 in the initial push. Now we are reaping the benefits with clean ups like this.

PR #140, opened 2024-01-03
Improve TestOidcToken

ebyhr:
Remove redundant catch block and add Language annotation.

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
@@ -13,46 +13,36 @@
  */
 package io.trino.gateway.ha.security;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.trino.gateway.ha.security.LbOAuthManager.OidcTokens;
+import org.intellij.lang.annotations.Language;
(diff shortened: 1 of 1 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
```diff
@@ -13,46 +13,33 @@
...
-                    + "}";
-
-            oidcTokens = objectMapper.readValue(jsonStr, OidcTokens.class);
-        }
-        catch (JsonProcessingException ex) {
-            log.error(ex.getMessage());
-            fail();
-        }
+        @Language("JSON")
+        String jsonStr = "{\"id_token\" : \"ABC235234\", "
```
mosabua, line 34: We should clean this up to a multi line string

mosabua (dismissed):

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOidcToken.java
ebyhr: I didn't use it intentionally as it inserts new lines - the value will be different from the original text.

mosabua: Fair enough

mosabua (approved):
Looks good.

PR #141, opened 2024-01-03
Skip spotbugs instead of false fail-spotbugs

ebyhr:
`air.check.skip-spotbugs` with true is same as Trino repository. 
`air.check.fail-spotbugs` with false leaves noisy error messages.


mosabua (approved):
Looks good. Ideally I would like to get to a level where we can leave it on and dont have any failures.. but that might be long way away.

PR #146, opened 2024-01-04
Add gateway runner with docker containers

ebyhr:


mosabua (dismissed):
I am not convinced we should get rid of the docker compose setup. The runner is great but only for Java developers...

wdyt @vishalya @willmostly @Chaho12 ?

ebyhr:
The runner is pretty easy to use compared to the existing steps. I believe even non-Java developers can use it.

mosabua:
Ok ..  lets get rid of docker compose .. but lets add info on how to use the runner somewhere.

docs/development.md
```diff
@@ -18,45 +18,15 @@
...
-
-`docker-compose exec postgres sh -c 'PGPASSWORD="P0stG&es" psql -h localhost -p 5432 -U trino_gateway_db_admin -d trino_gateway_db -f /etc/postgresql/gateway-ha-persistence-postgres.sql'`
-
-#### Add the backends for postgres
-
-`docker-compose exec postgres sh -c 'PGPASSWORD="P0stG&es" psql -h localhost -p 5432 -U trino_gateway_db_admin -d trino_gateway_db -f /etc/postgresql/add_backends_postgres.sql'`
-
-It would add 2 trino backend records which can be used for the development and
-testing
+The simplest way to run Trino Gateway for development is to run the `TrinoGatewayRunner` class
```
mosabua, line 29: Add more details here for running in IDE and in command line

mosabua:
```suggestion
The best way to run Trino Gateway for development is to run the `TrinoGatewayRunner` class
```

ebyhr: Addressed comments.

rdsarvar:
> The runner is pretty easy to use compared to the existing steps. I believe even non-Java developers can use it. Note that the existing step requires maven build either.

One drawback I can think of from this new approach is it's built more like a manual integration test. Modifications and state that might've been applied to containers (ex: postgres) that you want persisted will spin down alongside the application run.

An example workflow being: You spin up the backend DB and initialize, spin up the gateway, run some queries, modify how the gateway handles query history in some fashion, and then attempt to spin the gateway back up again. In the current setup the DB state would clear and you'd be required to either: re-do all the previous steps, change the code and mount your changes as an init script, update the backend container to mount it's data to a volume, etc.

I think that's the core gain from leveraging a compose file is that you get a bunch of functionality (+life cycle management) out of the box through a few YAML tags.

Note: The way I see it is once the gateway is dockerized, it'd be nice to provide a compose file which includes the latest Gateway image and allows for new coming users to hop into a README, run `docker compose -f localdev/docker-compose.yml up -d` and then they'd have a detached but connected running ecosystem where they can freely spin a gateway container up/down.

Feel free to ignore this if you'd like 😄 I'll be away from the internet for the next week or so and unfortunately won't be able to respond

ebyhr:
We can simply provide a flag to reuse containers like `TESTCONTAINERS_REUSE_ENABLE` in Trino. No need to redo all steps.

docs/development.md
```diff
@@ -18,46 +18,20 @@
 
 # Development
 
-## How to setup a dev environment
+## Build requirements
 
-Step 1: setup mysql. Install docker with docker-compose and run the below
-command when setting up first time:
+* Mac OS X or Linux
+* Java 17+, 64-bit
```
mosabua, line 24: 17+ is only true after https://github.com/trinodb/trino-gateway/pull/156 is merged.

docs/development.md
```diff
@@ -18,46 +18,20 @@
...
-- This setup helps you develop and test any routing rules for the trino
-- Both trino services would have a single `system` catalog
-- Add the catalog properties files in ` bin/localdev/coordinator/` for
-  additional catalogs
-
-```
-cd localdev
-docker-compose up -d
+The best way to run Trino Gateway for development is to run the `TrinoGatewayRunner` class.
+You need to run `io.trino.gateway.TrinoGatewayRunner.main()` method on IDE or execute the following command:
```
mosabua, line 30:
```suggestion
You need to run `io.trino.gateway.TrinoGatewayRunner.main()` method on your IDE or execute the following command:

```

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.testcontainers.utility.MountableFile.forClasspathResource;
+
+public final class TrinoGatewayRunner
+{
+    private TrinoGatewayRunner() {}
+
+    public static void main(String[] args)
+            throws Exception
+    {
+        TrinoContainer trino1 = new TrinoContainer("trinodb/trino");
```
mosabua, line 32: Should we fix the version or provide an option to set it .. maybe even different ones for the two containers?

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
```diff
@@ -0,0 +1,59 @@
...
+        trino1.setPortBindings(List.of("8081:8080"));
+        trino1.start();
+        TrinoContainer trino2 = new TrinoContainer("trinodb/trino");
+        trino2.setPortBindings(List.of("8082:8080"));
+        trino2.start();
+
+        PostgreSQLContainer<?> postgres = new PostgreSQLContainer("postgres");
```
mosabua, line 39: Should we fix version here as well?

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
```diff
@@ -0,0 +1,59 @@
...
+        postgres.withUsername("trino_gateway_db_admin");
+        postgres.withPassword("P0stG&es");
+        postgres.withDatabaseName("trino_gateway_db");
+        postgres.withCopyFileToContainer(forClasspathResource("gateway-ha-persistence-postgres.sql"), "/docker-entrypoint-initdb.d/1-gateway-ha-persistence-postgres.sql");
+        postgres.withCopyFileToContainer(forClasspathResource("add_backends_postgres.sql"), "/docker-entrypoint-initdb.d/2-add_backends_postgres.sql");
+        postgres.setPortBindings(List.of("5432:5432"));
+        postgres.start();
+
+        MySQLContainer<?> mysql = new MySQLContainer("mysql:5.7");
```
mosabua, line 48: maybe pull version out into optional config

mosabua (changes requested):
Lets also make sure that any references to the updated sql files are correct.. the installation and quickstart guides might use and mention them. Please check.

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
Chaho12: Fixing version sounds more stable. 
Then we could write sth like `compatible with Trino 435+`
- I rarely think think there would be any compatibility issue, but some drastic API change might happen

What would be reasons for having different versions? A testbed to check if it is allright to upgrade version?

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
ebyhr: What's the "optional config"?

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
Chaho12: I agree. Am not too familiar with postgres versions, but mysql versions DO have issues (like field size or primary key index issue) so it might not be compatible with old versions when developing.

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
mosabua: I think we should use some sort of variable for the version that has a default and can be overridden by a passed parameter to the overall class .. and we should do the same for postgresql and trino

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
mosabua: Yeah .. we have to add this to the docs yet as well .. on my plate..

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
ebyhr: Modifying the code is fine for me. At least, we don't provide such a parameter in query runners of Trino.

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
ebyhr: This query runner is just for local testing and isn't ensured in CI. We shouldn't document the version.

ebyhr:
Addressed comments.

gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java
mosabua: We can do that as a later enhancement

docs/development.md
```diff
@@ -18,46 +18,20 @@
...
 
-- This setup helps you develop and test any routing rules for the trino
-- Both trino services would have a single `system` catalog
-- Add the catalog properties files in ` bin/localdev/coordinator/` for
-  additional catalogs
-
-```
-cd localdev
-docker-compose up -d
+The best way to run Trino Gateway for development is to run the `TrinoGatewayRunner` class.
```
mosabua, line 29: Wrap

docs/development.md
```diff
@@ -18,46 +18,20 @@
...
+You need to run `io.trino.gateway.TrinoGatewayRunner.main()` method on your IDE or execute the following command:
+```sh
```
mosabua, line 31: Empty line above snippet

mosabua (approved):
I think we should merge this and improve later. If we find people still want docker compose setups we can introduce them again later when we get a container going and work towards helm charts/operator or whatever ..

Issue #79, opened 2023-10-24
Update to Java 21


mosabua:
Follow along the Java 21 requirement update for Trino itself. Currently the codebase does not compile with Java 21. Probably needs more dependency updates as well.

mosabua:
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project gateway-ha: Fatal error compiling: java.lang.NoSuchFieldError: Class com.sun.tools.javac.tree.JCTree$JCImport does not have member field 'com.sun.tools.javac.tree.JCTree qualid' -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project gateway-ha: Fatal error compiling

mosabua:
Might be just need for jvm config update or so

flavioiov:
I had the same issue ...and the solution was update lombok dependecy :

       <dependency>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok</artifactId>
            <version>1.18.30</version>
            <type>jar</type>
        </dependency>

mosabua:
We removed lombok usage completely, but there are still build/test failures with Java 21.

mosabua:
As of 2024-01-05 I get these test errors. 

[ERROR] Errors:
[ERROR]   TestRoutingGroupSelector.testByRoutingRulesEngine:70 » NoClassDefFound Could not initialize class org.mvel2.compiler.ExpressionCompiler
[ERROR]   TestRoutingGroupSelector.testByRoutingRulesEngine:70 » NoClassDefFound Could not initialize class org.mvel2.compiler.ExpressionCompiler
[ERROR]   TestRoutingGroupSelector.testByRoutingRulesEngine:70 » NoClassDefFound Could not initialize class org.mvel2.compiler.ExpressionCompiler
[ERROR]   TestRoutingGroupSelector.testByRoutingRulesEngine:70 » NoClassDefFound Could not initialize class org.mvel2.compiler.ExpressionCompiler
[ERROR]   TestRoutingGroupSelector.testByRoutingRulesEngineFileChange:128 » NoClassDefFound Could not initialize class org.mvel2.compiler.ExpressionCompiler
[...]

PR #149, opened 2024-01-05
Upgrade to Java 21

ebyhr:
Fixes https://github.com/trinodb/trino-gateway/issues/79
https://github.com/trinodb/trino/pull/20010 upgraded Java to 21 in Trino

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,10 +23,10 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v3
-      - name: Set up JDK 17
+      - name: Set up JDK 21
         uses: actions/setup-java@v3
         with:
-          java-version: '17'
+          java-version: '21'
           distribution: 'temurin'
       - name: Build with Maven
         run: $MAVEN clean verify
--- a/docs/development.md
+++ b/docs/development.md
@@ -60,7 +60,7 @@ testing
 
 ### Build and run
 
-This project requires Java 17. Note that higher version of Java have not been
+This project requires Java 21. Note that higher version of Java have not been
 verified and may run into unexpected issues.
 
 Run `./mvnw clean install` to build `trino-gateway`. VM options required for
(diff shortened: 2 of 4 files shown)


mosabua:
Nice .. do we want to upgrade now, wait until Trino is on Java 21 or wait even longer? Wdyt @vishalya @willmostly @Chaho12 

Maybe we can change it to be able to build with Java 21 (upgrade dependency) .. but not require it yet.

mosabua (dismissed):
Imho we should at least wait until Trino requires Java 21 to require Java 21 for Trino Gateway.

docs/release-notes.md
```diff
@@ -50,6 +50,6 @@ Gateway](https://github.com/lyft/presto-gateway/) 1.9.5 codebase
 * Add PostgreSQL support for backend database. (https://github.com/trinodb/trino-gateway/pull/13)
 * Allow routing of `/v1/node` endpoint URL. (https://github.com/trinodb/trino-gateway/pull/27)
 * Filter logs for sensitive information. (https://github.com/trinodb/trino-gateway/pull/9)
-* Require Java 17 for build and runtime. (https://github.com/trinodb/trino-gateway/pull/16)
+* Require Java 21 for build and runtime. (https://github.com/trinodb/trino-gateway/pull/16)
```
mosabua, line 53: This is wrong

Chaho12:
I agree, Trino doesn't officially require Java 21 as of now, and not many users are using java 21 so I think we should wait for a while.
Are there any big performance improvement with new version?

mosabua:
I think we should pull this PR apart and merge the update of the dependency to allow building with Java 21 .. and then update to Java 21 after a release or two

mosabua:
Cherri picked mvel upgrade into https://github.com/trinodb/trino-gateway/pull/156

mosabua:
We should look at updating and verifying dependencies first .. or at least rebase and test this. Team discussion is positive about going to Java 21 in line with Trino soon though.

PR #152, opened 2024-01-06
Replace MockWebServer with dockerized Trino

ebyhr:


mosabua (approved):
Looks good and covers everything the old test really achieved, but cleaner. Nice!

Issue #123, opened 2023-12-30
Flaky TestRoutingGroupSelector.testByRoutingRulesEngineFileChange


ebyhr:
https://github.com/trinodb/trino-gateway/actions/runs/7361226191/job/20038470483?pr=117
```
Error:  io.trino.gateway.ha.router.TestRoutingGroupSelector.testByRoutingRulesEngineFileChange -- Time elapsed: 0.007 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <etl2> but was: <etl>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1141)
	at io.trino.gateway.ha.router.TestRoutingGroupSelector.testByRoutingRulesEngineFileChange(TestRoutingGroupSelector.java:138)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

ebyhr:
https://github.com/trinodb/trino-gateway/actions/runs/7373251027/job/20062272299

ebyhr:
https://github.com/trinodb/trino-gateway/actions/runs/7373338926/job/20062427799?pr=131

ebyhr:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/File.html#lastModified()
> While the unit of time of the return value is milliseconds, the granularity of the value depends on the underlying file system and may be larger. For example, some file systems use time stamps in units of seconds.

```
2024-01-11T02:07:11.3585818Z INFO  [2024-01-11 02:07:11,349] io.trino.gateway.ha.router.TestRoutingGroupSelector: etl rule file created at: 1704938831345
[...]

PR #157, opened 2024-01-10
Add logging for flaky test

willmostly:
To address [123](https://github.com/trinodb/trino-gateway/issues/123), we need to first add logging to determine if the file change is detectable by comparing timestamps with millisecond precision.

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
@@ -19,6 +19,8 @@ import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FileWriter;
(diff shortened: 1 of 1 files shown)


mosabua (approved):
Good for now .. over time we should fix and also change assertions to use assertj

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -123,6 +127,9 @@ public void testByRoutingRulesEngineFileChange()
                         + "actions:\n"
                         + "  - \"result.put(\\\"routingGroup\\\", \\\"etl\\\")\"");
         fw.close();
+        file.getPath();
+        long etlLastModified = file.lastModified();
+        log.info("etl rule file created at: " + etlLastModified);
```
ebyhr, line 132: No need to change as temporal code, but we cloud use logger template:
```java
        log.info("etl rule file created at: {}", etlLastModified);
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -147,6 +154,9 @@ public void testByRoutingRulesEngineFileChange()
                         + "actions:\n"
                         + "  - \"result.put(\\\"routingGroup\\\", \\\"etl2\\\")\""); // change from etl to etl2
         fw.close();
+        long etl2LastModified = file.lastModified();
+        log.info("etl2 rule file created at: " + etl2LastModified);
+        assertNotEquals(etl2LastModified, etlLastModified);
```
ebyhr, line 159: nit: We could include message in assertion method instead of logging:
```java
assertNotEquals(long unexpected, long actual, String message)
```

PR #158, opened 2024-01-11
Minor cleanup in tests

ebyhr:


mosabua (approved):
Looks good. Should we just get rid of junit assertion usage altogether over time. I think that would follow the test guide we are currently getting together for Trino .. right @martint ?

PR #160, opened 2024-01-11
Fix rule file test

willmostly:
Fixes https://github.com/trinodb/trino-gateway/issues/123

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
@@ -19,8 +19,6 @@ import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FileWriter;
@@ -28,8 +26,8 @@ import java.util.stream.Stream;
 
 import static io.trino.gateway.ha.router.RoutingGroupSelector.ROUTING_GROUP_HEADER;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
@@ -37,7 +35,6 @@ import static org.mockito.Mockito.when;
 @TestInstance(Lifecycle.PER_CLASS)
 public class TestRoutingGroupSelector
 {
-    private static final Logger log = LoggerFactory.getLogger(TestRoutingGroupSelector.class);
     public static final String TRINO_SOURCE_HEADER = "X-Trino-Source";
     public static final String TRINO_CLIENT_TAGS_HEADER = "X-Trino-Client-Tags";
 
@@ -111,7 +108,6 @@ public class TestRoutingGroupSelector
         assertNull(routingGroupSelector.findRoutingGroup(mockRequest));
     }
 
-    //Todo: The functionality of reading the file before every request needs to be smarter
     @Test
     public void testByRoutingRulesEngineFileChange()
             throws Exception
@@ -127,9 +123,7 @@ public class TestRoutingGroupSelector
                         + "actions:\n"
                         + "  - \"result.put(\\\"routingGroup\\\", \\\"etl\\\")\"");
         fw.close();
-        file.getPath();
-        long etlLastModified = file.lastModified();
-        log.info("etl rule file created at: " + etlLastModified);
+        long lastModifed = file.lastModified();
 
         RoutingGroupSelector routingGroupSelector =
                 RoutingGroupSelector.byRoutingRulesEngine(file.getPath());
@@ -140,11 +134,6 @@ public class TestRoutingGroupSelector
         assertEquals("etl",
                 routingGroupSelector.findRoutingGroup(mockRequest));
 
-        Thread.sleep(1);
-        // java.nio.file.attribute.FileTime offers a maximum precision of 1 ms. If the first
-        // half of this test runs in <1ms, the gateway may not recognize that the file
-        // has changed.
-
         fw = new FileWriter(file, UTF_8);
         fw.write(
                 "---\n"
@@ -154,9 +143,7 @@ public class TestRoutingGroupSelector
                         + "actions:\n"
                         + "  - \"result.put(\\\"routingGroup\\\", \\\"etl2\\\")\""); // change from etl to etl2
         fw.close();
-        long etl2LastModified = file.lastModified();
-        log.info("etl2 rule file created at: " + etl2LastModified);
-        assertNotEquals(etl2LastModified, etlLastModified);
+        assertThat(file.setLastModified(lastModifed + 1000)).isTrue();
 
         when(mockRequest.getHeader(TRINO_SOURCE_HEADER)).thenReturn("airflow");
         assertEquals("etl2",


gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -140,21 +138,26 @@ public void testByRoutingRulesEngineFileChange()
         assertEquals("etl",
                 routingGroupSelector.findRoutingGroup(mockRequest));
 
-        Thread.sleep(1);
-        // java.nio.file.attribute.FileTime offers a maximum precision of 1 ms. If the first
-        // half of this test runs in <1ms, the gateway may not recognize that the file
-        // has changed.
+        long etl2LastModified = etlLastModified;
+        while (!(etl2LastModified > etlLastModified)) {
```
mosabua, line 142: So now you are writing the file again and again until the lastModified shows a difference .. hmm .. just waiting and writing it once does not work?

Other thought ... could we fake the date of the first write to just be a bit older or does the OS/Java not allow that cheating.. I guess not but maybe..

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -140,21 +138,26 @@ public void testByRoutingRulesEngineFileChange()
...
-                "---\n"
-                        + "name: \"airflow2\"\n"
-                        + "description: \"updated rule\"\n"
-                        + "condition: \"request.getHeader(\\\"X-Trino-Source\\\") == \\\"airflow\\\"\"\n"
-                        + "actions:\n"
-                        + "  - \"result.put(\\\"routingGroup\\\", \\\"etl2\\\")\""); // change from etl to etl2
-        fw.close();
-        long etl2LastModified = file.lastModified();
         log.info("etl2 rule file created at: " + etl2LastModified);
         assertNotEquals(etl2LastModified, etlLastModified);
```
mosabua, line 162: Also .. not that seems pointless .. we are basically forcing it to be different .. and then we asserting that it is .. at that stage the test might be useless altogether.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -111,7 +111,6 @@ void testByRoutingRulesEngineNoMatch(String rulesConfigPath)
         assertNull(routingGroupSelector.findRoutingGroup(mockRequest));
     }
 
-    //Todo: The functionality of reading the file before every request needs to be smarter
```
mosabua, line 114: Maybe we need to find a way to NOT read the file before every request and instead have some mechanism to reload it into memory when it gets changed .. or maybe the change triggers the reload

willmostly: I removed this comment because it isn't accurate. Tl;dr, a change does trigger a reload. A little history - 

The Presto Gateway reloaded the rules on every new request. Besides being inefficient, this was actually broken, because the `snakeyaml` parser is not thread safe. Routing would randomly fail as request concurrency increased.

I submitted a fix last year that added locking to prevent multiple threads from attempting to parse the rules file simultaneously. Making every request wait for a read lock could limit performance, so the timestamp check was added to check if the file itself was updated. If the file hasn't changed, the file doesn't need to be parsed and the current rules can be used. This is why the file timestamp needs to increment for this test to work.

We could add an API method to reload rules - but this would force every Gateway user to modify their operations, so I think the decision should be carefully considered. I also don't think this is a real world problem, CICD is the only place where you would modify the rules file < 1 ms after loading it.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
willmostly: The `sleep(1)` wasn't fully effective at preventing test failures. I think the failure frequency decreased, but didn't resolve them completely as issue #123 shows. We could just sleep longer, but [given this note](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/File.html#lastModified())
>While the unit of time of the return value is milliseconds, the granularity of the value depends on the underlying file system and may be larger. For example, some file systems use time stamps in units of seconds.

this approached seemed safer.

willmostly: Not sure where your comment went off to... but you suggested setting lastUpdatedTime automatically, which is possible through the `nio` API. So this logic is now removed. Thanks!

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
mosabua: Thanks for the background .. and agreed.

mosabua:
Once the conflict is resolved this should be good to go.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -123,6 +125,17 @@ public void testByRoutingRulesEngineFileChange()
                         + "actions:\n"
                         + "  - \"result.put(\\\"routingGroup\\\", \\\"etl\\\")\"");
         fw.close();
+        boolean successfullySetLastModifiedTime = file.setLastModified(0);
+        long etlLastModified = file.lastModified();
+        log.info("etl rule file created at: " + etlLastModified);
```
ebyhr, line 130: Avoid string concat in logger. Same for another usage.

ebyhr:
I think setting the last modified time https://github.com/trinodb/trino-gateway/pull/161 is better.

willmostly:
> I think setting the last modified time #161 is better.

this fix sets lastModifiedTime [here](https://github.com/trinodb/trino-gateway/pull/160/commits/3e8185ec463aff27787f41ef63410264e0a071d0#diff-973162668135a5cdba92480352158a4562d54c8815725d9fc0f852703202f4d8R128)

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -127,9 +125,17 @@ public void testByRoutingRulesEngineFileChange()
                         + "actions:\n"
                         + "  - \"result.put(\\\"routingGroup\\\", \\\"etl\\\")\"");
         fw.close();
-        file.getPath();
+        boolean successfullySetLastModifiedTime = file.setLastModified(0);
```
ebyhr, line 128: Setting 0 doesn't make sense. Please the last modified time + 1. 
Add `assertThat(file.setLastModified(0)).isTrue()`

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -127,9 +125,17 @@ public void testByRoutingRulesEngineFileChange()
...
         long etlLastModified = file.lastModified();
         log.info("etl rule file created at: " + etlLastModified);
+        if (!successfullySetLastModifiedTime) {
+            log.warn("File update time could not be set, sleep 10s");
+            Thread.sleep(10_000);
+            // RuleReloadingRoutingGroupSelector compares file timestamps to see if the rules file
+            // has been updated, and java.nio.file.attribute.FileTime offers a maximum precision of 1 ms.
+            // If the first half of this test runs in <1ms, and the lastModifiedTime cannot be set,
+            // then the gateway may not recognize that the file has changed.
+        }
```
ebyhr, line 138: Remove redundant block.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -156,7 +157,6 @@ public void testByRoutingRulesEngineFileChange()
         fw.close();
         long etl2LastModified = file.lastModified();
         log.info("etl2 rule file created at: " + etl2LastModified);
```
ebyhr, line 159: Remove unhelpful logger messages.

ebyhr (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -29,15 +27,14 @@
 import static io.trino.gateway.ha.router.RoutingGroupSelector.ROUTING_GROUP_HEADER;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
ebyhr, line 31: Please use AssertJ assertions.

willmostly: Sorry, I must have missed that we were migrating these

ebyhr (approved):

ebyhr:
Merged, thanks!

PR #163, opened 2024-01-17
introduced 'Pending' state for backend's health

andythsu:
This commit adds a 'Pending' state to keep track of backend's health.

Resolves: https://github.com/trinodb/trino-gateway/issues/80

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/BackendStatus.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/BackendStatus.java
@@ -13,23 +13,13 @@
  */
 package io.trino.gateway.ha.clustermonitor;
 
-import io.trino.gateway.ha.router.RoutingManager;
-
-public class HealthCheckObserver
-        implements TrinoClusterStatsObserver
-{
-    private final RoutingManager routingManager;
-
-    public HealthCheckObserver(RoutingManager routingManager)
-    {
-        this.routingManager = routingManager;
-    }
-
-    @Override
-    public void observe(java.util.List<ClusterStats> clustersStats)
-    {
-        for (ClusterStats clusterStats : clustersStats) {
-            routingManager.updateBackendHealth(clusterStats.getClusterId(), clusterStats.getBackendStatus());
-        }
-    }
+/**
+ * Healthy state is defined as backends are healthy and ready to be served
+ * Pending state is defined as when the backend is switched from inactive to active. It will wait until healthcheck returns success before turning backend to healthy
+ * Unhealthy state is defined as backends are returning error or not responding to healthcheck.
+ */
+public enum BackendStatus {
+  HEALTHY,
+  PENDING,
+  UNHEALTHY
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
@@ -24,7 +24,7 @@ public class ClusterStats
     private int queuedQueryCount;
     private int blockedQueryCount;
     private int numWorkerNodes;
-    private boolean healthy;
+    private BackendStatus backendStatus = BackendStatus.UNHEALTHY;
     private String clusterId;
     private String proxyTo;
     private String externalUrl;
@@ -73,14 +73,14 @@ public class ClusterStats
         this.numWorkerNodes = numWorkerNodes;
     }
 
-    public boolean isHealthy()
+    public BackendStatus getBackendStatus()
     {
-        return this.healthy;
+        return this.backendStatus;
     }
 
-    public void setHealthy(boolean healthy)
+    public void setBackendStatus(BackendStatus backendStatus)
     {
-        this.healthy = healthy;
+        this.backendStatus = backendStatus;
     }
 
     public String getClusterId()
@@ -147,7 +147,7 @@ public class ClusterStats
                 queuedQueryCount == stats.queuedQueryCount &&
                 blockedQueryCount == stats.blockedQueryCount &&
                 numWorkerNodes == stats.numWorkerNodes &&
-                healthy == stats.healthy &&
+                backendStatus == stats.backendStatus &&
                 Objects.equals(clusterId, stats.clusterId) &&
                 Objects.equals(proxyTo, stats.proxyTo) &&
                 Objects.equals(externalUrl, stats.externalUrl) &&
@@ -158,7 +158,7 @@ public class ClusterStats
     @Override
     public int hashCode()
     {
-        return Objects.hash(runningQueryCount, queuedQueryCount, blockedQueryCount, numWorkerNodes, healthy, clusterId, proxyTo, externalUrl, routingGroup, userQueuedCount);
+        return Objects.hash(runningQueryCount, queuedQueryCount, blockedQueryCount, numWorkerNodes, backendStatus, clusterId, proxyTo, externalUrl, routingGroup, userQueuedCount);
     }
 
     @Override
@@ -169,7 +169,7 @@ public class ClusterStats
                 .add("queuedQueryCount", queuedQueryCount)
                 .add("blockedQueryCount", blockedQueryCount)
                 .add("numWorkerNodes", numWorkerNodes)
-                .add("healthy", healthy)
+                .add("healthy", backendStatus)
                 .add("clusterId", clusterId)
                 .add("proxyTo", proxyTo)
                 .add("externalUrl", externalUrl)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
@@ -72,7 +72,7 @@ public class ClusterStatsHttpMonitor
             clusterStats.setQueuedQueryCount((int) result.get("queuedQueries"));
             clusterStats.setRunningQueryCount((int) result.get("runningQueries"));
             clusterStats.setBlockedQueryCount((int) result.get("blockedQueries"));
-            clusterStats.setHealthy(clusterStats.getNumWorkerNodes() > 0);
+            clusterStats.setBackendStatus(clusterStats.getNumWorkerNodes() > 0 ? BackendStatus.HEALTHY : BackendStatus.UNHEALTHY);
             clusterStats.setProxyTo(backend.getProxyTo());
             clusterStats.setExternalUrl(backend.getExternalUrl());
             clusterStats.setRoutingGroup(backend.getRoutingGroup());
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
@@ -47,6 +47,10 @@ public class ClusterStatsJdbcMonitor
             + "WHERE user != ? AND date_diff('hour',created,now()) <= 1 "
             + "GROUP BY state";
 
+    private static final String NODES_QUERY = "SELECT COUNT(*) as count "
+            + "FROM runtime.nodes "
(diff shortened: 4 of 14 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/BackendHealthState.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+public enum BackendHealthState {
```
ebyhr, line 16: Rename to BackendStatus.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -24,7 +24,7 @@ public class ClusterStats
     private int queuedQueryCount;
     private int blockedQueryCount;
     private int numWorkerNodes;
-    private boolean healthy;
+    private BackendHealthState healthy = BackendHealthState.UNHEALTHY;
```
ebyhr, line 27: Rename the variable to backendStatus.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -75,14 +75,29 @@ public void setNumWorkerNodes(int numWorkerNodes)
...
+        return this.healthy == BackendHealthState.HEALTHY;
+    }
+
+    public boolean isPending()
+    {
+        return this.healthy == BackendHealthState.PENDING;
     }
 
-    public void setHealthy(boolean healthy)
+    public boolean isUnhealthy()
```
ebyhr, line 86: Remove above isHealthy, isPending & isUnhealthy method.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -75,14 +75,29 @@ public void setNumWorkerNodes(int numWorkerNodes)
...
+    {
+        return this.healthy == BackendHealthState.UNHEALTHY;
+    }
+
+    public void setHealthy(BackendHealthState healthy)
     {
         this.healthy = healthy;
     }
 
+    public BackendHealthState getHealthy()
```
ebyhr, line 96: Move to before the setter.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
```diff
@@ -68,11 +68,11 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
 
-            clusterStats.setNumWorkerNodes((int) result.get("activeWorkers"));
-            clusterStats.setQueuedQueryCount((int) result.get("queuedQueries"));
-            clusterStats.setRunningQueryCount((int) result.get("runningQueries"));
-            clusterStats.setBlockedQueryCount((int) result.get("blockedQueries"));
-            clusterStats.setHealthy(clusterStats.getNumWorkerNodes() > 0);
+            clusterStats.setNumWorkerNodes((int) result.getOrDefault("activeWorkers", 0));
+            clusterStats.setQueuedQueryCount((int) result.getOrDefault("queuedQueries", 0));
+            clusterStats.setRunningQueryCount((int) result.getOrDefault("runningQueries", 0));
+            clusterStats.setBlockedQueryCount((int) result.getOrDefault("blockedQueries", 0));
```
ebyhr, line 74: Why these changes are required for introducing pending status?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
```diff
@@ -98,16 +98,16 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
             clusterStats.setRunningQueryCount(partialState.getOrDefault("RUNNING", 0));
             return clusterStats;
         }
         catch (TimeoutException e) {
-            log.error("timed out fetching status for {} backend", url, e);
+            log.error("timed out fetching status for {} backend", jdbcUrl, e);
         }
         catch (Exception e) {
-            log.error("could not fetch status for {} backend", url, e);
+            log.error("could not fetch status for {} backend", jdbcUrl, e);
```
ebyhr, line 110: Extract a commit. These changes are unrelated to pending status.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthCheckObserver.java
```diff
@@ -29,8 +29,7 @@ public HealthCheckObserver(RoutingManager routingManager)
     public void observe(java.util.List<ClusterStats> clustersStats)
     {
         for (ClusterStats clusterStats : clustersStats) {
-            routingManager.upateBackEndHealth(clusterStats.getClusterId(), clusterStats.isHealthy());
-            routingManager.updateBackEndHealthDB(clusterStats);
```
ebyhr, line 33: Why did you remove this line?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -66,7 +66,7 @@ public String load(String queryId)
                                     }
                                 });
 
-        this.backendToHealth = new ConcurrentHashMap<String, Boolean>();
+        this.backendToHealth = new ConcurrentHashMap();
```
ebyhr, line 69: Avoid raw usage of parameterized map.

andythsu: is it necessary to put `<String, BackendHealthState>` or just `<>()` is enough?

ebyhr: `<>()` is enough.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthCheckObserver.java
andythsu: this is the line that caused the problem. It directly manipulates the DB, which is unwanted. This method has been deleted in `routingManager`

willmostly (changes requested):
This is moving in the right direction, but as stated in the comments I think this change needs to include an update to the DB model as well.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/BackendStatus.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+public enum BackendStatus {
+  HEALTHY,
+  PENDING,
```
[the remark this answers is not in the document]
andythsu: sure, will add a javadoc

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthCheckObserver.java
Chaho12: > having multiple gateways could lead to inconsistent states for sure

In deploying using Helm Chart and for HA purpose for gateway server itself, people would use multiple gateways. Our team too are using HPA for the gateway itself, with 3 in minimum. 

If the status is local with serveral servers, then everytime I refresh the page I could get different Healthy status right?

I think a consistent sync is necessary, like 
- if status is inactive then the monitor could query and check if it is healthy
  - if the result is unhealthy, show unhealthy (meaning that it will do health check later) 
  - if the result is pending, show pending (similar to unhealthy but pending could also mean it is in checking state?)
  - if the result is healthy, then update DB so no other gateway server linked to same db would do unnecessary querying
- if status is active, then monitor does nothing

andythsu: by status do you mean backend status? I think you have active and inactive cases reversed..?

Chaho12: sry for confusion. status meant by active field in db

andythsu: Technically if you refresh the page you shouldn't get different results because if the pod is up and running, the healthcheck should be performed automatically. Therefore, you should get fairly consistent result given that the clusters are not flaky.

However, there will still be inconsistent states if the clusters are flaky and the gateway is not picking up yet. 

I see the value of having a central place to store this info to avoid inconsistent states. While this is great, I think it might lead to some kind of race condition, and could potentially cause requests to be routed to an unhealthy cluster. 

For example, if gateway A sees cluster A is bad and tries to mark it unhealthy, in the meantime gateway B looks at DB and thinks cluster A is healthy, and hence routes requests to cluster A. 

If this information is only available locally, then each gateway will manage its own clusters' states. Then there won't be any race condition.

Any thoughts?

Chaho12: > Technically if you refresh the page you shouldn't get different results because if the pod is up and running, the healthcheck should be performed automatically.  Therefore, you should get fairly consistent results given that the clusters are not flaky.

True, this does not happen that often. 

> For example, if gateway A sees cluster A is bad and tries to mark it unhealthy, in the meantime gateway B looks at DB and thinks cluster A is healthy, and hence routes requests to cluster A.

If the DB has value, i don't see any cases where gateway A, B sees cluster A active status differently.

Or maybe have a lock field? 
Assuming cluster A is inactive, and cluster B active.
- Gateway A scans if there is any inactive cluster. Finds cluster A, check for lock, if there isn't any lock, locks cluster A
  - Update (querying to see number of nodes) and releases lock afteward.
- Gateway B, also scans, and checks for lock. Cluster A is locked so skip.

andythsu: the current behavior is that if any cluster is **inactive** (meaning it's switched off manually), then the healthcheck will not be performed, since it doesn't make sense to perform healthcheck on **inactive** cluster.

If the cluster is **active**, then the healthcheck will perform. The cluster could be healthy, pending, or unhealthy. 

In your case, if Gateway A finds any inactive clusters, should skip them. It should only check on active clusters.

andythsu: The issue with storing healthcheck states in DB is that the healthcheck is done every minute and it's supposed to be a fast/small ping test. If we store this result in DB, we will need to look at DB everytime when healthcheck is performed, and hence it's adding more overhead. Also, when routingManager is choosing which cluster to route, it will need to look at DB to ensure the cluster is actually healthy. This could slow down the operation time.

vishalya:
On a related topic, the pain point is that each gateway instance runs the background job for collecting the stats and the status from the trino clusters. IMO just one instance should do run the background job looking at the timestamp when the last one was run.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthCheckObserver.java
willmostly: I'll start a Slack thread to discuss design options

andythsu:
Closing this PR for a new design: https://github.com/trinodb/trino-gateway/issues/222

PR #172, opened 2024-01-20
Remove redundant code

ebyhr:
These codes are unused.


mosabua:
Can you add in the PR or even better in the commit message what was redundant

ebyhr:
Updated PR description.

gateway-ha/src/main/java/io/trino/gateway/ha/module/ClusterStateListenerModule.java
```diff
@@ -53,7 +52,6 @@ public List<TrinoClusterStatsObserver> getClusterStatsObservers(
             BackendStateManager backendStateManager)
     {
         observers = new ArrayList<>();
-        NotifierConfiguration notifierConfiguration = getConfiguration().getNotifier();
```
mosabua, line 56: So notifierConfiugration is never access or used? And same for all the others?

ebyhr: NotifierConfiguration is still used from NotifierModule.

mosabua: Sure .. but no in this class .. hence the removal. And the same logic applies to other removals.. right?

mosabua (approved):
Lets ship it!

PR #176, opened 2024-01-20
Minor cleanup in TestTcpChecks

ebyhr:


gateway-ha/src/test/java/io/trino/gateway/ha/router/TestTcpChecks.java
```diff
@@ -36,13 +35,8 @@
 @TestInstance(PER_CLASS)
 public class TestTcpChecks
 {
-    public static final String TRINO_SOURCE_HEADER = "X-Trino-Source";
```
mosabua, line 39: So those were not even used?

mosabua (approved):
Looks good, and definitely cleaner than before.

PR #177, opened 2024-01-22
Fix the Clusters tab rendering

vishalya:
The `BackendState` went from being `@data` class to be a `record`


mosabua (approved):
Looks good. Thanks for testing and fixing @vishalya

PR #181, opened 2024-01-25
Update Maven plugins

mosabua:
Removing configs since they are managed by airbase anyway and we dont want to use these old versions. 

I confirmed that the new setup produces a identical JAR (apart from timestamp and git commit info and irrelevant empty lines in metadata files).


mosabua:
Any input @willmostly @vishalya @Chaho12

wendigo (approved):

Issue #165, opened 2024-01-17
OAuth handshake to Web UI fails with multiple backends


willmostly:
When connecting to a Trino webpage secured by OAuth from a browser, the OAuth handshake will fail if the Gateway forwards requests to different backends. This can be worked around by setting up a routing group with a single backend.

oneonestar:
Fixed by #188

PR #188, opened 2024-01-25
Add cookie based routing for selected paths

willmostly:
This adds an additional routing strategy for the Gateway. Currently the Gateway checks all incoming requests for a query ID. If one is found, the cluster running that query is looked up and the query is routed to that cluster. All other requests are sent to a semi random backend cluster in the routing group chosen by the `RoutingGroupSelector` based on the information present in the request. These routing strategies are not sufficient to solve issues such as https://github.com/trinodb/trino-gateway/issues/165. For https://github.com/trinodb/trino-gateway/issues/165, the Gateway needs the capability to route a request to the same backend that a previous request was sent to.

This adds the ability to add a session cookie, and to map that session cookie to a specific backend. This is lower precedence than the queryID based routing. Cookie based routing is performed only for paths in a configurable list. OAuth with multiple backends in a routing group (#165) can be solved by configuring `cookiePaths: ['/oauth2/callback']`. Finally, a cookie will be deleted if a request is made to a patch in the `logOutCookiePaths` list.

### DIFF
--- a/docs/routing-logic.md
+++ b/docs/routing-logic.md
@@ -1,0 +1,111 @@
+**Trino Gateway documentation**
+
+<table>
+  <tr>
+    <td>
+      <img src="./assets/logos/trino-gateway-v.png"/>
+    </td>
+    <td>
+      <ul>
+        <li><a href="quickstart.md">Quickstart</a></li>
+        <li><a href="installation.md">Installation</a></li>
+        <li><a href="security.md">Security</a></li>
+        <li><a href="operation.md">Operation</a></li>
+      </ul>
+    </td>
+    <td>
+      <ul>
+        <li><a href="gateway-api.md">Gateway API</a></li>
+        <li><a href="resource-groups-api.md">Resource groups API</a></li>
+        <li><a href="routing-rules.md">Routing rules</a></li>
+        <li><a href="routing-logic.md">Routing logic</a></li>
+      </ul>
+    </td>
+    <td>
+      <ul>
+        <li><a href="design.md">Design</a></li>
+        <li><a href="development.md">Development</a></li>
+        <li><a href="release-notes.md">Release notes</a></li>
+        <li><a href="references.md">References</a></li>
+      </ul>
+    </td>
+  </tr>
+</table>
+
+# Routing Logic
+
+## Overview
+
+Trino Gateway checks incoming requests to see if they're related to previous 
+ones it handled. If they are, then Trino Gateway sends them to the same backend 
+that dealt with the earlier requests.
+
+If it is a new request, the Trino Gateway refers to [Routing rules](routing-rules.md) 
+to decide which group of backends, called a 'Routing Group,' should handle it. 
+It then picks a backend from that Routing Group to handle the request using 
+either an adaptive or round-robin strategy.
+
+![Request Routing Flow](assets/gateway-routing-flow.svg)
+
+## Sticky routing
+
+A request related to an ongoing process, or to state maintained on a single 
+backend cluster, must be routed to that backend for proper handling. Two 
+mechanisms for identifying related requests are currently implemented. By default,
+only routing based on query identifier is enabled.
+
+### Routing based on query identifier (default)
+
+When a query is initiated through the Trino Gateway, the query id will be 
+extracted from the response and mapped to the backend that provided the 
+response. Any subsequent request containing that query id will be forwarded 
+to that backend. For example, to retrieve query results, the trino client 
+polls a URI of the form 
+`v1/statement/executing/queryid/nonce/counter`. The Trino Gateway will extract
+the queryid from this URI.
+
+### Routing based on cookies
+
+OAuth2 authentication requires that the same backend is used for each step of 
+the handshake. When `gatewayCookieConfiguration.enabled` is set to true, a cookie 
+will be added to requests made to paths beginning with `/oauth2` unless they already have 
+a cookie present, which is used to route further `/oauth2/*` requests to the correct backend. 
+Cookies are not added to requests to `v1/*` and other Trino endpoints.
+
+Trino Gateway signs its cookies to ensure that they are not tampered with. You 
+must set a `cookieSigningSecret` string in your configuration
+```yaml
+gatewayCookieConfiguration:
+    enabled: true
+    cookieSigningSecret: "ahighentropystring"
+```
+when making use of this feature. If you load balance request across multiple Trino Gateway
+instances, ensure each instance has the same `cookieSigningSecret`.
+
+The Trino Gateway handles standard Trino OAuth2 handshakes with no additional 
+configuration. If you are using a customized or commercial Trino distribution, then
+the paths used to define the OAuth handshake may be modified.
+
+`routingPaths`: If the request URI starts with a path in this list, then
+* If no cookie is present, add a routing cookie
+* If a cookie is present, route the request to the backend defined by that cookie
+
+`deletePaths`: If the request URI starts with a path in this list,
+return a response that instructs the client to delete the cookie.
+
+Additionally, the `lifetime` property sets the duration for which a cookie remains in 
+effect after creation. Ensure that it is greater than
+the time required to complete the handshake. Default `lifetime` is 10 minutes.
+
+These properties are defined under the `oauth2GatewayCookieConfiguration` node: 
+
+```yaml
+oauth2GatewayCookieConfiguration:
+  routingPaths:
+    - "/oauth2"
+    - "/custom/oauth2/callback"
+    - "/alternative/oauth2/initiate"
+  deletePaths:
+    - "/custom/logout"
+  lifetime: "5m"
+```
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/GatewayCookieConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/GatewayCookieConfiguration.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.config;
+
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class GatewayCookieConfiguration
+{
+    private SecretKey cookieSigningKey;
+    private boolean enabled;
+
+    public boolean isEnabled()
+    {
+        return enabled;
+    }
+
+    public void setEnabled(boolean enabled)
+    {
+        this.enabled = enabled;
+    }
+
+    public SecretKey getCookieSigningKey()
+    {
+        return cookieSigningKey;
+    }
+
+    public void setCookieSigningSecret(String cookieSigningSecret)
+    {
+        cookieSigningKey = new SecretKeySpec(cookieSigningSecret.getBytes(UTF_8), "HmacSHA256");
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieConfiguration.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.config;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.units.Duration;
+
+import java.util.List;
+
+public class OAuth2GatewayCookieConfiguration
+{
+    // Configuration initialization using dropwizard requires
+    // instance method setters. Values are global, and can be accessed using static getters
+    private List<String> routingPaths = ImmutableList.of("/oauth2");
+    private List<String> deletePaths = ImmutableList.of("/logout", "/oauth2/logout");
+    private Duration lifetime = Duration.valueOf("10m");
+
+    public List<String> getDeletePaths()
+    {
+        return deletePaths;
+    }
+
+    public void setDeletePaths(List<String> deletePaths)
+    {
+        this.deletePaths = deletePaths;
+    }
+
+    public List<String> getRoutingPaths()
+    {
+        return routingPaths;
+    }
+
+    public void setRoutingPaths(List<String> routingPaths)
+    {
+        this.routingPaths = routingPaths;
+    }
+
+    public Duration getLifetime()
+    {
+        return lifetime;
+    }
+
+    public void setLifetime(String lifetime)
+    {
+        this.lifetime = Duration.valueOf(lifetime);
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
@@ -15,13 +15,18 @@ package io.trino.gateway.ha.handler;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
 import io.airlift.log.Logger;
+import io.trino.gateway.ha.config.GatewayCookieConfigurationPropertiesProvider;
+import io.trino.gateway.ha.router.GatewayCookie;
+import io.trino.gateway.ha.router.OAuth2GatewayCookie;
 import io.trino.gateway.ha.router.QueryHistoryManager;
 import io.trino.gateway.ha.router.RoutingGroupSelector;
 import io.trino.gateway.ha.router.RoutingManager;
 import io.trino.gateway.proxyserver.ProxyHandler;
 import io.trino.gateway.proxyserver.wrapper.MultiReadHttpServletRequest;
+import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import jakarta.ws.rs.HttpMethod;
@@ -32,9 +37,11 @@ import org.eclipse.jetty.util.Callback;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -84,6 +91,7 @@ public class QueryIdCachingProxyHandler
     private final ProxyHandlerStats proxyHandlerStats;
     private final List<String> extraWhitelistPaths;
     private final String applicationEndpoint;
+    private final boolean cookiesEnabled;
 
     public QueryIdCachingProxyHandler(
             QueryHistoryManager queryHistoryManager,
@@ -99,6 +107,7 @@ public class QueryIdCachingProxyHandler
         this.queryHistoryManager = queryHistoryManager;
         this.extraWhitelistPaths = extraWhitelistPaths;
         this.applicationEndpoint = "http://localhost:" + serverApplicationPort;
+        cookiesEnabled = GatewayCookieConfigurationPropertiesProvider.getInstance().isEnabled();
     }
 
     protected static String extractQueryIdIfPresent(String path, String queryParams)
@@ -269,9 +278,23 @@ public class QueryIdCachingProxyHandler
                 || extraWhitelistPaths.stream().anyMatch(s -> path.startsWith(s));
     }
 
-    public boolean handleAuthRequest(HttpServletRequest request)
+    @Override
+    public List<Cookie> generateDeleteCookieList(HttpServletRequest clientRequest)
     {
-        return true;
+        if (!cookiesEnabled || clientRequest.getCookies() == null) {
+            return ImmutableList.of();
+        }
+
+        return Arrays.stream(clientRequest.getCookies())
+                .filter(c -> c.getName().startsWith(GatewayCookie.PREFIX))
+                .map(GatewayCookie::fromCookie)
+                .filter(c -> !c.isValid() || c.matchesDeletePath(clientRequest.getRequestURI()))
+                .map(GatewayCookie::toCookie)
+                .peek(c -> {
+                    c.setValue("delete");
+                    c.setMaxAge(0);
+                })
+                .toList();
     }
 
     @Override
@@ -281,10 +304,10 @@ public class QueryIdCachingProxyHandler
             return buildUriWithNewBackend(applicationEndpoint, request);
         }
 
-        if (isUseStickyRouting(request)) {
-            String backend = getPreviousBackend(request);
-            logRewrite(backend, request);
-            return buildUriWithNewBackend(backend, request);
+        Optional<String> previousBackend = getPreviousBackend(request);
+        if (previousBackend.isPresent()) {
+            logRewrite(previousBackend.orElseThrow(), request);
+            return previousBackend.map(b -> buildUriWithNewBackend(b, request)).orElseThrow();
         }
 
         String backend = getBackendFromRoutingGroup(request);
@@ -322,21 +345,27 @@ public class QueryIdCachingProxyHandler
         return routingManager.provideAdhocBackend(user);
     }
 
-    private String getPreviousBackend(HttpServletRequest request)
+    private Optional<String> getPreviousBackend(HttpServletRequest request)
     {
         String queryId = extractQueryIdIfPresent(request);
-
-        // Find query id and get url from cache
         if (!isNullOrEmpty(queryId)) {
-            return routingManager.findBackendForQueryId(queryId);
+            return Optional.of(routingManager.findBackendForQueryId(queryId));
         }
-        log.error("No backend found for queryId %s", queryId);
-        return getBackendFromRoutingGroup(request);
-    }
+        if (cookiesEnabled && request.getCookies() != null) {
+            List<GatewayCookie> cookies = Arrays.stream(request.getCookies())
+                    .filter(c -> c.getName().startsWith(GatewayCookie.PREFIX))
+                    .map(GatewayCookie::fromCookie)
+                    .filter(GatewayCookie::isValid)
+                    .filter(c -> !isNullOrEmpty(c.getBackend()))
+                    .filter(c -> c.matchesRoutingPath(request.getRequestURI()))
+                    .sorted()
+                    .toList();
+            if (!cookies.isEmpty()) {
+                return Optional.of(cookies.getFirst().getBackend());
+            }
+        }
 
-    private boolean isUseStickyRouting(HttpServletRequest request)
-    {
-        return !isNullOrEmpty(extractQueryIdIfPresent(request));
+        return Optional.empty();
     }
 
     @Override
@@ -349,56 +378,56 @@ public class QueryIdCachingProxyHandler
             Callback callback)
     {
         try {
-            String requestPath = request.getRequestURI();
-            if (requestPath.startsWith(V1_STATEMENT_PATH)
-                    && request.getMethod().equals(HttpMethod.POST)) {
-                String output;
-                boolean isGZipEncoding = isGZipEncoding(response);
-                if (isGZipEncoding) {
-                    output = plainTextFromGz(buffer);
-                }
-                else {
-                    output = new String(buffer);
-                }
-                log.debug("For Request [%s] got Response output [%s]", request.getRequestURI(), output);
-
-                QueryHistoryManager.QueryDetail queryDetail = getQueryDetailsFromRequest(request);
-                log.debug("Extracting Proxy destination : [%s] for request : [%s]",
-                        queryDetail.getBackendUrl(), request.getRequestURI());
-
-                if (response.getStatus() == HttpStatus.OK_200) {
-                    HashMap<String, String> results = OBJECT_MAPPER.readValue(output, HashMap.class);
-                    queryDetail.setQueryId(results.get("id"));
-
-                    if (!isNullOrEmpty(queryDetail.getQueryId())) {
-                        routingManager.setBackendForQueryId(
-                                queryDetail.getQueryId(), queryDetail.getBackendUrl());
-                        log.debug(
-                                "QueryId [%s] mapped with proxy [%s]",
-                                queryDetail.getQueryId(),
-                                queryDetail.getBackendUrl());
-                    }
-                    else {
-                        log.debug("QueryId [%s] could not be cached", queryDetail.getQueryId());
-                    }
-                }
-                else {
-                    log.error(
-                            "Non OK HTTP Status code with response [%s] , Status code [%s]",
-                            output,
-                            response.getStatus());
-                }
-                // Saving history at gateway.
-                queryHistoryManager.submitQueryDetail(queryDetail);
+            if (request.getRequestURI().startsWith(V1_STATEMENT_PATH) && request.getMethod().equals(HttpMethod.POST)) {
+                recordBackendForQueryId(request, response, buffer);
             }
-            else {
-                log.debug("SKIPPING For %s", requestPath);
+            else if (cookiesEnabled && request.getRequestURI().startsWith(OAuth2GatewayCookie.OAUTH2_PATH)
+                    && !(request.getCookies() != null
+                    && Arrays.stream(request.getCookies()).anyMatch(c -> c.getName().equals(OAuth2GatewayCookie.NAME)))) {
+                GatewayCookie oauth2Cookie = new OAuth2GatewayCookie(request.getHeader(PROXY_TARGET_HEADER));
+                response.addCookie(oauth2Cookie.toCookie());
             }
         }
         catch (Exception e) {
             log.error(e, "Error in proxying falling back to super call");
         }
         super.postConnectionHook(request, response, buffer, offset, length, callback);
+    }
+
+    void recordBackendForQueryId(HttpServletRequest request, HttpServletResponse response, byte[] buffer)
+            throws IOException
+    {
+        String output;
+        boolean isGZipEncoding = isGZipEncoding(response);
+        if (isGZipEncoding) {
+            output = plainTextFromGz(buffer);
+        }
+        else {
+            output = new String(buffer);
+        }
+        log.debug("For Request [%s] got Response output [%s]", request.getRequestURI(), output);
+
+        QueryHistoryManager.QueryDetail queryDetail = getQueryDetailsFromRequest(request);
+
+        if (queryDetail.getBackendUrl() == null) {
+            log.error("Server response to request %s does not contain proxytarget header", request.getRequestURI());
+        }
+        log.debug("Extracting Proxy destination : [%s] for request : [%s]", queryDetail.getBackendUrl(), request.getRequestURI());
+
+        if (response.getStatus() == HttpStatus.OK_200) {
+            HashMap<String, String> results = OBJECT_MAPPER.readValue(output, HashMap.class);
+            queryDetail.setQueryId(results.get("id"));
+
+            if (!isNullOrEmpty(queryDetail.getQueryId())) {
+                routingManager.setBackendForQueryId(queryDetail.getQueryId(), queryDetail.getBackendUrl());
+                log.debug("QueryId [%s] mapped with proxy [%s]", queryDetail.getQueryId(), queryDetail.getBackendUrl());
+            }
+        }
+        else {
+            log.error("Non OK HTTP Status code with response [%s] , Status code [%s]", output, response.getStatus());
+        }
+        // Save history in Trino Gateway.
+        queryHistoryManager.submitQueryDetail(queryDetail);
     }
 
     private QueryHistoryManager.QueryDetail getQueryDetailsFromRequest(HttpServletRequest request)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
@@ -1,0 +1,289 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import com.google.common.hash.Hashing;
+import io.airlift.json.JsonCodec;
+import io.airlift.log.Logger;
+import io.airlift.units.Duration;
+import io.trino.gateway.ha.config.GatewayCookieConfigurationPropertiesProvider;
+import jakarta.servlet.http.Cookie;
+
+import java.util.Base64;
+import java.util.List;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+
+@JsonPropertyOrder(alphabetic = true)
+public class GatewayCookie
+        implements Comparable<GatewayCookie>
+{
+    private static final Logger log = Logger.get(GatewayCookie.class);
+    private String signature;
+    private final UnsignedGatewayCookie unsignedGatewayCookie;
+    private final GatewayCookieConfigurationPropertiesProvider gatewayCookieConfigurationPropertiesProvider = GatewayCookieConfigurationPropertiesProvider.getInstance();
+
+    public static final String PREFIX = "TG.";
+
+    public static final JsonCodec<GatewayCookie> CODEC = JsonCodec.jsonCodec(GatewayCookie.class);
+
+    @JsonCreator
+    public GatewayCookie(
+            @JsonProperty("ts") Long ts,
+            @JsonProperty("name") String name,
+            @JsonProperty("payload") String payload,
+            @JsonProperty("routingPaths") List<String> routingPaths,
+            @JsonProperty("deletePaths") List<String> deletePaths,
+            @JsonProperty("backend") String backend,
+            @JsonProperty("priority") Integer priority,
+            @JsonProperty("ttl") Duration ttl,
+            @JsonProperty("signature") String signature)
+    {
+        this.unsignedGatewayCookie = new UnsignedGatewayCookie(
+                requireNonNull(ts),
+                requireNonNull(name),
+                payload,
+                backend,
+                requireNonNull(routingPaths),
+                requireNonNull(deletePaths),
+                requireNonNull(ttl),
+                priority);
+        this.signature = signature;
+    }
+
+    public GatewayCookie(String name, String payload, String backend, List<String> routingPaths, List<String> deletePaths, Duration ttl, int priority)
+    {
+        this.unsignedGatewayCookie = new UnsignedGatewayCookie(
+                System.currentTimeMillis(),
+                requireNonNull(name.startsWith(PREFIX) ? name : PREFIX + name),
+                payload,
+                backend,
+                requireNonNull(routingPaths),
+                requireNonNull(deletePaths),
+                requireNonNull(ttl),
+                priority);
+        signature = computeSignature();
+    }
+
+    @JsonProperty
+    public Long getTs()
+    {
+        return unsignedGatewayCookie.getTs();
+    }
+
+    @JsonProperty
+    public String getName()
+    {
+        return unsignedGatewayCookie.getName();
+    }
+
+    @JsonProperty
+    public String getPayload()
+    {
+        return unsignedGatewayCookie.getPayload();
+    }
+
+    @JsonProperty
+    public String getBackend()
+    {
+        return unsignedGatewayCookie.getBackend();
+    }
+
+    @JsonProperty
+    public int getPriority()
+    {
+        return unsignedGatewayCookie.getPriority();
+    }
+
+    @JsonProperty
+    public List<String> getRoutingPaths()
+    {
+        return unsignedGatewayCookie.getRoutingPaths();
+    }
+
+    @JsonProperty
+    public List<String> getDeletePaths()
+    {
+        return unsignedGatewayCookie.getDeletePaths();
+    }
+
+    @JsonProperty
+    public Duration getTtl()
+    {
+        return unsignedGatewayCookie.getTtl();
+    }
+
+    @JsonProperty
+    public String getSignature()
+    {
+        return signature;
+    }
+
+    public void setTs(Long ts)
+    {
+        unsignedGatewayCookie.setTs(ts);
+    }
+
+    private String computeSignature()
+    {
+        return Hashing.hmacSha256(gatewayCookieConfigurationPropertiesProvider.getCookieSigningKey())
+                .hashString(UnsignedGatewayCookie.CODEC.toJson(unsignedGatewayCookie), UTF_8)
+                .toString();
+    }
+
+    @Override
+    public int compareTo(GatewayCookie o)
+    {
+        int priorityDelta = unsignedGatewayCookie.getPriority() - o.getPriority();
+        return priorityDelta != 0 ? priorityDelta : (int) (unsignedGatewayCookie.getTs() - o.getTs());
+    }
+
+    public Cookie toCookie()
+    {
+        Cookie cookie = new Cookie(unsignedGatewayCookie.getName(), Base64.getUrlEncoder().encodeToString(CODEC.toJson(this).getBytes(UTF_8)));
+        cookie.setMaxAge((int) unsignedGatewayCookie.getTtl().toMillis() / 1000);
+        return cookie;
+    }
+
+    public static GatewayCookie fromCookie(Cookie cookie)
+    {
+        return GatewayCookie.CODEC.fromJson(Base64.getUrlDecoder().decode(cookie.getValue()));
+    }
+
+    public boolean matchesRoutingPath(String path)
+    {
+        if (matchesDeletePath(path)) {
+            return false;
+        }
+
+        return unsignedGatewayCookie.getRoutingPaths().stream().anyMatch(path::startsWith);
+    }
+
+    public boolean matchesDeletePath(String path)
+    {
+        return unsignedGatewayCookie.getDeletePaths().contains(path);
+    }
+
+    public boolean isValid()
+    {
+        if (System.currentTimeMillis() > unsignedGatewayCookie.getTs() + unsignedGatewayCookie.getTtl().toMillis()) {
+            return false;
+        }
+
+        if (isNullOrEmpty(signature) || !signature.equals(computeSignature())) {
+            log.error("Invalid cookie: %s", CODEC.toJson(this));
+            throw new IllegalArgumentException("Invalid cookie signature");
+        }
+
+        return true;
+    }
+
+    @JsonPropertyOrder(alphabetic = true)
+    public static class UnsignedGatewayCookie
+    {
+        public static final JsonCodec<UnsignedGatewayCookie> CODEC = JsonCodec.jsonCodec(UnsignedGatewayCookie.class);
+        private Long ts; // timestamp. The shortened name saves 8 bytes of cookie size
+        private final String name;
+        private final String payload;
+        private final List<String> routingPaths;
+
+        private final List<String> deletePaths;
+
+        private final int priority;
+        private final Duration ttl;
+        private final String backend;
+
+        public UnsignedGatewayCookie(GatewayCookie gatewayCookie)
+        {
+            this.ts = gatewayCookie.getTs();
+            this.name = gatewayCookie.getName();
+            this.payload = gatewayCookie.getPayload();
+            this.routingPaths = gatewayCookie.getRoutingPaths();
+            this.deletePaths = gatewayCookie.getDeletePaths();
+            this.priority = gatewayCookie.getPriority();
+            this.ttl = gatewayCookie.getTtl();
+            this.backend = gatewayCookie.getBackend();
+        }
+
+        public UnsignedGatewayCookie(Long ts, String name, String payload, String backend, List<String> routingPaths, List<String> deletePaths, Duration ttl, int priority)
+        {
+            this.name = name.startsWith(PREFIX) ? name : PREFIX + name;
+            this.payload = payload;
+            this.backend = backend;
+            this.routingPaths = routingPaths;
+            this.deletePaths = deletePaths;
+            this.ttl = ttl;
+            this.ts = ts;
+            this.priority = priority;
+        }
+
+        @JsonProperty
+        public Long getTs()
+        {
+            return ts;
+        }
+
+        public void setTs(Long ts)
+        {
+            this.ts = ts;
+        }
+
+        @JsonProperty
+        public String getName()
+        {
+            return name;
+        }
+
+        @JsonProperty
+        public String getPayload()
+        {
+            return payload;
+        }
+
+        @JsonProperty
+        public String getBackend()
+        {
+            return backend;
+        }
+
+        @JsonProperty
+        public int getPriority()
+        {
+            return priority;
+        }
+
+        @JsonProperty
+        public List<String> getRoutingPaths()
+        {
+            return routingPaths;
+        }
+
+        @JsonProperty
+        public List<String> getDeletePaths()
+        {
+            return deletePaths;
+        }
+
+        @JsonProperty
+        public Duration getTtl()
+        {
+            return ttl;
+        }
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyHandler.java
@@ -13,7 +13,9 @@
  */
 package io.trino.gateway.proxyserver;
 
+import com.google.common.collect.ImmutableList;
 import io.airlift.log.Logger;
+import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import org.eclipse.jetty.client.api.Request;
@@ -27,6 +29,7 @@ import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.List;
 import java.util.zip.GZIPInputStream;
 
 /* Order of control => rewriteTarget, preConnectionHook, postConnectionHook. */
@@ -113,5 +116,10 @@ public class ProxyHandler
     {
         return (compressed[0] == (byte) GZIPInputStream.GZIP_MAGIC)
                 && (compressed[1] == (byte) (GZIPInputStream.GZIP_MAGIC >> 8));
+    }
+
+    public List<Cookie> generateDeleteCookieList(HttpServletRequest clientRequest)
+    {
+        return ImmutableList.of();
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyServletImpl.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyServletImpl.java
@@ -98,6 +98,16 @@ public class ProxyServletImpl
         return target;
     }
 
+    @Override
+    protected void onServerResponseHeaders(
+            HttpServletRequest clientRequest,
+            HttpServletResponse proxyResponse,
+            Response serverResponse)
+    {
+        this.proxyHandler.generateDeleteCookieList(clientRequest).forEach(proxyResponse::addCookie);
+        super.onServerResponseHeaders(clientRequest, proxyResponse, serverResponse);
+    }
+
     /**
      * Customize the response returned from remote server.
      */
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
@@ -14,20 +14,29 @@
 package io.trino.gateway.ha;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableMap;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import io.trino.gateway.ha.router.GatewayCookie;
+import io.trino.gateway.ha.router.OAuth2GatewayCookie;
+import okhttp3.Cookie;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.mockwebserver.MockWebServer;
-import okhttp3.mockwebserver.RecordedRequest;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.testcontainers.containers.TrinoContainer;
+
+import java.io.IOException;
+import java.util.Base64;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -39,6 +48,11 @@ public class TestGatewayHaMultipleBackend
 
     private TrinoContainer adhocTrino;
     private TrinoContainer scheduledTrino;
+
+    public static String oauthInitiatePath = OAuth2GatewayCookie.OAUTH2_PATH;
+    public static String oauthCallbackPath = oauthInitiatePath + "/callback";
+    public static String oauthInitialResponse = "abc";
+    public static String oauthCallbackResponse = "xyz";
 
     final int routerPort = 20000 + (int) (Math.random() * 1000);
     final int customBackendPort = 21000 + (int) (Math.random() * 1000);
@@ -59,7 +73,11 @@ public class TestGatewayHaMultipleBackend
         int backend1Port = adhocTrino.getMappedPort(8080);
         int backend2Port = scheduledTrino.getMappedPort(8080);
 
-        HaGatewayTestUtils.prepareMockBackend(customBackend, customBackendPort, CUSTOM_RESPONSE);
+        HaGatewayTestUtils.prepareMockBackend(customBackend, customBackendPort, "default custom response");
+        HaGatewayTestUtils.setPathSpecificResponses(customBackend, ImmutableMap.of(
+                oauthInitiatePath, oauthInitialResponse,
+                oauthCallbackPath, oauthCallbackResponse,
+                CUSTOM_PATH, CUSTOM_RESPONSE));
 
         // seed database
         HaGatewayTestUtils.TestConfig testConfig =
@@ -93,9 +111,6 @@ public class TestGatewayHaMultipleBackend
                         .build();
         Response response1 = httpClient.newCall(request1).execute();
         assertThat(response1.body().string()).isEqualTo(CUSTOM_RESPONSE);
-        RecordedRequest recordedRequest = customBackend.takeRequest();
-        assertThat(recordedRequest.getMethod()).isEqualTo("POST");
-        assertThat(recordedRequest.getPath()).isEqualTo(CUSTOM_PATH);
 
         Request request2 =
                 new Request.Builder()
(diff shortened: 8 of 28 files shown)


mosabua:
I have not reviewed the code yet but the description of this PR should be improved and added to the docs. 

Also we should consider how this relates to the work from @vishalya in terms of pluggable strategies.

Imho we should have a default ordered list of strategies that are used out of the box, but it should be possible to activate and deactivate any strategy and also potentially change the order.

vishalya:
Does the rule engine take a lower precedence for a cookie path?

Chaho12:
Please rebase after https://github.com/trinodb/trino-gateway/pull/265 gets merged.
This PR is also important as it fixes bug to handle multiple backends in same group.

willmostly:
> Does the rule engine take a lower precedence for a cookie path?

Yes. Similar to a queryId, the presence of a cookie along with a path in `cookiePaths` signifies that a request is related to a previous one, so you don't want to send the request to any backend besides the one the previous requests were sent to. As noted in the TODO comments, I'd like to support more sophisticated decision making on whether or not sticky routing is used, but for now looking at the `path` is enough to support the oauth handshake and potentially other use cases.

Hopefully the relationships between the different routing strategies are more clear now that this is rebased and includes the refactoring I did of of `QueryIdCachingProxyHandler`.

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+        <li><a href="release-notes.md">Release notes</a></li>
+        <li><a href="references.md">References</a></li>
+      </ul>
+    </td>
+  </tr>
+</table>
+
+# Routing Logic
+
+## Overview
```
mosabua, line 37: remove redundant title or add intro sentence above

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+Incoming requests are evaluated by the Gateway to determine if they
```
mosabua, line 38:
```suggestion
Incoming requests are evaluated by Trino Gateway to determine if they
```

mosabua: and everywhere else .. full name

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+are a follow-up to a request previously handled by the Gateway. If the 
+Gateway determines that they are, they are forwarded to whatever the
```
mosabua, line 40:
```suggestion
Gateway determines that a request is a follow up to a previous request, it is forwarded to the same backend as before. 
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Routing Logic
+
+## Overview
+Incoming requests are evaluated by the Gateway to determine if they
+are a follow-up to a request previously handled by the Gateway. If the 
+Gateway determines that they are, they are forwarded to whatever the 
+backend that handled the previous request.
+
+If the request does not have an antecedent, the Gateway first uses [Routing Rules](routing-rules.md)
```
mosabua, line 43:
```suggestion
If the request is new, Trino Gateway first uses [Routing rules](routing-rules.md)
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+to determine which logical group of clusters, termed a "Routing Group",
+should handle the request. A member of the Routing Group, termed a backend, 
+is selected to handle the request using either an adaptive or round-robin 
+strategy.
+
+![Request Routing Flow](assets/gateway-routing-flow.svg)
+
+## Sticky Routing
+A request related to an ongoing process or state maintained on a single
```
mosabua, line 52: empty line after title here and elsewhere

mosabua, line 52:
```suggestion
A request related to an ongoing process or state, maintained on a single 
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+to determine which logical group of clusters, termed a "Routing Group",
+should handle the request. A member of the Routing Group, termed a backend, 
+is selected to handle the request using either an adaptive or round-robin 
+strategy.
+
+![Request Routing Flow](assets/gateway-routing-flow.svg)
+
+## Sticky Routing
+A request related to an ongoing process or state maintained on a single 
+backend cluster must be routed to that backend for proper handling. Two
```
mosabua, line 53:
```suggestion
backend cluster, must be routed to that backend for proper handling. Two 
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+should handle the request. A member of the Routing Group, termed a backend, 
+is selected to handle the request using either an adaptive or round-robin 
+strategy.
+
+![Request Routing Flow](assets/gateway-routing-flow.svg)
+
+## Sticky Routing
+A request related to an ongoing process or state maintained on a single 
+backend cluster must be routed to that backend for proper handling. Two 
+mechanisms for identifying related requests are currently implemented.
```
mosabua, line 54:
```suggestion
mechanisms for identifying related requests are implemented and detailed in the following sections.
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Query ID Based Routing
```
mosabua, line 56: Routing based on query identifier

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+When a query is initiated through the Gateway, the query id will be 
+extracted from the response and mapped to the backend that provided the 
+response. Any subsequent request containing that query id will be forwarded 
+to that backend. For example, to retrieve query results, the trino client 
+polls a URI of the form 
+`v1/statement/executing/queryid/nonce/counter`. The Gateway will extract
+the queryid from this URI.
+
+### Cookie Based Routing
```
mosabua, line 65: Routing based on cookies

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+The Gateway can add a session cookie to identify related requests. This is 
+useful for exchanges such as the OAuth handshake, where it isn't possible 
+to determine which backend handled the previous request based solely on
+the request URI and body.
+
+#### Cookie Based Routing Configuration
```
mosabua, line 71: Drop that title and replace with sentence about configuring .. adding where to configure this .. .. and say .. with the following properties

maybe have an example snippet

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+When a query is initiated through the Gateway, the query id will be 
+extracted from the response and mapped to the backend that provided the 
+response. Any subsequent request containing that query id will be forwarded 
+to that backend. For example, to retrieve query results, the trino client 
+polls a URI of the form 
+`v1/statement/executing/queryid/nonce/counter`. The Gateway will extract
+the queryid from this URI.
+
+### Cookie Based Routing
+The Gateway can add a session cookie to identify related requests. This is
```
mosabua, line 66:
```suggestion
Trino Gateway can add a session cookie to new request and use it to identify related requests. This behavior is 
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+extracted from the response and mapped to the backend that provided the 
+response. Any subsequent request containing that query id will be forwarded 
+to that backend. For example, to retrieve query results, the trino client 
+polls a URI of the form 
+`v1/statement/executing/queryid/nonce/counter`. The Gateway will extract
+the queryid from this URI.
+
+### Cookie Based Routing
+The Gateway can add a session cookie to identify related requests. This is 
+useful for exchanges such as the OAuth handshake, where it isn't possible
```
mosabua, line 67: Change this to talk about it being necessary for Trino cluster that use OAuth authentication ... so that it hits the same cluster based on the limited info

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+the queryid from this URI.
+
+### Cookie Based Routing
+The Gateway can add a session cookie to identify related requests. This is 
+useful for exchanges such as the OAuth handshake, where it isn't possible 
+to determine which backend handled the previous request based solely on
+the request URI and body.
+
+#### Cookie Based Routing Configuration
+`cookiePaths`: If the request URI starts with a path in this list, add a
```
mosabua, line 72: list? comma separated? add example

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+useful for exchanges such as the OAuth handshake, where it isn't possible 
+to determine which backend handled the previous request based solely on
+the request URI and body.
+
+#### Cookie Based Routing Configuration
+`cookiePaths`: If the request URI starts with a path in this list, add a 
+session cookie if none exists. If a cookie exists, route the request to
+the backend associated with that cookie.
+
+`removeCookiePaths`: If the request URI starts with a path in this list,
```
mosabua, line 76: why do you need that? In what situation should a user configure that .. to what value

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+return a response that instructs the client to delete the cookie.
+
+#### Cookie Based Routing for OAuth2
+The default OAuth2 handshake for the Trino REST API has the following steps 
+1. The client requests https://trino/v1/endpoint. The response includes
```
mosabua, line 81: empty line before list

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+```
+HTTP/2 401
+www-authenticate: Bearer x_redirect_server="https://trino/oauth2/token/initiate
+, x_token_server="https://trino/oauth2/token/encoded_token"
+```
+which prompts the client to use oauth2 authentication.
```
mosabua, line 87:
```suggestion
which prompts the client to use OAuth2 authentication.
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+2. GET https://trino/oauth2/token/encoded_token. The response redirects to the IdP's auth endpoint.
```
mosabua, line 88: need to spell out IdP since this is first usage on this page

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+3. GET https://idp/auth_endpoint. The response includes an authorization code and
```
mosabua, line 89: better example and explain where that comes from

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+which prompts the client to use oauth2 authentication.
+2. GET https://trino/oauth2/token/encoded_token. The response redirects to the IdP's auth endpoint.
+3. GET https://idp/auth_endpoint. The response includes an authorization code and 
+redirects the client back to the coordinator's callback endpoint.
+4. GET https://trino/oauth2/callback?authroization_code_parameters. The coordinator 
+sends the authorization code to the IdP in combination with secrets generated in 
+step (1). The IdP returns an access token to the coordinator, which encodes it and sends
+it to the client.
+
+Adding the configuration
```
mosabua, line 96: Use the following configuration (where) to include...

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+```yaml
+cookiePaths: /oauth2
+```
+will cause the Gateway to include a session cookie in the response to the
+request to `/oauth2/token` endpoint. This session cookie will be carried through the client's request
```
mosabua, line 101:
```suggestion
request to `/oauth2/token` endpoint. The session cookie is carried through the client's request
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+to the IdP auth endpoint and `/oauth2/callback`. The session cookie ensures that the request
```
mosabua, line 102:
```suggestion
to the IdP authentication endpoint and `/oauth2/callback`. The session cookie ensures that the request
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+to `/oauth2/callback` is made to the backend cluster that initiated the exchange, which
+is the only backend which has the secrets that allow it to use the authorization code. 
+
+The handshake for the Web UI is slightly different, but because it still uses endpoints
+under `/oauth2`, the same configuration supports a Web UI secured with OAuth2. Adding
+```yaml
```
mosabua, line 108: Add the following configuration to force removal...

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+deleteCookiePaths: /ui/logout
+```
+will force removal of the session cookie when a user logs out.
+
+This configuration will not affect requests to any `v1/*` or other Trino endpoints, so
```
mosabua, line 113: Which configuration ? The cookie-based routing configuration?

```suggestion
The cookie  configuration does not affect requests to any `v1/*` or other Trino endpoints.
```

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+queries and other requests will still be routed according to Routing Rules.
```
mosabua, line 114: Drop

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -20,14 +20,16 @@ public class DataStoreConfiguration
...
 
-    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, Integer queryHistoryHoursRetention)
+    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, Integer queryHistoryHoursRetention, Integer cookieHoursRetention)
     {
         this.jdbcUrl = jdbcUrl;
         this.user = user;
         this.password = password;
         this.driver = driver;
         this.queryHistoryHoursRetention = queryHistoryHoursRetention;
+        this.cookieHoursRetention = cookieHoursRetention;
```
mosabua, line 32: cookieRetentionTime

or 

cookieDuration

I think this should use the airlift "duration" as unit ideally .. not "hours"

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -122,6 +135,26 @@ else if (path.startsWith(TRINO_UI_PATH)) {
...
+        else {
+            return routingManager.provideAdhocBackend(user);
+        }
+    }
+
+    private boolean doRecordQueryId(HttpServletRequest request)
+    {
+        String requestPath = request.getRequestURI();
+        return requestPath.startsWith(V1_STATEMENT_PATH) && request.getMethod().equals(HttpMethod.POST);
+        //TODO: add queryPaths config
```
mosabua, line 155: ??

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -122,6 +135,26 @@ else if (path.startsWith(TRINO_UI_PATH)) {
...
+        if (!Strings.isNullOrEmpty(routingGroup)) {
+            // This falls back on adhoc backend if there are no cluster found for the routing group.
+            return routingManager.provideBackendForRoutingGroup(routingGroup, user);
+        }
+        else {
+            return routingManager.provideAdhocBackend(user);
+        }
+    }
+
+    private boolean doRecordQueryId(HttpServletRequest request)
```
mosabua, line 151: name is not clear what this method should do

recordQueryId

trackQueryId 

or so? 

having "do" and "record" is weird

mosabua: maybe think of what you would write for javadoc

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -252,13 +285,26 @@ private boolean isPathWhiteListed(String path)
                 || extraWhitelistPaths.stream().anyMatch(s -> path.startsWith(s));
     }
 
-    public boolean handleAuthRequest(HttpServletRequest request)
+    @Override
+    public Optional<Cookie> deleteCookieIfNeeded(HttpServletRequest clientRequest)
     {
-        return true;
+        if (!removeCookiePaths.contains(clientRequest.getRequestURI())) {
+            // TODO: support custom rules for determining if a SessionId cookie should be removed and provide current
```
mosabua, line 292: ??

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -252,13 +285,26 @@ private boolean isPathWhiteListed(String path)
                 || extraWhitelistPaths.stream().anyMatch(s -> path.startsWith(s));
     }
 
-    public boolean handleAuthRequest(HttpServletRequest request)
+    @Override
+    public Optional<Cookie> deleteCookieIfNeeded(HttpServletRequest clientRequest)
```
mosabua, line 289: why not also use "remove" .. like in the config

mosabua: also drop "ifNeeded"

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -271,13 +317,25 @@ public String rewriteTarget(HttpServletRequest request)
...
 
         String backend = getBackendFromRoutingGroup(request);
-        // set target backend so that we could save queryId to backend mapping later.
         ((MultiReadHttpServletRequest) request).addHeader(PROXY_TARGET_HEADER, backend);
         logRewrite(backend, request);
 
         return buildUriWithNewBackend(backend, request);
     }
 
+    private void setCookieForStickyRoutingIfNeeded(HttpServletRequest request, HttpServletResponse response, String backend)
```
mosabua, line 326:
```suggestion
    private void setCookie(HttpServletRequest request, HttpServletResponse response, String backend)
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -271,13 +317,25 @@ public String rewriteTarget(HttpServletRequest request)
...
+    {
+        // TODO: support custom rules for determining if a SessionId cookie should be added and provide current
```
mosabua, line 328: ? .. probably better to file issues for future work

mosabua (changes requested):
Started review .. will continue later

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+
+If the request does not have an antecedent, the Gateway first uses [Routing Rules](routing-rules.md)
+to determine which logical group of clusters, termed a "Routing Group",
+should handle the request. A member of the Routing Group, termed a backend, 
+is selected to handle the request using either an adaptive or round-robin 
+strategy.
+
+![Request Routing Flow](assets/gateway-routing-flow.svg)
+
+## Sticky Routing
```
mosabua, line 51:
```suggestion
## Sticky routing
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -252,13 +285,26 @@ private boolean isPathWhiteListed(String path)
...
+    public Optional<Cookie> deleteCookieIfNeeded(HttpServletRequest clientRequest)
     {
-        return true;
+        if (!removeCookiePaths.contains(clientRequest.getRequestURI())) {
+            // TODO: support custom rules for determining if a SessionId cookie should be removed and provide current
+            // path based rule as default option
+            return Optional.empty();
+        }
+        Optional<Cookie> cookie = Arrays.stream(clientRequest.getCookies()).filter(
+                c -> c.getName().equalsIgnoreCase("JSESSIONID")).findAny();
```
mosabua, line 297: do we want to set JSESSIONID as static somewhere .. also do we need a separate one to avoid collisions with other cookies potentially?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -308,74 +366,38 @@ private String getBackendFromRoutingGroup(HttpServletRequest request)
     private String getPreviousBackend(HttpServletRequest request)
     {
         String queryId = extractQueryIdIfPresent(request);
-
-        // Find query id and get url from cache
         if (!isNullOrEmpty(queryId)) {
             return routingManager.findBackendForQueryId(queryId);
         }
+        else if (!Strings.isNullOrEmpty(request.getRequestedSessionId())) {
+            //pin browser sessions to the same backend based on jsessionid, but load balance queries
```
mosabua, line 373: ?? how is that related to load balancing here .. thats just what happens in fall through right?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -308,74 +366,38 @@ private String getBackendFromRoutingGroup(HttpServletRequest request)
...
+            return routingManager.findBackendForCookie(request.getRequestedSessionId().split("\\.")[0]);
+        }
         log.error("No backend found for queryId {}", queryId);
```
mosabua, line 376: update .. maybe we want to log inside the if statement instead .. or in addition?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -384,6 +406,54 @@ protected void postConnectionHook(
...
+                        queryDetail.getBackendUrl());
+            }
+        }
+        else {
+            log.error(
+                    "Non OK HTTP Status code with response [{}] , Status code [{}]",
+                    output,
+                    response.getStatus());
+        }
+        // Saving history at gateway.
```
mosabua, line 453:
```suggestion
        // Save history in Trino Gateway.
```

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -105,5 +106,23 @@ private void startCleanUps()
                 1,
                 120,
                 TimeUnit.MINUTES);
+
+        executorService.scheduleWithFixedDelay(
```
mosabua, line 110: Follow up task for later maybe .. we need a feature in the UI to wipe stale cookies maybe?

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/CookieBackend.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.persistence.dao;
+
+import org.javalite.activejdbc.Model;
```
mosabua, line 16: Can we implement this with JDBI from the start instead of piling more activejdbc on? cc @ebyhr

gateway-ha/src/main/java/io/trino/gateway/ha/router/CookieCacheManager.java
```diff
@@ -0,0 +1,74 @@
...
+{
+    private static final Logger log = LoggerFactory.getLogger(CookieCacheManager.class);
+    private JdbcConnectionManager connectionManager;
+
+    public CookieCacheManager(JdbcConnectionManager connectionManager)
+    {
+        this.connectionManager = connectionManager;
+    }
+
+    public void submitCookieBackend(String cookie, String backend)
```
mosabua, line 33: why not just "addCookie" .. you are using "removeCookie"

gateway-ha/src/main/java/io/trino/gateway/ha/router/CookieCacheManager.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+        catch (Exception e) {
+            log.warn(String.format("Error saving cookie %s for backend %s: %s", cookie, backend, e));
+        }
+        finally {
+            connectionManager.close();
+        }
+    }
+
+    public Optional<String> getBackendForCookie(String cookie)
```
mosabua, line 48: g
```suggestion
    public Optional<String> getBackend(String cookie)
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -66,6 +70,20 @@ public String load(String queryId)
                                     }
                                 });
 
+        cookieBackendCache =
+                CacheBuilder.newBuilder()
+                        .maximumSize(10000)
```
mosabua, line 75: Do we need to make these various values configurable? Now .. or later?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -79,6 +97,17 @@ public void setBackendForQueryId(String queryId, String backend)
         queryIdBackendCache.put(queryId, backend);
     }
 
+    public void setBackendForCookie(String cookie, String backend)
+    {
+        cacheManager.submitCookieBackend(cookie, backend);
+    }
+
+    public boolean deleteCookie(String cookie)
```
mosabua, line 105: delete vs remove .. lets just pick one

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -125,6 +154,17 @@ public String findBackendForQueryId(String queryId)
         return backendAddress;
     }
 
+    public String findBackendForCookie(String cookie)
```
mosabua, line 157:
```suggestion
    public String findBackend(String cookie)
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -185,6 +225,12 @@ protected String findBackendForUnknownQueryId(String queryId)
...
 
+    protected String lookupBackendForCookie(String cookie)
+    {
+        Optional<String> backend = cacheManager.getBackendForCookie(cookie);
+        return backend.orElseGet(() -> provideAdhocBackend(""));
+    }
+
     // Predicate helper function to remove the backends from the list
     // We are returning the unhealthy (not healthy)
     private boolean isBackendNotHealthy(String backendId)
```
mosabua, line 236: Sidenode .. this would be better as isBackendHealthy ..

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -185,6 +225,12 @@ protected String findBackendForUnknownQueryId(String queryId)
         return gatewayBackendManager.getActiveAdhocBackends().get(0).getProxyTo();
     }
 
+    protected String lookupBackendForCookie(String cookie)
```
mosabua, line 228: hm... whats the difference to findBackend? not clear from name

gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql
```diff
@@ -75,4 +75,10 @@ CREATE TABLE IF NOT EXISTS exact_match_source_selectors (
 
     PRIMARY KEY (environment, source, query_type),
     UNIQUE (source, environment, query_type, resource_group_id)
-);
\ No newline at end of file
+);
+
+CREATE TABLE IF NOT EXISTS cookie_backend_lookup (
+    cookie VARCHAR(256) NOT NULL PRIMARY KEY,
```
mosabua, line 81: 256 is long enough?

mosabua (changes requested):

mosabua:
First review round completed. Looks like its in pretty good shape .. please rebase, update naming and such and let me know when its ready again.

docs/routing-logic.md
Chaho12: Why would you need user to configure this? Other than logout?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
Chaho12: In which cases would you need custom rules?

docs/routing-logic.md
```diff
@@ -0,0 +1,114 @@
...
+Gateway determines that they are, they are forwarded to whatever the 
+backend that handled the previous request.
+
+If the request does not have an antecedent, the Gateway first uses [Routing Rules](routing-rules.md)
+to determine which logical group of clusters, termed a "Routing Group",
+should handle the request. A member of the Routing Group, termed a backend, 
+is selected to handle the request using either an adaptive or round-robin 
+strategy.
+
+![Request Routing Flow](assets/gateway-routing-flow.svg)
```
Chaho12, line 49: Can you add a start/user in the flow? I assume it's "is request descended from a previous request"

docs/routing-logic.md
Chaho12: How about adding `(default)` at the end?

Chaho12 (commented):
Is cookiepath config

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: this note is here b/c I was thinking about supporting things like the Trino memory connector. For example
```
CREATE TABLE memory.default.cached_result AS SELECT ...
```
The admin in this case might want to add a cookie to make sure every subsequent query in this session goes to the cluster where this user has created their memory connector tables, and remove the cookie once they have all been `DROP`'d. 

I'll remove this TODO to an issue, after all we may choose a different approach if we find a need for this functionality some day.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: That comment was leftover from earlier development, removed

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: refactored the logic since this statement was unreachable due to a prior check

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
willmostly: i'll open an issue once this is in

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
willmostly: We also have a `findBackendForQueryId` in `RoutingManager` that takes a string, so we need to disambiguate them

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
willmostly: @andythsu ☝️

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
willmostly: the difference was that this method looks up the cookie in the db, while findBackend gets it from the local cache. However this was only used in the cache builder so it makes more sense just to inline it.

gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql
willmostly: It should be. The default session manager, which we are using, uses a long as session id: https://github.com/ganqing2017/jetty/blob/79b822339c3940679ac69a51eaed21b7af3e71c1/jetty-server/src/main/java/org/eclipse/jetty/server/session/DefaultSessionIdManager.java#L253

docs/routing-logic.md
willmostly: Not sure why I'm unable to respond to these individually

> Why would you need user to configure this? Other than logout?
There is no generic reason. However, the cookie routing might be leveraged in use case specific ways. For example, maybe some workflow using [Dune's table API](https://docs.dune.com/api-reference/tables/endpoint/create) would benefit from pinning to a specific cluster, then unpinning after a certain request.

Are you looking for more specific guidance on this in the docs?

> In which cases would you need custom rules?
Again, no generic reason, but there could be a use case specific reason. For example, Starburst has a few different APIs for managing access control, caching and data virtualization. There may be some case where you want to unpin depending on something in the response instead of just the path. However, this is out of scope for the current work so I've removed the TODO

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -269,32 +296,55 @@ private boolean isPathWhiteListed(String path)
                 || extraWhitelistPaths.stream().anyMatch(s -> path.startsWith(s));
     }
 
-    public boolean handleAuthRequest(HttpServletRequest request)
+    @Override
+    public Optional<Cookie> removeCookie(HttpServletRequest clientRequest)
     {
-        return true;
+        if (!removeCookiePaths.contains(clientRequest.getRequestURI())) {
```
oneonestar, line 302: > cookiePaths: If the request URI starts with a path in this list, add a session cookie if none exists. If a cookie exists, route the request to the backend associated with that cookie.

Docs states this is a  startsWith check, but the actual logic is doing a full match check?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -269,32 +296,55 @@ private boolean isPathWhiteListed(String path)
...
-        // set target backend so that we could save queryId to backend mapping later.
         ((MultiReadHttpServletRequest) request).addHeader(PROXY_TARGET_HEADER, backend);
         logRewrite(backend, request);
 
         return buildUriWithNewBackend(backend, request);
     }
 
+    private void setCookie(HttpServletRequest request, HttpServletResponse response, String backend)
+    {
+        if (!cookiePaths.contains(request.getRequestURI())) {
```
oneonestar, line 339: Docs states this is a startsWith check, but the actual logic is doing a full match check?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -269,32 +296,55 @@ private boolean isPathWhiteListed(String path)
...
+    public Optional<Cookie> removeCookie(HttpServletRequest clientRequest)
     {
-        return true;
+        if (!removeCookiePaths.contains(clientRequest.getRequestURI())) {
+            return Optional.empty();
+        }
+        Optional<Cookie> cookie = Arrays.stream(clientRequest.getCookies()).filter(
+                c -> c.getName().equalsIgnoreCase(COOKIE_NAME)).findAny();
+        clientRequest.getSession().invalidate();
+        cookie.ifPresent(c -> routingManager.removeCookie(c.getValue().split("\\.")[0]));
```
oneonestar, line 308: Extract `.split("\\.")[0]` using `Splitter`

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -322,77 +372,33 @@ private String getBackendFromRoutingGroup(HttpServletRequest request)
...
         }
-        log.error("No backend found for queryId %s", queryId);
-        return getBackendFromRoutingGroup(request);
-    }
-
-    private boolean isUseStickyRouting(HttpServletRequest request)
-    {
-        return !isNullOrEmpty(extractQueryIdIfPresent(request));
+        else if (!Strings.isNullOrEmpty(request.getRequestedSessionId())) {
+            return Optional.of(routingManager.findBackendForCookie(request.getRequestedSessionId().split("\\.")[0]));
```
oneonestar, line 382: Replace `.split("\\.")[0]` with `Splitter`.

oneonestar (commented):
Skimming the first half

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -20,14 +20,16 @@ public class DataStoreConfiguration
     private String password;
     private String driver;
     private Integer queryHistoryHoursRetention = 4;
+    private Integer cookieHoursRetention = 24;
```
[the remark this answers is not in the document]
willmostly: no reason other than maintaining uniformity with `queryHistoryHoursRetention`. I see we use a mix of `Integer` and `int` in the `*Configuration` classes, but `int` is a bit more popular. I'll use `int` here, at some point we should clean up the other configs.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: good catch, same for `setCookie` as well

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -85,19 +92,20 @@ public class QueryIdCachingProxyHandler
     private final Meter requestMeter;
     private final List<String> extraWhitelistPaths;
     private final String applicationEndpoint;
+    private final Set<String> cookiePaths;
+    private final Set<String> removeCookiePaths;
+    private final DefaultSessionIdManager sessionIdManager = new DefaultSessionIdManager(new Server());
+
+    public QueryIdCachingProxyHandler(QueryHistoryManager queryHistoryManager, RoutingManager routingManager, RoutingGroupSelector routingGroupSelector, int serverApplicationPort, Meter requestMeter, List<String> extraWhitelistPaths, Set<String> cookiePaths, Set<String> removeCookiePaths)
```
oneonestar, line 99: @willmostly Looks like you applied a wrong formatter on this file.

oneonestar:
I saw a NPE error but the test passed. Is this expected?
```
./mvnw clean test -Dtest=TestGatewayHaMultipleBackend
...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!    THIS APPLICATION HAS NO HEALTHCHECKS. THIS MEANS YOU WILL NEVER KNOW      !
!     IF IT DIES IN PRODUCTION, WHICH MEANS YOU WILL NEVER KNOW IF YOU'RE      !
!    LETTING YOUR USERS DOWN. YOU SHOULD ADD A HEALTHCHECK FOR EACH OF YOUR    !
!         APPLICATION'S DEPENDENCIES WHICH FULLY (BUT LIGHTLY) TESTS IT.       !
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
2024-03-14T05:59:05.553Z	INFO	main	org.eclipse.jetty.server.handler.ContextHandler	Started i.d.j.MutableServletContextHandler@14f35726{/,null,AVAILABLE}
2024-03-14T05:59:05.553Z	INFO	main	org.eclipse.jetty.server.AbstractConnector	Started application@30b16aa6{HTTP/1.1, (http/1.1)}{0.0.0.0:30575}
2024-03-14T05:59:05.554Z	INFO	main	org.eclipse.jetty.server.AbstractConnector	Started admin@1228a579{HTTP/1.1, (http/1.1)}{0.0.0.0:31088}
2024-03-14T05:59:05.554Z	INFO	main	org.eclipse.jetty.server.Server	Started Server@6a1b51a7{STARTING}[11.0.19,sto=30000] @17768ms
2024-03-14T05:59:05.556Z	INFO	main	io.trino.gateway.ha.HaGatewayTestUtils	okhttp3.RequestBody$2@72b16e26
2024-03-14T05:59:05.691Z	INFO	pool-8-thread-7 - POST /entity?entityType=GATEWAY_BACKEND	io.trino.gateway.ha.resource.EntityEditorResource	Setting up the backend trino1 with healthy state
2024-03-14T05:59:05.691Z	INFO	pool-8-thread-7 - POST /entity?entityType=GATEWAY_BACKEND	io.trino.gateway.ha.router.RoutingManager	backend trino1 isHealthy true
2024-03-14T05:59:05.716Z	INFO	main	io.trino.gateway.ha.HaGatewayTestUtils	okhttp3.RequestBody$2@5575b4c2
2024-03-14T05:59:05.732Z	INFO	pool-8-thread-2 - POST /entity?entityType=GATEWAY_BACKEND	io.trino.gateway.ha.resource.EntityEditorResource	Setting up the backend trino2 with healthy state
2024-03-14T05:59:05.732Z	INFO	pool-8-thread-2 - POST /entity?entityType=GATEWAY_BACKEND	io.trino.gateway.ha.router.RoutingManager	backend trino2 isHealthy true
2024-03-14T05:59:05.732Z	INFO	main	io.trino.gateway.ha.HaGatewayTestUtils	okhttp3.RequestBody$2@472cf6af
2024-03-14T05:59:05.747Z	INFO	pool-8-thread-8 - POST /entity?entityType=GATEWAY_BACKEND	io.trino.gateway.ha.resource.EntityEditorResource	Setting up the backend custom with healthy state
2024-03-14T05:59:05.747Z	INFO	pool-8-thread-8 - POST /entity?entityType=GATEWAY_BACKEND	io.trino.gateway.ha.router.RoutingManager	backend custom isHealthy true
2024-03-14T05:59:05.763Z	INFO	qtp1976110789-76	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Rerouting [http://127.0.0.1:20110/v1/statement]--> [http://localhost:32791/v1/statement]
2024-03-14T05:59:05.763Z	INFO	qtp1976110789-76	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Processing request endpoint: [/v1/statement], payload: [SELECT 1]
2024-03-14T05:59:05.773Z	INFO	qtp1976110789-93	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	user from X-Trino-User
2024-03-14T05:59:05.802Z	INFO	qtp1976110789-91	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Rerouting [http://127.0.0.1:20110/v1/statement]--> [http://localhost:32792/v1/statement]
2024-03-14T05:59:05.802Z	INFO	qtp1976110789-91	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Processing request endpoint: [/v1/statement], payload: [SELECT 1]
2024-03-14T05:59:05.819Z	INFO	qtp1976110789-72	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	user from X-Trino-User
2024-03-14T05:59:05.835Z	INFO	qtp1976110789-77	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Rerouting [http://127.0.0.1:20110/oauth2]--> [http://localhost:21561/oauth2]
2024-03-14T05:59:05.880Z	ERROR	qtp1976110789-91	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Error in proxying falling back to super call
java.lang.NullPointerException: Cannot read the array length because "array" is null
	at java.base/java.util.Arrays.stream(Arrays.java:5528)
	at io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.setCookie(QueryIdCachingProxyHandler.java:306)
	at io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.postConnectionHook(QueryIdCachingProxyHandler.java:350)
	at io.trino.gateway.proxyserver.ProxyServletImpl.onResponseContent(ProxyServletImpl.java:133)
	at org.eclipse.jetty.proxy.ProxyServlet$ProxyResponseListener.onContent(ProxyServlet.java:217)
	at org.eclipse.jetty.client.api.Response$AsyncContentListener.onContent(Response.java:197)
	at org.eclipse.jetty.client.ResponseNotifier.notifyContent(ResponseNotifier.java:150)
	at org.eclipse.jetty.client.ResponseNotifier.notifyContent(ResponseNotifier.java:134)
	at org.eclipse.jetty.client.HttpReceiver$ContentListeners.notifyContent(HttpReceiver.java:711)
	at org.eclipse.jetty.client.HttpReceiver.plainResponseContent(HttpReceiver.java:381)
	at org.eclipse.jetty.client.HttpReceiver.responseContent(HttpReceiver.java:364)
	at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.content(HttpReceiverOverHTTP.java:344)
	at org.eclipse.jetty.http.HttpParser.parseContent(HttpParser.java:1854)
	at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:1565)
	at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.parse(HttpReceiverOverHTTP.java:221)
	at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.process(HttpReceiverOverHTTP.java:160)
	at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.receive(HttpReceiverOverHTTP.java:91)
	at org.eclipse.jetty.client.http.HttpChannelOverHTTP.receive(HttpChannelOverHTTP.java:97)
	at org.eclipse.jetty.client.http.HttpConnectionOverHTTP.onFillable(HttpConnectionOverHTTP.java:207)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.runTask(AdaptiveExecutionStrategy.java:421)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.consumeTask(AdaptiveExecutionStrategy.java:390)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.tryProduce(AdaptiveExecutionStrategy.java:277)
	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.run(AdaptiveExecutionStrategy.java:199)
	at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:411)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:969)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1194)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1149)
	at java.base/java.lang.Thread.run(Thread.java:1583)


2024-03-14T05:59:05.912Z	INFO	qtp1976110789-78	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Rerouting [http://127.0.0.1:20110/oauth2/callback]--> [http://localhost:21561/oauth2/callback]
2024-03-14T05:59:05.912Z	WARN	qtp1976110789-78	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Proxy Target not set on request, unable to decipher HOST header
2024-03-14T05:59:05.919Z	WARN	qtp1976110789-76	io.trino.gateway.ha.router.CookieCacheManager	Error saving cookie node0s8y2ljvxnnxi10fedegmdxxqj0 for backend null: org.javalite.activejdbc.DBException: org.h2.jdbc.JdbcSQLException: Unique index or primary key violation: "PRIMARY_KEY_2 ON PUBLIC.COOKIE_BACKEND_LOOKUP(COOKIE) VALUES ('node0s8y2ljvxnnxi10fedegmdxxqj0', 1)"; SQL statement:
INSERT INTO cookie_backend_lookup (cookie, created_timestamp) VALUES (?, ?) [23505-192], query: INSERT INTO cookie_backend_lookup (cookie, created_timestamp) VALUES (?, ?), params: node0s8y2ljvxnnxi10fedegmdxxqj0, 1710395945917
2024-03-14T05:59:05.944Z	INFO	qtp1976110789-74	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Rerouting [http://127.0.0.1:20110/oauth2/callback]--> [http://localhost:32791/oauth2/callback]
2024-03-14T05:59:05.944Z	WARN	qtp1976110789-74	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Proxy Target not set on request, unable to decipher HOST header
2024-03-14T05:59:05.998Z	INFO	qtp1976110789-91	io.trino.gateway.ha.handler.QueryIdCachingProxyHandler	Rerouting [http://127.0.0.1:20110/v1/custom/extra]--> [http://localhost:21561/v1/custom/extra]
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 18.43 s -- in io.trino.gateway.ha.TestGatewayHaMultipleBackend
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for trino-gateway-parent 7-SNAPSHOT:
[INFO]
[INFO] trino-gateway-parent ............................... SUCCESS [  3.014 s]
[INFO] gateway-ha ......................................... SUCCESS [ 47.937 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  51.126 s
[INFO] Finished at: 2024-03-14T14:59:07+09:00
[INFO] ------------------------------------------------------------------------
```

oneonestar:
I don't understand why we have to store the cookie into database.
If the cookie contains enough information, we could decide which backend to go by just looking at it right?

willmostly:
> I don't understand why we have to store the cookie into database. If the cookie contains enough information, we could decide which backend to go by just looking at it right?

The cookie only contains a session id, it does not identify the backend. Are you suggesting using a custom cookie instead of the Jetty session cookie?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: Could you elaborate? I use IntelliJ with the [Airlift codestyle](https://github.com/airlift/codestyle) for formatting.

oneonestar:
> The cookie only contains a session id, it does not identify the backend. Are you suggesting using a custom cookie instead of the Jetty session cookie?

Yea. I'm not familiar with HTTP cookie. Could a cookie be `trino_routing_cookie=<this_goes_to_backend1>`?

willmostly:
> Yea. I'm not familiar with HTTP cookie. Could a cookie be `trino_routing_cookie=<this_goes_to_backend1>`?

It could, the content of a cookie can be anything under 4kB. They also may have attributes such as `max-age`.

IMO the decision between using Jetty's ootb session management, as I'm doing here, and a custom cookie like `GATEWAYSESSIONCOOKIE backend=trino.xyz.com` comes down to whether Jetty's built in support offers maintainability advantages that outweight the additional burden on the DB. I could go either way.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
oneonestar: I thought the codestyle was `Method declaration parameters: Chop down if long + New line after '('`, but looks like they were not set. I think https://github.com/airlift/codestyle needs an update.

Examples:
https://github.com/trinodb/trino/blob/9410a6db2d8d152658f04424777a6f4e54ce7d4b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java#L50-L58
https://github.com/trinodb/trino/blob/9410a6db2d8d152658f04424777a6f4e54ce7d4b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/orc/OrcFileWriterFactory.java#L123-L134

---

This also doesn't look right (basically unreadable). `Binary expressions` might needs an update.
https://github.com/trinodb/trino-gateway/blob/5a659c3b2931709f28ede8a27fa8a8688a5d67ca/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java#L260

oneonestar: https://github.com/airlift/codestyle/pull/13#issuecomment-885611067
Seems there are side effects for these rules, so they were not added. We might have to enforce it manually.

oneonestar:
Could you please explain what features do Jetty's session management provided here?
I saw we are setting and deleting the cookie manually.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: Thanks, i didn't realize this rule wasn't in effect - I will avoid allowing IntelliJ to remove custom line breaks going forward!

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -269,26 +273,38 @@ private boolean isPathWhiteListed(String path)
...
+        if (removeCookiePaths.isEmpty() || removeCookiePaths.stream().noneMatch(c -> clientRequest.getRequestURI().startsWith(c))) {
+            return Optional.empty();
+        }
+
+        Optional<Cookie> cookie = Arrays.stream(clientRequest.getCookies()).filter(c -> c.getName().equalsIgnoreCase(COOKIE_NAME)).findAny();
+        clientRequest.getSession().invalidate();
+        cookie.ifPresent(c -> routingManager.removeCookie(sessionIdManager.getId(c.getValue())));
+
+        return cookie.map(c -> {
+            c.setMaxAge(0);
```
vishalya, line 288: can we have utility function to delete the cookies which we can reuse with the same values? UI also sets up some cookies.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -269,26 +273,38 @@ private boolean isPathWhiteListed(String path)
...
     {
         if (!isPathWhiteListed(request.getRequestURI())) {
             return buildUriWithNewBackend(applicationEndpoint, request);
         }
 
-        if (isUseStickyRouting(request)) {
-            String backend = getPreviousBackend(request);
-            logRewrite(backend, request);
-            return buildUriWithNewBackend(backend, request);
+        Optional<String> previousBackend = getPreviousBackend(request);
```
vishalya, line 301: Does the OAuth request land here?

vishalya (commented):
Not much from my side, overall it's good to go.

willmostly:
The approach has been updated - all session information is now stored in cookies. The cookies include metadata that should provide similar capabilities to server side session management. 

Cookies are used to support routing for the OAuth2 handshake. Cookies are not used for any other routing strategy at this time.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: i updated the function to delete cookies based on the metadata set in the cookie. It will only delete the cookies it sets up itself - is that what you need?

willmostly:
This needs a fixup before reviewing, please hold off for the moment

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -349,50 +372,14 @@ protected void postConnectionHook(
             Callback callback)
     {
         try {
-            String requestPath = request.getRequestURI();
```
willmostly, line 352: this code block was extracted to `recordBackendForQueryId`

willmostly:
@mosabua I will update the docs once I get confirmation from @oneonestar, @Chaho12 and @vishalya that the new approach is directionally correct

mosabua:
Sounds good

oneonestar:
Excellent! Signed cookie and stateless server are exactly what I thought of.
I'll review it closely later.

gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
```diff
@@ -0,0 +1,241 @@
...
+        if (System.currentTimeMillis() > ts + lifetime.toMillis()) {
+            return false;
+        }
+
+        if (GatewayCookieConfiguration.isValidate()) {
+            if (isNullOrEmpty(signature)) {
+                return false;
+            }
+            return signature.equals(computeSignature());
+        }
```
oneonestar, line 237: Why not just always sign and verify the cookie?

gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyServletImpl.java
```diff
@@ -98,6 +98,22 @@ protected String rewriteTarget(HttpServletRequest request)
...
+            Response serverResponse)
+    {
+        if (clientRequest.getCookies() == null) {
+            super.onServerResponseHeaders(clientRequest, proxyResponse, serverResponse);
+            return;
+        }
+
+        this.proxyHandler.deleteCookies(clientRequest).forEach(proxyResponse::addCookie);
+
+        super.onServerResponseHeaders(clientRequest, proxyResponse, serverResponse);
```
oneonestar, line 114: This logic could be simplified.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
```diff
@@ -162,6 +176,90 @@ public void testBackendConfiguration()
...
+                        .addHeader("X-Trino-Routing-Group", "custom")
+                        .build();
+        Response initiateResponse = httpClient.newCall(initiateRequest).execute();
+        String cookieHeader = initiateResponse.header("set-cookie");
+        assertThat(isNullOrEmpty(cookieHeader)).isFalse();
+        GatewayCookie cookie = GatewayCookie.fromCookie(new Cookie(cookieHeader.split("=")[0], cookieHeader.split("=")[1].split(";")[0]));
+        assertThat(isNullOrEmpty(cookie.getSignature())).isFalse();
+
+        // Tamper with values. This will cause the cookie to be ignored because its values will not match the signature,
+        // causing the request will be routed to the adhoc backend
```
oneonestar, line 250: If cookie is invalid due to incorrect signature, we ignore the cookie and route it according to default routing logic?
Doesn't feel right about this. Any bug / mis-configuration could result in query being routed to random backend without error message. IMO, we should reject the request with invalid cookie.

Chaho12: I think we should consider this in a more broader and consistent perspective, as how Gateway should do on invalid requests, soley as a load balancer routing with default conf on errors? or with some internal logic that checks and returns error without routing. 

Invalid requests can be from two cases, by Gateway and by Trino.
- The latter, of course, is bypassed without any decisions made by gateway.
- The former, as of now, seems to bypass and/or route to adhoc
  - e.g.) if user sends query request with wrong header, it is sent to adhoc (default)
  - e.g.) any NON HTTP Post request in white list path, including extra path by user, is forwarded. Other than that, it is considered an API for the gateway and could fail.
  
REST API is slighlty exception cases, but other than that i think sending to adhoc seems fine, keeping in mind that Gateway is a router.

gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.google.common.collect.ImmutableSet;
+import io.airlift.units.Duration;
+
+import java.util.Set;
+
+public class OAuth2GatewayCookieConfiguration
+{
+    // Configuration initialization using dropwizard requires
+    // instance method setters. Values are global, and can be accessed using static getters
```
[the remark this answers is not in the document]
willmostly: I can revert this to the standard approach if desired. While working on this I have found that plumbing the configurations through `HaGatewayProviderModule` to be cumbersome and time consuming. Unlike Trino, where you can have multiple copies of a plugin with different configurations, all of the Gateway config is global. So I figured we could do away with the plumbing.

gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    private static Set<String> routingPaths = ImmutableSet.of("/oauth2");
+    private static Set<String> deletePaths = ImmutableSet.of("/logout", "/oauth2/logout");
```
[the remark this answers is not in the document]
willmostly: Someone might be running a custom web app with a different logout URL. The `/oauth2/logout` path is not from OS Trino, but is used by Starburst Insights for example. I do not know what Dune and other Trino based products use.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -269,9 +275,20 @@ private boolean isPathWhiteListed(String path)
                 || extraWhitelistPaths.stream().anyMatch(s -> path.startsWith(s));
     }
 
-    public boolean handleAuthRequest(HttpServletRequest request)
+    @Override
+    public List<Cookie> deleteCookies(HttpServletRequest clientRequest)
```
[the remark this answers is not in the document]
willmostly: What would you suggest? `expireCookie`?

gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
willmostly: I wouldn't mind removing this option. I figured some people may complain about overhead.

gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyHandler.java
```diff
@@ -114,4 +117,9 @@ protected boolean isCompressed(final byte[] compressed)
         return (compressed[0] == (byte) GZIPInputStream.GZIP_MAGIC)
                 && (compressed[1] == (byte) (GZIPInputStream.GZIP_MAGIC >> 8));
     }
+
+    public List<Cookie> deleteCookies(HttpServletRequest clientRequest)
+    {
+        return ImmutableList.of();
```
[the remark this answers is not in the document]
willmostly: throwing here causes `TestProxyServer.testProxyServer` to fail

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -322,21 +339,27 @@ private String getBackendFromRoutingGroup(HttpServletRequest request)
...
-        // Find query id and get url from cache
         if (!isNullOrEmpty(queryId)) {
-            return routingManager.findBackendForQueryId(queryId);
+            return Optional.of(routingManager.findBackendForQueryId(queryId));
+        }
+        if (request.getCookies() != null) {
+            List<GatewayCookie> cookies = Arrays.stream(request.getCookies())
+                    .filter(c -> c.getName().startsWith(GatewayCookie.PREFIX))
+                    .map(GatewayCookie::fromCookie)
+                    .filter(c -> c.isValid())
```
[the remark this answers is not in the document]
willmostly: I'm using statics to avoid the configuration plumbing

gateway-ha/src/main/java/io/trino/gateway/ha/config/GatewayCookieConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class GatewayCookieConfiguration
+{
+    private static SecretKey cookieSigningKey;
+    private static boolean validate;
```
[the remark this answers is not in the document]
willmostly: Yeah I agree its weird. It was required since dropWizard uses the setter. Lets continue the conversation in the comment below.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
willmostly: I agree with @oneonestar in this case, since the cookie will only have an invalid or missing signature due to misbehavior by the client. There are no cases where the client should need to interact with it, so I don't think we should try to handle this case gracefully.

gateway-ha/src/main/java/io/trino/gateway/ha/config/GatewayCookieConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+    public static SecretKey getCookieSigningKey()
+    {
+        return cookieSigningKey;
+    }
+
+    public void setCookieSigningSecret(String cookieSigningSecret)
+    {
+        cookieSigningKey = new SecretKeySpec(cookieSigningSecret.getBytes(UTF_8), "HmacSHA256");
+    }
+}
```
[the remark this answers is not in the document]
willmostly: It is used by Dropwizard during config initialization

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
Chaho12: Sounds fair enough :)

gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieConfiguration.java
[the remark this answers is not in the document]
willmostly: I reworked this to use a singleton instead of statics, which looks better imo. LMK if you guys like the approach. I find the current way that configuration values are passed through constructors to be very cumbersome.

Please consider the naming. I decided on `*ConfigurationPropertiesProvider`, since it provides the values of the config properties of the corresponding `*Configuration` classes. Open to suggestions...

gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
willmostly: @Chaho12 wdyt? Any complaints about always validating? There is the cost of computing a cryptographic hash, which isn't free but probably a small fraction of the total cost per request.

Chaho12: Do you know how small the fraction is?

willmostly: It should be roughly similar to the cost of initiating a TLS handshake

Chaho12: Then i think it is small enough to be ignored :)

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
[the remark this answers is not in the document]
Chaho12: `expireCookieList` maybe?

gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieProperties.java
```diff
@@ -0,0 +1,63 @@
...
+    public Duration getLifetime()
+    {
+        ensureInitialized();
+        return oAuth2GatewayCookieConfiguration.getLifetime();
+    }
+
+    private void ensureInitialized()
+    {
+        if (oAuth2GatewayCookieConfiguration == null) {
+            throw new IllegalStateException("GatewayCookieConfiguration not initialized before use");
```
Chaho12, line 60: this should be `oAuth2GatewayCookieConfiguration` ?
Btw in which cases would oAuth2GatewayCookieConfiguration not be initialized?

docs/routing-logic.md
Chaho12: Please do add examples :)

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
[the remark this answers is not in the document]
oneonestar: Maybe `prepareCookiesForDeletion()` or `generateDeleteCookieList()`?

gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieProperties.java
willmostly: It is initialized during startup, so this error shouldn't be reachable unless we introduce a bug.

willmostly: >this should be oAuth2GatewayCookieConfiguration ?
can you clarify?

Chaho12: Oh i meant the message `"GatewayCookieConfiguration not initialized before use"` shouldn't it be `"oAuth2GatewayCookieConfiguration not initialized before use"`

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
```diff
@@ -162,6 +175,90 @@ public void testBackendConfiguration()
...
+
+        Request logoutRequest =
+                new Request.Builder()
+                        .url("http://localhost:" + routerPort + "/custom/logout")
+                        .post(requestBody)
+                        .addHeader("Cookie", initiateResponse.header("set-cookie"))
+                        .build();
+        Response logoutResponse = httpClient.newCall(logoutRequest).execute();
+
+        assertThat(isNullOrEmpty(logoutResponse.header("set-cookie"))).isFalse();
```
oneonestar, line 219: Test it's actually a deleted cookie.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
```diff
@@ -162,6 +175,90 @@ public void testBackendConfiguration()
...
+                        .build();
+        Response initiateResponse = httpClient.newCall(initiateRequest).execute();
+        String cookieHeader = initiateResponse.header("set-cookie");
+        assertThat(isNullOrEmpty(cookieHeader)).isFalse();
+        GatewayCookie cookie = GatewayCookie.fromCookie(new Cookie(cookieHeader.split("=")[0], cookieHeader.split("=")[1].split(";")[0]));
+        assertThat(isNullOrEmpty(cookie.getSignature())).isFalse();
+
+        // Tamper with values. This will cause the cookie to be ignored because its values will not match the signature,
+        // causing the request will be routed to the adhoc backend
+        cookie.setTs(System.currentTimeMillis());
```
oneonestar, line 250: This test sometimes fail in my local environment.
Shell we set another field other than ts?

```
[ERROR] Failures: 
[ERROR]   TestGatewayHaMultipleBackend.testCookieSigning:259 
expected: 500
 but was: 200
[INFO] 
[ERROR] Tests run: 95, Failures: 1, Errors: 0, Skipped: 0
```

gateway-ha/src/main/java/io/trino/gateway/ha/config/OAuth2GatewayCookieProperties.java
willmostly: Ah yes, thanks! I added some extra guidance to make the error message more descriptive

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
willmostly: `setTs` is the only setter, since all the other fields are final. `gatewayCookie.setTs(gatewayCookie.getTs() + 1000)` should fail deterministically - do you think this is ok or should we construct a new cookie with a different backend?

oneonestar: `gatewayCookie.setTs(gatewayCookie.getTs() + 1000)` should be good.

mosabua (approved):
I think this is in a good enough shape now and we should merge. Any objections @oneonestar @vishalya @Chaho12 ?

Chaho12 (approved):
LGTM~

mosabua:
Merging this. Any further concerns can be addressed in follow up PRs.

PR #189, opened 2024-01-25
Add github milestone labeling

### LINKED ISSUE #187: Add milestones
Port the github action workflow from trino and potentially add/update links to the release notes page

mosabua:
This is the same flow as used in the trino repo. Should work the same since the pom setup and such are identical. I am not sure if there is a way to test apart from merging it.

Fixes #187

### DIFF
--- a/.github/workflows/milestone.yml
+++ b/.github/workflows/milestone.yml
@@ -1,0 +1,83 @@
+name: milestone
+
+on:
+  push:
+    branches: [main]
+
+defaults:
+  run:
+    shell: bash --noprofile --norc -euo pipefail {0}
+
+jobs:
+  set-milestone:
+    if: github.repository_owner == 'trinodb'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Get milestone from pom.xml
+        run: |
+          .github/bin/retry ./mvnw -v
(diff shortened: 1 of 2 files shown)


.github/workflows/milestone.yml
```diff
@@ -0,0 +1,83 @@
+name: milestone
+
+on:
+  push:
+    branches: [master]
```
ebyhr, line 5: We don't use master branch in this repository.

.github/workflows/milestone.yml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  set-milestone:
+    if: github.repository_owner == 'trinodb'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Get milestone from pom.xml
+        run: |
+          .github/bin/retry ./mvnw -v
```
ebyhr, line 20: retry script doesn't exist in this repository.

ebyhr (changes requested):

.github/workflows/milestone.yml
mosabua: dang .. good spotting

.github/workflows/milestone.yml
mosabua: will add ..

.github/workflows/milestone.yml
mosabua: Updated to main

ebyhr (approved):

Issue #184, opened 2024-01-25
Fix dependency issues


mosabua:
To allow dependency plugin analyze to be enabled. For development you can set

```
<air.check.skip-dependency>false</air.check.skip-dependency>
```

in root pom and the use. 

```
mvn install
```

Or you can use

```
mvn dependency:analyze
```

For now that produces LOTS of failures. PRs will remove problems one at a time. 

Some improvements (also towards Java 21 support) will come from updating dependencies.

```
mvn versions:display-dependency-updates
```

At the end we will remove the skip setting in the root pom.

mosabua:
Ping @wendigo as the dependency master ;-)

mosabua:
Made considerable progress. Current analyze report is less than half as large as it was:

```
[INFO] --- dependency:3.6.0:analyze (default-cli) @ gateway-ha ---
[ERROR] Used undeclared dependencies found:
[ERROR]    ch.qos.logback:logback-classic:jar:1.4.8:compile
[ERROR]    com.fasterxml.jackson.core:jackson-databind:jar:2.16.0:compile
[ERROR]    org.apache.commons:commons-pool2:jar:2.11.1:compile
[ERROR]    org.bouncycastle:bcprov-jdk15on:jar:1.70:compile
[ERROR]    ch.qos.logback:logback-core:jar:1.4.8:compile
[ERROR]    org.slf4j:slf4j-api:jar:2.0.9:compile
[ERROR]    org.glassfish.jersey.core:jersey-server:jar:3.0.12:compile
[ERROR] Unused declared dependencies found:
[ERROR]    io.trino:trino-jdbc:jar:433:compile
[ERROR]    org.bouncycastle:bcpkix-jdk15on:jar:1.70:compile
[ERROR]    org.postgresql:postgresql:jar:42.6.0:compile
[ERROR]    io.airlift:http-client:jar:239:compile
[ERROR]    org.ehcache:ehcache:jar:3.8.1:compile
[ERROR]    org.mvel:mvel2:jar:2.5.1.Final:compile
[ERROR]    joda-time:joda-time:jar:2.12.5:compile
[ERROR]    mysql:mysql-connector-java:jar:8.0.17:compile
```

mosabua:
A bunch more improvements are in. Latest status:

```
[INFO] --- dependency:3.6.0:analyze (default-cli) @ gateway-ha ---
[ERROR] Used undeclared dependencies found:
[...]

PR #192, opened 2024-01-26
Use Jakarta Nonnull

mosabua:
Part of #184


gateway-ha/pom.xml
```diff
@@ -42,6 +42,11 @@
             <artifactId>jackson-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
```
ebyhr, line 47: It's fine for temporary change, but we should remove this dependency eventually. https://github.com/trinodb/trino/pull/18047

ebyhr (approved):

mosabua:
Let me leave this open .. maybe I fix this up now .. there is so much cruft in this codebase..

gateway-ha/pom.xml
mosabua: Updated to PR .. was a quick pivot ;-)

Issue #114, opened 2023-12-20
[Bug] trino-gateway v4 oidc redirecting to http://localhost:8090/sso


jeffreychucASAPP:
Hi, we originally had oidc to okta working with v3 of trino-gateway with the following config

```
authentication:
 defaultType: oauth
 oauth:
   issuer: ${ISSUER}
   clientId: ${TRINO_GATEWAY_OIDC_CLIENT_ID}
   clientSecret: ${TRINO_GATEWAY_OIDC_CLIENT_SECRET}
   tokenEndpoint: ${ISSUER}/oauth2/v1/token
   authorizationEndpoint: ${ISSUER}/oauth2/v1/authorize
   jwkEndpoint: ${ISSUER}/oauth2/v1/keys
   redirectUrl: https://${URL_OF_GATEWAY}/oidc/callback
   userIdField: email
   scopes:
     - openid
     - email
     - offline_access
```

After we updated to latest version 4 [latest commit of main on trino-gateway at the time]  https://github.com/trinodb/trino-gateway/commit/b6bbb1423f935c36e49e3f2bb83bc44b5336474f when we hit the url of the gateway we had deployed it would redirect to `http://localhost:8090/sso` instead of `http://${URL_OF_GATEWAY}/sso`

I believe the update of `jetty`/`dropwizard` introduced this
https://github.com/trinodb/trino-gateway/commit/434eec9d77465b9373dd07683449102e4d793b83

I tested that theory by checking out the latest v3 and applying all commits one at a time and that was the commit that introduced that behavior.

Hopefully theres just a config option I'm missing but I couldn't get the redirect fixed.

mosabua:
The Lombok removal is not yet in any release and I am not sure how the jetty removal could cause the issue.

jeffreychucASAPP:
@mosabua ok so is the suggestion to stay on the v3 release for now?  We were at a loss at how the jetty update would cause the issue as well :\

mosabua:
We will have to find out one way or another. I just dont have time to look at it at this stage.

willmostly:
@jeffreychucASAPP were you able to identify the RC here?

jeffreychucASAPP:
@willmostly sorry I totally forgot to respond here.  When I was testing I believe the commit that causes the issue is this commit https://github.com/trinodb/trino-gateway/commit/434eec9d77465b9373dd07683449102e4d793b83

[...]

PR #194, opened 2024-01-26
Set the defaults to forward the headers

vishalya:
This is should be default value, dropwizard changed the defaults from 1.0 to some higher version. It takes care of the issues like https://github.com/trinodb/trino-gateway/issues/114

### DIFF
--- a/gateway-ha/gateway-ha-config.yml
+++ b/gateway-ha/gateway-ha-config.yml
@@ -26,9 +26,11 @@ server:
   applicationConnectors:
     - type: http
       port: 8090
+      useForwardedHeaders: true
   adminConnectors:
     - type: http
       port: 8091
+      useForwardedHeaders: true
 
 # This can be adjusted based on the coordinator state
 monitor:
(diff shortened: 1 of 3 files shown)


willmostly (approved):
Approved. I would love to see some e2e tests including oauth on the backends to give us a better chance to catch changes in defaults, but I don't think it is possible with our current test framework. I'll open an issue for that.

mosabua:
Can you add it to the docker config file as well ... then I am good with it.

Chaho12 (approved):
Thx for quick fix. LGTM 👍

gateway-ha/gateway-ha-config.yml
```diff
@@ -26,9 +26,11 @@ server:
   applicationConnectors:
     - type: http
       port: 8090
+      useForwardedHeaders: true
```
ebyhr, line 29: Please capitalize the commit title. https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#format-git-commit-messages

vishalya: Fixed!

mosabua (approved):
Thanks @vishalya

mosabua:
Squashed, reworded commit message, and merged.

Issue #195, opened 2024-01-26
Refactor LogFilterFactory


mosabua:
To get rid of usage of `ch.qos.logback.classic.spi.ILoggingEvent`;

That would allow us to get rid of the dependency  `ch.qos.logback:logback-classic:jar:1.4.8:compile`

Part of #184

andythsu:
I think logback is the framework used by dropwizard and LogFilterFactory is used to filter out messages that contain "Authorization" keyword. Do we want to get rid of filters in general? Or switch to another framework

mosabua:
@andythsu .. logback is fine. What we want to get rid of is the old logback-classic usage .. there is only one class that uses it and it pull in the old jar. I assume there is a modern replacemnt in the normal logback and then we can get rid of that jar as dependency.

andythsu:
I see. I can give it a try. You can assign it to me

mosabua:
Decided to fix dependency declaration instead. @andythsu will send PR.

PR #202, opened 2024-01-28
get rid of the dependency ch.qos.logback:logback-classic:jar:1.4.8:compile

andythsu:
resolves https://github.com/trinodb/trino-gateway/issues/195


gateway-ha/pom.xml
```diff
@@ -187,6 +187,12 @@
             <version>1.70</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.codehaus.janino</groupId>
+            <artifactId>janino</artifactId>
+            <version>${dep.janino.version}</version>
+        </dependency>
```
ebyhr, line 194: I don't feel much benefit to replace the existing dependency with new one. We're going to use airlift as the base plugin that uses JUL for logger.

ebyhr (changes requested):

gateway-ha/src/main/resources/logback.xml
```diff
@@ -0,0 +1,24 @@
...
+<!DOCTYPE configuration>
+
+<configuration>
+    <import class="ch.qos.logback.core.ConsoleAppender" />
+    <import class="ch.qos.logback.core.filter.EvaluatorFilter" />
+
+    <appender name="STDOUT" class="ConsoleAppender">
+        <filter class="EvaluatorFilter">
+            <evaluator>
+                <expression>event.getMessage().contains("Authorization")</expression>
```
ebyhr, line 11: What happens if we use broken Java expressions here? Does CI catch such mistake?

andythsu: I would assume no. It looks like it print the error message in the console, but CI won't fail

gateway-ha/pom.xml
andythsu: Just wondering why is trino using JUL for logger? Looks like slf4j is more powerful than JUL

mosabua:
Discussed with @andythsu .. we will just declare the dependency since this is approach is not really helping us improve the overall setup. Andy will send a PR for the dependency declaration.

PR #209, opened 2024-01-29
Improve docker setup and docs

mosabua:


docker/README.md
```diff
@@ -1,39 +1,92 @@
...
 * jq
 
-### Run the Trino Gateway server
+# Building a custom Docker image
+
+To build an image for a locally modified version of Trino Gateway, run the Maven
+build and then build the image for your desired processor architecture.
+
+```bash
+./build.sh arm64
```
rdsarvar, line 22: i think this is missing `-a`

docker/README.md
```diff
@@ -1,39 +1,92 @@
...
+
+The release process publishes images to DockerHub for Trino Gateway 6 and newer.
+
+Set the default image
+```bash
+export TRINO_GATEWAY_IMAGE=${TEST_GATEWAY_IMAGE} 
+export DOCKER_DEFAULT_PLATFORM=${TEST_PLATFORM}
+
 TEST_GATEWAY_IMAGE="trinodb/trino-gateway:latest"
 TEST_PLATFORM="amd64"
```
rdsarvar, line 80: nitpick: duplicative, can be deleted

docker/README.md
```diff
@@ -1,39 +1,92 @@
...
 
-TRINO_GATEWAY_IMAGE=${TEST_GATEWAY_IMAGE} DOCKER_DEFAULT_PLATFORM=${TEST_PLATFORM} \
-    docker compose -f minimal-compose.yml \
-    up --wait
+```
```
rdsarvar, line 82: i think this might be an extra set of backticks which will break the rendering

docker/README.md
```diff
@@ -51,13 +106,16 @@ Or, visit it in your browser by opening http://localhost:8080
...
 ## Health Checking
 
-By default the container health checking is done by the [/usr/lib/trino/bin/health-check](./bin/health-check)
-script which simply expects a 2XX response from the server at `/api/public/backends`.
+By default the container health checking is done by
+the [/usr/lib/trino/bin/health-check](./bin/health-check)
+script which simply expects a 2XX response from the server
+at `/api/public/backends`.
 
 ## Building a custom Docker image
```
rdsarvar, line 120: todo: can delete this section as it's been reworded+moved to be the first

docker/README.md
```diff
@@ -1,39 +1,92 @@
...
+export TEST_GATEWAY_IMAGE="trinodb/trino-gateway:latest"
+export TEST_PLATFORM="amd64"
+```
+
+The release process publishes images to DockerHub for Trino Gateway 6 and newer.
+
+Set the default image
+```bash
+export TRINO_GATEWAY_IMAGE=${TEST_GATEWAY_IMAGE} 
+export DOCKER_DEFAULT_PLATFORM=${TEST_PLATFORM}
```
rdsarvar, line 77: note(non-blocking): i don't know if i forgot to mention this anywhere but once there's a `trinodb/trino-gateway:latest` neither of these technically need to be exported for people to spin up locally with the minimal-compose.yml file. Not sure if it's worth mentioning or not anywhere, I'll leave it to you 😄 

setting `DOCKER_DEFAULT_PLATFORM` -> Enables you to do cross platform runs when doing `docker compose`, but the default of not being set should be sufficient for local testing. We use it during the build process to test the containers as `docker compose` doesn't provide a `--platform` option

setting `TRINO_GATEWAY_IMAGE` -> Allows you to run a local image version you might've just built. (As you've already shown the readers)

rdsarvar (commented):
couple nitpicks and minor changes, otherwise the improvements lg2m

docker/README.md
mosabua: Thanks .. I have been wondering about the indirection. So from what I understand now we could just get rid of the usage of TEST_GATEWAY_IMAGE and TEST_PLATFORM just explain how to define TRINO_GATEWAY_IMAGE and DOCKER_DEFAULT_PLATFORM only .. right @rdsarvar ?

docker/README.md
mosabua: Done

docker/README.md
mosabua: Also .. once we published the first container to DockerHub we will separate the development docs from the docs about running it locally maybe .. or maybe even morph the docker compose setup into something suitable for spinning it up in production.. we will see

docker/README.md
mosabua: This are all alternative configs so each one has to be complete..

docker/README.md
rdsarvar: >Thanks .. I have been wondering about the indirection. So from what I understand now we could just get rid of the usage of TEST_GATEWAY_IMAGE and TEST_PLATFORM just explain how to define TRINO_GATEWAY_IMAGE and DOCKER_DEFAULT_PLATFORM only .. right @rdsarvar ?

correct, it's only really needed in the scenario that you want to run a different architecture
ex: running arm64 and for whatever reason you want to run amd64 instead using `docker compose`

explaining how to define it makes sense 👍

one additional thing to note though: if you haven't run `build.sh` without specifying the `-a` flag you might not have pulled the different platform for the postgres image, example from the `container-test.sh` file:
```bash
    # Ensure the platform specific image has been pulled locally for Postgres
    DOCKER_DEFAULT_PLATFORM=${PLATFORM} \
        docker compose -f "${COMPOSE_PATH}" \
        pull ${COMPOSE_POSTGRES_NAME}
```

as by default if you're on arm64 you'd only have that specific image pulled, you might not have seen the errors because you ran `./docker/build.sh` which pulled all of the platform images for you


example -- 

lets run it with our arch first:
```
docker compose -f docker/minimal-compose.yml up -d
[+] Running 15/15
✔ postgres 14 layers [⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿]      0B/0B      Pulled                                                                                                               25.7s
 <LAYERS PULLED HERE>
[+] Running 3/3
✔ Network docker_default       Created                                                                                                                                      0.1s
✔ Container docker-postgres-1  Healthy                                                                                                                                      0.2s
✔ Container docker-gateway-1   Started                                                                                                                                      0.0s


docker compose -f docker/minimal-compose.yml down
[+] Running 3/3
✔ Container docker-gateway-1   Removed                                                                                                                                      0.2s
✔ Container docker-postgres-1  Removed                                                                                                                                      0.1s
✔ Network docker_default       Removed                                                                                                                                      0.1s
```

then lets run it with a different arch:
```
TRINO_GATEWAY_IMAGE="trino-gateway:6-SNAPSHOT-arm64" DOCKER_DEFAULT_PLATFORM="linux/arm64" docker compose -f docker/minimal-compose.yml up -d
[+] Running 1/0
✔ Network docker_default       Created                                                                                                                                      0.0s
⠋ Container docker-postgres-1  Creating                                                                                                                                     0.0s
Error response from daemon: image with reference postgres was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64
```
pretty much it complains because it has the postgres image locally but it's not for the right platform

so you might also need to acknowledge an explicit pull of the image:
```
DOCKER_DEFAULT_PLATFORM="linux/arm64" docker compose -f docker/minimal-compose.yml pull postgres
[+] Pulling 15/15
✔ postgres 14 layers [⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿]      0B/0B      Pulled                                                                                                               24.2s
 <LAYERS PULLED HERE>

TRINO_GATEWAY_IMAGE="trino-gateway:6-SNAPSHOT-arm64" DOCKER_DEFAULT_PLATFORM="linux/arm64" docker compose -f docker/minimal-compose.yml up -d
[+] Running 2/2
✔ Container docker-postgres-1  Healthy                                                                                                                                      0.2s
✔ Container docker-gateway-1   Started                                                                                                                                      0.0s
```

rdsarvar: though... the above is probably an edge case of usage? and hopefully those doing the above scenario would know to pull the image.. so i'll leave it up to you if you want to note the above information or not

mosabua: I ran into exactly that thing on macos. It complained about having linux/amd64 but needing darwin/arm64 .. so I had to delete the image and restart.

mosabua: Also .. does it make sense to set the variables each time or better to just export them once and call it a day?

mosabua: Developers could even export them in their bashrc file or whatever.

mosabua: At this stage I think we should get rid of TEST_GATEWAY_IMAGE and TEST_PLATFORM amnd just use TRINO_GATEWAY_IMAGE and DOCKER_DEFAULT_PLATFORM ..

does that make sense?

mosabua:
@raghavsethi I pushed a cleaned up version but I think I still want to get rid of TEST_GATEWAY_IMAGE and TEST_PLATFORM ...

mosabua:
Also .. should we rename the file minimal-compose.yml to docker-compose.yml 

then we can remove the -f flag usage in the command and just go

docker compose up --wait
docker compose down

docker/build.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -xeuo pipefail
+set -euo pipefail
```
mosabua, line 3: -x is way too verbose for default config .. even the current setup still is rather loud..

docker/README.md
rdsarvar: >At this stage I think we should get rid of TEST_GATEWAY_IMAGE and TEST_PLATFORM amnd just use TRINO_GATEWAY_IMAGE and DOCKER_DEFAULT_PLATFORM ..

yup makes sense

>Also .. does it make sense to set the variables each time or better to just export them once and call it a day?

Without going through this it's hard for me to best judge what would be preferable. For now I suppose we educate them on it and let them decide?

probably not the best to set `DOCKER_DEFAULT_PLATFORM` in their shell bootup unless they're always working on that platform :thinking:

mosabua:
I will get rid of the variable duplication ..

mosabua:
> I will get rid of the variable duplication ..

Done ..

andythsu (approved):

Chaho12 (approved):
build + docker compose works fine on my mac :)

mosabua:
Ping @wendigo @ebyhr ..

docker/README.md
```diff
@@ -1,93 +1,158 @@
...
+
+By default the container health check uses the file `docker/bin/health-check`
+mounted at `/usr/lib/trino/bin/health-check`. The scripts expects a 2XX response
+from the server  at `/api/public/backends`.
+
+## Release
+
+After the container images are built locally, a maintainer can login to
+DockerHub and use the `release-docker.sh` script to publish the locally built
+images.
```
ebyhr, line 154: This section is unrelated to most users. Not sure if worth mentioning.

docker/README.md
```diff
@@ -1,93 +1,158 @@
...
+
+
+>>>>>>> 8280976 (Improve docker setup and docs)
```
ebyhr, line 157: Remove.

docker/README.md
```diff
@@ -1,93 +1,158 @@
...
+```
 
-```bash
-./build.sh
+Use a published image from DockerHub.
+
+```shell
+export TRINO_GATEWAY_IMAGE="trinodb/trino-gateway:latest"
 ```
+The release process publishes images for Trino Gateway 6 and newer to DockerHub.
```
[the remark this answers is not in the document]
mosabua: Because currently none is available and that command wont work for releases before 6 ever.

docker/README.md
mosabua: Because currently this is the readme file and it contains a mixture of user, developer and other docs. In a follow up we will split it apart and move into the proper docs

docker/README.md
mosabua: done

docker/README.md
```diff
@@ -1,93 +1,158 @@
...
 ```
 
-Set the environment variable `TRINO_GATEWAY_BASE_IMAGE` to use a specific base image
-to build Trino Gateway image.
+Inspect the logs for progress and troubleshooting:
 
 ```bash
-export TRINO_GATEWAY_BASE_IMAGE=<image>
-./build.sh
+docker compose logs gateway
```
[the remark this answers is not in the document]
mosabua: renamed .. also changed filename

docker/README.md
```diff
@@ -1,93 +1,158 @@
...
 
-To build an image for a specific released version of Trino Gateway,
-specify the `-r` option, and the build script will download
-all the required artifacts:
+Start Trino Gateway and its PostgreSQL backend database, and wait until the 
+health check is successful:
 
 ```bash
-./build.sh -r 4
+docker compose up --wait
```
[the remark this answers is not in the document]
mosabua: can you make sure you pull the latest of the branch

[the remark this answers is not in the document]
mosabua: The command works if the filename is docker-compose.yml .. which it is .. but was not in the refactoring .. so maybe you had an outdated branch

mosabua:
Rename to trino-gateway and such all done.

docker/build.sh
```diff
@@ -140,7 +142,7 @@ source container-test.sh
 for arch in "${ARCHITECTURES[@]}"; do
     # TODO: remove when https://github.com/multiarch/qemu-user-static/issues/128 is fixed
     if [[ "$arch" != "ppc64le" ]]; then
-        test_container "${TAG_PREFIX}-$arch" "linux/$arch" "${SCRIPT_DIR}/minimal-compose.yml" "gateway" "postgres"
+        test_container "${TAG_PREFIX}-$arch" "linux/$arch" "${SCRIPT_DIR}/docker-compose.yml" "gateway" "postgres"
```
ebyhr, line 145: This script is broken. Rename gateway to trino-gateway.

docker/README.md
```diff
@@ -1,93 +1,153 @@
...
+configured backends:
+
+```bash
+curl localhost:8080/api/public/backends
+```
+
+The Trino Gateway is available at [http://localhost:8080](http://localhost:8080).
+
+The PostgreSQL backend database for Trino Gateway runs at
+[http://localhost:5432](http://localhost:5432).
```
ebyhr, line 136: Does PostgreSQL accept HTTP access?

docker/README.md
```diff
@@ -1,93 +1,153 @@
...
+## Configuration
+
+The image uses the configuration file `gateway-ha/gateway-ha-config-docker.yml`
+from the project checkout, and mounts it at `/opt/trino/gateway-ha-config.yml`.
+
+## Health check
+
+By default the container health check uses the file `docker/bin/health-check`
+mounted at `/usr/lib/trino/bin/health-check`. The scripts expects a 2XX response
+from the server  at `/api/public/backends`.
```
ebyhr, line 147:
```suggestion
from the server at `/api/public/backends`.
```

docker/README.md
mosabua: Rejigged to talk about querying the db via that host and port.

docker/build.sh
mosabua: Fixed

mosabua:
Still figuring out why it works locally but not on GHA

docker/build.sh
mosabua: Renamed passed parameter and section in chart. Testing again.

mosabua:
> Still figuring out why it works locally but not on GHA

Had to update the retries count for the postgres backend since it timed out on GHA. All good now @ebyhr

mosabua:
Can you maybe also have a look @willmostly and @vishalya  .. I think its in really good shape finally .. thanks to the help from @ebyhr and @rdsarvar

mosabua:
Fyi .. I would like to get this in asap so we can cut a release with it.

docker/README.md
```diff
@@ -1,93 +1,154 @@
...
 * [Docker Compose V2](https://docs.docker.com/compose/)
 * jq
 
-### Run the Trino Gateway server
+# Building a custom Docker image
+
+Use the following steps to build a Docker image from your local Trino Gateway
+codebase
+
+First, run the Maven build in the project root.
```
Chaho12, line 21:
```suggestion
First, run the Maven build in the project root as guided in [development](../docs/development.md)
```
nit. am not sure how familiar they are with maven but command is not `mvn clean install` but `./mvnw clean install` which might be confusing for first timer in trino

docker/README.md
```diff
@@ -1,93 +1,154 @@
...
 
-## Quickstart
-
-### Dependencies
-
-This docker build process requires:
+This docker build process requires the following software:
 
 * [Docker Compose V2](https://docs.docker.com/compose/)
 * jq
```
Chaho12, line 14: how about adding java 17, mvn?

Chaho12 (approved):

docker/README.md
```diff
@@ -1,93 +1,154 @@
...
 
 ```bash
-export TRINO_GATEWAY_BASE_IMAGE=<image>
-./build.sh
+docker compose logs trino-gateway
 ```
 
-## Getting Help
+Typically your operating system automatically sets the default Docker platform
+value. Try to In some cases it can be useful to explicitly set it.
```
martint, line 108: Broken sentence: "Try to In some cases it can ..."

docker/README.md
```diff
@@ -1,93 +1,154 @@
...
+
+For example on ARM64-based MacOS you can set it to use `linux` because it 
+otherwise is potentially set to `darwin` and there are no PostgreSQL images 
+available for `darwin` and this can prevent starting the with docker compose.
```
martint, line 112: "starting the ??? with"

docker/README.md
mosabua: Add inline link

docker/README.md
mosabua: No .. thats in the development docs .. we will ultimately merge them but dont want to duplicate now.

ebyhr (approved):

PR #210, opened 2024-01-29
Joda

mosabua:


gateway-ha/pom.xml
```diff
@@ -163,11 +163,6 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
         </dependency>
 
-        <dependency>
```
ebyhr, line 166: We should change the scope to runtime instead of removal in my opinion.

wendigo:
https://github.com/trinodb/trino-gateway/pull/224

PR #212, opened 2024-01-30
Upgrade and fix wiremock dependency

mosabua:
The dependency moved in terms of groupid to `org.wiremock` and we have a duplicate declaration for that already.

Weirdly the package of the java classes did NOT move so there is no code to update.

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -304,13 +304,6 @@
 
         <!-- Test deps -->
         <dependency>
-            <groupId>com.github.tomakehurst</groupId>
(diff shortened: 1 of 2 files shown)


gateway-ha/pom.xml
```diff
@@ -303,13 +303,6 @@
         </dependency>
 
         <!-- Test deps -->
-        <dependency>
```
mosabua, line 306: These are the old group id coordinates and org.wiremock is already declared as duplicate in the pom.

ebyhr (approved):

Issue #213, opened 2024-01-30
Fix UI render failure


mosabua:
Somehow I get the following failure now when starting Trino Gateway and trying to access the login screen.

```
io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 0c1704d55dbb0f4e
! io.dropwizard.views.common.ViewRenderException: Unable to find a renderer for /template/query-history-view.ftl
! at io.dropwizard.views.common.ViewMessageBodyWriter.writeTo(ViewMessageBodyWriter.java:77)
! ... 66 common frames omitted
```

This might have to do with the recent removal of the dependency

```
        <dependency>
            <groupId>io.dropwizard</groupId>
            <artifactId>dropwizard-views-freemarker</artifactId>
        </dependency>
```

but my initial testing of adding it back in did not fix the issue yet. I need to do more testing I guess.

mosabua:
Could we related to https://github.com/trinodb/trino-gateway/pull/190

mosabua:
Or https://github.com/trinodb/trino-gateway/pull/185

PR #215, opened 2024-01-30
Fix ViewRenderer missing dependency

andythsu:
Fixes https://github.com/trinodb/trino-gateway/issues/213


ebyhr (approved):

mosabua:
Thanks for the fix @andythsu and @ebyhr  ... will see how that affects the dependency analysis

PR #217, opened 2024-01-30
Documentation update 

ruschecker:
This config required to let trino know that request has been proxied from gateway.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -43,6 +43,12 @@ application locally.
 Following are instructions for installing Trino Gateway for production
 environments.
 
+You also have to update `config.properties` with
+```commandline
+http-server.process-forwarded=true
+```
+
+
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


Chaho12 (commented):
idk this will be necessary as it will be added by default.
@vishalya what do you think?

- https://github.com/trinodb/trino-gateway/pull/194

mosabua:
The Trino Gateway configuration is separate from the Trino configuration.. we might need both. This PR would have to move the changes into a section that talks about the Trino cluster configuration.

And @vishalya is still updating his PR to add the same config to the docker example config and then we will merge that PR separately.

mosabua:
👋 @ruschecker .. are you going to update this PR or should I send a replacement addressing the concerns raised in the review.

mosabua:
Replaced by #297

PR #224, opened 2024-01-31
Update airbase to 149 and airlift to 240 and enable more checks

wendigo:
Closes https://github.com/trinodb/trino-gateway/issues/184

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -342,6 +342,12 @@
 
         <dependency>
             <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -85,11 +85,6 @@
         </dependency>
 
         <dependency>
-            <groupId>io.airlift</groupId>
-            <artifactId>http-client</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-assets</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


gateway-ha/pom.xml
```diff
@@ -137,18 +193,19 @@
...
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-views</artifactId>
         </dependency>
 
-        <dependency>
```
mosabua, line 147: We found that this is needed for the UI to work .. we might have to experiment with a different scope or somehow figure out how the dependency analyse run can be satisfied.

gateway-ha/pom.xml
```diff
@@ -280,23 +326,29 @@
             <version>${dep.jeasy.version}</version>
         </dependency>
 
-        <dependency>
```
mosabua, line 283: This dep was added to fix an issue as well .. we need to verify this does not come up again.

gateway-ha/pom.xml
mosabua: See https://github.com/trinodb/trino-gateway/issues/213

gateway-ha/pom.xml
```diff
@@ -312,6 +364,20 @@
...
+            <artifactId>ehcache</artifactId>
+            <version>${dep.ehcache.version}</version>
+            <scope>runtime</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.mvel</groupId>
+            <artifactId>mvel2</artifactId>
+            <version>2.5.1.Final</version>
+            <scope>runtime</scope>
```
mosabua, line 378: oh.. cool .. I see

mosabua (dismissed):

gateway-ha/pom.xml
mosabua: I see.. you just moved it down to runtime scope.

wendigo: Yep

wendigo:
I've tested UI locally. It renders. No errors in the logs.

mosabua:
So good.. you rock @wendigo ! I was gonna get to it eventually but you were faster.. nice!

mosabua (approved):
Awesome .. thanks for all the clean up. The final jar with dependencies probably shrunk quite a bit more again. Love it.

PR #226, opened 2024-01-31
Bunch of dependencies updates

wendigo:


mosabua (commented):
In terms of the changes this looks good. But we have to verify that the mysql and postgres driver updates dont mess up the backend db connection.

ebyhr (approved):

PR #227, opened 2024-01-31
Allow custom base image

andythsu:


docker/Dockerfile
```diff
@@ -12,8 +12,8 @@
 # limitations under the License.
 
 # syntax=docker/dockerfile:1
-
-FROM registry.access.redhat.com/ubi9/ubi-minimal:latest AS jdk-download
+ARG BASE_IMAGE
```
ebyhr, line 15: Please mention the motivation in PR description.

mosabua: Done

docker/build.sh
```diff
@@ -114,6 +114,9 @@ cp "${SCRIPT_DIR}/Dockerfile" "${WORK_DIR}"
 
 TAG_PREFIX="trino-gateway:${TRINO_GATEWAY_VERSION}"
 
+# Use ubi9 minimal as it's more secure
```
mosabua, line 117: remove this comment .. more secure than what?

docker/build.sh
```diff
@@ -114,6 +114,9 @@ cp "${SCRIPT_DIR}/Dockerfile" "${WORK_DIR}"
...
+BASE_IMAGE=${BASE_IMAGE:-'registry.access.redhat.com/ubi9/ubi-minimal:latest'}
```
mosabua, line 118: Rename property to TRINO_GATEWAY_BASE_IMAGE

mosabua:
Please also add this to the README in the docker folder

docker/README.md
```diff
@@ -80,6 +80,13 @@ all the required artifacts:
 ./build.sh -r 4
 ```
 
+Set the environment variable `TRINO_GATEWAY_BASE_IMAGE` to use a specific base image to build Trino Gateway image.
+
+```bash
+export TRINO_GATEWAY_BASE_IMAGE=<image>
+./build.sh
```
ebyhr, line 87: build.sh already offers some arguments h,a,r,j to changer the behavior. Why do we need to use an environment variable for the base image?

andythsu: do we want to use these arguments to change the base image then? Because right now the base image is using the one on dockerhub, but we will need to use our own base image to build the docker image.

mosabua: The script also uses a bunch of env variables.. so its kinda inconsistent .. I think of the trino image more like the java home path to be honest

docker/README.md
```diff
@@ -80,6 +80,13 @@ all the required artifacts:
 ./build.sh -r 4
 ```
 
+Set the environment variable `TRINO_GATEWAY_BASE_IMAGE` to use a specific base image to build Trino Gateway image.
```
mosabua, line 83: 80 char width wrap please

andythsu: fixed

mosabua (approved):
Looks good now.

Issue #229, opened 2024-02-01
Replace ActiveJDBC with JDBI


ebyhr:
The example PR is https://github.com/trinodb/trino-gateway/pull/145

- [x] query_history #145
- [ ] selectors https://github.com/trinodb/trino-gateway/pull/284
- [x] exact_match_source_selectors https://github.com/trinodb/trino-gateway/pull/285
- [x] gateway_backend #231
- [ ] resource_groups
- [x] resource_groups_global_properties #230

mosabua:
Thanks for filing this issue. It will help with tracking and also potentially others helping out.

PR #230, opened 2024-02-02
Use JDBI for resource_groups_global_properties

ebyhr:
Relates to #229


Chaho12 (approved):
wow so neat. 👍

mosabua (approved):
Lets roll with this for now and see. Worst case we have to refactor.

PR #232, opened 2024-02-02
Use builder for ClusterStats

ebyhr:


gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -13,168 +13,113 @@
...
-import static com.google.common.base.MoreObjects.toStringHelper;
+import java.util.Map;
 
-public class ClusterStats
+import static java.util.Objects.requireNonNull;
+
+public record ClusterStats(
+        String clusterId, int runningQueryCount,
+        int queuedQueryCount,
+        int blockedQueryCount,
```
willmostly, line 25: `blockedQueryCount` is not used anywhere. It should correspond to `BasicQueryStats.isFullyBlocked()`, as mentioned in https://github.com/trinodb/trino-gateway/issues/233. 

I think we should remove it, as it doesn't seem like an important dimension of cluster health. What is your opinion?

willmostly (approved):
This looks like a good change. Ptal at my comment on the `blocked` field - just thought i'd note it in case it is more convenient to remove it here vs a  separate PR.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
ebyhr: Good catch. Let's remove the field in a follow-up.

ebyhr:
@wendigo @mosabua Gentle reminder.

mosabua (approved):
Looks good. 

Just our of curiosity .. are you aiming to update other record classes to the same pattern as well?

ebyhr:
No. We don't need to change the existing recrod which don't require setters.

mosabua:
> No. We don't need to change the existing record which don't require setters.

Cool.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
ebyhr: https://github.com/trinodb/trino-gateway/pull/241

PR #235, opened 2024-02-02
Inline dependency version

ebyhr:
Follows dependency declaration style in Trino project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,28 +44,6 @@
         <air.java.version>17</air.java.version>
         <air.check.skip-spotbugs>true</air.check.skip-spotbugs>
         <air.check.skip-pmd>true</air.check.skip-pmd>
-
-        <!-- dependency versions -->
-        <dep.activejdbc.version>2.3</dep.activejdbc.version>
-        <dep.apache.commons.version>3.13.0</dep.apache.commons.version>
-        <dep.apache.httpcomponents>4.5.13</dep.apache.httpcomponents>
-        <dep.bouncycastle.version>1.70</dep.bouncycastle.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -44,28 +44,6 @@
...
-        <dep.mysqlconnector.version>8.0.33</dep.mysqlconnector.version>
-        <dep.okhttp.version>3.14.9</dep.okhttp.version>
-        <dep.postgresql.version>42.7.1</dep.postgresql.version>
-        <dep.reflections.version>0.9.10</dep.reflections.version>
-        <dep.trino.version>433</dep.trino.version>
-        <dep.wiremock.version>3.3.1</dep.wiremock.version>
-
     </properties>
 
     <dependencyManagement>
```
mosabua, line 49: If we stick with one module .. we could get rid of dependency management altogether

pom.xml
```diff
@@ -44,28 +44,6 @@
         <air.java.version>17</air.java.version>
         <air.check.skip-spotbugs>true</air.check.skip-spotbugs>
         <air.check.skip-pmd>true</air.check.skip-pmd>
-
-        <!-- dependency versions -->
```
mosabua, line 48: Some of these were needed as overrides since there were different versions coming in between dropwizard and airbase .. and now also airlift

mosabua:
If we do this clean up we might as well clean up dependencyManagement as well

wendigo (approved):

pom.xml
mosabua: And we might have to check that our override works and potentially enforce via depMgt

wendigo: Airlift comes with latest :)

PR #237, opened 2024-02-05
Avoid using TestMethodOrder in TestHaGatewayManager

ebyhr:
We shouldn't use `TestMethodOrder` as much as possible.


mosabua (approved):
Looks good.. I was concerned about the order stuff when I saw it originally but never got around to cleaning up.

PR #238, opened 2024-02-07
Suggest changing health check conf in docker compose

Chaho12:
~It seems unncessary as of now to have healthcheck confs in both minimal compose and each extended compose files.~
~IDE shows `Missing required key(s): 'test'` error too so unless a different health check is needed for running minimal compose, it seems better to get rid of one conf (makes it more confusing)~
- ~[gateway-compose yaml](https://github.com/trinodb/trino-gateway/blob/main/docker/gateway-compose.yml) and [postgres compose yaml](https://github.com/trinodb/trino-gateway/blob/main/docker/postgres-backend-compose.yml)~

Since docker compose files is merged into one + docker compose conf is updated,
This PR 
- adds a test as IDE shows `Missing required key(s): 'test' warning
- suggests users to change in health check depending on the workload in production usage

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -6,6 +6,11 @@ The Docker image of Trino Gateway is designed for the following use cases:
 * Automated usage with an orchestration system like Kubernetes to simplify
   deployment
 
+## Production setup
+
+The healthcheck configurations in the `docker-compose` file is for suitable for development and testing purposes only.
+Change the configuration for your production deployment based on the workload and your specific requirements.
+
 ## Build requirements
 
 This docker build process requires the following software:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -6,6 +6,7 @@ services:
       postgres:
         condition: service_healthy
     healthcheck:
+      test: ["CMD-SHELL", "/usr/lib/trino/bin/health-check || exit 1"]
       interval: 5s
       timeout: 5s
       retries: 60


docker/minimal-compose.yml
```diff
@@ -4,17 +4,7 @@ services:
     extends:
       file: ./gateway-compose.yml
       service: gateway
-    healthcheck:
```
rdsarvar, line 7: note: for context the `minimal-compose.yml` file originally had a different configuration than the `localdev/docker-compose.yml` that was removed

the goal of overwriting these values was to give sufficient timeouts during the `test-container.sh` script without needing to have overly long health checks in real usage


Open to this removal and updating the defaults to be larger, i guess it asks the expectations of these files:
As long as we tell end users that the values aren't meant for production and just local dev + release I'd say we just make the defaults long enough to be sufficient for the test-container.sh case?

mosabua: yes .. I agree that these files should be sufficient for local usage only. But we do need to document some advice for production setup.

Chaho12: @rdsarvar what's your suggested times? is current conf in minimal-compose suggested onces?

After removing overwrites in minimal-compose, how about these confs?

## local dev + release testing

gateway-compose | before | after
-- | -- | --
interval(s) | 10 | 5
timeout(s) | 5 | 5
retries | 3 | 60
start_period(s) | 10 | 20

postgres-backend-compose | before | after
-- | -- | --
interval(s) | 10 | 1
timeout(s) | 3 | 1
retries | 3 | 60
start_period(s) | | 10

## production setup for docs

  | gateway-compose | postgres-backend-compose
-- | -- | --
interval(s) | 10 | 10
timeout(s) | 3 | 3
retries | 3 | 3
start_period(s) | 10 | 10

rdsarvar: Those values seem reasonable to me

Chaho12: @rdsarvar one more question. in healthcheck for gateway-compose.yml file does not have test defined so how about adding 
`test: ["CMD-SHELL", "/usr/lib/trino/bin/health-check || exit 1"]`

docker/gateway-compose.yml
```diff
@@ -6,10 +6,11 @@ services:
       postgres:
         condition: service_healthy
     healthcheck:
-      interval: 10s
+      test: ["CMD-SHELL", "/usr/lib/trino/bin/health-check || exit 1"]
```
Chaho12, line 9: Added healthcheck here @rdsarvar

docker/README.md
```diff
@@ -7,6 +7,18 @@ This Docker image is designed to:
...
+Current healthcheck configurations for docker compose files is for develop purpose.  
+So we advise changing it to settings as following: 
+
+|  configuration  | [gateway-compose.yml](./gateway-compose.yml) | [postgres-backend-compose.yml](./postgres-backend-compose.yml) |
+|:---------------:|:--------------------------------------------:|:--------------------------------------------------------------:|
+|   interval(s)   |                      10                      |                               10                               |
+|   timeout(s)    |                      3                       |                               3                                |
+|     retries     |                      3                       |                               3                                |
+| start_period(s) |                      10                      |                               10                               |
+
```
Chaho12, line 21: added production setup here PTAL @mosabua

docker/gateway-compose.yml
rdsarvar: By default it’s using the health check defined in the Dockerfile, is your preference to be explicit?

mosabua:
Please rebase

docker/README.md
ebyhr: The basis for these numbers are unclear to me. I believe this config should be determined based on the workload. I don't think mentioning these numbers is a good idea. 

If this config fits all environments, we should change the yml files instead of docs.

ebyhr (changes requested):

docker/README.md
Chaho12: Well as @rdsarvar [saids](https://github.com/trinodb/trino-gateway/pull/238/files#r1481547210), current config is for internal test purpose and not really suitable for production.
I agree to the point that the numbers depend based on the workload, then how about just simply mentioning that current setup is for develop purpose and user's should change it when using in production?

Chaho12:
rebased (notice some config changes has already been applied) 

Since docker compose files is merged into one + docker compose conf is updated,
This PR 
- adds a test as IDE shows `Missing required key(s): 'test' warning
- suggests users to change in health check depending on the workload in production usage

ebyhr (commented):
I would leave the final review to @mosabua

docker/README.md
```diff
@@ -6,6 +6,11 @@ The Docker image of Trino Gateway is designed for the following use cases:
 * Automated usage with an orchestration system like Kubernetes to simplify
   deployment
 
+## Production Setup
+
+Current healthcheck configurations for docker compose file is for `develop` purpose.
```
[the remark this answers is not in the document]
Chaho12: done :)

docker/README.md
```diff
@@ -6,6 +6,11 @@ The Docker image of Trino Gateway is designed for the following use cases:
 * Automated usage with an orchestration system like Kubernetes to simplify
   deployment
 
+## Production Setup
```
mosabua, line 9:
```suggestion
## Production setup
```

docker/README.md
```diff
@@ -6,6 +6,11 @@ The Docker image of Trino Gateway is designed for the following use cases:
...
+
+Current healthcheck configurations for docker compose file is for develop purpose.
```
mosabua, line 11:
```suggestion
The healthcheck configurations in the `docker-compose` file is for suitable for development and testing purposes only.
```

docker/README.md
```diff
@@ -6,6 +6,11 @@ The Docker image of Trino Gateway is designed for the following use cases:
...
+So we suggest changing it based on the workload for production.
```
mosabua, line 12:
```suggestion
Change the configuration for your production deployment based on the workload and your specific requirements.
```

mosabua (approved):
Just some rewording we should apply. Then good to go. Ultimately we need to provide more guidance for production deployement, but that can follow later.

Chaho12:
@mosabua also PTAL for merge

mosabua:
Looks good now. Thanks for adjusting.

PR #241, opened 2024-02-09
Remove unused field

ebyhr:

PR #243, opened 2024-02-10
Remove unused field

ebyhr:
Same as https://github.com/trinodb/trino-gateway/pull/241


ebyhr:
Merging as it's already approved in #241

PR #245, opened 2024-02-12
Add helm chart

### LINKED ISSUE #87: Add Helm chart
It might be useful to add a Helm chart to manage Trino Gateway deployment on k8s.

luismacosta:
[trino-gateway] add helm chart:

- secret-bootstrap: database username + password
- secret-extra: can be used for more secrets
- deployment: initContainer added to initialize database, engines postgres + mysql. it's working for postgres. mysql, I'm getting the error _"ERROR 1061 (42000) at line 17: Duplicate key name 'query_history_created_idx'"_
- ingress: expose request + admin UIs
- gateway-config.yml: this was added to docker image, contains env vars. [dockerize](https://github.com/jwilder/dockerize) will replace the content using values.yaml.  docker-image folder added just to explain this logic.

Fixes #87


helm/docker-image/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN set -eux && \
+    apt-get -qq update && apt-get --no-install-recommends --no-install-suggests -yqq install curl && \
+    curl --silent --location --retry 3 -o /tmp/gateway-ha.jar https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha/$GATEWAY_VERSION/gateway-ha-$GATEWAY_VERSION-jar-with-dependencies.jar && \
+    curl --silent --location --retry 3 https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz | tar xz -C /usr/bin && \
+    curl --silent --location --retry 3 -o /tmp/gateway-ha-persistence-postgres.sql https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql
+
+FROM ${FROM_REPO}:${FROM_TAG} as target
+
+ARG GATEWAY_UID=1001
```
luismacosta, line 19:
```suggestion
ARG GATEWAY_UID=1000
```
```suggestion
ARG GATEWAY_UID=1001
```

helm/docker-image/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ARG GATEWAY_GID=1001
```
luismacosta, line 20:
```suggestion
ARG GATEWAY_GID=1000
```

fabjesus (approved):

helm/templates/configmap-bootstrap.yaml
```diff
@@ -0,0 +1,166 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: db-bootstrap
+  labels:
+    {{- include "gateway.labels" . | nindent 4 }}
+data:
+  gateway-ha-persistence-postgres.sql: |
```
Chaho12, line 9: hmm is it necessary to have sql as configmap? How about copying during dockerfile as it does not change that often

helm/values.yaml
```diff
@@ -0,0 +1,92 @@
+bootstrap:
+  image:
+    #must be postgres or mysql
+    repository: postgres
+    pullPolicy: IfNotPresent
+    tag: 16.1-bullseye
```
Chaho12, line 6: Here we do not really specify postgres version, so I think we should make this same; let user to define, use latest, or fix version. 
- https://github.com/trinodb/trino-gateway/blob/main/docker/postgres-backend-compose.yml#L4

helm/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    level: INFO
+  log:
+    level: INFO
+  tls:
+    enabled: false
+    keystorePath: ""
+    keystorePassword: ""
+  routingRules:
+    enabled: false
+    configPath: ""
```
Chaho12, line 27: How about adding a configmap for this and provide default routing rules as routing rules is almost mandatory to use gateway.

helm/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+  queryHistorySize: 1000
+  rerouteRequestsToApplication: true
```
Chaho12, line 29: what this config for? i don't see this being used

helm/values.yaml
luismacosta: configmap added

helm/values.yaml
luismacosta: removed

helm/values.yaml
luismacosta: sure

helm/templates/configmap-bootstrap.yaml
luismacosta: well, it depends.
in that case user needs a customised postgres/mysql image.
or, using these images, for instance, in initContainer first install curl + get sql scripts and then run psql/mysql commands.

helm/docker-image/templates/gateway-ha-config.yml
```diff
@@ -0,0 +1,64 @@
...
+  name: trinoRouter
+  historySize: {{ default .Env.QUERY_HISTORY_SIZE "1000" }}
+  requestBufferSize: 8192
+
+dataStore:
+  jdbcUrl: {{ default .Env.JDBC_URL "" }}
+  user: {{ default .Env.DB_USER "" }}
+  password: {{ default .Env.DB_PASSWORD "" }}
+  driver: {{ default .Env.JDBC_DRIVER "" }}
+  queryHistoryHoursRetention: 24
```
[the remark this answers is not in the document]
luismacosta: done

helm/templates/configmap-bootstrap.yaml
Chaho12: I see. How about adding such explanation to readme? so that user can customize to fit their needs

luismacosta: README updated

luismacosta:
Hello @Chaho12 

The requested changes have been done
Helm chart successfully deployed in eks cluster v1.27 with RDS Postgresql 16.1

Chaho12:
Have you signed & sent the CLA?

Chaho12:
@vishalya @willmostly Can you guys also take a look on this?

rdsarvar (commented):
did a really quick skim, will try to give it a test if I can find some free time

willmostly (changes requested):
In addition to the comments, I believe the chart should include a way to configure arbitrary parameters. Otherwise we'll need to ensure this chart provides full coverage for all the various config options the Gateway supports.

willmostly:
@luismacosta I see a number of my comments are marked outdated. Could you please squash your commits?

helm/docker-image/Dockerfile
willmostly: What is the reason for this change? Usually its considered more secure to run as UID 10001 or greater

mosabua: The file is already removed @willmostly

mosabua (commented):
I have reviewed in detail but just a first initial input. Can we cut down the complexity and get rid of having all these separate values in values.yaml that just mirror config properties that go into the config files. I think it would be much simpler and more flexible if we just have a minimum amount of values that are basically the content of the needed config files.. similar to the config files in the Trino helm chart but even cleaner and less overhead..

And maybe even have no defaults in the chart that go beyond what the docker container already has.

willmostly:
> Can we cut down the complexity and get rid of having all these separate values in values.yaml that just mirror config properties that go into the config files. I think it would be much simpler and more flexible if we just have a minimum amount of values that are basically the content of the needed config files.. similar to the config files in the Trino helm chart but even cleaner and less overhead..
> 
> And maybe even have no defaults in the chart that go beyond what the docker container already has.

I second this approach. Providing the gateway configuration as a configmap seems like a nice clean approach, possibly with a mechanism for interpolating usernames and passwords from secrets.

alaturqua:
Suggestion: Please consider including the following additions:
- Labels: Some Kubernetes setups require defining labels for successful deployment.
- Annotations: Annotations provide valuable metadata and enhance compatibility with various Kubernetes configurations.

AlonHarmon:
Suggestion:
add templates of statefullset + service for PostgreSQL (that can be disabled)

Then we will have all in one chart without external deps (except the trino clusters)

mosabua:
Replaced by https://github.com/trinodb/trino-gateway/pull/323

Michalosu:
@mosabua, @willmostly 

it looks like #323 it is not fully covering scope of this PR. There is no support for bootstrapping DB. Perhaps, initContainer is not best solution for DB bootstrap, imho better to have k8s Job with pre-install, pre-upgrade Helm hook and also I would consider to have dependency to psql/mysql charts in case someone wants to quickly bootstrap some PoC on k8s (configured via values). With current state of Helm chart it requires a few manual steps to get Trino gateway working. Install psql/mysql and run manually SQL bootstrap.

Does this project support (ORM/sql module) DB migrations?

mosabua:
We kept the Helm chart simple on purpose and therefore did not include the db. We have docker compose and other setups for that but in production this will vary widely. We might do more depending on community contributions and what we observe from using the chart.

Also we dont have any db migration at this stage but discussed wanting to add this in the future potentially. Current the schema is easy and if needed we add info to the release notes and docs.

mosabua:
Feel free to create separate issues or PR if you want to discuss more. Also find us on trino slack

PR #246, opened 2024-02-13
Remove unused classes and file

ebyhr:


mosabua:
Doesnt that remove actual functionality of using such a rule as in the example @willmostly @vishalya @Chaho12 ... 

I think we probably want this kind of feature in the pluggable router modules you are working on @vishalya right?

willmostly:
> Doesnt that remove actual functionality of using such a rule as in the example @willmostly @vishalya @Chaho12 ...

The test doesn't accurately reflect current functionality. In particular, the test adds this connection checker class to the `request` as an attribute, which makes it available for use in a rule. There are no implementations of `RoutingGroupSelector.findRoutingGroup` that do the same however.
```
HttpServletRequest mockRequest = mock(HttpServletRequest.class);
when(mockRequest.getAttribute("connectionChecker")).thenReturn(checker);
``` 
The proper way to add this would be [as a fact](https://github.com/trinodb/trino-gateway/blob/301c87d6b17a23609d44caa44429fbd64c368e43/gateway-ha/src/main/java/io/trino/gateway/ha/router/RuleReloadingRoutingGroupSelector.java#L90) to the rules engine instead of adding it to the `request`.

@vishalya I don't see this code in the old lyft gateway, but it was in the initial commit. Were you guys leveraging it at some point?

mosabua (approved):
Thanks @willmostly for confirming that it was needed / on our horizon but we can kill it off for now.

PR #247, opened 2024-02-13
Update quickstart doc

Chaho12:
- Udate quickstart doc to version ~5~ 6 with db checking code
- add var to gitignore in all directory hierarchy (for local dev) like trino https://github.com/trinodb/trino/commit/45bf374d6139406a83472ccd9314381a3280d955

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -40,14 +40,16 @@ Gateway server running in the host operating system.
 ## Start Trino Gateway server
 
 The following script starts a Trino Gateway server using the 
-[Quickstart Configuration](quickstart-config.yaml) with the request service
+[Quickstart configuration](quickstart-config.yaml) with the request service
 at http://localhost:9080, the application service at http://localhost:9081,
 and the Admin service at http://localhost:9082. It also starts a dockerized
 PostgreSQL database at localhost:5432.
 
 To start the server, copy the script below to a temporary directory 
 under the project root folder, and run it at the temporary directory.
-It will copy necessary files to current directory, which consists of the following:
+
+It  copies the following, necessary files to current directory:
+
 - gateway-ha.jar
 - gateway-ha-persistence-postgres.sql
 - quickstart-config.yaml
@@ -57,7 +59,9 @@ It will copy necessary files to current directory, which consists of the followi
 
 VERSION=7
 
-#Check and get the Gateway Jar
+# Copy necessary files to current directory
+
+# Check and get the Gateway Jar
 if [[ -f "gateway-ha.jar" ]]; then
     echo "Found gateway-har.jar file in current directory."
 else
@@ -65,9 +69,19 @@ else
     curl https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha/${VERSION}/gateway-ha-${VERSION}-jar-with-dependencies.jar -o ./gateway-ha.jar
 fi
 
-# Copy necessary files to current directory
-cp ../docs/quickstart-config.yaml ./quickstart-config.yaml
-cp ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql ./gateway-ha-persistence-postgres.sql
+# Check and get the Config.yaml
+if [[ -f "quickstart-config.yaml" ]]; then
+    echo "Found quickstart-config.yaml file in current directory."
+else
+    cp ../docs/quickstart-config.yaml ./quickstart-config.yaml
+fi
+
+# Check and get the postgres.sql
+if [[ -f "gateway-ha-persistence-postgres.sql" ]]; then
+    echo "Found gateway-ha-persistence-postgres.sql file in current directory."
+else
+    cp ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql ./gateway-ha-persistence-postgres.sql
+fi
 
 #Check if DB is running
 if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
@@ -87,7 +101,6 @@ fi
 
 #Start Trino Gateway server.
 java -Xmx1g --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar ./gateway-ha.jar server ./quickstart-config.yaml
-
 ```
 
 You can clean up by running
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -35,9 +35,7 @@
 
 The scripts from this quickstart guide set up a local environment consisting of
 two Trino servers and a PostgreSQL database running in Docker, and a Trino
-Gateway server running in the host operating system. If you are using a Mac and
-have the `brew` package manager installed, this script attempts to install
-`psql`, if not you must install `psql` through other means.
+Gateway server running in the host operating system. 
 
 ## Start Trino Gateway server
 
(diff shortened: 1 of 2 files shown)


docs/quickstart.md
```diff
@@ -38,6 +38,8 @@ two Trino servers and a PostgreSQL database running in Docker, and a Trino
 Gateway server running in the host operating system. If you are using a Mac and
 have the `brew` package manager installed, this script attempts to install
 `psql`, if not you must install `psql` through other means.
+- If you use `brew` package, you might have to set libpq first in your PATH, run:
+- `echo 'export PATH="/opt/homebrew/opt/libpq/bin:$PATH"' >> ~/.zshrc`
```
mosabua, line 42: This assume usage of zsh ... dont do that. Just say what folder has to be added to the PATH. Something like 

Use the following command in your shell startup script or manually: 

```
export PATH="/opt/homebrew/opt/libpq/bin:$PATH"
```

docs/quickstart.md
```diff
@@ -65,15 +67,20 @@ if ! $(command -v psql >/dev/null); then
...
-#make sure the DB has time to initialize
-sleep 5
-
-#Initialize the DB
-curl https://raw.githubusercontent.com/trinodb/trino-gateway/main/gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql > ./gateway-ha-persistence-postgres.sql 
-psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
-psql  -U postgres -h localhost -d gateway -f ./gateway-ha-persistence-postgres.sql 
+#Check if DB is running
+if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
+    echo "Postgres DB 'localhost-postgre' container is already running, so only run gateway server."
```
mosabua, line 72:
```suggestion
    echo "PostgreSQL database container 'localhost-postgre' is already running. Only starting Trino Gateway."
```

Chaho12: Fixed :)

docs/quickstart.md
```diff
@@ -37,7 +37,10 @@ The scripts from this quickstart guide set up a local environment consisting of
 two Trino servers and a PostgreSQL database running in Docker, and a Trino
 Gateway server running in the host operating system. If you are using a Mac and
 have the `brew` package manager installed, this script attempts to install
-`psql`, if not you must install `psql` through other means.
+`psql`, if not you must install `psql` through other means. 
+
+Use the following command in your shell startup script or manually:
+- `export PATH="/opt/homebrew/opt/libpq/bin:$PATH"`
```
ebyhr, line 43: This section should be rewritten to use psql in a container instead of local psql. Suggesting brew install doesn't make sense as it already exists in the container. 

Also, we already have TrinoGatewayRunner that initialize the database automatically. I feel this section is increasing managing costs with less benefits.

ebyhr (changes requested):

docs/quickstart.md
Chaho12: I think this section is for users to play with gateway without any mvn builds, just use public gateway image.


> This section should be rewritten to use psql in a container instead of local psql. Suggesting brew install doesn't make sense as it already exists in the container.

Now that you point out, we can change this by running psql within container. What do you think?
```bash
# before
psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
psql  -U postgres -h localhost -d gateway -f gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql
# after
docker exec local-postgres psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
docker cp gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql local-postgres:/tmp
docker exec local-postgres psql -U postgres -h localhost -d gateway -f /tmp/gateway-ha-persistence-postgres.sql

```

Chaho12: @willmostly any thoughts?

ebyhr: > I think this section is for users to play with gateway without any mvn builds, just use public gateway image with postgre image

Mentioning docker compose yml of `docker` directory should be sufficient. We should avoid managing duplicated information.

mosabua:
Just to clarify .. I think we will need to clearly separate the different usage aspects: 

* developer vs platform admin tasks
* using jar vs docker vs helm

Currently its all mixed about and thats the root cause of the problem. Like I mentioned in the last dev sync I am hoping to get a doc site going and then break up the docs around all these aspects. With that in mind .. lets keep this PR simple so we can merge and do follow ups as needed.

Chaho12:
As I moved psql command to be executed within postgresql docker, user does now does not need to have homebrew or libpq.
So I cleaned up bash script for easier use.

docs/quickstart.md
```diff
@@ -50,30 +48,31 @@ PostgreSQL database at localhost:5432.
...
+else
+    echo "PostgreSQL database container 'localhost-postgre' is not running, initialize and run database"
+    export PGPASSWORD=mysecretpassword
+    docker run --name local-postgres -p 5432:5432 -e POSTGRES_PASSWORD=$PGPASSWORD -d postgres:latest
+    #Make sure the DB has time to initialize
+    sleep 5
+
+    #Initialize the DB
+    docker exec local-postgres psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
+    docker cp ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql local-postgres:/tmp
```
willmostly, line 73: could you start the container with `docker run -v ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql:/tmp/gateway-ha-persistence-postgres.sql ...` instead of copying?

willmostly (approved):
Looks good, modulo @mosabua 's comments on overall organization.

docs/quickstart.md
willmostly: This predated the gateway docker image, so its worthwhile to consider if a play environment that uses the bare jars is still worthwhile to maintain. However as @mosabua says below we should evaluate this as part of a strategic reorganization.

ebyhr: This command downloads jar from maven central. It means no-difference from using docker container. 
I still disagree with having this commands in quickstart.

mosabua: We have to pull this apart for quick start with docker and quick start with jar as soon as possible. However that does not necessarily have to be part of this PR.

docs/quickstart.md
Chaho12: changed to use mount volume command :)

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
-#make sure the DB has time to initialize
-sleep 5
-
-#Initialize the DB
-curl https://raw.githubusercontent.com/trinodb/trino-gateway/main/gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql > ./gateway-ha-persistence-postgres.sql 
-psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
-psql  -U postgres -h localhost -d gateway -f ./gateway-ha-persistence-postgres.sql 
+#Check if DB is running
+if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
+    echo "PostgreSQL database container 'localhost-postgre' is already running. Only starting Trino Gateway."
```
mosabua, line 63: localhost-postgres ?

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
-if ! $(command -v psql >/dev/null); then
-    if ! $(command -v brew >/dev/null); then
-        echo 'Please install psql and rerun';
-        exit 1
-    else
-        brew install libpq
-    fi
+#Check and get the Gateway Jar
+if [[ -f "gateway-ha.jar" ]]; then
+    echo "Trino Gateway Jar already exists so skip fetching."
```
mosabua, line 55: Maybe better to say

Found gateway-har.jar file in current directory.

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
+else
+    echo "Trino Gateway Jar does not exist, try to fetch from maven repository."
```
mosabua, line 57: Failed to find gateway-ha.jar in current directory. Fetching version $VERSION from Maven Central repository.

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
-
-#Initialize the DB
-curl https://raw.githubusercontent.com/trinodb/trino-gateway/main/gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql > ./gateway-ha-persistence-postgres.sql 
-psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
-psql  -U postgres -h localhost -d gateway -f ./gateway-ha-persistence-postgres.sql 
+#Check if DB is running
+if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
+    echo "PostgreSQL database container 'localhost-postgre' is already running. Only starting Trino Gateway."
+else
+    echo "PostgreSQL database container 'localhost-postgre' is not running, initialize and run database"
```
mosabua, line 65:
```suggestion
    echo "PostgreSQL database container 'localhost-postgre' is not running. Proceeding to initialize and run database server."
```

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
+    #Make sure the DB has time to initialize
+    sleep 5
+
+    #Initialize the DB
+    docker exec local-postgres psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
+    docker exec local-postgres psql -U postgres -h localhost -d gateway -f /tmp/gateway-ha-persistence-postgres.sql
+fi
 
-#Start the DB
+#Start the Gateway server. Please check #Build and run in development doc if you don't have jar file built yet.
```
mosabua, line 76:
```suggestion
#Start Trino Gateway server.
```

mosabua (changes requested):

Chaho12:
> Currently its all mixed about and thats the root cause of the problem. Like I mentioned in the last dev sync I am hoping to get a doc site going and then break up the docs around all these aspects. With that in mind .. lets keep this PR simple so we can merge and do follow ups as needed.

rebased and fixed wordings

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
-#Start the persistence DB
-if ! $(command -v psql >/dev/null); then
-    if ! $(command -v brew >/dev/null); then
-        echo 'Please install psql and rerun';
-        exit 1
-    else
-        brew install libpq
-    fi
+#Check and get the Gateway Jar
+if [[ -f "gateway-ha.jar" ]]; then
```
mosabua, line 54: this checks in the current folder of the script

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
-curl https://raw.githubusercontent.com/trinodb/trino-gateway/main/gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql > ./gateway-ha-persistence-postgres.sql 
-psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
-psql  -U postgres -h localhost -d gateway -f ./gateway-ha-persistence-postgres.sql 
+#Check if DB is running
+if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
+    echo "PostgreSQL database container 'localhost-postgres' is already running. Only starting Trino Gateway."
+else
+    echo "PostgreSQL database container 'localhost-postgres' is not running. Proceeding to initialize and run database server."
+    export PGPASSWORD=mysecretpassword
+    docker run -v ./gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql:/tmp/gateway-ha-persistence-postgres.sql --name local-postgres -p 5432:5432 -e POSTGRES_PASSWORD=$PGPASSWORD -d postgres:latest
```
mosabua, line 67: this assumes the project checkout root folder `trino-gateway` is the location of the quickstart script

docs/quickstart.md
```diff
@@ -50,32 +48,32 @@ PostgreSQL database at localhost:5432.
...
+    sleep 5
+
+    #Initialize the DB
+    docker exec local-postgres psql -U postgres -h localhost -c 'CREATE DATABASE gateway'
+    docker exec local-postgres psql -U postgres -h localhost -d gateway -f /tmp/gateway-ha-persistence-postgres.sql
+fi
 
-#Start the DB
+#Start Trino Gateway server.
 java -Xmx1g --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar ./gateway-ha.jar server ./quickstart-config.yaml
```
mosabua, line 77: this assumes the script is run in the docs folder .. or the config is copied

mosabua (changes requested):
I tried this now and as it stands the script does not work. Or at least it is not clear to me where the script should be created and run from. See my comments .. 

We should add the info on what files are needed where and then make it work for that setup. 

In the longer run we have to separate dev, testing, and productions setup docs .. but for now this is fine .. as long as it actually works and can be followed. Please work together @willmostly and @Chaho12 to figure out what direction you want to take in terms of specified folder. Maybe it would be even cleaner if we have a list of files to copy into a random new folder.

mosabua (approved):
Looks good now. Thanks for addressing the comments.

PR #258, opened 2024-02-17
Import String.format method statically

ebyhr:


gateway-ha/src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java
```diff
@@ -111,7 +110,7 @@ public Response getAuthorizationCode()
         String clientId = oauthConfig.getClientId();
         String redirectUrl = oauthConfig.getRedirectUrl();
         String scopes = String.join("+", oauthConfig.getScopes());
-        String url = String.format(
+        String url = format(
                 "%s?client_id=%s&response_type=code&redirect_uri=%s&scope=%s",
```
willmostly, line 114: Should these be unwrapped? Line length should be 146 characters, which is less than [RIGHT_MARGIN](https://github.com/airlift/codestyle/blob/3fbde21e2761e7d985c55bb3ec53b5da31eac459/IntelliJIdea2019/Airlift.xml#L19)

willmostly (approved):
LGTM. I suggested one additional place to unwrap lines, but feel free to ignore it if it makes the code less readable.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java
ebyhr: Yes, it should be, but I didn't touch as unrelated to import.

PR #259, opened 2024-02-17
Run docker build conditionally on CI

ebyhr:


mosabua:
What is the reason for doing this? Do we really have such a busy pipeline that we need to exclude this step? Or is there some other reason I am missing?

mosabua:
Ping @ebyhr

PR #262, opened 2024-02-17
Remove unused variable

ebyhr:


mosabua (approved):
Looks good .. but it sure does make me wonder why this was added in the first place.

PR #265, opened 2024-02-27
Fix query id extraction from request query

willmostly:
The logic for extracting a query id from the request query was broken. The test for this functionality was also broken, since it passed the `query` to the extraction method as part of the `path`.

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
@@ -60,7 +60,19 @@ public class QueryIdCachingProxyHandler
     public static final String SOURCE_HEADER = "X-Trino-Source";
     public static final String HOST_HEADER = "Host";
     private static final int QUERY_TEXT_LENGTH_FOR_HISTORY = 200;
-    private static final Pattern QUERY_ID_PATTERN = Pattern.compile(".*[/=?](\\d+_\\d+_\\d+_\\w+).*");
+    /**
+     * This regular expression matches query ids as they appear in the path of a URL. The query id must be preceded
+     * by a "/". A query id is defined as three groups of digits separated by underscores, with a final group
+     * consisting of any alphanumeric characters.
+     */
+    private static final Pattern QUERY_ID_PATH_PATTERN = Pattern.compile(".*/(\\d+_\\d+_\\d+_\\w+).*");
+    /**
+     * This regular expression matches query ids as they appear in the query parameters of a URL. The query id is
+     * defined as in QUERY_TEXT_LENGTH_FOR_HISTORY. The query id must either be at the beginning of the query parameter
+     * string, or be preceded by %2F (a URL-encoded "/"), or  "query_id=", with or without the underscore and any
+     * capitalization.
+     */
+    private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
 
     private static final Pattern EXTRACT_BETWEEN_SINGLE_QUOTES = Pattern.compile("'([^\\s']+)'");
 
@@ -113,7 +125,13 @@ public class QueryIdCachingProxyHandler
             }
         }
         else if (path.startsWith(TRINO_UI_PATH)) {
-            Matcher matcher = QUERY_ID_PATTERN.matcher(path);
+            Matcher matcher = QUERY_ID_PATH_PATTERN.matcher(path);
+            if (matcher.matches()) {
+                queryId = matcher.group(1);
+            }
+        }
+        if (!isNullOrEmpty(queryParams)) {
+            Matcher matcher = QUERY_ID_PARAM_PATTERN.matcher(queryParams);
             if (matcher.matches()) {
                 queryId = matcher.group(1);
             }
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
@@ -23,6 +23,7 @@ import org.mockito.Mockito;
 
 import java.io.IOException;
 
+import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.extractQueryIdIfPresent;
 import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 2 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -33,20 +34,20 @@ public class TestQueryIdCachingProxyHandler
...
-                "/ui/api/query/20200416_160256_03078_6b4yt?bogus_fictional_param",
-                "/ui/api/query?query_id=20200416_160256_03078_6b4yt",
-                "/ui/api/query.html?20200416_160256_03078_6b4yt"};
-        for (String path : paths) {
-            String queryId = QueryIdCachingProxyHandler.extractQueryIdIfPresent(path, null);
+        URI[] uris = {
+                URI.create("http://example.com/ui/api/query/20200416_160256_03078_6b4yt"),
+                URI.create("http://example.com/ui/api/query/20200416_160256_03078_6b4yt?bogus_fictional_param"),
+                URI.create("http://example.com/ui/api/query?query_id=20200416_160256_03078_6b4yt"),
+                URI.create("http://example.com/ui/api/query.html?20200416_160256_03078_6b4yt")};
```
mosabua, line 41: also .. will the path really always be ui/api/query .. and we depend on that.. if not .. should we add some other valid ones?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -61,6 +61,7 @@ public class QueryIdCachingProxyHandler
     public static final String HOST_HEADER = "Host";
     private static final int QUERY_TEXT_LENGTH_FOR_HISTORY = 200;
     private static final Pattern QUERY_ID_PATTERN = Pattern.compile(".*[/=?](\\d+_\\d+_\\d+_\\w+).*");
+    private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|=|^)(\\d+_\\d+_\\d+_\\w+).*");
```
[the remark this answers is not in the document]
vishalya: I am curious if trino also uses the same pattern.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
vishalya: a few other cases with `https` and `www.example.com` would be good.

vishalya (changes requested):
Overall looks good, apart from minor comments.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -118,6 +119,12 @@ else if (path.startsWith(TRINO_UI_PATH)) {
                 queryId = matcher.group(1);
             }
         }
+        if (!isNullOrEmpty(queryParams)) {
+            Matcher matcher = QUERY_ID_PARAM_PATTERN.matcher(queryParams);
+            if (matcher.matches()) {
+                queryId = matcher.group(1);
+            }
+        }
```
Chaho12, line 127: Assuming that queryparam comes after `/ui` only, 
this code can be simplified with changing line 117 (`Matcher matcher = QUERY_ID_PATTERN.matcher(path);`) with below. No need for new regex pattern, which is very similiar to `QUERY_ID_PATTERN`.
I've tested using QUERY_ID_PATTERN matcher on your test cases and it passes fine
```java
Matcher matcher = QUERY_ID_PATTERN.matcher(
                    path + (isNullOrEmpty(queryParams) ? "" : "?" + queryParams));

File: gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java (hunk 1 above)
Comment (Chaho12) [reply gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java:41]: How about adding other tests for these apis without queryid?
- ui/api/cluster
- ui/api/query
- ui/api/stats

File: gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java (hunk 2 above)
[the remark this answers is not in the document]
Comment (willmostly) [reply gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java:64]: Trino uses a [lightweight test](https://github.com/trinodb/trino/blob/0fd6170841bbdcefb2a1636d50c3f17f3aad719d/core/trino-spi/src/main/java/io/trino/spi/QueryId.java#L81) that only checks for characters in the set `[_a-z0-9]`. A queryId could technically be something like `"testqueryid"`, as you see in some tests, but the [generator imposes a specific format](https://github.com/trinodb/trino/blob/f83854687229df1ae3325249e7bcd29f03d0943c/core/trino-main/src/main/java/io/trino/execution/QueryIdGenerator.java#L120)

These patterns impose additional restrictions on where a queryid can appear to reduce false positives.

File: gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java [hunk 4]
```diff
@@ -33,20 +34,20 @@ public class TestQueryIdCachingProxyHandler
...
             throws IOException
     {
-        String[] paths = {
-                "/ui/api/query/20200416_160256_03078_6b4yt",
-                "/ui/api/query/20200416_160256_03078_6b4yt?bogus_fictional_param",
-                "/ui/api/query?query_id=20200416_160256_03078_6b4yt",
-                "/ui/api/query.html?20200416_160256_03078_6b4yt"};
-        for (String path : paths) {
-            String queryId = QueryIdCachingProxyHandler.extractQueryIdIfPresent(path, null);
+        URI[] uris = {
```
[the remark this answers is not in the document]
willmostly: the `protocol://host` part of the URI is not passed on to the method being tested, so I'll remove it to avoid confusion

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
willmostly: @Chaho12 are these negative tests to add under `nonMatchingUris`?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
willmostly: added a test case for redirects to illustrate a case where the queryid may not follow `/ui`. I will narrow the scope of the path matcher since it includes some patterns that will not appear in a path.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
Chaho12: I was thinking more about how it will route in such cases, then i just realized that test only checks if it extracts queryid or not.
so sure, adding it at nonMatchingUris is good (but now i don't think it's much meaningful test case though)

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -33,20 +34,28 @@ public class TestQueryIdCachingProxyHandler
...
+                URI.create("/v1/statement/queued/20200416_160256_03078_6b4yt/y0d7620a6941e78d3950798a1085383234258a566/1"),
+                URI.create("/ui/api/query/20200416_160256_03078_6b4yt"),
+                URI.create("/ui/api/query/20200416_160256_03078_6b4yt/killed"),
+                URI.create("/ui/api/query/20200416_160256_03078_6b4yt/preempted"),
+                URI.create("/v1/query/20200416_160256_03078_6b4yt?pretty"),
+                URI.create("/ui/troubleshooting?queryId=20200416_160256_03078_6b4yt"),
+                URI.create("/ui/query.html?20200416_160256_03078_6b4yt"),
+                URI.create("/login?redirect=%2Fui%2Fapi%2Fquery%2F20200416_160256_03078_6b4yt")};
+
+        for (URI uri : uris) {
```
ebyhr, line 48: I would recommend removing for-loop in a preparatory commit: 
```java
        assertThat(extractQueryIdIfPresent("/ui/api/query/20200416_160256_03078_6b4yt", null))
                .isEqualTo("20200416_160256_03078_6b4yt");
        assertThat(extractQueryIdIfPresent("/ui/api/query/20200416_160256_03078_6b4yt?bogus_fictional_param", null))
                .isEqualTo("20200416_160256_03078_6b4yt");
...
```

Same for the following `nonMatchingUris` loop.

willmostly: @ebyhr I read this quickly and made the suggested change in this PR, were you suggesting to do it separately?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -60,7 +60,8 @@ public class QueryIdCachingProxyHandler
     public static final String SOURCE_HEADER = "X-Trino-Source";
     public static final String HOST_HEADER = "Host";
     private static final int QUERY_TEXT_LENGTH_FOR_HISTORY = 200;
-    private static final Pattern QUERY_ID_PATTERN = Pattern.compile(".*[/=?](\\d+_\\d+_\\d+_\\w+).*");
+    private static final Pattern QUERY_ID_PATH_PATTERN = Pattern.compile(".*/(\\d+_\\d+_\\d+_\\w+).*");
+    private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
```
mosabua, line 64: oh how I hate regex like this .. should we somehow add some Java doc that explains it a little bit .. or do we just treat the test as the docs?

mosabua (approved):
Looks good even if the regex is really not easily readable. Feel free to add some javadoc if you think it helps .. but otherwise also feel free to ship as it is now

gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java
```diff
@@ -113,7 +114,13 @@ protected static String extractQueryIdIfPresent(String path, String queryParams)
...
         }
         else if (path.startsWith(TRINO_UI_PATH)) {
-            Matcher matcher = QUERY_ID_PATTERN.matcher(path);
+            Matcher matcher = QUERY_ID_PATH_PATTERN.matcher(path);
+            if (matcher.matches()) {
+                queryId = matcher.group(1);
+            }
+        }
+        if (!isNullOrEmpty(queryParams)) {
+            Matcher matcher = QUERY_ID_PARAM_PATTERN.matcher(queryParams);
```
ebyhr, line 123: Is it possible to replace `HttpServletRequest#getQueryString` at L199 with  `ServletRequest#getParameter` to avoid regexp?

willmostly: The cases we would need to handle are
* the query details page (tested for [here](https://github.com/trinodb/trino-gateway/pull/265/commits/c89ea0d2d78119aa1eca7e071ba8d0da03b248a9#diff-131197136faa3df44808d73f7ba20ef93e4f8b5ada84d3b8ee21bcd1246dab42R51)), e.g. `/ui/query.html?20200416_160256_03078_6b4yt` 
* redirects, represented by [this case](https://github.com/trinodb/trino-gateway/pull/265/commits/c89ea0d2d78119aa1eca7e071ba8d0da03b248a9#diff-131197136faa3df44808d73f7ba20ef93e4f8b5ada84d3b8ee21bcd1246dab42R53), e.g. `/login?redirect=%2Fui%2Fapi%2Fquery%2F20200416_160256_03078_6b4yt`
* `queryId=foo` and variants like `query_id=foo`. I don't think Trino has this, but Starburst does and I wouldn't be surprise if other extended versions of Trino do as well.

You can special case each of these, e.g. for `/ui/query.html` always get the first parameter. For the redirect parameters, you could url decode it and pass it to `extractQueryIdIfPresent` as the path. 

@ebyhr & @mosabua - do you think the extra complexity is worth it to avoid the regex?

mosabua: As discussed.. no .. lets stick with the regex pattern and your additional docs.

mosabua:
@Chaho12 also confirmed in the dev sync on the 6th of March that this PR is good to go.

mosabua (approved):
Thanks for adding the comments.

PR #266, opened 2024-02-27
allow customization of test containers image name

andythsu:
Problem: Some tests download trino docker images from public dockerhub. For example: https://github.com/trinodb/trino-gateway/blob/e2aaa0e9c71ae69767f842a3f4834f9d531d6620/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java#L55
However, this will fail when running behind a proxy. Our in-house images have different names from the names on dockerhub.  Therefore, we need a way to override `trinodb/trino` to a completely different name when necessary. This will eventually expand to other public images as they are added to the codebase.

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway;
+
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.ImageNameSubstitutor;
+
+/**
+ * We can use this class to override the images used by test containers, if the image has
+ * a different name from dockerhub. This class is used in testcontainers.properties file.
+ * TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE can be set as an environment variable.
+ */
+public class CustomTrinoImageNameSubstitutor
+        extends ImageNameSubstitutor
+{
+    private static final String TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE = "TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE";
+    private static final String TRINO_IMAGE = "trinodb/trino";
+
+    @Override
+    public DockerImageName apply(DockerImageName dockerImageName)
+    {
+        String trinoImageSubstitute = System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE);
+        if (dockerImageName.getUnversionedPart().equals(TRINO_IMAGE) && trinoImageSubstitute != null) {
+            return DockerImageName.parse(trinoImageSubstitute).asCompatibleSubstituteFor(TRINO_IMAGE);
+        }
+        return dockerImageName;
+    }
+
+    @Override
+    protected String getDescription()
+    {
+        return "custom Trino image name substitutor";
+    }
+}
--- a/gateway-ha/src/test/resources/testcontainers.properties
+++ b/gateway-ha/src/test/resources/testcontainers.properties
@@ -1,0 +1,1 @@
+image.substitutor=io.trino.gateway.CustomTrinoImageNameSubstitutor


gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,38 @@
+/*
```
ebyhr, line 1: > allow customization of test containers image name

Please capitalize the commit title. https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#format-git-commit-messages

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,38 @@
...
+package io.trino.gateway;
+
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.ImageNameSubstitutor;
+
+public class CustomImageNameSubstitutor
+        extends ImageNameSubstitutor
+{
+    private final String TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE = "TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE";
+    @Override
```
ebyhr: Add an empty line.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package io.trino.gateway;
+
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.ImageNameSubstitutor;
+
+public class CustomImageNameSubstitutor
+        extends ImageNameSubstitutor
+{
+    private static final String TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE = "TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE";
```
ebyhr, line 22: Rename to `TESTCONTAINER_DOCKER_PULL_THROUGH_REGISTRY`.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public DockerImageName apply(DockerImageName dockerImageName)
+    {
+        if (dockerImageName.asCanonicalNameString().equals("trinodb/trino:latest") && System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE) != null) {
```
ebyhr, line 27: Remove the 1st condition.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,39 @@
...
+            String imageSubstitute = System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE);
+            return DockerImageName.parse(imageSubstitute).asCompatibleSubstituteFor("trinodb/trino");
```
ebyhr, line 29: Please avoid hard-coding the image name. We will add other docker images soon.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+        return dockerImageName;
+    }
+
+    @Override
+    protected String getDescription()
+    {
+        return "custom image name substitutor";
+    }
+}
```
ebyhr, line 39: `testcontainers.properties` isn't required?

andythsu: i was going to use env variables to configure this. https://java.testcontainers.org/features/configuration/
It looks like we can just choose one

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
andythsu: first condition makes sure we are overriding a specific image. If we don't specify it will just substitute all images

ebyhr: See https://github.com/trinodb/trino-gateway/pull/266#discussion_r1505078523

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
ebyhr: If we use env variables, why this class is required? I'm looking the following sentence:

> Setting environment variables TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX=registry.mycompany.com/mirror/
https://java.testcontainers.org/features/image_name_substitution/#automatically-modifying-docker-hub-image-names

andythsu: this may not work because it needs image name to be predictable

andythsu: I was looking at this https://java.testcontainers.org/features/image_name_substitution/#developing-a-custom-function-for-transforming-image-names-on-the-fly

ebyhr: > it needs image name to be predictable

I don't understand this. Can you elaborate on that?

andythsu: > If we use env variables, why this class is required? I'm looking the following sentence:
> 
> > Setting environment variables TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX=registry.mycompany.com/mirror/
> > https://java.testcontainers.org/features/image_name_substitution/#automatically-modifying-docker-hub-image-names

The link you are looking at is specific to add prefix to image name. It has an assumption:

>Consider this if:
>>Your private registry has copies of images from Docker Hub where the names are predictable, and just adding a prefix is enough. For example, registry.mycompany.com/mirror/mysql:8.0.36 can be derived from the original Docker Hub image name (mysql:8.0.36) with a consistent prefix string: registry.mycompany.com/mirror/

Also, it looks like even though we can change the registry, the names of the image have to match. In our case, we have different image names from the dockerhub.

ebyhr: Thanks for your explanation. 

> we have different image names from the dockerhub.

This is the important fact I didn't know. Please leave the code comment.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway;
+
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.ImageNameSubstitutor;
+
+public class CustomImageNameSubstitutor
```
ebyhr, line 19: The class name is misleading. Please rename to `CustomTrinoImageNameSubstitutor`.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
ebyhr: I understand you want to change only Trino image names. I suppose we should tweak the condition to exclude the version.

andythsu: we can probably use this `getUnversionedPart()` from https://javadoc.io/doc/org.testcontainers/testcontainers/1.15.1/org/testcontainers/utility/DockerImageName.html#getUnversionedPart--

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
andythsu: From [the doc](https://java.testcontainers.org/features/image_name_substitution/#developing-a-custom-function-for-transforming-image-names-on-the-fly) I suppose we can only have one ImageNameSubstitutor, then we use `TESTCONTAINERS_IMAGE_SUBSTITUTOR=com.mycompany.testcontainers.ExampleImageNameSubstitutor` to tell testcontainer to use this class. Therefore, I think this class should have a generic name as this is the only substitutor that testcontainer will look at.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
andythsu: added.

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
ebyhr: I still disagree with the current name.

Chaho12: If the class CustomImageNameSubstitutor is used for other images, `CustomImageNameSubstitutor` seems right but since this is only applied to trino image, I think it's better off with what ebyhr saids. We can change the name to a more general one if there are other images to substitute in future

andythsu: > I still disagree with the current name. Even CustomImageNameSubstitutor doesn't explain whether TESTCONTAINERS_IMAGE_SUBSTITUTOR supports single substitutor or several substitutors either.
> 
> I don't understand your concern to be honest. There's only one class extending ImageNameSubstitutor in this repository. Who will try setting several substitutors?

The concern is we can only have one ImageSubstitutor. Naming it to `CustomTrinoImageNameSubstitutor` is misleading because it might lead to thinking it's only for substituting Trino image. However, this class should be flexible to substitute other images as well. I agree with @Chaho12's suggestion. We can change the name to be generic later as we need.

ebyhr:
> We probably want to add how to use this class to the documentation

Most people won't use this feature. Please leave javadoc in the class.

gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class CustomTrinoImageNameSubstitutor
+        extends ImageNameSubstitutor
+{
+    private static final String TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE = "TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE";
+
+    @Override
+    public DockerImageName apply(DockerImageName dockerImageName)
+    {
+        String trinoImage = "trinodb/trino";
```
ebyhr, line 27: Move to a constant.

gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
```diff
@@ -0,0 +1,40 @@
...
+            String imageSubstitute = System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE);
+            return DockerImageName.parse(imageSubstitute).asCompatibleSubstituteFor(trinoImage);
+        }
+        return dockerImageName;
+    }
+
+    @Override
+    protected String getDescription()
+    {
+        return "custom image name substitutor";
```
ebyhr, line 38:
```suggestion
        return "custom Trino image name substitutor";
```

gateway-ha/src/test/java/io/trino/gateway/CustomImageNameSubstitutor.java
ebyhr: > added.

I can't find it.

gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway;
+
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.ImageNameSubstitutor;
+
+public class CustomTrinoImageNameSubstitutor
```
wendigo, line 19: you should add this to a testcontainers.properties service loader file

gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
```diff
@@ -0,0 +1,40 @@
...
+        extends ImageNameSubstitutor
+{
+    private static final String TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE = "TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE";
+
+    @Override
+    public DockerImageName apply(DockerImageName dockerImageName)
+    {
+        String trinoImage = "trinodb/trino";
+        if (dockerImageName.getUnversionedPart().equals(trinoImage) && System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE) != null) {
```
wendigo, line 28: extract `System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE)` to a variable before if and then use inside if

gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
andythsu: added

wendigo:
```
0:16:09.097 [main] INFO org.testcontainers.utility.ImageNameSubstitutor -- Found configured ImageNameSubstitutor: custom Trino image name substitutor
20:16:09.098 [main] INFO org.testcontainers.utility.ImageNameSubstitutor -- Image name substitution will be performed by: Chained substitutor of 'DefaultImageNameSubstitutor (composite of 'ConfigurationFileImageNameSubstitutor' and 'PrefixingImageNameSubstitutor')' and then 'custom Trino image name substitutor'
```

gateway-ha/src/test/java/io/trino/gateway/CustomTrinoImageNameSubstitutor.java
```diff
@@ -0,0 +1,46 @@
...
+{
+    private static final String TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE = "TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE";
+    private static final String TRINO_IMAGE = "trinodb/trino";
+
+    @Override
+    public DockerImageName apply(DockerImageName dockerImageName)
+    {
+        String trinoImageSubstitute = System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE);
+        if (dockerImageName.getUnversionedPart().equals(TRINO_IMAGE) && trinoImageSubstitute != null) {
+            String imageSubstitute = System.getenv(TESTCONTAINERS_TRINO_IMAGE_SUBSTITUTE);
```
wendigo, line 35: redundant variable. You have it above

wendigo (changes requested):
Small change requested

wendigo (approved):

PR #267, opened 2024-02-29
Disallow JUnit assertions

ebyhr:


gateway-ha/pom.xml
```diff
@@ -568,6 +568,21 @@
                     </dependency>
                 </dependencies>
             </plugin>
+
+            <plugin>
```
mosabua, line 572: Cant we just inherit that from airbase instead or is that not defined there? Also we need to make sure this adds and does not override other rules.

mosabua: Also .. maybe this should go in pluginMgt in the parent

mosabua (commented):
In principal I think this is great. Just have to look at impl and refine a bit maybe.

gateway-ha/pom.xml
ebyhr: Airbase doesn't have this rule https://github.com/airlift/airbase/blob/f9cb1aafcbe1392d45414ddecab186468e7e573c/airbase/pom.xml#L688-L738

> Also we need to make sure this adds and does not override other rules.

I already confirmed it's append.

mosabua (approved):
Lets ship this, probably worth adding to trino some time

PR #270, opened 2024-03-05
Update airbase to 151

ebyhr:


mosabua:
Is this ok if we are not yet on Java 21?

mosabua:
@wendigo ? 

Also .. will that pull in a bunch of dependency updates (like Jetty 12??)

wendigo:
No. These come with Airlift, not Airbase @mosabua

wendigo (approved):

PR #273, opened 2024-03-05
Merge yml files into docker-compose.yml

ebyhr:
No need to separate yml file because they're not reused. We had to jump other files to confirm the service definition.


mosabua (approved):
Looks good. Thanks for also updating the docs.

PR #282, opened 2024-03-14
Add .editorconfig

Chaho12:
## Description

While I was reviewing code, i noticed trino has [.editorconfig](https://github.com/trinodb/trino/blob/master/.editorconfig) for coding convention.
Even if we are using [Airlift style](https://github.com/airlift/codestyle/blob/master/IntelliJIdea2019/Airlift.xml), i think we should also add editorconfig for same [reason](https://github.com/trinodb/trino/pull/2151#issuecomment-560019098)

https://github.com/trinodb/trino/pull/2151

## Release notes

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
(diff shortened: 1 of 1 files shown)


oneonestar:
LGTM

ebyhr (approved):

mosabua:
Cool.... over time we should update for markdown as well .. but that can come later

Issue #41, opened 2023-09-15
Adopt airlift as base framework


mosabua:
Currently the project uses an old version of Dropwizard. Airlift is a superior alternative that is actively maintained and also used in Trino itself. 

Rather than upgrading Dropwizard it might be better to move to Airlift and then take advantage of aspects such as easy tarball creation, config system, opentelemety support and much more. This is the preferred approach by @martint. 

Potential an initial upgrade of dropwizard to fix the worst issues might still be a good step.

mosabua:
This is done and part of version 5 onwards

mosabua:
Partially using airlift already .. but not as full base framework

oneonestar:
Preparation phase (splittable):
- [ ] Remove dropwizard.auth: #289 
- [ ] Remove dropwizard.configuration
- [ ] (Maybe) Refactor module injection

Migration phase (all-at-once):
- [ ] Remove dropwizard.*
- [ ] Update to Jetty 12.0.7 EE10 (Ref: https://github.com/airlift/airlift/pull/1089)
- [ ] com.codahale.metrics to opentelemetry
- [ ] Use airlift for bootstrap
- [ ] Migrate to airlift.configuration

Uncertain things:
* Configuration: .properties or .yaml
Currently Trino and Airlift are using .properties, but gateway is using YAML.
IMO, YAML is a better configuration format.
Adding support on yaml in airlift could be an option (see https://github.com/airlift/airlift/issues/738#issuecomment-508444448).
* Listen to multiple ports using Airlift
Request service at :9080, application service at :9081, and the Admin service at :9082
Not sure how to do this using Airlift.

PR #289, opened 2024-03-24
Remove dependency on Dropwizard Auth

oneonestar:
## Description
Relates to #41

## Release notes


gateway-ha/src/main/java/io/trino/gateway/ha/security/util/BasicCredentials.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Base64;
+import java.util.List;
+import java.util.Optional;
+
+import static com.google.common.base.Strings.nullToEmpty;
+import static com.google.common.net.HttpHeaders.AUTHORIZATION;
+import static java.nio.charset.StandardCharsets.ISO_8859_1;
+import static java.util.Objects.requireNonNull;
+
+public record BasicCredentials(String username, String password)
```
willmostly, line 28: Would it make sense to use `io.trino.server.security.BasicAuthCredentials` here?

willmostly (commented):
Just skimmed and looks good, I will take another pass next week. 

Have you tested this change against an OAuth2 server? I'm not overly confident in the mock tests.

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/BasicCredentials.java
oneonestar: I tried that before. There were few problems:

Some Trino authentication logic are mixed into BasicAuthCredentials. The username must not be empty, otherwise it will throw ([BasicAuthCredentials.java#L61-L63](https://github.com/trinodb/trino/blob/f769e9d3c35d319d16dfe8bb8d2f9f51b4223ca2/core/trino-main/src/main/java/io/trino/server/security/BasicAuthCredentials.java#L61-L63)). That's not required by RFC 2617/7617.

It mixed AuthenticationException and Optional.Empty() to indicate different parsing errors, in order to satisfy different use cases in Trino. That's not needed by gateway and makes the caller's logic become complicated.

It doesn't have unit test coverage in Trino. I don't feel safe to extract and reuse that particular class.

oneonestar:
> Have you tested this change against an OAuth2 server? I'm not overly confident in the mock tests.

I didn't. Overall I don't feel we have enough tests to cover the auth part. That's why I'm holding this PR as a draft.
I planned to add some tests in a separate PR to ensure the auth doesn't break during migration.
In particular I'd like to test `/findQueryHistory` and `/userinfo` endpoint, LDAP authentication and authorization, and OAuth2.

oneonestar:
Test on `/userinfo` added in #321.
Done a manual test on LDAP and OAuth2.
Not sure how to automate the test / mock the required components.

oneonestar:
Found some security issues in the current OIDC implementation while I was adding the test using Hydra.
#344 #345 should fix the issues.

mosabua:
Lets chat about all this in the sync tomorrow. I think we should add them all in.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ResourceSecurityDynamicFeature.java
```diff
@@ -0,0 +1,45 @@
...
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.container.DynamicFeature;
+import jakarta.ws.rs.container.ResourceInfo;
+import jakarta.ws.rs.core.FeatureContext;
+
+import static jakarta.ws.rs.Priorities.AUTHENTICATION;
+import static java.util.Objects.requireNonNull;
+
+@Priority(AUTHENTICATION)
+public class ResourceSecurityDynamicFeature
```
mosabua, line 27: I am hoping you can think of a better class name here .. but maybe its just my unfamiliarity with the jax-rs api

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFilter.java
```diff
@@ -13,32 +13,45 @@
...
 import java.util.Optional;
 
-@Priority(Priorities.AUTHENTICATION)
-public class LbFilter<P extends Principal>
-        extends AuthFilter<String, P>
+import static jakarta.ws.rs.Priorities.AUTHENTICATION;
+import static java.util.Objects.requireNonNull;
+
+@Priority(AUTHENTICATION)
+public class LbFilter
```
mosabua, line 35: for all these classes I think you are implying lb = Load Balancer = Trino Gateway .. right? If so ... I am wondering if we really want that ... our naming is already a bit all over the place .. and introducing lb when we already have gateway and hagateway and Trino Gateway in use seems bad

mosabua: However .. we should fix that in another PR later since this is already in the current codebase..

gateway-ha/src/main/java/io/trino/gateway/ha/security/ResourceSecurityDynamicFeature.java
oneonestar: https://github.com/trinodb/trino/blob/569b045f3bb228703f93a4784411274df5f597d7/core/trino-main/src/main/java/io/trino/server/security/ResourceSecurityDynamicFeature.java
I use the same class name as Trino.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFilter.java
oneonestar: I'd like to keep the original naming during refactoring. Otherwise it'll mess up the diff.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ResourceSecurityDynamicFeature.java
mosabua: ok .. that works then.

mosabua (approved):
Looks good now. Merging this so we can proceed with further testing in main and work towards complete dropwizard removal

PR #291, opened 2024-03-27
Replace WireMock with okhttp3:mockwebserver

oneonestar:
## Description
Part of #41 

Replace WireMock with okhttp3:mockwebserver.
Trino is also using `okhttp3:mockwebserver` for testing.
WireMock depends on jetty 11 which is incompatible with Airlift (jetty 12 & EE 10).

## TODO
Update okhttp3 to the latest version (`3.14.9` -> `4.12.0`).

Side effect:
New version of okhttp3 won't send Authorization header after a redirect. (See: https://github.com/trinodb/trino/issues/21026)
This will break credentials-based authentication if anyone is using:
`Trino-gateway -> (some proxy with redirect mode) -> Trino`.
Needs to override this behavior and always send Authorization header.

## Release notes

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -24,6 +24,7 @@
         <dep.jdbi.version>3.44.0</dep.jdbi.version>
         <dep.jeasy.version>4.1.0</dep.jeasy.version>
         <dep.mockito.version>5.8.0</dep.mockito.version>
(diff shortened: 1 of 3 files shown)


wendigo (approved):

mosabua:
VERY NICE!

PR #292, opened 2024-03-29
refactoring ldap groups to map to trino gateway roles

cghyzel:
## Description

Refactor of the LDAP Auth to map ldap groups Trino Gateway Roles (ADMIN, USER, API). 

Previously the class was simply taking the LDAP attribute information just returning it. This broke the logic of the restUserinfo function in the LoginResource class, which simply splits the memberOf information by an '_'. Furthermore, only "ADMIN", "API", and "USER" are expected by the webapp.
https://github.com/trinodb/trino-gateway/blob/739ec5e67799dd5df0ab8aaec7934e21a63507fa/gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java#L110
https://github.com/trinodb/trino-gateway/blob/739ec5e67799dd5df0ab8aaec7934e21a63507fa/webapp/src/store/access.ts#L6C1-L11C1

## Release notes

```markdown
* Refactored LDAP Group configuration to use a ldap group => gateway role mapping
```


gateway-ha/src/test/java/io/trino/gateway/ha/security/LbLdapClientTest.java
```diff
@@ -135,7 +135,7 @@ public void testMemberof()
         String ret = lbLdapClient.getMemberOf(user);
 
         log.info("ret is %s", ret);
-        assertThat(ret).isEqualTo("Admin,User");
+        assertThat(ret).isEqualTo("ADMIN_USER");
```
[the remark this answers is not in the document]
cghyzel: It could if this in LoginResource was also refactored to process a comma delimiter

https://github.com/trinodb/trino-gateway/blob/739ec5e67799dd5df0ab8aaec7934e21a63507fa/gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java#L110

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -107,12 +108,24 @@ public String getMemberOf(String user)
...
+        String ldapGroups = "";
         if (list != null && !list.isEmpty()) {
-            memberOf = list.listIterator().next().getMemberOf();
-            log.debug("Member of %s", memberOf);
+            ldapGroups = list.listIterator().next().getMemberOf();
+            log.debug("LdapGroups: %s", ldapGroups);
         }
-        return memberOf;
+        ArrayList<String> memberOfArray = new ArrayList<String>();
+        if (ldapGroups.matches(config.getGroupAdminRoleMap())) {
```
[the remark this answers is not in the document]
cghyzel: I've added logic for handling nulls and added early failure invalid regex patterns in the latest commit

docs/security.md
```diff
@@ -154,8 +149,15 @@ authorization:
   ldapConfigPath: '<ldap_config_path>'
 ```
 
-The LDAP config file should have the following contents:
+**Caution:**
+The regex pattern matching for roles only works for API calls, and not on the webapp. 
+The webapp only supports ADMIN, USER, and API roles. 
+For instance, in the case that a user has `privileges: SUPER_USER_API` and the authorization section includes `admin: (.*)SUPER(.*)` the user will *not* have ADMIN role privileges in the webapp.
```
oneonestar, line 155: What I understood is the the result of `lbLdapClient.getMemberOf()` will override the privilege setting in presetUsers.
https://github.com/trinodb/trino-gateway/blob/acca2ea637fd673af1bef84fe27a0b482ea1ada0/gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java#L46-L51

docs/security.md
```diff
@@ -154,8 +149,15 @@ authorization:
   ldapConfigPath: '<ldap_config_path>'
 ```
 
-The LDAP config file should have the following contents:
+**Caution:**
+The regex pattern matching for roles only works for API calls, and not on the webapp. 
+The webapp only supports ADMIN, USER, and API roles.
```
oneonestar, line 154: Webapp is using API, so the security setting should be consistent between webapp and API.
APIs are annotated by `@RolesAllowed("ADMIN/User/API")`. No other roles are supported.

gateway-ha/src/test/java/io/trino/gateway/ha/security/LbLdapClientTest.java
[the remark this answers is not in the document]
oneonestar: I saw `.split("_")` being hard coded everywhere. Let's stay in `_` for now.

docs/security.md
```diff
@@ -130,21 +130,16 @@ authentication:
 
 ## Authorization
 
-Trino Gateway supports the following roles in regex string format:
+Trino Gateway supports the following roles in regex string format (for API calls):
```
mosabua, line 133: why .. those groups are supported by the UI as well 

```suggestion
Trino Gateway supports the following roles in regex string format:
```

mosabua: and if not .. thats a bug in the UI code

docs/security.md
```diff
@@ -130,21 +130,16 @@ authentication:
...
 - admin : Allows access to the Editor tab, which can be used to configure the
   backends
-
 - user : Allows access to the rest of the website
-
 - api : Allows access to rest apis to configure the backends
 
-Users with attributes next to the role will be giving those privileges the
-users. You can use the preset users defined in the yaml file. 
+Users with attributes matching each role regex will be giving the privileges of that role (admin, user, or api) for Api calls. You can use the preset users defined in the yaml file.
```
mosabua, line 140: Please wrap at 80 char columns width so I can review .. but again .. it should be for UI as well

docs/security.md
```diff
@@ -154,8 +149,15 @@ authorization:
   ldapConfigPath: '<ldap_config_path>'
 ```
 
-The LDAP config file should have the following contents:
+**Caution:**
+The regex pattern matching for roles only works for API calls, and not on the webapp.
```
mosabua, line 153: Thats a bug in my opinion

gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
```diff
@@ -36,6 +38,9 @@ public class LdapConfiguration
     private Integer poolMaxTotal;
     private Integer poolMinIdle;
     private boolean poolTestOnBorrow;
+    private String groupAdminRoleMap;
```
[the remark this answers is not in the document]
mosabua: good idea

mosabua (commented):
I have a feeling we are adding something new here instead of fixing a bug in the UI code. We might have to organize a chat to discuss, but my understanding was that the UI and the API should be using the same role setup.

Chaho12 (commented):
If I understood correctly, the ldap returns `uid:<ldap id>` and not `user` or `admin` or `api` in role right? i notice that this works fine for `presetUsers`, but not for ldap

docs/security.md
cghyzel: You would need to refactor the `hasRole` call to get the UI and the web API to function the same.

https://github.com/trinodb/trino-gateway/blob/739ec5e67799dd5df0ab8aaec7934e21a63507fa/webapp/src/store/access.ts#L73C9-L76C9

See the following places where it is used:

https://github.com/trinodb/trino-gateway/blob/ac407dcbd02c3e00655a2ce0a63dd21be429585b/webapp/src/components/selector.tsx#L80
https://github.com/trinodb/trino-gateway/blob/ac407dcbd02c3e00655a2ce0a63dd21be429585b/webapp/src/components/cluster.tsx#L89
https://github.com/trinodb/trino-gateway/blob/ac407dcbd02c3e00655a2ce0a63dd21be429585b/webapp/src/components/history.tsx#L93
https://github.com/trinodb/trino-gateway/blob/ac407dcbd02c3e00655a2ce0a63dd21be429585b/webapp/src/components/resource-group.tsx#L88

cghyzel:
> If I understood correctly, the ldap returns `uid:<ldap id>` and not `user` or `admin` or `api` in role right? i notice that this works fine for `presetUsers`, but not for ldap

It doesn't work for present users either, you can see for yourself in the following example:

authorization:
    admin: .*FOO.*
    user: .*BAR.*
    api: .*BAZ.*

presetUsers:
    admin1:
        password: password
        privileges: FOO_BAR
    user1:
        password: password
        privileges: BAR_BAZ

cghyzel:
I found a much more elegant solution, will open a new PR.

PR #297, opened 2024-04-03
Add Trino configuration docs

mosabua:
## Description

- with process forwarding config only for now

OAuth2 and other tips will follow in separate PRs from @willmostly and others.

Replaces #217 

## Release notes

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -89,6 +89,32 @@ various cloud platforms, such as Amazon EKS also work. However Amazon Athena
 does not work since it uses alternative, custom protocols and lacks the concept
 of individual clusters.
 
+### Trino configuration
+
+From a users perspective Trino Gateway acts as a transparent proxy for one 
+or more Trino clusters. The following Trino configuration tips should be 
+taken into account for all clusters behind the Trino Gateway.
+
+Process forwarding must be enabled:
+
+```commandline
+http-server.process-forwarded=true
+```
+
+Without this setting, first requests go from the user to Trino Gateway and then
+to Trino correctly. However, the URL for subsequent next URIs for more results
+in a query provided by Trino is then using the local URL of the Trino cluster,
+and not the URL of the Trino Gateway. This circumvents the Trino Gateway for all
+these requests, and is contrary to the purpose of using the Trino Gateway. In
+scenarios, where the local URL of the Trino cluster is private to the Trino
+cluster on the DNS/network level, these following calls might not work at all
+for users.
+
+This setting is also required for Trino to authenticate in the case TLS is 
+terminated at the Trino Gateway. Normally it refuses to authenticate plain HTTP 
+requests, but if `http-server.process-forwarded=true` it authenticates over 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -89,6 +89,27 @@ various cloud platforms, such as Amazon EKS also work. However Amazon Athena
...
+
+Without this setting, first requests go from the user to Trino Gateway and then
+to Trino correctly. However, the URL for subsequent next URIs for more results
+in a query provided by Trino is then using the local URL of the Trino cluster,
+and not the URL of the Trino Gateway. This circumvents the Trino Gateway for all
+these requests, and is contrary to the purpose of using the Trino Gateway. In
+scenarios, where the local URL of the Trino  cluster is private to the Trino
+cluster on the DNS/network level, these  following calls might not work at all
+for users.
+
```
willmostly, line 112: This setting is also required for Trino to authenticate in the case TLS is terminated at the Trino Gateway. Normally it refuses to authenticate plain HTTP requests, but if `http-server.process-forwarded=true` it will authenticate over HTTP if the request includes `X-Forwarded-Proto: HTTPS`. I doubt this is a common set up, but perhaps worth mentioning.

docs/installation.md
```diff
@@ -89,6 +89,27 @@ various cloud platforms, such as Amazon EKS also work. However Amazon Athena
...
+```commandline
+http-server.process-forwarded=true
+```
+
+Without this setting, first requests go from the user to Trino Gateway and then
+to Trino correctly. However, the URL for subsequent next URIs for more results
+in a query provided by Trino is then using the local URL of the Trino cluster,
+and not the URL of the Trino Gateway. This circumvents the Trino Gateway for all
+these requests, and is contrary to the purpose of using the Trino Gateway. In
+scenarios, where the local URL of the Trino  cluster is private to the Trino
```
willmostly, line 109: nit
```suggestion
scenarios, where the local URL of the Trino cluster is private to the Trino
```

docs/installation.md
```diff
@@ -89,6 +89,27 @@ various cloud platforms, such as Amazon EKS also work. However Amazon Athena
...
+cluster on the DNS/network level, these  following calls might not work at all
```
willmostly, line 110:
```suggestion
cluster on the DNS/network level, these following calls might not work at all
```

willmostly (approved):
Placement in the installation section looks good, and the content is accurate. I'm not sure if the additional information on TLS termination is worth mentioning or if it is distracting - up to your judgement.

docs/installation.md
mosabua: fixed

docs/installation.md
mosabua: fixed

docs/installation.md
mosabua: Good point... I added this as well. Maybe we should also add to Trino docs honestly.. hahah

mosabua: I will create a ticket to do that .. just looked .. the docs for that in Trino are bad..

PR #302, opened 2024-04-05
Remove dropwizard configuration and jackson

oneonestar:
## Description
Part of #41.
Remove dropwizard configuration and jackson.

Dropwizard configuration support override values using system properties.
For example, `-Dlb.ldapHost=test.com` can override the ldapHost value in `SomeLdapConfig.yml`.
Not sure is anyone relying on this feature.

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
@@ -13,14 +13,17 @@
  */
 package io.trino.gateway.ha.config;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.airlift.log.Logger;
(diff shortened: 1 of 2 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
```diff
@@ -13,14 +13,17 @@
...
-import io.dropwizard.configuration.YamlConfigurationFactory;
-import io.dropwizard.jackson.Jackson;
+
+import java.io.File;
+import java.io.IOException;
 
 public class LdapConfiguration
 {
     private static final Logger log = Logger.get(LdapConfiguration.class);
+    private static final ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
```
ebyhr, line 26: nit: Rename to OBJECT_MAPPER.

ebyhr (approved):

Chaho12 (approved):
I think it's ok to remove it as passing argument is not preferred way of handling confs

PR #304, opened 2024-04-06
Add /static folder to gitignore and remove /static from repository

andythsu:
## Description

We should be able to add `/static` to gitignore because this folder is a copy of `/dist` folder (see https://github.com/trinodb/trino-gateway/blob/5735aa1e8ae1003202d376009c9e311607c110bb/gateway-ha/pom.xml#L556-L577), which is generated by `webapp` during build time. If `/dist` is ignored, we should be fine to ignore `/static` too.

## Release notes

```markdown
* 
```


.gitignore
```diff
@@ -3,3 +3,5 @@ target/**
 .idea/**
 *.iml
 /**/var/
+
+**/static/**
```
ytwp, line 7: `gateway-ha/**/static/**` seems better.

andythsu: indeed. fixed

ebyhr (approved):

PR #305, opened 2024-04-06
Fix pnpm-lock.yaml having hardcoded dependency path in version

andythsu:
## Description

Get rid of the hardcoded dependency path in `pnpm-lock.yaml` to allow download from other URLs. This is a follow-up of https://github.com/trinodb/trino-gateway/pull/303

## Release notes

```markdown
* 
```

### DIFF
--- a/webapp/pnpm-lock.yaml
+++ b/webapp/pnpm-lock.yaml
@@ -7,553 +7,470 @@ settings:
 dependencies:
   '@douyinfe/semi-icons':
     specifier: ^2.47.1
-    version: registry.npmmirror.com/@douyinfe/semi-icons@2.47.1(react@18.2.0)
+    version: 2.47.1(react@18.2.0)
   '@douyinfe/semi-icons-lab':
     specifier: ^2.47.1
-    version: registry.npmmirror.com/@douyinfe/semi-icons-lab@2.47.1(react-dom@18.2.0)(react@18.2.0)
+    version: 2.47.1(react-dom@18.2.0)(react@18.2.0)
   '@douyinfe/semi-illustrations':
     specifier: ^2.47.1
-    version: registry.npmmirror.com/@douyinfe/semi-illustrations@2.47.1(react@18.2.0)
+    version: 2.47.1(react@18.2.0)
   '@douyinfe/semi-ui':
     specifier: ^2.47.1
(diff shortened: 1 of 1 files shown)


ytwp:
Looks great👍

oneonestar:
I'm not sure how to review this kind of changes.
Can we reproduce the same output using some command?

andythsu:
I'm assuming everyone would come up with somewhat different minor/patch versions since it depends on the version. I think these changes are one-time only and as long as the majority of reviewers can also build it using this updated `pnpm-lock`, we should be fine

mosabua:
Isnt the lock file required to ensure that the build is repeatable and the same no matter where you build it .. at least thats how lock files work in rubygems and other tools.. so I think we should NOT remove the file

andythsu:
You are correct about the intent of the lock file, but I'm NOT removing the file.

Instead, I'm removing the hardcoded path.

It looks like this now:
```
dependencies:
  '@douyinfe/semi-icons':
    specifier: ^2.47.1
    version: registry.npmmirror.com/@douyinfe/semi-icons@2.47.1(react@18.2.0)
```

The `registry.npmmirror.com` specifies the download path, which is unnecessary

With current PR and https://github.com/trinodb/trino-gateway/pull/303, we can just use 

```
dependencies:
  '@douyinfe/semi-icons':
    specifier: ^2.47.1
    version: 2.47.1(react@18.2.0)
```

mosabua:
Oh .. I get it. Is markup supported by pnpm .. if yes.. I am good with the change

andythsu:
what do you mean by is markup supported by pnpm?

the pnpm registry url is defined in the pom file, which shouldn't be touched that often unless we add more dependencies. 

how do we ensure they don't sneak back in... i'm not sure.. maybe we just need to be more careful if we see a pnpm-lock change?

andythsu:
not sure if there's such tool to validate the integrity of the lock file 🤔

andythsu:
@ytwp do you know any?

mosabua:
Well... is it complaining about errors with the modified file or does it just work .. if it works.. we are good. I have not had a chance to test....

andythsu:
nope. There's no complaints for me with the modified file. It works as expected.

mosabua (approved):
Lets ship it!

ytwp:
If `registry.npmmirror.com` is specified, it will be downloaded from there. If not specified, it will be downloaded from `pnpm config get registry`.

PR #317, opened 2024-04-12
Replace Dropwizard with Airlift

### LINKED ISSUE #41: Adopt airlift as base framework

oneonestar:
## Description
Resolve #41

Commits cannot be built and tested individually. This is caused by the mass dependency conflict between Dropwizard and Airlift. It is impossible to replace Dropwizard part by part and keep each commit buildable. Also, it will be very difficult to review if all changes are squashed into one commit.

## TODO in this PR
- [x] Fix docker build & health-check
- [x] Find out if there are better ways to handle the configuration (yaml vs .properties)
- [x] HTTPS support

## Follow-up later in other PRs
- Merge requestRouter, applicationConnectors and adminConnectors ports (#382)
- Update docs (#382)
- Refactor and make the use of dependency injection consistent

## How to run in Intellij
VM options: `-Xmx1g -Djdk.attach.allowAttachSelf=true`
Main class: ` io.trino.gateway.ha.HaGatewayLauncher`
Argument: `docs/quickstart-config.yaml`
Working directory: `$ProjectFileDir$`

## Release notes

```markdown
* Replace Dropwizard with Airlift
* (TODO) Document the incompatible changes
```


gateway-ha/etc/config.properties
```diff
@@ -0,0 +1,3 @@
+node.id=ffffffff-ffff-ffff-ffff-ffffffffffff
+node.environment=test
```
willmostly, line 2: The gateway is not clustered, why does it need a `node.id` and `node.environment`?

oneonestar: That's because HttpServerModule depends on NodeInfo which needs NodeConfig which requires non-null node.environment and node.id.
I should be able to set these in Java and eliminate these from the config file.
Might need to dig deeper to see if there are drawbacks of doing it.

mosabua: There is no harm in having these value though .. right?

gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
```diff
@@ -68,21 +68,19 @@ public abstract class BaseApp
...
-            // Modules must have exactly one constructor. The signature must be either one:
+            // Modules must have exactly one constructor. The signature must be:
             // public Module constructor(HaGatewayConfiguration)
-            // public Module constructor(HaGatewayConfiguration, Environment)
             Constructor<?>[] constructors = Class.forName(clazz).getConstructors();
             if (constructors.length != 1) {
                 throw new RuntimeException(format("Failed to load module [%s]. Multiple constructors exist.", clazz));
             }
             Constructor<?> constructor = constructors[0];
             Object module = switch (constructor.getParameterCount()) {
```
willmostly, line 66: could replace this `switch` with an `if`

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -20,6 +20,7 @@
 
 public class HaGatewayConfiguration
 {
+    private Map<String, String> airliftConfig = new HashMap<>();
```
willmostly, line 23: `airliftConfig` may not be intuitive to users unfamiliar with the implementation details. If we set `node.environment` in the code, then something like `httpConfig` would be more obvious

Chaho12:
I assume this is ready for PR right?

docker/Dockerfile
```diff
@@ -47,7 +47,7 @@ COPY --chown=trino:trino gateway-ha /usr/lib/trino
 
 EXPOSE 8080
 USER trino:trino
-CMD java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar /usr/lib/trino/gateway-ha-jar-with-dependencies.jar "server" "/opt/trino/gateway-ha-config.yml"
+CMD java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED -jar /usr/lib/trino/gateway-ha-jar-with-dependencies.jar "/opt/trino/gateway-ha-config.yml"
```
Chaho12, line 50: Can you also update
https://github.com/trinodb/trino-gateway/blob/main/docs/development.md?plain=1#L35 too?

docs/quickstart-config.yaml
```diff
@@ -32,6 +26,3 @@ modules:
 managedApps:
   - io.trino.gateway.ha.GatewayManagedApp
   - io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor
-
-logging:
-    type: external
```
Chaho12, line 37: Can you update installation doc too?
https://github.com/trinodb/trino-gateway/blob/main/docs/installation.md?plain=1#L103-L112

Chaho12: Question. How is this different to httpConfig properties that you can set with https://trino.io/docs/current/admin/properties-logging.html properties using `-Dconfig=var/config.properties`?

oneonestar:
> I assume this is ready for PR right?

Yes. the all-or-nothing changes are done in this PR. Reviews are welcome.
I'm still working on merging different ports into one. The changes are big and they can be tested individually. I'd like to create another PR for that.
Docs will come afterward.

docs/quickstart-config.yaml
oneonestar: `httpConfig` works the same as setting `-Dconfig`. There was a discussion on Slack about the pros and cons of moving from YAML to properties, mixing YAML and properties, and keeping everything in YAML. We agreed to keep everything in YAML for now.

Since we chose to use `httpConfig`, `-Dconfig` won't work. They are mutually exclusive: https://github.com/airlift/airlift/blob/538e1ff498bca145c5ec4d582692367694a94fc8/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java#L151-L156

gateway-ha/src/main/java/io/trino/gateway/baseapp/WebUIStaticResource.java
```diff
@@ -0,0 +1,74 @@
...
+import jakarta.ws.rs.core.Response;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import static jakarta.ws.rs.core.Response.Status.NOT_FOUND;
+
+@Path("/")
```
willmostly, line 30: this will prevent forwarding `/` to a backend correct? In some cases I forward `/`, and access the TGW UI on the app port. IMO we should put all the UI assets behind a prefix such as `/tgw`, and make the prefix configurable if possible

willmostly (approved):
This looks ready to me, mod the discussion on the UI path. Awesome work, this is massive!

gateway-ha/src/main/java/io/trino/gateway/ha/log/GatewayRequestLog.java
```diff
@@ -1,32 +0,0 @@
...
-
-public class GatewayRequestLog
-        implements RequestLog
-{
-    private static final Logger log = Logger.get(GatewayRequestLog.class);
-
-    @Override
-    public void log(Request request, Response response)
-    {
-        // Logging without filter as both request and response don't contain sensitive information
```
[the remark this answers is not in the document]
oneonestar: https://github.com/oneonestar/trino-gateway/blob/93cc98f58b7306f75da7a085f84fb3e608d37423/docs/migration-to-airlift.md#request-logging
I've done it in another PR.

gateway-ha/src/main/java/io/trino/gateway/baseapp/WebUIStaticResource.java
oneonestar: That's not an easy task. IIUC, the react webapp hard coded `/` as root path in many places. Airlift doesn't support setting the ContextPath for ServletContextHandler, which makes configuring the prefix difficult.

IMO forwarding `/`, `/v1/info`, etc to an arbitrary Trino cluster is a bit confusing. Something like `/route/{trinoClusterName}/ui` might make more sense, except this break the drop-in transparent proxy function. Could you explain more about the use case?

willmostly: Starburst has a web UI with multi-cluster query history support, which also includes a SQL IDE. I agree that it is not useful to forward to an arbitrary OS Trino WebUI. I forward `/` because this is redirected to the webUI, its more user friendly for a user to type into their browser than remembering the path.

This may also apply to products such as DuneSQL, which I'm less familiar with.

oneonestar:
I've address the UI path issue in #382 with the following changes:

### Path change for WebUI
The root path for Trino Gateway WebUI has been changed from `/` to `/tgw`. By
default, access to `/` will be redirected to `/tgw`. This behavior can be
changed by adding `/` to `extraWhitelistPaths`, which will route `/` to
Trino.

### Format of `extraWhitelistPaths`
The path setting in `extraWhitelistPaths` is now a regex matches the full
URI in request. The old version will forward requests with a URI prefix that
matches any path in `extraWhitelistPaths`. The new version takes regexes from
`extraWhitelistPaths` and forwards only those requests with a URI that exactly
matches any of the regexes. This changes is required because in the old 
version, `/` will match and forward every request.

Be sure to use single-quoted strings so that escaping is not required. The
following configurations are equal.

Old config:
```yaml
extraWhitelistPaths:
    - "/ui"
    - "/v1/custom"
```

New config:
```yaml
extraWhitelistPaths:
    - '/ui.*'
    - '/v1/custom.*'
```

rdsarvar:
> I've address the UI path issue in #382 with the following changes:
> 
> ### Path change for WebUI
> The root path for Trino Gateway WebUI has been changed from `/` to `/tgw`. By default, access to `/` will be redirected to `/tgw`. This behavior can be changed by adding `/` to `extraWhitelistPaths`, which will route `/` to Trino.
> 
> ### Format of `extraWhitelistPaths`
> The path setting in `extraWhitelistPaths` is now a regex matches the full URI in request. The old version will forward requests with a URI prefix that matches any path in `extraWhitelistPaths`. The new version takes regexes from `extraWhitelistPaths` and forwards only those requests with a URI that exactly matches any of the regexes. This changes is required because in the old version, `/` will match and forward every request.
> 
> Be sure to use single-quoted strings so that escaping is not required. The following configurations are equal.
> 
> Old config:
> 
> ```yaml
> extraWhitelistPaths:
>     - "/ui"
>     - "/v1/custom"
> ```
> 
> New config:
> 
> ```yaml
> extraWhitelistPaths:
>     - '/ui.*'
>     - '/v1/custom.*'
> ```

is the naming of `tgw` up for discussion? personally i prefer the explicit noun based naming `/trino-gateway/vX` where you'd abstract away the clustering from the user with `<?more than 1 gateway?>.trinodb.<org domain>.com/trino-gateway/`

is the issue with it the potential lengths of the URL? i see added readability as `tgw` isn't a well known abbreviation

mosabua:
We will need to ensure we add a breaking release notes entry with upgrade info (or link to docs with the info would be even better).

Also I think the longer path would actually be nice ..

oneonestar:
I'm fine with `/trino-gateway`. I am not aware of any practical limitations for having a longer URL path. We can finalize the decision in today's [meeting](https://github.com/trinodb/trino-gateway/wiki/Contributor-meetings#trino-gateway-dev-sync-12-jun-2024).

pom.xml
```diff
@@ -44,22 +44,32 @@
         <air.check.skip-spotbugs>true</air.check.skip-spotbugs>
         <air.check.skip-pmd>true</air.check.skip-pmd>
         <air.release.preparation-goals>clean verify -DskipTests</air.release.preparation-goals>
+        <dep.jetty.version>12.0.9</dep.jetty.version>
```
wendigo, line 47: Is this needed? This is a default version in latest airlift. Please drop it.

pom.xml
```diff
@@ -44,22 +44,32 @@
...
     </properties>
 
     <dependencyManagement>
         <dependencies>
+
```
wendigo, line 52: Revert

pom.xml
```diff
@@ -44,22 +44,32 @@
...
             <dependency>
                 <groupId>io.airlift</groupId>
                 <artifactId>bom</artifactId>
-                <version>240</version>
+                <version>248</version>
```
wendigo, line 56: 248 comes with jetty 12.0.9

pom.xml
```diff
@@ -44,22 +44,32 @@
...
+                <version>${dep.jetty.version}</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
 
             <dependency>
-                <groupId>io.dropwizard</groupId>
-                <artifactId>dropwizard-dependencies</artifactId>
-                <version>4.0.6</version>
+                <groupId>org.eclipse.jetty.ee10</groupId>
```
wendigo, line 70: Why do you need to explicitly define jetty dependencies?

pom.xml
```diff
@@ -81,9 +91,15 @@
...
             <dependency>
-                <groupId>jakarta.activation</groupId>
-                <artifactId>jakarta.activation-api</artifactId>
-                <version>2.1.2</version>
+                <groupId>io.airlift</groupId>
+                <artifactId>units</artifactId>
+                <version>1.10</version>
+            </dependency>
+
+            <dependency>
```
wendigo, line 99: This is provided by airlift as well.

gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
```diff
@@ -159,68 +107,58 @@ protected List<Module> addModules(HaGatewayConfiguration configuration)
...
-                                LifecycleEnvironment lifecycle = environment.lifecycle();
-                                lifecycle.manage((Managed) injector.getInstance(c));
-                                logger.info("op=register type=managed item=%s", c);
-                            }
-                            catch (Exception e) {
-                                logger.error(e, "Error loading managed app");
-                            }
-                        });
-        return managedApps;
+        configuration.getManagedApps().forEach(
```
wendigo, line 126: Do we need it? In the configuration classes these are always:
``managedApps:
  - io.trino.gateway.ha.GatewayManagedApp
  - io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor
``

We can bind GatewayManagedApp.class and ActiveClusterMonitor.class

gateway-ha/src/main/java/io/trino/gateway/baseapp/WebUIStaticResource.java
```diff
@@ -0,0 +1,74 @@
...
+public class WebUIStaticResource
```
wendigo, line 31: You should use Airlift's `HttpServerBinder.bindResource` instead  :)

gateway-ha/src/main/java/io/trino/gateway/ha/HaGatewayLauncher.java
```diff
@@ -33,6 +111,9 @@ public static void main(String[] args)
         Logging logging = Logging.initialize();
         logging.configure(configuration);
 
-        new HaGatewayLauncher().run(args);
+        ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+        HaGatewayConfiguration haGatewayConfiguration = objectMapper.readValue(new File(args[0]), HaGatewayConfiguration.class);
```
wendigo, line 103: You should check args.length first :)

wendigo (commented):
Overall looks good

pom.xml
oneonestar: Please correct me if I'm wrong.
`<dep.jetty.version>` is being defined in the root pom.xml in Airlift.
It doesn't exist in `io.airlift:bom` nor `io.airlift:airbase`, so I cannot reuse it.

pom.xml
oneonestar: Same as above. It doesn't exist in `io.airlift:bom` nor `io.airlift:airbase`.
Also, all the `jetty-ee10-*` dependencies will be removed in #382.
We don't need to worry too much about this.

pom.xml
oneonestar: Same as above.

pom.xml
oneonestar: Done

gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
oneonestar: I agree we should use static binding, but that changes will also affect `modules` and others.
I'd like to remove `modules` and `managedApps` in the configuration in the future.
We'll need to update the design of configuration and that requires further discussion.

Since these are nice to have changes and not a must for the migration to Airlift, I'd like to leave it as is.

gateway-ha/src/main/java/io/trino/gateway/ha/HaGatewayLauncher.java
oneonestar: Done

wendigo:
Let's land this and improve in follow-ups.

wendigo:
@mosabua thoughts?

docs/quickstart-config.yaml
```diff
@@ -1,19 +1,14 @@
+httpConfig:
```
mosabua, line 1: I think we can rename and then reuse it for all key-value config like config.properties and node.properties in trino.  But we can do that in a follow up.

gateway-ha/src/main/java/io/trino/gateway/baseapp/WebUIStaticResource.java
mosabua: We can sort this out in a follow up

pom.xml
mosabua: you can test that ..  but we can also sort that out in a follow up

mosabua (approved):
Amazing work! Definitely good enough to merge. We will follow up with minor updates.

PR #318, opened 2024-04-12
Standardize the path for the gateway jar on the docker image

willmostly:
## Description

This removes the version number from the gateway jar path in order to prepare for Helm chart development. 

## Additional context and related issues

The docker image locates the gateway executable jar at
```
"/usr/lib/trino/gateway-ha-${TRINO_GATEWAY_VERSION}-jar-with-dependencies.jar"
```
This is inconvenient for implementing a Helm chart that will not use the image's `CMD` directive.

## Release notes

( x) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #323, opened 2024-04-17
Add Helm chart for k8s deployment

willmostly:
## Description

Add a Helm chart. This chart passes configuration directly from the values.yaml to the gateway config.yaml, loosely coupling the chart structure with the configuration of the gateway itself. Secrets for configurations that contain sensitive data are also supported.

This chart is only compatible with version 7 of the Trino Gateway docker image. Before this chart is released, we should release a public image with https://github.com/trinodb/trino-gateway/pull/318. This chart can then be updated to remove the version number from the executable path, making it compatible with all versions of the docker image going forward.

## Release notes

(X ) Release notes are required, with the following suggested text:

```markdown
Add a Helm Chart for deployment
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -143,3 +143,95 @@ java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 \
     --add-opens=java.base/java.net=ALL-UNNAMED \
     -jar gateway-ha.jar server gateway-config.yml
 ```
+
+### Helm
+
+Helm manages the deployment of Kubernetes applications by templating Kubernetes
+resources with a set of Helm charts. The Trino Gateway Helm chart consists 
+of the following compnents:
+
+* A `config` node for general configuration
+* `dataStoreSecret`, `backendStateSecret` and `authenticationSecret` for 
+  providing sensitive configurations through Kubernetes secrets, 
+* Standard Helm options such as `replicaCount`, `resources` and `ingress`.
+
+The default [values.yaml](../helm/trino-gateway/values.yaml) includes basic
+configuration options as an example. For a simple deployment, proceed with 
+the following steps:
+
+Create a yaml file containing the configuration for your `datastore`:
+
+```shell
+cat << EOF > datastore.yaml
+dataStore:
+   jdbcUrl: jdbc:postgresql://yourdatabasehost:5432/gateway
+   user: postgres
+   password: secretpassword
+   driver: org.postgresql.Driver
+EOF
+```
+Create a Kubernetes secret from this file:
+
+```shell
+kubectl create secret generic datastore-yaml --from-file datastore.yaml --dry-run=client -o yaml | kubectl apply -f -
+```
+
+Create a values override with a name such as `values-override.yaml` and
+reference this secret in the `backendStateSecret` node:
+
+```yaml
+backendStateSecret:
+    name: "datastore-yaml"
+    key: "datastore.yaml"
+```
+
+When a Secret is created with the `--from-file` option, the filename is used as
+the key. Finally, you can deploy Trino Gateway with the chart from the root 
+of this repository:
+
+```shell
+helm install tg --values values-override.yaml helm/trino-gateway 
+```
+
+Secrets for `authenticationSecret` and `backendState` can be provisioned
+similarly. Alternatively,  you can directly define the `config.backEndState` 
+node in `values-override.yaml` and leave `backendStateSecret` undefined. 
+However, a [Secret](https://kubernetes.
+io/docs/concepts/configuration/secret/)
+is recommended to protect the  database credentials required for this 
+configuration.
+
+#### Additional options
+
+To implement routing rules, create a ConfigMap from your routing rules yaml
+definition:
+
+```shell
+kubectl create cm routing-rules --from-file your-routing-rules.yaml
+```
+
+Then mount it to your container:
+
+```yaml
+volumes:
+    - name: routing-rules
+      configMap:
+          name: routing-rules
+          items:
+              name: your-routing-rules.yaml
+              path: your-routing-rules.yaml
+
+volumeMounts:
+    - name: routing-rules
+      mountPath: "/etc/routing-rules/your-routing-rules.yaml"
+      subPath: your-routing-rules.yaml
+```
+
+Ensure that the `mountPath` matches the `rulesConfigPath` specified in your
+configuration. Note that the `subPath` is not strictly necessary, and if it 
+is not specified the file is mounted at `mountPath/<configMap key>`. 
+Kubernetes updates the mounted file when the ConfigMap is updated.
+
+Standard Helm oœptions such as `replicaCount`, `image`, `imagePullSecrets`, 
+`service`, `ingress` and `resources` are supported. These are defined in 
+`helm/values.yaml`. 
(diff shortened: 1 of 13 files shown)


helm/trino-gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+              - "java"
+              - "--add-opens=java.base/java.lang=ALL-UNNAMED"
+              - "--add-opens=java.base/java.net=ALL-UNNAMED"
+              - "-jar"
+              - "/usr/lib/trino/gateway-ha-7-jar-with-dependencies.jar"
```
willmostly, line 48: this will be changed to `"/usr/lib/trino/gateway-ha-jar-with-dependencies.jar"` once an image is published for release 8

helm/trino-gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
Chaho12, line 102: is this config necessary to be added as default?

helm/trino-gateway/values.yaml
```diff
@@ -0,0 +1,126 @@
...
+  #    hosts:
+  #      - chart-example.local
+
+resources: {}
+  # limits:
+  #   cpu: 1
+  #   memory: 1Gi
+  # requests:
+  #   cpu: 1
+  #   memory: 1Gi
```
Chaho12, line 83: Do you suggest scale up or scale out in general?
For us, i notice that there is cpu throttle issue if cpu is set too low. so we prefer to scale up before scaling out. 
I haven't tested with recent version though

ebyhr:
> A straightforward helm chart for the trino gateway

Please follow the commit message guideline: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#format-git-commit-messages

Just "Add helm chart" should be fine.

helm/trino-gateway/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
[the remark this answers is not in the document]
willmostly: Must have been inserted when I did `helm create`. We should decide on a versioning scheme though, good eye. We can increment versioning along with the Trino Gateway, however there is nothing version specific in the chart so we could separate them. @mosabua wdyt?

helm/trino-gateway/values.yaml
willmostly: I haven't experimented with running on less than 2 CPU. I run with at least 2 instances to avoid a SPOF, and usually use a c7g.xlarge or similar. 

We could set some reasonable limits by default, like 2 CPU/ 4Gi

helm/trino-gateway/Chart.yaml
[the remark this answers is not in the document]
willmostly: My concern about using the same version number as the Trino Gateway release is that someone might assume they *have* to update the chart version in tandem with the gateway version.

helm/trino-gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+            {{- end }}
+            {{- range $index, $connector := $.Values.config.server.applicationConnectors}}
+            - name: {{print "app-" $index }}
+              protocol: TCP
+              containerPort: {{ $connector.port }}
+            {{- end }}
+          livenessProbe:
+            httpGet:
+                path: /
+                port: {{ (index .Values.config.server.adminConnectors 0).port}}
```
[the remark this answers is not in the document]
willmostly: Added configurations for these. With the defaults now set in values.yaml the readiness probe starts polling at 5 seconds, and polls every 5 seconds for one minute before marking the pod `unready`. The liveness probe will begin polling after 30 seconds.

helm/trino-gateway/values.yaml
Chaho12: Sure sounds good

helm/trino-gateway/templates/deployment.yaml
Chaho12: Sounds fair enough :)

Chaho12 (approved):

helm/trino-gateway/Chart.yaml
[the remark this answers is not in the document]
mosabua: I think we should keep chart and app version in sync for now ..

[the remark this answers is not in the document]
mosabua: Easier for us to test and manage .. if we really need to split it .. we can deal with that later, for now I see no reason

helm/trino-gateway/values.yaml
mosabua: Yeah .. higher, more realistic values for a production setup are probably better

helm/trino-gateway/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
mosabua, line 18: can we use 8 here

helm/trino-gateway/templates/deployment.yaml
mosabua: can you change that now so the chart will work for 8 and onwards

helm/trino-gateway/Chart.yaml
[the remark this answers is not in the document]
mosabua: Use 8 here

mosabua (changes requested):

docs/installation.md
```diff
@@ -143,3 +143,95 @@ java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 \
...
+
+When a Secret is created with the `--from-file` option, the filename is used as
+the key. Finally, you can deploy Trino Gateway with the chart from the root 
+of this repository:
+
+```shell
+helm install tg --values values-override.yaml helm/trino-gateway 
+```
+
+Secrets for `autenticationSecret` and `backendState` can be provisioned
```
willmostly, line 196:
```suggestion
Secrets for `authenticationSecret` and `backendState` can be provisioned
```

mosabua (approved):
This looks good enough to run with in the 8 release and improve after.

PR #325, opened 2024-04-24
Process Request for routing rules

willmostly:
## Description

Extract information from the HttpServeletRequest into a class more convenient for usage in routing logic.

Routing rules are authored in the MVEL language used by the jeasy rules engine. MVEL has limited functionality in comparison to standard Java, and cannot use libraries not included with the JDK. This PR passes an additional `processedRequest` object to the rules engine with getters for the user, query type, and SQL details such as referenced tables and schemas.

## Release notes

( x) Release notes are required, with the following suggested text:

```markdown
* Process request to facilitate rule writing
```


gateway-ha/pom.xml
```diff
@@ -221,6 +227,12 @@
             <artifactId>metrics-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-parser</artifactId>
+            <version>440</version>
```
ebyhr, line 233: Use 445.

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -36,6 +36,9 @@ public class HaGatewayConfiguration
     private ClusterStatsConfiguration clusterStatsConfiguration;
     private List<String> extraWhitelistPaths = new ArrayList<>();
 
+    private ProcessedRequestConfig processedRequestConfig = new ProcessedRequestConfig();
+    //TODO: add getter/setter for processedrequestconfig
```
ebyhr, line 40: Is this TODO comment left-over?

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -183,4 +186,14 @@ public void setManagedApps(List<String> managedApps)
...
+
+    public ProcessedRequestConfig getProcessedRequestConfig()
+    {
+        return processedRequestConfig;
+    }
+
+    public void setProcessedRequestConfig(ProcessedRequestConfig processedRequestConfig)
+    {
+        this.processedRequestConfig = processedRequestConfig;
+    }
```
ebyhr, line 198: Move to before `getModules` method for consistency with the field orders.

gateway-ha/src/main/java/io/trino/gateway/ha/config/ProcessedRequestConfig.java
```diff
@@ -0,0 +1,77 @@
...
+package io.trino.gateway.ha.config;
+
+public class ProcessedRequestConfig
+{
+    Integer maxBodySize = 1_000_000;
+
+    boolean isClientsUseV2Format;
+    String tokenUserField = "email";
+    String oauthTokenInfoUrl;
+    boolean isGetUserFromOauth;
```
ebyhr, line 23: Add `private`

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,524 @@
...
+import static java.util.Objects.requireNonNullElse;
+
+public class ProcessedRequest
+{
+    public static final String UNSET = "__UNSET__";
+
+    private final Logger log = Logger.get(ProcessedRequest.class);
+    private String oauthTokenInfoUrl;
+    private boolean isGetUserFromOAuth;
+    private boolean isClientsUseV2Format;
```
ebyhr, line 78: Add `final`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,524 @@
...
+    private String body;
+    private String queryType;
+    private Set<QualifiedName> tables = new HashSet<>();
```
ebyhr, line 81: Use ImmutableSet builder. Same for catalogs, schemas, catalogSchemas.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,524 @@
...
+            }
+            if (schemas.isEmpty()) {
+                schemas = tables.stream().map(q -> q.getParts().get(1)).collect(Collectors.toSet());
+            }
+            catalogSchemas = tables.stream()
+                    .map(qualifiedName -> String.format("%s.%s", qualifiedName.getParts().getFirst(), qualifiedName.getParts().get(1)))
+                    .collect(Collectors.toSet());
+        }
+        catch (IOException e) {
+            log.warn("Error extracting request body for rules processing: " + e.getMessage());
```
ebyhr, line 170: Avoid string concat in logger. Same for others.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,524 @@
...
+
+        String encoded = headerValue.substring(prefix.length());
+        byte[] compressed = base64Url().decode(encoded);
+
+        byte[] preparedStatement = new byte[toIntExact(ZstdDecompressor.getDecompressedSize(compressed, 0, compressed.length))];
+        new ZstdDecompressor().decompress(compressed, 0, compressed.length, preparedStatement, 0, preparedStatement.length);
+        return new String(preparedStatement, UTF_8);
+    }
+
+    private void getTables(Node node, Set<QualifiedName> tables)
```
ebyhr, line 213: Please add a unit test for this method.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,524 @@
...
+
+    public Set<String> getSchemas()
+    {
+        return schemas;
+    }
+
+    public Set<String> getCatalogSchemas()
+    {
+        return catalogSchemas;
+    }
```
ebyhr, line 475: Remove unused methods.

ebyhr (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,524 @@
...
+                catch (JsonProcessingException ex) {
+                    log.debug("Could not deserialize token info response to json: " + responseBody);
+                }
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private String googleUserInfo(String token)
```
sajjoseph, line 312: why googleUserInfo? 
Token can be from any IDP - right?

willmostly: yep - leftover from early development

willmostly: any IDP that supports the userinfo endpoint should work

gateway-ha/pom.xml
willmostly: 445 causes `RequireUpperBoundDeps` errors when building, which I think is because gateway is using an earlier version of airlift. @oneonestar would upgrading airlift interfere with the work you are doing to remove Dropwizard?

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
willmostly: yes missed that one

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
willmostly: This method is tested by `TestRoutingGroupSelector.testRequestProcessorQueryDetails`. For the SQL
```
SELECT x.*, y.*, z.* FROM catx.schemx.tblx a, schemy.tbly y, tblz z
```
The matched rule checks
```
processedRequest.tablesContains("cat_default.\"schem_\\\"default\".tblz")
  && processedRequest.tablesContains("cat_default.schemy.tbly")
  && processedRequest.tablesContains("catx.schemx.tblx")
  && processedRequest.getSchemas().contains("schemy")
  && processedRequest.getCatalogs().contains("catx")
```

What would you like to see in a unit test?

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
willmostly: These are intended for use in routing rules, but are unused in the Trino Gateway code. I added more cases to `routing_rules_processed_request.yml` to cover `getDefaultCatalog()` and `getCatalogSchemas()` and provide usage examples.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,533 @@
...
+                body = "";
+                return;
+            }
+
+            Map<String, String> preparedStatements = getPreparedStatements(request);
+            SqlParser parser = new SqlParser();
+            reader.mark(config.getMaxBodySize());
+            body = CharStreams.toString(request.getReader());
+
+            if (isClientsUseV2Format) {
```
vishalya, line 127: can this be determined per request basis? looking at the url or any other mechanism?

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,533 @@
...
+            if (reader == null) {
+                log.warn("HTTP request returned null reader");
+                body = "";
+                return;
+            }
+
+            Map<String, String> preparedStatements = getPreparedStatements(request);
+            SqlParser parser = new SqlParser();
+            reader.mark(config.getMaxBodySize());
+            body = CharStreams.toString(request.getReader());
```
vishalya, line 125: Can we not work with the same reader? I don't see that the reader is used again.

vishalya (commented):
I have my first round of comments and questions in.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RuleReloadingRoutingGroupSelector.java
```diff
@@ -43,10 +44,12 @@ public class RuleReloadingRoutingGroupSelector
     private volatile Rules rules = new Rules();
     private volatile long lastUpdatedTime;
     private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock(true);
+    private final ProcessedRequestConfig processedRequestConfig;
```
vishalya, line 47: I was wondering what's the best way to make this flow optional. At least till it gets some mileage on it. Overall it will be a good idea to skip the processing of the request if we want have a more performant flow.

vishalya:
Are you handling the DDL statements like create and drop table? They are useful in case of `memory` catalog.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RuleReloadingRoutingGroupSelector.java
```diff
@@ -84,9 +87,13 @@ public String findRoutingGroup(HttpServletRequest request)
                     writeLock.unlock();
                 }
             }
+            ProcessedRequest processedRequest = new ProcessedRequest(request, processedRequestConfig);
```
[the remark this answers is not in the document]
willmostly: The body processing is restricted to [POST](https://github.com/trinodb/trino-gateway/pull/325/commits/c100ecdf150c6cad7a465765699d3016d4a2173e#diff-973162668135a5cdba92480352158a4562d54c8815725d9fc0f852703202f4d8R297). The `processedRequest` object needs to be passed to the rules engine everytime it is invoked, because a rule referencing a nonexistent object will produce a runtime error. However, I could add a method like `boolean isPost()` that could be checked in a rule - would that be helpful?

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,533 @@
...
+            ImmutableSet.Builder<String> schemaBuilder = ImmutableSet.builder();
+            ImmutableSet.Builder<String> catalogSchemaBuilder = ImmutableSet.builder();
+
+            getTables(statement, tableBuilder, catalogBuilder, schemaBuilder, catalogSchemaBuilder);
+            tables = tableBuilder.build();
+            catalogs = catalogBuilder.build();
+            schemas = schemaBuilder.build();
+            catalogSchemas = catalogSchemaBuilder.build();
+
+            if (catalogs.isEmpty()) {
```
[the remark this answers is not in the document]
willmostly: Catalogs and schemas will be populated directly for `SHOW [SCHEMAS, CATALOGS]`. For any query that references a catalog or schema directly, the `[catalog, schema]Builder` will be used, and `isEmpty()` will be false. For queries where the catalog and schema are part of a table QualifiedName, either explicitly or not, the catalogs and schemas will be extracted from the `tables` set.

However, i think it is simpler if I just take the union of the two sets. This conversation also reminds me that I should handle CREATE, ALTER, DROP SCHEMA as well as SHOW.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
```diff
@@ -0,0 +1,533 @@
...
+import java.util.stream.Collectors;
+
+import static com.google.common.io.BaseEncoding.base64Url;
+import static io.airlift.json.JsonCodec.jsonCodec;
+import static java.lang.Math.toIntExact;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElse;
+
+public class ProcessedRequest
```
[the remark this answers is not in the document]
willmostly: the capability to use the OAuth userInfo endpoint would apply to non-query requests as well. I could split that out into a separate object, then pass `user` and `parsedQuery`  objects to the rules engine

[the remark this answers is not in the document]
willmostly: changed to `TrinoQueryProperties` and `TrinoRequestUser`, wdyt?

gateway-ha/pom.xml
willmostly: Unfortunately upgrading Airlift can't be done here. It looks like it will require removing dropwizard. 

Moving to airbase 156 required upgrading `jakarta.activation-api` to 3.0.0. This caused errors in CI like the below. I wasn't able to work around these, and some googling indicates that Dropwizard is simply incompatible with modern versions of jakarta. `SingleThreadModel` was [deprecated in 2.4](https://docs.oracle.com/cd/E17802_01/products/products/servlet/2.5/docs/servlet-2_5-mr2/javax/servlet/SingleThreadModel.html).
```
2024-05-22T04:24:30.916Z	INFO	main	stderr	java.lang.NoClassDefFoundError: jakarta/servlet/SingleThreadModel
2024-05-22T04:24:30.916Z	INFO	main	stderr		at org.eclipse.jetty.servlet.ServletHolder.setServlet(ServletHolder.java:173)
2024-05-22T04:24:30.917Z	INFO	main	stderr		at org.eclipse.jetty.servlet.ServletHolder.<init>(ServletHolder.java:146)
2024-05-22T04:24:30.917Z	INFO	main	stderr		at io.dropwizard.jetty.setup.ServletEnvironment.addServlet(ServletEnvironment.java:46)
2024-05-22T04:24:30.917Z	INFO	main	stderr		at io.dropwizard.core.setup.AdminEnvironment.<init>(AdminEnvironment.java:45)
2024-05-22T04:24:30.917Z	INFO	main	stderr		at io.dropwizard.core.setup.Environment.<init>(Environment.java:85)
2024-05-22T04:24:30.917Z	INFO	main	stderr		at io.dropwizard.core.cli.EnvironmentCommand.run(EnvironmentCommand.java:52)
2024-05-22T04:24:30.917Z	INFO	main	stderr		at io.dropwizard.core.cli.ConfiguredCommand.run(ConfiguredCommand.java:98)
2024-05-22T04:24:30.918Z	INFO	main	stderr		at io.dropwizard.core.cli.Cli.run(Cli.java:78)
2024-05-22T04:24:30.918Z	INFO	main	stderr		at io.dropwizard.core.Application.run(Application.java:94)
2024-05-22T04:24:30.918Z	INFO	main	stderr		at io.trino.gateway.ha.HaGatewayLauncher.main(HaGatewayLauncher.java:49)
2024-05-22T04:24:30.918Z	INFO	main	stderr		at io.trino.gateway.ha.security.TestAuthorization.setup(TestAuthorization.java:55)
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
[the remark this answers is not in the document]
vishalya: that's better, thanks!

willmostly:
@vishalya DDL statements are handled, please check [provideTableExtractionQueries](https://github.com/trinodb/trino-gateway/pull/325/commits/8f2e51004914ccf69e786dea547b4f60dbc0ebe5#diff-973162668135a5cdba92480352158a4562d54c8815725d9fc0f852703202f4d8R306) for the full list of statement types. LMK if I missed any.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ProcessedRequest.java
willmostly: @ebyhr , any feedback?

ebyhr: Please add `@SuppressWarnings("unused")` annotation to unused methods.

gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.config;
+
+public class RequestAnalyzerConfig
```
vishalya, line 16: Are you adding the document for the configuration?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,192 @@
...
+
+    private Optional<String> user = Optional.empty();
+
+    private String oauthTokenInfoUrl;
+    private final boolean isGetUserFromOAuth;
+    private static final Logger log = Logger.get(TrinoRequestUser.class);
+
+    public TrinoRequestUser(HttpServletRequest request, RequestAnalyzerConfig config)
+    {
+        isGetUserFromOAuth = config.isGetUserFromOauth();
```
vishalya, line 51: Please clarify how does this flow work, the backend trino can have multiple authentication mechanisms. so are you setting the flag for an individual request? Also gateway and trino can in theory have different IDPs.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,192 @@
...
+                }
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private String getUserInfo(String token)
+    {
+        OkHttpClient client = new OkHttpClient();
```
vishalya, line 177: Just wondering if the connection is going to happen for each processed request, if so do we need any further optimization?

vishalya (commented):
I think we are getting there.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
willmostly: good call out. I've added a cache with a 10 minute ttl to avoid bumping into rate limits

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,192 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.airlift.log.Logger;
+import io.trino.gateway.ha.config.RequestAnalyzerConfig;
+import jakarta.servlet.http.Cookie;
+import jakarta.servlet.http.HttpServletRequest;
+import okhttp3.Call;
+import okhttp3.HttpUrl;
+import okhttp3.OkHttpClient;
```
ebyhr, line 25: Can we use Airlift HTTP client instead?

willmostly: This now uses Nimbus. I considered using `NimbusAirliftHttpClient` but it did not seem to add any benefits to this simple usage.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
willmostly: This is now documented under `### TrinoRequestUser` in `routing-rules.md`. Essentially it tries multiple methods, and uses whichever works. The userinfo url is configurable, it does not reuse the url from the authentication config.

gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
willmostly: This is now documented in `routing-rules.md`, ptal

mosabua:
@willmostly please rebase.. then we are ready to merge I think.

willmostly:
@mosabua rebased

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -148,4 +302,123 @@ public void testByRoutingRulesEngineFileChange()
...
+        HttpServletRequest mockRequest = mock(HttpServletRequest.class);
+        when(mockRequest.getMethod()).thenReturn(HttpMethod.POST);
+        return mockRequest;
+    }
+
+    @Test
+    public void testLongQuery()
+            throws IOException
+    {
+        BufferedReader bufferedReader = new BufferedReader(new FileReader("src/test/resources/wide_select.sql"));
```
oneonestar, line 417:
```suggestion
        BufferedReader bufferedReader = new BufferedReader(new FileReader("src/test/resources/wide_select.sql", UTF_8));
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -66,23 +90,153 @@ public void testByRoutingGroupHeader()
...
+            {
+                return index < elements.length;
+            }
+
+            @Override
+            public String nextElement()
+            {
+                return elements[index++];
+            }
+        };
```
oneonestar, line 216: Can be done by `Collections.enumeration(Arrays.asList(encodedStatements.split(",")))`

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -31,10 +46,19 @@
 import static org.mockito.Mockito.when;
 
 @TestInstance(Lifecycle.PER_CLASS)
+@SuppressModernizer
+// HttpServelet request uses an Enumeration for getHeaders, and constructing an Enumeration for testing violates modernizer rules
```
oneonestar, line 50: We can get rid of this after addressing the issues below.

gateway-ha/pom.xml
```diff
@@ -316,6 +328,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.gaul</groupId>
+            <artifactId>modernizer-maven-annotations</artifactId>
+            <version>2.7.0</version>
+            <scope>test</scope>
+        </dependency>
+
```
oneonestar, line 337: We can avoid this.
See comments on `TestRoutingGroupSelector.java`.

gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public boolean isClientsUseV2Format()
+    {
+        return isClientsUseV2Format;
+    }
+
+    public void setClientsUseV2Format(boolean clientsUseV2Format)
+    {
+        isClientsUseV2Format = clientsUseV2Format;
+    }
```
oneonestar, line 66: Is the v2 protocol actually being implemented in Trino?

oneonestar (commented):
Skimmed through the first half.

gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
willmostly: Not in OS, but it is in use by commercial versions

oneonestar: OK. I think it's worth adding explanations to the code comments and user documentation to prevent confusion.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+            ImmutableSet.Builder<String> catalogSchemaBuilder)
+    {
+        if (schemaOptional.isEmpty()) {
+            schemaBuilder.add(defaultSchema);
+            catalogBuilder.add(defaultCatalog);
+            catalogSchemaBuilder.add(String.format("%s.%s", defaultCatalog, defaultSchema));
+        }
+        else {
+            QualifiedName schema = schemaOptional.orElseThrow();
+            switch (schema.getParts().size()) {
```
oneonestar, line 311: NIT: Add default branch and log error

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    private QualifiedName qualifyName(QualifiedName table)
+    {
+        List<String> tableParts = table.getParts();
+        return switch (tableParts.size()) {
+            case 1:
+                yield QualifiedName.of(defaultCatalog, defaultSchema, tableParts.getFirst());
+            case 2:
+                yield QualifiedName.of(defaultCatalog, tableParts.getFirst(), tableParts.get(1));
```
oneonestar, line 332: `tableParts.get(1)` -> `tableParts.getLast()`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+            case 3:
+            default:
+                yield QualifiedName.of(tableParts.getFirst(), tableParts.get(1), tableParts.get(2));
+        };
+    }
+
+    public String getBody()
```
oneonestar, line 339: `@SuppressWarnings("unused")`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+    {
+        return body;
+    }
+
+    public String getQueryType()
```
oneonestar, line 344: `@SuppressWarnings("unused")`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    public boolean isQuery()
+    {
+        return isQuery;
+    }
+
+    public static class AlternateStatementRequestBodyFormat
+    {
+        // Based on https://github.com/prestodb/presto/wiki/Presto-v2-client-protocol, without session
```
oneonestar, line 435: Link to trino instead?
Also, add an explanation that this is only implemented in the private fork.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+            catch (JsonProcessingException e) {
+                log.warn("Could not deserialize bearer token as json");
+            }
+        }
+
+        if (!isNullOrEmpty(oauthUserInfoUrl)) {
+            try {
+                UserInfo userInfo = userInfoCache.get(token);
+                this.userInfo = Optional.of(userInfo);
+                return Optional.of(userInfo.getSubject().toString());
```
oneonestar, line 176: Avoid variable hiding.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+    private Optional<UserInfo> userInfo = Optional.empty();
+
+    private final String oauthUserInfoUrl;
+    private static final Logger log = Logger.get(TrinoRequestUser.class);
+
+    private final LoadingCache<String, UserInfo> userInfoCache =
+            CacheBuilder.newBuilder()
+                    .maximumSize(10000)
+                    .expireAfterAccess(10, TimeUnit.MINUTES)
+                    .build(CacheLoader.from((accessToken) -> getUserInfo(accessToken)));
```
oneonestar, line 61: I don't think this cache works since we create new instance of TrinoRequestUser per request in `RuleReloadingRoutingGroupSelector`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+public class TrinoRequestUser
+{
+    public static final String TRINO_USER_HEADER_NAME = "X-Trino-User";
+    public static final String TRINO_UI_TOKEN_NAME = "Trino-UI-Token";
+    public static final String TRINO_SECURE_UI_TOKEN_NAME = "__Secure-Trino-ID-Token";
+
+    private Optional<String> user = Optional.empty();
+    private Optional<UserInfo> userInfo = Optional.empty();
+
+    private final String oauthUserInfoUrl;
```
oneonestar, line 54: Let's use `Optional` instead of null.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+        oauthUserInfoUrl = config.getOauthTokenInfoUrl();
+        user = extractUser(request, config.getTokenUserField());
+    }
+
+    public Optional<String> getUser()
+    {
+        return user;
+    }
+
+    public Optional<UserInfo> getUserInfo()
```
oneonestar, line 74: `@SuppressWarnings("unused")`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+                    .expireAfterAccess(10, TimeUnit.MINUTES)
+                    .build(CacheLoader.from((accessToken) -> getUserInfo(accessToken)));
+
+    public TrinoRequestUser(HttpServletRequest request, RequestAnalyzerConfig config)
+    {
+        oauthUserInfoUrl = config.getOauthTokenInfoUrl();
+        user = extractUser(request, config.getTokenUserField());
+    }
+
+    public Optional<String> getUser()
```
oneonestar, line 69: `@SuppressWarnings("unused")`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+        return user;
+    }
+
+    public Optional<UserInfo> getUserInfo()
+    {
+        return userInfo;
+    }
+
+    @SuppressWarnings("unused")
+    public boolean userEquals(String testUser)
```
oneonestar, line 80: Maybe `userExistsAndEquals`?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+                    if (node.has(userField)) {
+                        log.debug("Got user from cookie");
+                        return node.get(userField).asText();
+                    }
+                }
+                catch (JsonProcessingException e) {
+                    log.warn("Could not deserialize bearer token as json");
+                }
+            }
+            return null;
```
oneonestar, line 110: Could use lib instead.
```
DecodedJWT jwt = JWT.decode(cookie);
...
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    private Optional<String> extractUserFromAuthorizationHeader(String header, String userField)
+    {
+        if (header == null) {
+            return Optional.empty();
+        }
+
+        if (header.contains("Basic")) {
+            return Optional.of(new String(Base64.getDecoder().decode(header.split(" ")[1]), StandardCharsets.UTF_8).split(":")[0]);
```
oneonestar, line 137: Could be malformed and trigger an RuntimeException.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,205 @@
...
+            try {
+                JsonNode node = mapper.readTree(new String(Base64.getDecoder().decode(token)));
+                if (node.has(userField)) {
+                    log.debug("Trying to extract user from JWS json. User: %s", node.get(userField).asText());
+                    return Optional.of(node.get(userField).asText());
+                }
+            }
+            catch (JsonProcessingException e) {
+                log.warn("Could not deserialize bearer token as json");
+            }
```
oneonestar, line 169: Could use lib instead.

oneonestar (commented):
Overall looks promising.

docs/routing-rules.md
```diff
@@ -59,6 +62,57 @@ condition: 'request.getHeader("X-Trino-Client-Tags") contains "label=foo"'
...
+if a token is successfully exchanged with the `oauthTokenInfoUrl`, and an empty `Optional` otherwise.
+
+User extraction is only available if enabled by configuring `requestAnalyzerConfig.isAnalyzeRequest = True`
+
+### TrinoQueryProperties
+
+This class attempts to parse the body of a request as SQL. Note that only a syntactic analysis is performed! If a query
+references a view, then that view will not be expanded, and tables referenced by the view will not be recognized.
+A routing rule can call the following methods on the `trinoQueryProperties` object:   
+* `boolean isQuery()`: is the request a valid SQL query. Note that this will return false for invalid SQL.
```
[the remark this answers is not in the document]
willmostly: makes sense. Your interpretation was correct btw, I was equating successful parsing to syntactically valid SQL

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+import static com.google.common.io.BaseEncoding.base64Url;
+import static io.airlift.json.JsonCodec.jsonCodec;
+import static java.lang.Math.toIntExact;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElse;
+
+public class TrinoQueryProperties
+{
+    public static final String UNSET = "__UNSET__";
```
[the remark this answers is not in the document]
willmostly: Initially I thought [MVEL](http://mvel.documentnode.com/) lacked support for `Optional`, but it turns out you can call methods like `trinoRequestUser.getUser().isPresent()` in MVEL even though you can't cast to `Optional` or use its static methods. This should be updated

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+                body = "";
+                return;
+            }
+
+            Map<String, String> preparedStatements = getPreparedStatements(request);
+            SqlParser parser = new SqlParser();
+            reader.mark(config.getMaxBodySize());
+            body = CharStreams.toString(reader);
+
+            if (isClientsUseV2Format) {
```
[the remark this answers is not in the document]
willmostly: it is not possible to use the path unfortunately. We could remove the flag and have it try to parse the body as JSON for every request, or use a lighter weight check like `if (body.startsWith("{"))...`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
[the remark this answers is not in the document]
oneonestar: It is possible that someone might reuse these codes beside MVEL.
I think `Optional` is more future proof.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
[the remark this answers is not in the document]
oneonestar: I see. The alternatives are inefficient and difficult to understand. Keeping the flag makes more sense.

oneonestar:
Could you also address the other comments?
I think they're being hidden in the UI.
image

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,470 @@
...
+            if (reader == null) {
+                log.warn("HTTP request returned null reader");
+                body = "";
+                return;
+            }
+
+            Map<String, String> preparedStatements = getPreparedStatements(request);
+            SqlParser parser = new SqlParser();
+            reader.mark(config.getMaxBodySize());
+            body = CharStreams.toString(reader);
```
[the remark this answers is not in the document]
willmostly: the body size limit is there because, if i understand the semantics of `BufferedReader` correctly, reading is a destructive operation. I use the `mark` and `reset` methods to ensure I leave the `HttpServletRequest` in the same state that I received it.

If you read more than the`readAheadLimit` argument to `mark`, `reset` is not guaranteed to get you back to the mark. 

Prepared statements are request headers, which are plain Strings, so they don't need this handling.

willmostly:
> Could you also address the other comments? I think they're being hidden in the UI. 
They were, thanks for the thorough review!

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -66,23 +89,137 @@ public void testByRoutingGroupHeader()
...
+        Enumeration<String> encodedStatementsEnumeration = Collections.enumeration(Arrays.asList(encodedStatements.split(",")));
+        String body = "EXECUTE statement4";
+        RoutingGroupSelector routingGroupSelector =
+                RoutingGroupSelector.byRoutingRulesEngine("src/test/resources/rules/routing_rules_trino_query_properties.yml", requestAnalyzerConfig);
+        Reader reader = new StringReader(body);
+        BufferedReader bufferedReader = new BufferedReader(reader);
+        HttpServletRequest mockRequest = prepareMockRequest();
+        when(mockRequest.getReader()).thenReturn(bufferedReader);
+        when(mockRequest.getHeader(TrinoQueryProperties.TRINO_PREPARED_STATEMENT_HEADER_NAME)).thenReturn(encodedStatements);
+        when(mockRequest.getHeaders(TrinoQueryProperties.TRINO_PREPARED_STATEMENT_HEADER_NAME)).thenReturn(encodedStatementsEnumeration);
```
oneonestar, line 208: NIT: inline `encodedStatementsEnumeration`

gateway-ha/src/main/java/io/trino/gateway/ha/router/StatementUtils.java
```diff
@@ -0,0 +1,246 @@
...
+    public static Optional<QueryType> getQueryType(Statement statement)
+    {
+        if (statement instanceof ExplainAnalyze) {
+            return getQueryType(((ExplainAnalyze) statement).getStatement());
+        }
+        return Optional.ofNullable(STATEMENT_QUERY_TYPES.get(statement.getClass()))
+                .map(StatementTypeInfo::getQueryType);
+    }
+
+    public static Set<Class<? extends Statement>> getNonDataDefinitionStatements()
```
oneonestar, line 207: Unused function

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -0,0 +1,512 @@
...
+                    body = requestBody.getQuery();
+                    preparedStatements = requestBody.getPreparedStatements();
+                }
+                catch (IllegalArgumentException e) {
+                    // Do nothing, request is using standard format
+                }
+            }
+
+            Statement statement = parser.createStatement(body);
+            isQueryParsingSuccessful = true;
```
oneonestar, line 142: I think it is better to move this to the last line of the try block.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,224 @@
...
+                }
+            }
+            catch (JWTDecodeException e) {
+                log.warn("Could not deserialize bearer token as json");
+            }
+        }
+
+        if (oauthUserInfoUrl.isPresent()) {
+            try {
+                this.userInfo = Optional.of(userInfoCache.get(new UserInfoRequestArgs(token, oauthUserInfoUrl.orElseThrow())));
```
oneonestar, line 173: No need `this`

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -0,0 +1,224 @@
...
+    public static final String TRINO_UI_TOKEN_NAME = "Trino-UI-Token";
+    public static final String TRINO_SECURE_UI_TOKEN_NAME = "__Secure-Trino-ID-Token";
+
+    private Optional<String> user = Optional.empty();
+    private Optional<UserInfo> userInfo = Optional.empty();
+
+    private final Optional<String> oauthUserInfoUrl;
+    private static final Logger log = Logger.get(TrinoRequestUser.class);
+
+    private static final LoadingCache<UserInfoRequestArgs, UserInfo> userInfoCache =
```
oneonestar, line 56: Why use `UserInfoRequestArgs` (ie. `oauthUserInfoUrl` + `accessToken`) instead of just `accessToken` as hash key?
`oauthUserInfoUrl` is a constant which stay the same across different TrinoRequestUser instances.

oneonestar (commented):
Just a few more comments.
I think it’s good to be merged soon.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
willmostly: I made `oauthUserInfoUrl` a static to facilitate this, lmk if you agree with the approach or would prefer something else

oneonestar: How about using a provider to get rid of the `static` and synchronization issue like this: https://github.com/trinodb/trino-gateway/commit/117636462daaf38ae4f978e80203686f74cb5ec4

oneonestar (approved):
LGTM =D

PR #326, opened 2024-04-25
Add option for custom v1/statement-like paths

willmostly:
## Description

Add an option to configure additional custom paths that should be treated like `/v1/statement`. This is required to support some features of Starburst Trino.

The trino gateway checks if a request path starts with `/v1/statement` to decide if it should use query processing logic. 

## Release notes

( x) Release notes are required, with the following suggested text:
Add option for custom v1/statement-like paths


willmostly:
The runtime changes introduced here are limited. However, `QueryIdCachingProxyHandler` units tests were expanded to cover instance methods, which required some significant refactoring. The functionality of `RoutingManager` and `ProxyHandlerStats` should not be changed, but interfaces were extracted so that testing versions of these classes could be introduced.

ebyhr:
> This is required to support some features of Starburst Trino.

Is it possible to test the feature with https://hub.docker.com/r/starburstdata/starburst-enterprise?

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -37,6 +37,7 @@ public class HaGatewayConfiguration
     private List<String> extraWhitelistPaths = new ArrayList<>();
     private OAuth2GatewayCookieConfiguration oauth2GatewayCookieConfiguration = new OAuth2GatewayCookieConfiguration();
     private GatewayCookieConfiguration gatewayCookieConfiguration = new GatewayCookieConfiguration();
+    private List<String> customStatementPaths = new ArrayList<>();
```
mosabua, line 40: I think it would be better to have a statementPaths setup and it defaults to `v1/statement` .. and users can override or expand to multiple paths

gateway-ha/src/main/java/io/trino/gateway/ha/handler/DropWizardProxyHandlerStats.java
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.handler;
+
+import com.codahale.metrics.Meter;
+import io.airlift.stats.CounterStat;
+import io.dropwizard.core.setup.Environment;
```
mosabua, line 18: lets not add new Dropwizard dependencies in new PRs

gateway-ha/src/main/java/io/trino/gateway/ha/handler/DropWizardProxyHandlerStats.java
```diff
@@ -0,0 +1,66 @@
...
+import org.weakref.jmx.MBeanExporter;
+import org.weakref.jmx.Managed;
+import org.weakref.jmx.Nested;
+
+import java.lang.management.ManagementFactory;
+
+import static java.util.Objects.requireNonNull;
+
+public final class DropWizardProxyHandlerStats
```
mosabua, line 27: rename to Jetty or airlift .. depending on how you change

mosabua (changes requested):
Just some initial feedback after taking a glance.

willmostly:
> > This is required to support some features of Starburst Trino.
> 
> Is it possible to test the feature with https://hub.docker.com/r/starburstdata/starburst-enterprise?

It is not, the Insights IDE feature that exposes the non-standard statement endpoint requires a license

gateway-ha/src/main/java/io/trino/gateway/ha/handler/DropWizardProxyHandlerStats.java
willmostly: Sorry - I should have explained that this is a refactor of existing code to allow for testing. 

This class was `ProxyHandlerStats`, which was a pain to use in unit tests because of the Dropwizard setup. So I made `ProxyHandlerStats` into an interface, made this class an implementation of it, and added the `TestingProxyHandlerStats` implementation.

mosabua:
> > > This is required to support some features of Starburst Trino.
> > 
> > 
> > Is it possible to test the feature with https://hub.docker.com/r/starburstdata/starburst-enterprise?
> 
> It is not, the Insights IDE feature that exposes the non-standard statement endpoint requires a license

Also we dont want to start testing with various distributions - that is something the distributors should do in their own CI setup.

mosabua:
Lets discuss in the dev sync tomorrow.. I think we should merge this after adding docs .. or get the airlift removal done first and refactor this after.

mosabua:
Another one to rebase and then merge @willmostly  ;-)

mosabua (commented):
I think the PR looks mostly good, however I think we need to add tests and potentially make the code more robust. I think we should allow something like the following as well

* `v1/statement`
* `somerandomcontext/v1/statement`

Or alternatively require the root `/` symbol and verify for it.

And with multiple also allows mixing like

* `/v1/statement`,  `v1/statement`, `blabla/v2/statement`, `some/custom/path`

Does that make sense?

Also.. we are not verifying at all if `v1/statement` is in the list and I think that is fine .. however do we want to log a warning about that? Or do we want to require is thinking that if that URL is not used its not a valid trino cluster or so .. I think thats wrong but feel free to convince me otherwise.

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -225,4 +231,16 @@ public void setManagedApps(List<String> managedApps)
...
+    public List<String> getStatementPaths()
+    {
+        return statementPaths;
+    }
+
+    public void setStatementPaths(List<String> statementPaths)
+    {
+        // remove trailing slashes to ensure predictable behavior when splitting on "/"
+        this.statementPaths =
+                ImmutableList.copyOf(Stream.concat(statementPaths.stream(), this.statementPaths.stream()).map(s -> s.replaceAll("/$", "")).toList());
```
oneonestar, line 244: This appends the received argument to `statementPaths` each time the function is called. This is not an usual behavior for a setter. This function could potentially be called multiple times, eg. during testing.
We could simply set `statementPaths` to `V1_STATEMENT_PATH + <received argument>`.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -116,28 +116,29 @@ public static String extractQueryIdIfPresent(HttpServletRequest request)
...
     {
         if (path == null) {
             return null;
         }
         String queryId = null;
-
         log.debug("Trying to extract query id from path [%s] or queryString [%s]", path, queryParams);
-        if (path.startsWith(V1_STATEMENT_PATH) || path.startsWith(V1_QUERY_PATH)) {
+        Optional<String> matchingStatementPath = statementPaths.stream().filter(path::startsWith).findAny();
+        if (matchingStatementPath.isPresent() || path.startsWith(V1_QUERY_PATH)) {
```
oneonestar, line 130: Could we add comments that show some example paths?

willmostly:
> Also.. we are not verifying at all if v1/statement is in the list and I think that is fine .. however do we want to log a warning about that? Or do we want to require is thinking that if that URL is not used its not a valid trino cluster or so .. I think thats wrong but feel free to convince me otherwise.

`/v1/statement` is always added to the list. I'll document that behavior. Do you think changing the config name to `additionalStatementPaths` would be helpful?

> I think we need to add tests and potentially make the code more robust

I have 
```
assertThat(extractQueryIdIfPresent("/custom/api/statement/executing/20200416_160256_03078_6b4yt/ya7e884929c67cdf86207a80e7a77ab2166fa2e7b/1368", null, statementPaths))
```
in `TestQueryIdCachingProxyHandler` and tests for the config validation. Would you like to see an end-to-end test as well? I can mock up a backend with a custom endpoint.

willmostly:
> Also.. we are not verifying at all if v1/statement is in the list and I think that is fine .. however do we want to log a warning about that? Or do we want to require is thinking that if that URL is not used its not a valid trino cluster or so .. I think thats wrong but feel free to convince me otherwise.

`/v1/statement` is always added to the list. I'll document that behavior. Do you think changing the config name to `additionalStatementPaths` would be helpful?

mosabua:
No need for a mock test imho.

willmostly:
@mosabua updated and added short doc section

mosabua (approved):
Looks good to me now. Please check on the build failure though..

willmostly:
> Please check on the build failure though..

I should have saved the error message, but the build failed because it downloaded a dependency and the archive file was corrupted. It succeeded on re-run

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -225,4 +231,36 @@ public void setManagedApps(List<String> managedApps)
...
+    public List<String> getStatementPaths()
+    {
+        return statementPaths;
+    }
+
+    public void setAdditionalStatementPaths(List<String> statementPaths)
+    {
+        // remove trailing slashes to ensure predictable behavior when splitting on "/"
+        this.statementPaths = Streams.concat(ImmutableList.of(V1_STATEMENT_PATH).stream(),
+                statementPaths.stream().peek(s -> validateStatementPath(s, statementPaths)).map(s -> s.replaceAll("/$", ""))).toList();
```
oneonestar, line 244: This only remove the last trailing slash, not all trailing slashes as stated in the comment.

mosabua:
I think we should merge this .. any objections @willmostly @vishalya @Chaho12 @oneonestar ?

oneonestar (approved):

docs/installation.md
```diff
@@ -138,6 +138,23 @@ extraWhitelistPaths:
   - '/ext/faster'
 ```
 
+### Configure additional v1/statement-like paths
+
+The Trino client protocol specifies that queries are initiated by a POST to `v1/statement`.
```
Chaho12, line 143:
```suggestion
The Trino client protocol specifies that queries are initiated by a POST to `/v1/statement`. 
```
Might be good to say that path should start with `/` as it will give an error anyway

Chaho12 (approved):
LGTM

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -225,4 +231,36 @@ public void setManagedApps(List<String> managedApps)
...
+        if (statementPath.startsWith(V1_STATEMENT_PATH) ||
+                statementPaths.stream().filter(s -> !s.equals(statementPath)).anyMatch(s -> s.startsWith(statementPath))) {
+            throw new HaGatewayConfigurationException("Statement paths cannot be prefixes of other statement paths");
+        }
+        if (!statementPath.startsWith("/")) {
+            throw new HaGatewayConfigurationException("Statement paths must be absolute");
+        }
+    }
+
+    public static class HaGatewayConfigurationException
```
ebyhr, line 258: An existing `IllegalArgumentException` is sufficient for me.

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -225,4 +231,36 @@ public void setManagedApps(List<String> managedApps)
...
+
+    private void validateStatementPath(String statementPath, List<String> statementPaths)
+    {
+        if (statementPath.startsWith(V1_STATEMENT_PATH) ||
+                statementPaths.stream().filter(s -> !s.equals(statementPath)).anyMatch(s -> s.startsWith(statementPath))) {
+            throw new HaGatewayConfigurationException("Statement paths cannot be prefixes of other statement paths");
+        }
+        if (!statementPath.startsWith("/")) {
+            throw new HaGatewayConfigurationException("Statement paths must be absolute");
+        }
```
ebyhr, line 255: I don't understand why a dedicated exception class was needed. Can we use `checkArgument` method instead?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -36,24 +36,26 @@
...
 
     public RoutingTargetHandler(
             RoutingManager routingManager,
             RoutingGroupSelector routingGroupSelector,
+            List<String> statementPaths,
             List<String> extraWhitelistPaths)
     {
         this.routingManager = requireNonNull(routingManager);
         this.routingGroupSelector = requireNonNull(routingGroupSelector);
+        this.statementPaths = requireNonNull(statementPaths);
```
ebyhr, line 58: Add `ImmutableList.copyOf`.

gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java
```diff
@@ -98,6 +99,7 @@ public ProxyRequestHandler(
         cookiesEnabled = GatewayCookieConfigurationPropertiesProvider.getInstance().isEnabled();
         asyncTimeout = haGatewayConfiguration.getRouting().getAsyncTimeout();
         addXForwardedHeaders = haGatewayConfiguration.getRouting().isAddXForwardedHeaders();
+        statementPaths = haGatewayConfiguration.getStatementPaths();
```
ebyhr, line 102: Add `ImmutableList.copyOf`.

gateway-ha/src/test/java/io/trino/gateway/ha/config/TestHaGatewayConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Test
+    void testAdditionalPath()
+    {
+        HaGatewayConfiguration haGatewayConfiguration = new HaGatewayConfiguration();
+        List<String> additionalPaths = ImmutableList.of("/ui/api/insights/ide/statement/", "/v2/statement");
+        haGatewayConfiguration.setAdditionalStatementPaths(additionalPaths);
+        assertThat(haGatewayConfiguration.getStatementPaths())
+                .containsAll(ImmutableList.of(V1_STATEMENT_PATH, "/ui/api/insights/ide/statement", "/v2/statement"));
```
ebyhr, line 41: Please consider using `containsExactly` or `containsOnly` without `ImmutableList.of`.

PR #332, opened 2024-05-03
Initialize the null userQueueCount map

vishalya:
## Description
The new QueryCountBasedRouter router fails to route any request due to uninitialized userQueryCount map attribute.

My guess is that the refactoring for Record to store the cluster stats removed the initialization.
This PR fixes the initialization of the userQueryCount map in the QueryCountBasedRouter. 
Also, modified the tests to take care of the scenario.

## Release notes

This is a bug fix
() Release notes are required, with the following suggested text:

```markdown
* 
```

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
@@ -47,7 +47,7 @@ public class TestQueryCountBasedRouter
     // Helper function to generate the ClusterStat list
     private static List<ClusterStats> getClusterStatsList(String routingGroup)
     {
-        ImmutableList.Builder<ClusterStats> clustersBuilder = new ImmutableList.Builder();
+        ImmutableList.Builder<ClusterStats> clustersBuilder = new ImmutableList.Builder<ClusterStats>();
         // Set Cluster1 stats
         {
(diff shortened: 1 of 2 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
```diff
@@ -130,11 +130,10 @@ static ClusterStats getClusterWithNoUserQueueAndMinQueueCount()
         cluster.routingGroup("adhoc");
         cluster.runningQueryCount(5);
         cluster.queuedQueryCount(LEAST_QUEUED_COUNT);
-        cluster.userQueuedCount(new HashMap<String, Integer>());
```
mosabua, line 133: I assume this is triggers the scenario in the test now .. do we need another non-null test?

mosabua (approved):
Looks good .. I do think we need to merge this for the release since the issue could hit users otherwise.. and we would need a release notes entry for this as a fix in the next release. 

Wdyt @willmostly and @Chaho12 ?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
vishalya: That's correct. This is sufficient, I verified that it does validate the scenario which is being fixed.

PR #343, opened 2024-05-14
Use Nimbus OAuth SDK for OIDC

oneonestar:
## Description
Use Nimbus OAuth SDK for OIDC. The same SDK is also used by Trino.
This is only a refactoring. Security issues related to OIDC will be addressed in another PR.

Ref: https://connect2id.com/products/nimbus-oauth-openid-connect-sdk/examples/openid-connect/token-request

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #345, opened 2024-05-15
Add state and nonce verification in OIDC

### LINKED ISSUE #339: OIDC login implementation isn't up to standard
Things that are missing:
* unique session token that holds state
* set and verify `nonce` and `state` during authentication

I think what I'm going to do is reimplement the OIDC login using OAuth library (nimbusds).

---
Ref:
https://github.com/trinodb/trino-gateway/blob/53ec04bd043fbb5a5d11c25472b091ba398b3608/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java#L119

```
$ curl -X POST http://localhost:9080/sso
{"code":200,"msg":"Ok","data":"http://localhost:4444/oauth2/auth?client_id=trino_client_id&response_type=code&redirect_uri=http://localhost:9080/oidc/callback&scope=openid"}
```

`state` should be included in the OAuth request redirect.
It is required by the spec and some OAuth provider reject the request if `state` field is missing.

For example, Hydra returns the following error:
`The state is missing or does not have enough characters and is therefore considered too weak. Request parameter "state" must be at least be 8 characters long to ensure sufficient entropy.`

Ref:
* https://developers.google.com/identity/openid-connect/openid-connect
* https://developer.okta.com/blog/2019/08/22/okta-authjs-pkce#openid-connect-and-oauth-20-overview
* https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics

Implementation in Trino:
https://github.com/trinodb/trino/blob/ae789c04f5995dcd87efe6b5e1862521c2ad6957/core/trino-main/src/main/java/io/trino/server/security/oauth2/OAuth2Service.java#L151

oneonestar:
## Description
Fix #339 
Depends on #343
Add state and nonce verification and integration test in OIDC.

### State
Ref: https://developers.google.com/identity/openid-connect/openid-connect#server-flow
Correct implementation of OIDC:
1. Create an anti-forgery state token (implemented by this PR using `OidcCookie`)
2. Send an authentication request to Google
3. Confirm the anti-forgery state token (implemented by this PR using `state`)
4. Exchange code for access token and ID token
5. Obtain user information from the ID token
6. Authenticate the user

#### Nonce
Ref: https://openid.net/specs/openid-connect-core-1_0-17_orig.html#NonceNotes
> The nonce parameter value needs to include per-session state and be unguessable to attackers. One method to achieve this for Web Server Clients is to store a cryptographically random value as an HttpOnly session cookie and use a cryptographic hash of the value as the nonce parameter. In that case, the nonce in the returned ID Token is compared to the hash of the session cookie to detect ID Token replay by third parties.

Implemented by this PR using `nonce`, `hashNonce()` and  `OidcCookie`.

## Release notes

```markdown
* 
```


mosabua (approved):
Thats quite the complex setup for functionality and testing. From my perspective this looks good though, since OIDC just is tricky.

PR #349, opened 2024-05-20
Centralize the place to log unhealthy clusters

andythsu:
## Description

As different routing rules are implemented, `routingManager.updateBackEndStats` method could be overridden. An example is https://github.com/trinodb/trino-gateway/blob/c052bb237450d7fc2f2f7af1b06f7fc8f4cdee48/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java#L238-L241

Therefore, it will be more useful to log unhealthy clusters at `ActiveClusterMonitor` level instead of `RoutingManager` level 

## Release notes

```markdown
* 
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
@@ -79,6 +79,10 @@ public class ActiveClusterMonitor
                             List<ClusterStats> stats = new ArrayList<>();
                             for (Future<ClusterStats> clusterStatsFuture : futures) {
                                 ClusterStats clusterStats = clusterStatsFuture.get();
+                                if (!clusterStats.healthy()) {
+                                    log.error("cluster %s is unhealthy", clusterStats.clusterId());
+                                    log.debug("cluster stats: %s", clusterStats);
+                                }
                                 stats.add(clusterStats);
                             }
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
@@ -39,9 +39,9 @@ public class ActiveClusterMonitor
 
     private final int taskDelaySeconds;
     private final ClusterStatsMonitor clusterStatsMonitor;
-    private volatile boolean monitorActive = true;
     private final ExecutorService executorService = Executors.newFixedThreadPool(DEFAULT_THREAD_POOL_SIZE);
     private final ExecutorService singleTaskExecutor = Executors.newSingleThreadExecutor();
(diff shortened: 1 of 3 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
```diff
@@ -39,9 +39,9 @@ public class ActiveClusterMonitor
 
     private final int taskDelaySeconds;
     private final ClusterStatsMonitor clusterStatsMonitor;
-    private volatile boolean monitorActive = true;
     private final ExecutorService executorService = Executors.newFixedThreadPool(DEFAULT_THREAD_POOL_SIZE);
     private final ExecutorService singleTaskExecutor = Executors.newSingleThreadExecutor();
+    private volatile boolean monitorActive = true;
```
andythsu, line 44: this change is done through auto format

ebyhr: Please separate a commit.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthCheckObserver.java
```diff
@@ -26,7 +28,7 @@ public HealthCheckObserver(RoutingManager routingManager)
     }
 
     @Override
-    public void observe(java.util.List<ClusterStats> clustersStats)
+    public void observe(List<ClusterStats> clustersStats)
```
ebyhr, line 31: Please separate a commit.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
```diff
@@ -79,6 +79,10 @@ public void start()
                             List<ClusterStats> stats = new ArrayList<>();
                             for (Future<ClusterStats> clusterStatsFuture : futures) {
                                 ClusterStats clusterStats = clusterStatsFuture.get();
+                                if (!clusterStats.healthy()) {
+                                    log.error("cluster %s is unhealthy", clusterStats.clusterId());
+                                    log.debug("cluster stats: %s", clusterStats);
```
willmostly, line 84: Will this print the stats, or just the default `Object:toString`?

willmostly (changes requested):
One question, otherwise this looks good to me

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
andythsu: on a second thought, I'm not sure whether this will add more value, or creating another observer will. I'm thinking we can create a new observer to just alert when no clusters are available instead of printing each cluster's stats. WDYT

PR #352, opened 2024-05-22
Capitalize page permission key conf

Chaho12:
## Description

Since both roles in [web & security context are in upper-case](https://github.com/trinodb/trino-gateway/commit/d4992c86ae31b5ebdba416f1635185815fb77a47), page permission should also be in upper case to [match with roles](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java#L157). 

Without this patch, user has to set upper-case key's in yaml in their yaml file, which is awkward as all other auth related confs are in lower-case. 

```
pagePermissions:
  ADMIN:
  USER: dashboard_history
  API:
```

## Release notes

```markdown
* 
```


ebyhr:
Please check CI failure:
```
Error:    TestLbAuthenticator.testNoLdapNoPresetUsers:122 » NullPointer Cannot invoke "java.util.Map.forEach(java.util.function.BiConsumer)" because "pagePermissions" is null
Error:    TestLbAuthenticator.testWrongLdapConfig:131 » NullPointer Cannot invoke "java.util.Map.forEach(java.util.function.BiConsumer)" because "pagePermissions" is null
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -48,7 +49,10 @@ public LbFormAuthManager(FormAuthConfiguration configuration,
             Map<String, String> pagePermissions)
     {
         this.presetUsers = presetUsers;
-        this.pagePermissions = pagePermissions;
+        this.pagePermissions = new HashMap<>();
+        if (pagePermissions != null) {
```
ebyhr, line 53: Is this really nullable in product code?

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -48,7 +49,10 @@ public LbFormAuthManager(FormAuthConfiguration configuration,
...
+            pagePermissions.forEach((key, value) -> this.pagePermissions.put(key.toUpperCase(), value));
```
ebyhr, line 54: Use immutable map.

Chaho12: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
Chaho12: I noticed that it is not nullable in production. (value though could be null so ignored it)
Tested with giving below conf.
```
pagePermissions:
  admin:
  user: dashboard_history
  api:
```

ebyhr:
Please fix CI failure:
```
Error:    TestLbAuthenticator.testNoLdapNoPresetUsers:122 » NullPointer Cannot invoke "java.util.Map.entrySet()" because "pagePermissions" is null
Error:    TestLbAuthenticator.testWrongLdapConfig:131 » NullPointer Cannot invoke "java.util.Map.entrySet()" because "pagePermissions" is null
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
Chaho12: should we rather change to test code so that it doesn't send null? or have sth like below?
```java
this.pagePermissions = Objects.requireNonNullElse(pagePermissions, Map.<String, String>of()).entrySet().stream()
            .filter(entry -> entry.getValue() != null)  // Filter out entries with null values
            .collect(toImmutableMap(
                entry -> entry.getKey().toUpperCase(),
                Map.Entry::getValue
            ));
```

ebyhr: We should change the test in this case.

Chaho12: didn't know that ide was giving a warnging about  `@NotNull` annotations too. i changed it to new HashMap 
image

ytwp (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
ebyhr: > i changed it to new HashMap

Please avoid mutable map.

Chaho12: i don't quite understand when you mean by `adding test with null values` if pagePermission cannot be null in anycase. what is the purpose of having null value??
ps. changed to use immutable map in test case

ebyhr: I mean covering the following line:
```java
.filter(entry -> entry.getValue() != null)  // Filter out entries with null values
```

Chaho12: if i add a test for map with null values, it fails to build with following error
```java
    @Test
    public void testNullValuesInPagePermission()
        throws Exception
    {
        Map<String, UserConfiguration> presetUsers = ImmutableMap.of(
                "user1", new UserConfiguration("admin, user, api", "pass1"));
        Map<String, String> pagePermission = ImmutableMap.of(
                "admin", "",
                "user", null,
                "api", "");
        LbFormAuthManager authentication = new LbFormAuthManager(null, presetUsers, pagePermission);
        assertThat(authentication.authenticate(new BasicCredentials("user1", "pass1")))
                .isTrue();
    }
```
```
/Users/user/work/trino-gateway-chaho/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbAuthenticator.java:144:25
java: [NullArgumentForNonNullParameter] Null is not permitted for this parameter.
    (see https://errorprone.info/bugpattern/NullArgumentForNonNullParameter)
```

ebyhr: Please use `unmodifiableMap` with `HashMap`.

Chaho12: fixed using unmodifiableMap & hashmap. 
- test code checks for null values for either key/value for page permission.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -48,7 +50,11 @@ public LbFormAuthManager(FormAuthConfiguration configuration,
             Map<String, String> pagePermissions)
     {
         this.presetUsers = presetUsers;
-        this.pagePermissions = pagePermissions;
+        this.pagePermissions = pagePermissions.entrySet().stream()
+                .filter(entry -> entry.getKey() != null && entry.getValue() != null)
```
ebyhr, line 54: When do we receive null keys?

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbAuthenticator.java
```diff
@@ -128,11 +129,28 @@ public void testNoLdapNoPresetUsers()
...
     {
-        LbFormAuthManager authentication = new LbFormAuthManager(null, null, null);
+        LbFormAuthManager authentication = new LbFormAuthManager(null, null, ImmutableMap.of());
         assertThat(authentication.authenticate(new BasicCredentials("user1", "pass1")))
                 .isFalse();
     }
 
+    @Test
+    public void testNullInPagePermission()
+        throws Exception
```
ebyhr, line 139: Indentation is wrong. Also, this line is redundant. Remove.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
ebyhr: Please throw an exception if it doesn't happen in a real situation.

ebyhr:
Please fix CI failure:
```
Error:  src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java:[52,1] (imports) ImportOrder: Wrong order for 'com.google.common.collect.ImmutableMap.toImmutableMap' import.

```

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java
```diff
@@ -67,7 +68,18 @@ public class LbOAuthManager
...
+                        entry -> entry.getKey().toUpperCase(),
+                        Map.Entry::getValue));
+    }
+
+    public static Response buildUnauthorizedResponse()
+    {
+        return Response.status(UNAUTHORIZED)
+                .cookie(OidcCookie.delete())
+                .build();
     }
```
Chaho12, line 83: Formatted code with airlift style changed function location

ebyhr: No need to move the method. Please revert.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
Chaho12: Sure, i added try-catch NPE when doing UpperCase for key.

ebyhr: No need to add try-catch block in my opinion. I would just throw an exception.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java
```diff
@@ -201,4 +215,5 @@ public static Response buildUnauthorizedResponse()
                 .cookie(OidcCookie.delete())
                 .build();
     }
+
```
ebyhr, line 218: Remove a redundant empty line.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
ebyhr: Reminder.

ebyhr: I pushed changes instead.

Chaho12: Missed it. removed try-catch

ebyhr (approved):

ebyhr:
I pushed small changes because the indentation was wrong.

Issue #359, opened 2024-05-24
Retire BackendStateManager


andythsu:
It looks like the purpose of BackendStateManager class is to provide states of clusters to frontend. We can use RouterManager to achieve the same result.

PR #360, opened 2024-05-24
Add PENDING type to healthchecks

andythsu:
## Description

Resolves https://github.com/trinodb/trino-gateway/issues/359 part 1

## Release notes

```markdown
* 
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
@@ -68,6 +68,9 @@ public abstract class BaseApp
 {
     private static final Logger logger = Logger.get(BaseApp.class);
     private final ImmutableList.Builder<Module> appModules = ImmutableList.builder();
+    // this injector reference is needed to use reflection in
+    // TestGatewayHaSingleBackend and TestGatewayMultipleBackend
+    private Injector injector;
 
     private Module newModule(String clazz, HaGatewayConfiguration configuration, Environment environment)
     {
@@ -129,7 +132,7 @@ public abstract class BaseApp
     {
         appModules.add(new MetricRegistryModule(environment.metrics()));
         appModules.addAll(addModules(configuration, environment));
-        Injector injector = Guice.createInjector(appModules.build());
+        injector = Guice.createInjector(appModules.build());
         injector.injectMembers(this);
         registerWithInjector(configuration, environment, injector);
     }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
@@ -24,7 +24,7 @@ public record ClusterStats(
         int runningQueryCount,
         int queuedQueryCount,
         int numWorkerNodes,
-        boolean healthy,
+        TrinoHealthStateType healthy,
         String proxyTo,
         String externalUrl,
         String routingGroup,
@@ -41,7 +41,7 @@ public record ClusterStats(
         private int runningQueryCount;
         private int queuedQueryCount;
         private int numWorkerNodes;
-        private boolean healthy;
+        private TrinoHealthStateType healthy;
         private String proxyTo;
         private String externalUrl;
         private String routingGroup;
@@ -70,7 +70,7 @@ public record ClusterStats(
             return this;
         }
 
-        public Builder healthy(boolean healthy)
+        public Builder healthy(TrinoHealthStateType healthy)
         {
             this.healthy = healthy;
             return this;
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
@@ -46,7 +46,7 @@ public class ClusterStatsInfoApiMonitor
                 .routingGroup(backend.getRoutingGroup()).build();
     }
 
-    private boolean isReadyStatus(String baseUrl)
+    private TrinoHealthStateType isReadyStatus(String baseUrl)
     {
         Request request = prepareGet()
                 .setUri(uriBuilderFrom(URI.create(baseUrl)).appendPath("/v1/info").build())
@@ -54,11 +54,11 @@ public class ClusterStatsInfoApiMonitor
 
         try {
             ServerInfo serverInfo = client.execute(request, SERVER_INFO_JSON_RESPONSE_HANDLER);
-            return !serverInfo.isStarting();
+            return serverInfo.isStarting() ? TrinoHealthStateType.UNHEALTHY : TrinoHealthStateType.HEALTHY;
         }
         catch (Exception e) {
             log.error("Exception checking {} for health: {} ", request.getUri(), e.getMessage());
         }
-        return false;
+        return TrinoHealthStateType.UNHEALTHY;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthChecker.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthChecker.java
@@ -34,7 +34,7 @@ public class HealthChecker
     public void observe(List<ClusterStats> clustersStats)
     {
         for (ClusterStats clusterStats : clustersStats) {
-            if (!clusterStats.healthy()) {
+            if (clusterStats.healthy() == TrinoHealthStateType.UNHEALTHY) {
                 notifyUnhealthyCluster(clusterStats);
             }
             else {
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
@@ -1,0 +1,21 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+public enum TrinoHealthStateType
+{
+    PENDING, // when clusters are switched from inactive to active
+    HEALTHY, // when health checks report clusters as up
+    UNHEALTHY // when health checks report clusters as down
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
@@ -91,7 +91,7 @@ public class EntityEditorResource
                     ProxyBackendConfiguration backend =
                             OBJECT_MAPPER.readValue(jsonPayload, ProxyBackendConfiguration.class);
                     gatewayBackendManager.updateBackend(backend);
-                    log.info("Setting up the backend %s with healthy state", backend.getName());
+                    log.info("Turning cluster %s %s", backend.getName(), backend.isActive() ? "on" : "off");
                     routingManager.updateBackEndHealth(backend.getName(), backend.isActive());
                     backendStateManager.updateStates(
                             backend.getName(),
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
@@ -18,7 +18,10 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
+import io.trino.gateway.ha.clustermonitor.ClusterStats;
+import io.trino.gateway.ha.clustermonitor.TrinoHealthStateType;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import io.trino.gateway.ha.router.BackendStateManager;
 import io.trino.gateway.ha.router.GatewayBackendManager;
 import io.trino.gateway.ha.router.ResourceGroupsManager;
 import io.trino.gateway.ha.router.RoutingManager;
@@ -52,13 +55,15 @@ public class EntityEditorResource
     private final GatewayBackendManager gatewayBackendManager;
     private final ResourceGroupsManager resourceGroupsManager;
     private final RoutingManager routingManager;
+    private final BackendStateManager backendStateManager;
 
     @Inject
-    public EntityEditorResource(GatewayBackendManager gatewayBackendManager, ResourceGroupsManager resourceGroupsManager, RoutingManager routingManager)
+    public EntityEditorResource(GatewayBackendManager gatewayBackendManager, ResourceGroupsManager resourceGroupsManager, RoutingManager routingManager, BackendStateManager backendStateManager)
     {
         this.gatewayBackendManager = requireNonNull(gatewayBackendManager, "gatewayBackendManager is null");
         this.resourceGroupsManager = requireNonNull(resourceGroupsManager, "resourceGroupsManager is null");
         this.routingManager = requireNonNull(routingManager, "routingManager is null");
+        this.backendStateManager = requireNonNull(backendStateManager, "backendStateManager is null");
     }
 
     @GET
@@ -88,6 +93,11 @@ public class EntityEditorResource
                     gatewayBackendManager.updateBackend(backend);
                     log.info("Setting up the backend %s with healthy state", backend.getName());
                     routingManager.updateBackEndHealth(backend.getName(), backend.isActive());
+                    backendStateManager.updateStates(
+                            backend.getName(),
+                            ClusterStats.builder(backend.getName())
+                                    .healthy(backend.isActive() ? TrinoHealthStateType.PENDING : TrinoHealthStateType.UNHEALTHY)
+                                    .build());
                     break;
                 case RESOURCE_GROUP:
                     ResourceGroupsDetail resourceGroupDetails = OBJECT_MAPPER.readValue(jsonPayload,
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
@@ -18,6 +18,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
+import io.trino.gateway.ha.clustermonitor.TrinoHealthStateType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -46,7 +47,7 @@ public class QueryCountBasedRouter
     {
         private int runningQueryCount;
         private int queuedQueryCount;
-        private boolean healthy;
+        private TrinoHealthStateType healthy;
         private String proxyTo;
         private String routingGroup;
         private String clusterId;
@@ -93,12 +94,12 @@ public class QueryCountBasedRouter
             this.queuedQueryCount = queuedQueryCount;
         }
 
-        public boolean healthy()
+        public TrinoHealthStateType healthy()
         {
             return this.healthy;
         }
 
-        public void healthy(boolean healthy)
+        public void healthy(TrinoHealthStateType healthy)
         {
             this.healthy = healthy;
         }
@@ -187,7 +188,7 @@ public class QueryCountBasedRouter
     {
         log.debug("sorting cluster stats for {} {}", user, routingGroup);
         List<LocalStats> filteredList = clusterStats.stream()
-                    .filter(stats -> stats.healthy())
+                    .filter(stats -> stats.healthy() == TrinoHealthStateType.HEALTHY)
                     .filter(stats -> routingGroup.equals(stats.routingGroup()))
                     .collect(Collectors.toList());
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
@@ -18,6 +18,7 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
+import io.trino.gateway.ha.clustermonitor.TrinoHealthStateType;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.proxyserver.ProxyServerConfiguration;
 import jakarta.ws.rs.HttpMethod;
@@ -133,7 +134,7 @@ public abstract class RoutingManager
     public void updateBackEndStats(List<ClusterStats> stats)
     {
         for (ClusterStats clusterStats : stats) {
-            updateBackEndHealth(clusterStats.clusterId(), clusterStats.healthy());
+            updateBackEndHealth(clusterStats.clusterId(), clusterStats.healthy() == TrinoHealthStateType.HEALTHY);
         }
     }
 
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
@@ -15,9 +15,12 @@ package io.trino.gateway.ha;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
+import com.google.inject.Injector;
+import io.trino.gateway.baseapp.BaseApp;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.router.GatewayCookie;
 import io.trino.gateway.ha.router.OAuth2GatewayCookie;
+import io.trino.gateway.ha.router.RoutingManager;
 import okhttp3.Cookie;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
@@ -33,6 +36,7 @@ import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.testcontainers.containers.TrinoContainer;
 
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.util.Base64;
 import java.util.List;
 import java.util.Optional;
@@ -85,7 +89,8 @@ public class TestGatewayHaMultipleBackend
 
         // Start Gateway
         String[] args = {"server", testConfig.configFilePath()};
-        HaGatewayLauncher.main(args);
+        HaGatewayLauncher haGatewayLauncher = new HaGatewayLauncher();
+        haGatewayLauncher.run(args);
         // Now populate the backend
         HaGatewayTestUtils.setUpBackend(
                 "trino1", "http://localhost:" + backend1Port, "externalUrl", true, "adhoc", routerPort);
@@ -95,6 +100,14 @@ public class TestGatewayHaMultipleBackend
         HaGatewayTestUtils.setUpBackend(
                 "custom", "http://localhost:" + customBackendPort, "externalUrl", true, "custom",
                 routerPort);
+
+        // use injector to manually set backends to healthy
+        Field privateInjectorField = BaseApp.class.getDeclaredField("injector");
+        privateInjectorField.setAccessible(true);
+        Injector injector = (Injector) privateInjectorField.get(haGatewayLauncher);
+        injector.getInstance(RoutingManager.class).updateBackEndHealth("trino1", true);
+        injector.getInstance(RoutingManager.class).updateBackEndHealth("trino2", true);
+        injector.getInstance(RoutingManager.class).updateBackEndHealth("custom", true);
     }
 
     @Test
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
@@ -71,7 +71,7 @@ public class ClusterStatsHttpMonitor
                     .numWorkerNodes(activeWorkers)
                     .queuedQueryCount((int) result.get("queuedQueries"))
                     .runningQueryCount((int) result.get("runningQueries"))
-                    .healthy(activeWorkers > 0)
+                    .healthy(activeWorkers > 0 ? TrinoHealthStateType.HEALTHY : TrinoHealthStateType.UNHEALTHY)
                     .proxyTo(backend.getProxyTo())
                     .externalUrl(backend.getExternalUrl())
                     .routingGroup(backend.getRoutingGroup());
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
@@ -82,7 +82,9 @@ public class ClusterStatsJdbcMonitor
                 partialState.put(rs.getString("state"), rs.getInt("count"));
             }
             return clusterStats
-                    .healthy(true)
+                    // at this point we can set cluster to healthy because otherwise
+                    // it wouldn't have gotten worker stats
+                    .healthy(TrinoHealthStateType.HEALTHY)
                     .queuedQueryCount(partialState.getOrDefault("QUEUED", 0))
                     .runningQueryCount(partialState.getOrDefault("RUNNING", 0))
                     .build();
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
@@ -14,7 +14,10 @@
 package io.trino.gateway.ha;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Injector;
+import io.trino.gateway.baseapp.BaseApp;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import io.trino.gateway.ha.router.RoutingManager;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -26,6 +29,8 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.testcontainers.containers.TrinoContainer;
+
+import java.lang.reflect.Field;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -50,10 +55,17 @@ public class TestGatewayHaSingleBackend
                 HaGatewayTestUtils.buildGatewayConfigAndSeedDb(routerPort, "test-config-template.yml");
         // Start Gateway
         String[] args = {"server", testConfig.configFilePath()};
-        HaGatewayLauncher.main(args);
+        HaGatewayLauncher haGatewayLauncher = new HaGatewayLauncher();
+        haGatewayLauncher.run(args);
         // Now populate the backend
         HaGatewayTestUtils.setUpBackend(
                 "trino1", "http://localhost:" + backendPort, "externalUrl", true, "adhoc", routerPort);
+
+        // use injector to manually set backends to healthy
+        Field privateInjectorField = BaseApp.class.getDeclaredField("injector");
+        privateInjectorField.setAccessible(true);
+        Injector injector = (Injector) privateInjectorField.get(haGatewayLauncher);
+        injector.getInstance(RoutingManager.class).updateBackEndHealth("trino1", true);
     }
 
     @Test
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
@@ -74,6 +74,6 @@ public class TestClusterStatsMonitor
 
         ClusterStats stats = monitor.monitor(proxyBackend);
         assertThat(stats.clusterId()).isEqualTo("test_cluster");
-        assertThat(stats.healthy()).isTrue();
+        assertThat(stats.healthy()).isEqualTo(TrinoHealthStateType.HEALTHY);
     }
 }
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryCountBasedRouter.java
@@ -15,6 +15,7 @@ package io.trino.gateway.ha.router;
 
 import com.google.common.collect.ImmutableList;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
+import io.trino.gateway.ha.clustermonitor.TrinoHealthStateType;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -52,7 +53,7 @@ public class TestQueryCountBasedRouter
         {
             ClusterStats.Builder cluster = ClusterStats.builder("c1");
             cluster.proxyTo(BACKEND_URL_1);
-            cluster.healthy(true);
+            cluster.healthy(TrinoHealthStateType.HEALTHY);
             cluster.routingGroup(routingGroup);
             cluster.runningQueryCount(50);
             cluster.queuedQueryCount(SAME_QUERY_COUNT);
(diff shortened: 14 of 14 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
```diff
@@ -86,8 +91,13 @@ public Response updateEntity(
                     ProxyBackendConfiguration backend =
                             OBJECT_MAPPER.readValue(jsonPayload, ProxyBackendConfiguration.class);
                     gatewayBackendManager.updateBackend(backend);
-                    log.info("Setting up the backend %s with healthy state", backend.getName());
+                    log.info("Turning cluster %s %s", backend.getName(), backend.isActive() ? "on" : "off");
                     routingManager.updateBackEndHealth(backend.getName(), backend.isActive());
+                    backendStateManager.updateStates(
```
andythsu, line 96: this is needed to update the states on the UI. We can refactor to get rid of BackendStateManager in the future. https://github.com/trinodb/trino-gateway/issues/359

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -46,7 +47,7 @@ static class LocalStats
     {
         private int runningQueryCount;
         private int queuedQueryCount;
-        private boolean healthy;
+        private TrinoHealthStateType healthy;
```
rdsarvar, line 50: question: what are your thoughts on having this be `heathState` instead of `healthy`?

as well as respective places where `healthy` is used

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthChecker.java
```diff
@@ -34,7 +34,7 @@ public HealthChecker(Notifier notifier)
     public void observe(List<ClusterStats> clustersStats)
     {
         for (ClusterStats clusterStats : clustersStats) {
-            if (!clusterStats.healthy()) {
+            if (clusterStats.healthy() == TrinoHealthStateType.UNHEALTHY) {
```
rdsarvar, line 37: nitpick: feels weird to read `does healthy() = UNHEALTHY ?`, would `healthState() == UNHEALTHY` read better?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+public enum TrinoHealthStateType
+{
+    PENDING, // when clusters are switched from inactive to active
```
rdsarvar, line 18: question (non-blocking): to better understand for myself, is the only usage of `PENDING` right now to differentiate `PENDING` and `UNHEALTHY` on the UI/observability?

but functionally `PENDING` would act as `UNHEALTHY` in the Gateway? (as most tests are X = HEALTHY)

(this is fine I'm just reaffirming)

rdsarvar (changes requested):
some small nitpicks, mostly around use of a 'healthy' variable that has > 2 states

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
andythsu: correct

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
Chaho12: how does the ui change after adding this state, pending? can you share screenshots?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthChecker.java
Chaho12: or maybe just `clusterStats.Health()` ?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -46,19 +46,19 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
+    private TrinoHealthStateType isReadyStatus(String baseUrl)
     {
         Request request = prepareGet()
                 .setUri(uriBuilderFrom(URI.create(baseUrl)).appendPath("/v1/info").build())
                 .build();
 
         try {
             ServerInfo serverInfo = client.execute(request, SERVER_INFO_JSON_RESPONSE_HANDLER);
-            return !serverInfo.isStarting();
+            return serverInfo.isStarting() ? TrinoHealthStateType.UNHEALTHY : TrinoHealthStateType.HEALTHY;
```
Chaho12, line 57: am not so sure about the exact response of serverInfo.isStarting, but if it is in the process of starting (trino backend going up), shouldn't `TrinoHealthStateType.UNHEALTH` be pending?

Chaho12 (commented):
How about `TrinoHealthStateType` to `TrinoHealthType` ? i find it quite a redundant to say healthState as health is state of some status.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
andythsu: good point. should be pending

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
andythsu: UI currently doesn't show the healthstate of each cluster. I can submit a PR to add that onto UI after this one

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/HealthChecker.java
rdsarvar: Sure, open to whichever naming - seeing the ‘y’ suffix (to me) on healthy implies boolean

andythsu: I like `healthState() == UNHEALTHY`

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
Chaho12: can you write this to a comment code maybe at line 15?

Chaho12 (approved):
Please resolve conversation after fixes has been made.
It makes it easier to PR (know that fix has been made for the commend)

LGTM 👍

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
andythsu: as in write the purpose of this enum class?

Chaho12: just to clarify that PENDING is there for ui/observability purpose and functionally is unhealthy ? unless you plan to have other meaning for pending

andythsu: sure, i will add that

gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
```diff
@@ -68,6 +68,9 @@ public abstract class BaseApp
 {
     private static final Logger logger = Logger.get(BaseApp.class);
     private final ImmutableList.Builder<Module> appModules = ImmutableList.builder();
+    // this injector reference is needed to use reflection in
+    // TestGatewayHaSingleBackend and TestGatewayMultipleBackend
```
ebyhr, line 72: Please avoid using reflection. We should find different ways.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
```diff
@@ -52,13 +55,15 @@ public class EntityEditorResource
     private final GatewayBackendManager gatewayBackendManager;
     private final ResourceGroupsManager resourceGroupsManager;
     private final RoutingManager routingManager;
+    private final BackendStateManager backendStateManager;
 
     @Inject
-    public EntityEditorResource(GatewayBackendManager gatewayBackendManager, ResourceGroupsManager resourceGroupsManager, RoutingManager routingManager)
+    public EntityEditorResource(GatewayBackendManager gatewayBackendManager, ResourceGroupsManager resourceGroupsManager, RoutingManager routingManager, BackendStateManager backendStateManager)
```
ebyhr, line 61: The line is too long now. Put parameters on separate lines.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -133,13 +134,13 @@ public void updateBackEndHealth(String backendId, Boolean value)
...
         for (ClusterStats clusterStats : stats) {
-            updateBackEndHealth(clusterStats.clusterId(), clusterStats.healthy());
+            updateBackEndHealth(clusterStats.clusterId(), clusterStats.healthState() == TrinoHealthStateType.HEALTHY);
         }
     }
 
     /**
      * This tries to find out which backend may have info about given query id. If not found returns
-     * the first healthy backend.
+     * the first healthState backend.
```
ebyhr, line 143: `the first healthState backend` looks wrong grammar.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -181,7 +182,7 @@ protected String findBackendForUnknownQueryId(String queryId)
     }
 
     // Predicate helper function to remove the backends from the list
-    // We are returning the unhealthy (not healthy)
+    // We are returning the unhealthy (not healthState)
```
ebyhr, line 185: ditto.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStochasticRoutingManager.java
```diff
@@ -54,11 +54,11 @@ public void addMockBackends()
             proxyBackend.setProxyTo(backend + ".trino.example.com");
             proxyBackend.setExternalUrl("trino.example.com");
             backendManager.addBackend(proxyBackend);
-            //set backend as healthy start with
+            //set backend as healthState start with
             haRoutingManager.updateBackEndHealth(backend, true);
         }
 
-        //Keep only 1st backend as healthy, mark all the others as unhealthy
+        //Keep only 1st backend as healthState, mark all the others as unhealthy
```
ebyhr, line 61: ditto.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStochasticRoutingManager.java
```diff
@@ -54,11 +54,11 @@ public void addMockBackends()
             proxyBackend.setProxyTo(backend + ".trino.example.com");
             proxyBackend.setExternalUrl("trino.example.com");
             backendManager.addBackend(proxyBackend);
-            //set backend as healthy start with
+            //set backend as healthState start with
```
ebyhr, line 57: Revert.

ebyhr (dismissed):

gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
andythsu: if we don't use reflection another way is to add a timeout to let healthcheck finish checking the health since `TestGatewayHaSingleBackend` and `TestGatewayMultipleBackend` are integration tests

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
rdsarvar: +1 on the chain

- the usage of 'healthy' here is okay as it's relating to the backend manager in most cases which is still a boolean healthy/unhealthy
- 'healthy' is now one of our states so you can say things like `first healthy backend` / `first pending backend` / `first unhealthy backend` etc (when leveraging the healthState variable that is, which is unrelated to this current usage)

andythsu: agreed. It's a result of refactoring. I'll fix it

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+/**
+ * PENDING is for ui/observability purpose and functionally it's unhealthy
+ * We should use PENDING when Trino clusters are still spinning up
+ * HEALTHY is when health checks report clusters as up
+ * UNHEALTHY is when health checks report clusters as down
+ */
```
willmostly, line 21: this should be added to the docs. As to placement I think a section on health checks should be added, and linked to from the routing logic and operation sections. Wdyt @mosabua ?

willmostly (changes requested):

andythsu:
@willmostly 
>TestGatewayMultipleBackend uses the TrinoContainer from TestContainers for trino1 and trino2, which does not finish its startup() until SELECT 1 returns. So the health check should succeed. customBackend is a MockWebServer so you will need to add an endpoint to satisfy the healthcheck. I do not believe you should need to set health status manually through injection

Sorry for the confusion.

In `TestGatewayMultipleBackend` and `TestGatewaySingleBackend`, the clusters are added by calling [post api](https://github.com/trinodb/trino-gateway/blob/d45c64f516062841339e9ce93e4675257e6e93e8/gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java#L154-L158)

The default healthstate when clusters are first added to the gateway should be `PENDING`. Because `PENDING` is functionally treated as unhealthy, the test cases will fail (for example, https://github.com/trinodb/trino-gateway/blob/d45c64f516062841339e9ce93e4675257e6e93e8/gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java#L65-L70) since all clusters' states are still in `PENDING` when the test cases run. Unless we wait until the first round of healthcheck kicks in and changes the states from `PENDING` to `HEALTHY`, no clusters are available.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
andythsu: Where do we want to put this in now that we have new doc?

vishalya (approved):
I think this good to be merged. Does anyone have any further comments? We can wait for a day or two and merge it.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,31 @@ public static void setUpBackend(
...
+        TrinoHealthStateType newClusterHealthState = TrinoHealthStateType.PENDING;
+        long startTime = System.currentTimeMillis();
+        long lastExecutionTime = System.currentTimeMillis();
+        // pull cluster health states for 10 seconds
+        // It should be enough as the healthcheck is run every second
+        int timeout = 10 * 1000;
+        while (newClusterHealthState != TrinoHealthStateType.HEALTHY && (lastExecutionTime - startTime) < timeout) {
+            // check the state of newly added cluster every second
+            if (System.currentTimeMillis() - lastExecutionTime <= 1000) {
+                continue;
```
rdsarvar, line 154: question: should we be adding some form of yield here so the while loop isn't burning through CPU cycles? maybe a 100ms sleep or something

rdsarvar (changes requested):
one potential change to the new test modifications, otherwise lg2m as a first step

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
andythsu: as in `Thread.sleep(100);`? In that case do we still need this while loop?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
```diff
@@ -86,8 +95,13 @@ public Response updateEntity(
                     ProxyBackendConfiguration backend =
                             OBJECT_MAPPER.readValue(jsonPayload, ProxyBackendConfiguration.class);
                     gatewayBackendManager.updateBackend(backend);
-                    log.info("Setting up the backend %s with healthy state", backend.getName());
-                    routingManager.updateBackEndHealth(backend.getName(), backend.isActive());
+                    log.info("Turning cluster %s %s", backend.getName(), backend.isActive() ? "on" : "off");
```
xkrogen, line 98: on/off may be a little confusing, let's stick to the same terminology used in the code? 
```java
                    TrinoHealthStateType healthState = backend.isActive() ? TrinoHealthStateType.PENDING : TrinoHealthStateType.UNHEALTHY;
                    log.info("Marking cluster '%s' with health state %s", backend.getName(), healthState);
                    routingManager.updateBackEndHealth(backend.getName(), healthState);
...
```

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
xkrogen: +1 you should refactor more like:
```java
        int timeout = 10 * 1000;
        while (newClusterHealthState != TrinoHealthStateType.HEALTHY && (System.currentTimeMillis() - startTime) < timeout) {
          // do whatever logic
          Thread.sleep(100);
        }
```

rdsarvar: you'd still need the while loop right, as you want to check the cluster healthy every 1 second or so?

you could update this to be:
```
            if (System.currentTimeMillis() - lastExecutionTime <= 1000) {
                Thread.sleep(System.currentTimeMillis() - lastExecutionTime)
                continue;
            }
```

Or later on for the `if (response.isSuccessful()) {` call you could have an else Thread.sleep(1000) and remove this current if-condition

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
andythsu: I think on/off is fine here because this code is triggered when user turns on/off cluster in the web UI. If we use the `healthState`, it's going to say "Marking cluster cluster_a with health state PENDING", which is not user-friendly IMO

rdsarvar (approved):
lg2m

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
mosabua: I agree with @xkrogen .. the error message is is misleading .. the cluster is not shut down or anything .. its just flagged with a specific status from the point of view of the Trino Gateway

gateway-ha/src/test/resources/test-config-template.yml
```diff
@@ -11,6 +11,17 @@ dataStore:
...
+  - io.trino.gateway.ha.module.ClusterStatsMonitorModule
+
+managedApps:
+  - io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor
+
+clusterStatsConfiguration:
+  monitorType: INFO_API
+
+monitor:
+  taskDelaySeconds: 1
```
mosabua, line 24: could we look at using Airlift duration instead?

mosabua (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+/**
+ * PENDING is for ui/observability purpose and functionally it's unhealthy
+ * We should use PENDING when Trino clusters are still spinning up
+ * HEALTHY is when health checks report clusters as up
+ * UNHEALTHY is when health checks report clusters as down
```
ebyhr, line 20: How about changing HEALTHY/UNHEALTHY to UP/DOWN?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public enum TrinoHealthStateType
```
ebyhr, line 22:
```suggestion
public enum TrinoStatus
```

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
```diff
@@ -86,8 +95,13 @@ public Response updateEntity(
...
+                    routingManager.updateBackEndHealth(backend.getName(), backend.isActive() ? TrinoHealthStateType.PENDING : TrinoHealthStateType.UNHEALTHY);
+                    backendStateManager.updateStates(
+                            backend.getName(),
+                            ClusterStats.builder(backend.getName())
+                                    .healthy(backend.isActive() ? TrinoHealthStateType.PENDING : TrinoHealthStateType.UNHEALTHY)
```
ebyhr, line 103: Extract a variable for repeated `backend.isActive() ? TrinoHealthStateType.PENDING : TrinoHealthStateType.UNHEALTHY`.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,29 @@ public static void setUpBackend(
...
+                JsonCodec<ClusterStats> responseCodec = JsonCodec.jsonCodec(ClusterStats.class);
+                ClusterStats clusterStats = responseCodec.fromJson(response.body().string());
+                newClusterHealthState = clusterStats.healthState();
+                log.info("health state for trino cluster %s is %s", name, newClusterHealthState);
+            }
+            else {
+                Thread.sleep(1000);
+            }
+        }
+        assertThat(newClusterHealthState).isEqualTo(TrinoHealthStateType.HEALTHY);
```
ebyhr, line 167: This is a utility class that manages helper methods. I don't understand why this check & assertion exist here. Could you extract into a dedicated test?

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaMultipleBackend.java
```diff
@@ -84,11 +90,26 @@ void setup()
         int backend2Port = scheduledTrino.getMappedPort(8080);
 
         HaGatewayTestUtils.prepareMockBackend(customBackend, customBackendPort, "default custom response");
-        HaGatewayTestUtils.setPathSpecificResponses(customBackend, ImmutableMap.of(
+        Map<String, String> pathResponseMap = ImmutableMap.of(
```
ebyhr, line 93: Rename `pathResponseMap` to `pathToResponse`.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,29 @@ public static void setUpBackend(
                         .build();
         Response response = httpClient.newCall(request).execute();
         assertThat(response.isSuccessful()).isTrue();
+        TrinoHealthStateType newClusterHealthState = TrinoHealthStateType.PENDING;
+        long startTime = System.currentTimeMillis();
```
ebyhr, line 146: We could use Guava's `Stopwatch` instead.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,29 @@ public static void setUpBackend(
...
+        // pull cluster health states for 10 seconds
+        // It should be enough as the healthcheck is run every second
+        int timeout = 10 * 1000;
+        while (newClusterHealthState != TrinoHealthStateType.HEALTHY && (System.currentTimeMillis() - startTime) < timeout) {
+            // check the state of newly added cluster every second
+            request = new Request.Builder()
+                    .url(String.format("http://localhost:%s/api/public/backends/%s/state", routerPort, name))
+                    .get()
+                    .build();
+            response = httpClient.newCall(request).execute();
```
ebyhr, line 156: No need to avoid reusing `request` and `response` variables.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,29 @@ public static void setUpBackend(
...
+            if (response.isSuccessful()) {
+                JsonCodec<ClusterStats> responseCodec = JsonCodec.jsonCodec(ClusterStats.class);
+                ClusterStats clusterStats = responseCodec.fromJson(response.body().string());
+                newClusterHealthState = clusterStats.healthState();
+                log.info("health state for trino cluster %s is %s", name, newClusterHealthState);
```
ebyhr, line 161: Change to debug level or remove it.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,29 @@ public static void setUpBackend(
                         .build();
         Response response = httpClient.newCall(request).execute();
         assertThat(response.isSuccessful()).isTrue();
+        TrinoHealthStateType newClusterHealthState = TrinoHealthStateType.PENDING;
```
ebyhr, line 145: Please fix all warning in the following code block. (NPE, try-with-resource and etc)

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -55,19 +55,19 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
-    private boolean isReadyStatus(String baseUrl, int retriesRemaining)
+    private TrinoHealthStateType isReadyStatus(String baseUrl, int retriesRemaining)
     {
         Request request = prepareGet()
                 .setUri(uriBuilderFrom(URI.create(baseUrl)).appendPath("/v1/info").build())
                 .build();
         try {
             ServerInfo serverInfo = client.execute(request, SERVER_INFO_JSON_RESPONSE_HANDLER);
-            return !serverInfo.isStarting();
+            return serverInfo.isStarting() ? TrinoHealthStateType.PENDING : TrinoHealthStateType.HEALTHY;
```
ebyhr, line 70: I don't understand why "pending" is "ready" status. It worth leaving a code comment.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +142,29 @@ public static void setUpBackend(
...
+        long startTime = System.currentTimeMillis();
+        // pull cluster health states for 10 seconds
+        // It should be enough as the healthcheck is run every second
+        int timeout = 10 * 1000;
+        while (newClusterHealthState != TrinoHealthStateType.HEALTHY && (System.currentTimeMillis() - startTime) < timeout) {
+            // check the state of newly added cluster every second
+            request = new Request.Builder()
+                    .url(String.format("http://localhost:%s/api/public/backends/%s/state", routerPort, name))
```
ebyhr, line 153: Static import `format` method. https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#prefer-string-formatting

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
andythsu: what about "Marking the cluster activated/deactivated"?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoHealthStateType.java
andythsu: I think UP/DOWN has the implication of server being up or down. HEALTHY could be a better candidate here because it means it passed the healthcheck. On the other hand, UNHEALTHY means it failed the healthcheck, but the trino cluster is still UP

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
andythsu: no need or need?

gateway-ha/src/test/resources/test-config-template.yml
andythsu: can I put this in a separate PR? This will require some code changes

mosabua: Sure

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
andythsu: Not sure if I follow. This check and assert is added to make sure when `setupBackend` is called, the backends state are healthy before returning to the caller.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
andythsu: if trino is still starting, its status should be "pending", otherwise its status will be healthy

gateway-ha/src/test/resources/test-config-template.yml
andythsu: This config has been there for a while now so it will be a breaking change for this or next release.

mosabua: Okay .. then it should definitely be a separate change. Please file an issue for that work so we dont forget.

andythsu: created https://github.com/trinodb/trino-gateway/issues/520

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -55,19 +55,19 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
                 .routingGroup(backend.getRoutingGroup()).build();
     }
 
-    private boolean isReadyStatus(String baseUrl)
+    private TrinoStatus isReadyStatus(String baseUrl)
```
ebyhr, line 58: Rename this method to `checkStatus`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -45,7 +46,7 @@ static class LocalStats
     {
         private int runningQueryCount;
         private int queuedQueryCount;
-        private boolean healthy;
+        private TrinoStatus healthState;
```
ebyhr, line 49: Rename variable.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -45,7 +46,7 @@ public abstract class RoutingManager
     private final LoadingCache<String, String> queryIdBackendCache;
     private final ExecutorService executorService = Executors.newFixedThreadPool(5);
     private final GatewayBackendManager gatewayBackendManager;
-    private final ConcurrentHashMap<String, Boolean> backendToHealth;
+    private final ConcurrentHashMap<String, TrinoStatus> backendToHealth;
```
ebyhr, line 49: Rename this variable to `backendToStatus`.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +146,31 @@ public static void setUpBackend(
                         .build();
         Response response = httpClient.newCall(request).execute();
         assertThat(response.isSuccessful()).isTrue();
+        TrinoStatus newClusterHealthState = TrinoStatus.PENDING;
```
ebyhr, line 149:
```suggestion
        TrinoStatus newTrinoStatus = TrinoStatus.PENDING;
```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -24,7 +24,7 @@ public record ClusterStats(
         int runningQueryCount,
         int queuedQueryCount,
         int numWorkerNodes,
-        boolean healthy,
+        TrinoStatus healthState,
```
ebyhr, line 27: Rename variable.

docs/routing-rules.md
```diff
@@ -134,6 +134,21 @@ condition: 'request.getHeader("X-Trino-Client-Tags") contains "label=foo"'
 
 If no rules match, then request is routed to adhoc.
 
+### TrinoState
```
ebyhr, line 137:
```suggestion
### TrinoStatus
```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -41,7 +41,7 @@ public static final class Builder
         private int runningQueryCount;
         private int queuedQueryCount;
         private int numWorkerNodes;
-        private boolean healthy;
+        private TrinoStatus healthState;
```
ebyhr, line 44: Rename variable.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -70,9 +70,9 @@ public Builder numWorkerNodes(int numWorkerNodes)
             return this;
         }
 
-        public Builder healthy(boolean healthy)
+        public Builder healthy(TrinoStatus healthState)
```
ebyhr, line 73: Rename variable.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -92,14 +93,14 @@ public void queuedQueryCount(int queuedQueryCount)
...
         {
-            return this.healthy;
+            return this.healthState;
         }
 
-        public void healthy(boolean healthy)
+        public void healthState(TrinoStatus healthState)
         {
-            this.healthy = healthy;
+            this.healthState = healthState;
```
ebyhr, line 103: Rename the method and variable.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -187,10 +188,10 @@ private boolean isBackendNotHealthy(String backendId)
             log.error("backends can not be empty");
             return true;
         }
-        Boolean isHealthy = backendToHealth.get(backendId);
-        if (isHealthy == null) {
+        TrinoStatus healthType = backendToHealth.get(backendId);
```
ebyhr, line 191: Rename the variable to `status`.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -70,9 +70,9 @@ public Builder numWorkerNodes(int numWorkerNodes)
             return this;
         }
 
-        public Builder healthy(boolean healthy)
+        public Builder healthy(TrinoStatus healthStatus)
```
ebyhr, line 73:
```suggestion
        public Builder healthy(TrinoStatus trinoStatus)
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -45,7 +46,7 @@ static class LocalStats
     {
         private int runningQueryCount;
         private int queuedQueryCount;
-        private boolean healthy;
+        private TrinoStatus healthStatus;
```
ebyhr, line 49:
```suggestion
        private TrinoStatus trinoStatus;
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -92,14 +93,14 @@ public void queuedQueryCount(int queuedQueryCount)
             this.queuedQueryCount = queuedQueryCount;
         }
 
-        public boolean healthy()
+        public TrinoStatus healthStatus()
```
ebyhr, line 96:
```suggestion
        public TrinoStatus trinoStatus()
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -92,14 +93,14 @@ public void queuedQueryCount(int queuedQueryCount)
...
         {
-            return this.healthy;
+            return this.healthStatus;
         }
 
-        public void healthy(boolean healthy)
+        public void healthStatus(TrinoStatus healthStatus)
```
ebyhr, line 101:
```suggestion
        public void trinoStatus(TrinoStatus trinoStatus)
```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStats.java
```diff
@@ -70,9 +70,9 @@ public Builder numWorkerNodes(int numWorkerNodes)
             return this;
         }
 
-        public Builder healthy(boolean healthy)
+        public Builder healthy(TrinoStatus trinoStatus)
```
ebyhr, line 73:
```suggestion
        public Builder trinoStatus(TrinoStatus trinoStatus)
```

docs/routing-rules.md
```diff
@@ -134,6 +134,21 @@ condition: 'request.getHeader("X-Trino-Client-Tags") contains "label=foo"'
 
 If no rules match, then request is routed to adhoc.
 
+### TrinoStatus
```
ebyhr, line 137: Please fix broken style: 
Screenshot 2024-10-08 at 8 23 35

andythsu: it should be fixed now. It looks like 
image
on my IDE's preview

mosabua: Ideally test with mkdocs locally.

andythsu: yep, tested it 
image

gateway-ha/src/main/java/io/trino/gateway/ha/resource/EntityEditorResource.java
```diff
@@ -86,8 +95,14 @@ public Response updateEntity(
                     ProxyBackendConfiguration backend =
                             OBJECT_MAPPER.readValue(jsonPayload, ProxyBackendConfiguration.class);
                     gatewayBackendManager.updateBackend(backend);
-                    log.info("Setting up the backend %s with healthy state", backend.getName());
-                    routingManager.updateBackEndHealth(backend.getName(), backend.isActive());
+                    log.info("Marking the cluster %s %s", backend.getName(), backend.isActive() ? "active" : "inactive");
+                    TrinoStatus trinoStatus = backend.isActive() ? TrinoStatus.PENDING : TrinoStatus.UNHEALTHY;
```
ebyhr, line 99: Why do we set PENDING, not HEALTHY here? It's worth leaving a code comment.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -159,6 +146,31 @@ public static void setUpBackend(
...
+                    if (getBackendStateResponseBody != null) {
+                        ClusterStats clusterStats = responseCodec.fromJson(getBackendStateResponseBody.string());
+                        newTrinoStatus = clusterStats.trinoStatus();
+                        log.debug("status for new trino cluster %s is %s", name, newTrinoStatus);
+                    }
+                }
+            }
+            Thread.sleep(1000);
+        }
+        assertThat(newTrinoStatus).isEqualTo(TrinoStatus.HEALTHY);
```
ebyhr, line 173: Can we simplify something like this? 
```java
    private static void verifyTrinoStatus(int port, String name)
            throws IOException
    {
        Request getBackendStateRequest = new Request.Builder()
                .url(format("http://localhost:%s/api/public/backends/%s/state", port, name))
                .get()
                .build();

        for (int i = 0; i < 10; i++) {
            // check the state of newly added cluster every second
            try (Response getBackendStateResponse = httpClient.newCall(getBackendStateRequest).execute()) {
                checkState(getBackendStateResponse.isSuccessful());
                JsonCodec<ClusterStats> responseCodec = JsonCodec.jsonCodec(ClusterStats.class);
                ResponseBody getBackendStateResponseBody = requireNonNull(getBackendStateResponse.body(), "getBackendStateResponse.body() is null");
                ClusterStats clusterStats = responseCodec.fromJson(getBackendStateResponseBody.string());
                if (clusterStats.trinoStatus() == TrinoStatus.HEALTHY) {
                    return;
                }
            }
            sleepUninterruptibly(1, TimeUnit.SECONDS);
        }
        throw new IllegalStateException("Trino cluster is not healthy");
    }
```

ebyhr (approved):

PR #367, opened 2024-05-29
Fix broken renderings of code blocks in docs

rdsarvar:
## Description

Two docs pages were leveraging the `$xslt` tag for syntax highlighting of fenced code blocks which broke mkdocs rendering. It appears that the **```$xslt** line is treated like a regular line of text. The syntax tags have been updated to leverage bash and json instead.

## Release notes

```markdown
* 
```

### DIFF
--- a/docs/gateway-api.md
+++ b/docs/gateway-api.md
@@ -2,7 +2,7 @@
 
 ## Add or update a backend
 
-```$xslt
+```bash
 curl -X POST http://localhost:8080/entity?entityType=GATEWAY_BACKEND \
  -d '{  "name": "trino-3",
         "proxyTo": "http://localhost:8083",
@@ -15,7 +15,7 @@ If the backend URL is different from the `proxyTo` URL (for example if they are
 internal vs. external hostnames). You can use the optional `externalUrl` field
 to override the link in the Active Backends page.
 
-```$xslt
+```bash
 curl -X POST http://localhost:8080/entity?entityType=GATEWAY_BACKEND \
  -d '{  "name": "trino-3",
         "proxyTo": "http://localhost:8083",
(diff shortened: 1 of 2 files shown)


docs/gateway-api.md
```diff
@@ -2,7 +2,7 @@
 
 ## Add or update a backend
 
-```$xslt
+```bash
```
mosabua, line 5: should it be `shell` instead .. I am not sure what mkdocs suggests

mosabua: if shell works I would prefer that

mosabua: looks like just `sh` might also be valid.. lets find out what is best and do that consistently.

For example interesting link .. not necessarily for mkdocs but still useful

https://pawamoy.github.io/markdown-exec/usage/shell/#exec-76--markdown

rdsarvar: Looks like `shell` is an available lexer for `mkdocs-material`: https://pygments.org/docs/lexers/#lexers-for-various-shells

Based on the lexer name and the relevant short names I think they all point to the same thing under the hood:
image

I'll try it on my local env in a bit and push if it renders

mosabua: I just tried as well. Lets see what is most commly used in the Trino docs (even though its sphinx vs mkdocs)

mosabua: All three definitely look the same

mosabua: Okay .. Trino docs use `shell` .. lets go with the same

rdsarvar: looks like there's 13 instances of `bash` and 18 of `shell` 😄 
```
trino % grep -nr '```bash' docs | wc -l
      13
trino % grep -nr '```sh' docs | wc -l
      18
trino % grep -nr '```shell' docs | wc -l
      18
```

All occurences of `shell` are part of the `sphinx` docs and all occurences of `bash` are in the `docs/README.md` file

mosabua (approved):
Awesome @rdsarvar

PR #375, opened 2024-05-31
Remove unnecessary CONCAT function

grahamplata:
## Description

Resolves the following error when using Spanner by Explicitly Cast text values in [CONCAT function](https://cloud.google.com/spanner/docs/reference/postgresql/functions-and-operators#string_functions). 

`SELECT CONCAT(FLOOR(created / 1000 / 60)::text) AS minute,`

<details><summary>Resolved the following error</summary>

```
trino-gateway org.jdbi.v3.core.statement.UnableToExecuteStatementException: org.postgresql.util.PSQLException: ERROR: Postgres function concat(double precision) is not supported - Statement: 'SELECT CONCAT(FLOOR(created / 1000 / 60)) AS minute,
trino-gateway        backend_url AS backend_url,
trino-gateway        COUNT(1) AS query_count
trino-gateway FROM query_history
trino-gateway WHERE created > $1
trino-gateway GROUP BY minute, backend_url
trino-gateway ' [statement:"SELECT CONCAT(FLOOR(created / 1000 / 60)) AS minute,
trino-gateway        backend_url AS backend_url,
trino-gateway        COUNT(1) AS query_count
trino-gateway FROM query_history
trino-gateway WHERE created > :created
trino-gateway GROUP BY minute, backend_url
trino-gateway ", arguments:{positional:{0:1717180785232}, named:{created:1717180785232}, finder:[]}]
```

</details>

## Release notes

```markdown
* 
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
@@ -62,7 +62,7 @@ public interface QueryHistoryDao
     Long count(@Define("condition") String condition);
 
     @SqlQuery("""
-            SELECT CONCAT(FLOOR(created / 1000 / 60)) AS minute,
+            SELECT CONCAT(FLOOR(created / 1000 / 60)::text) AS minute,
                    backend_url AS backend_url,
                    COUNT(1) AS query_count
             FROM query_history
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
@@ -62,7 +62,7 @@ public interface QueryHistoryDao
     Long count(@Define("condition") String condition);
 
     @SqlQuery("""
-            SELECT CONCAT(FLOOR(created / 1000 / 60)::text) AS minute,
+            SELECT FLOOR(created / 1000 / 60) AS minute,
                    backend_url AS backend_url,
                    COUNT(1) AS query_count
             FROM query_history
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
@@ -122,7 +122,7 @@ public class HaQueryHistoryManager
         List<DistributionResponse.LineChart> resList = new ArrayList<>();
         for (Map<String, Object> model : results) {
             DistributionResponse.LineChart lineChart = new DistributionResponse.LineChart();
-            long minute = Long.parseLong(model.get("minute").toString());
+            long minute = (long) Float.parseFloat(model.get("minute").toString());
             Instant instant = Instant.ofEpochSecond(minute * 60L);
             LocalDateTime dateTime = LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
             DateTimeFormatter formatter = DateTimeFormatter.ofPattern("HH:mm");
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryHistoryManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestQueryHistoryManager.java
@@ -15,12 +15,14 @@ package io.trino.gateway.ha.router;
 
 import io.trino.gateway.ha.domain.response.DistributionResponse;
 import io.trino.gateway.ha.persistence.JdbcConnectionManager;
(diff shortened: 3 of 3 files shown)


mosabua (commented):
I don't think this is valid syntax that will work in MySQL and PostgreSQL. Have you tested this in both?

ebyhr (changes requested):
This isn't a valid syntax in case of MySQL.

grahamplata:
> This isn't a valid syntax in case of MySQL.

Ahh, Yep. I see that. My apologies. In that case is CONCAT needed for this since its a singular input? If it is not a hard requirement dropping that should meet the need

mosabua:
Also note that Spanner is not really something we use or test. Currently only MySQL and PostgreSQL are used.

mosabua:
> > This isn't a valid syntax in case of MySQL.
> 
> Ahh, Yep. I see that. My apologies. In that case is CONCAT needed for this since its a singular input? If it is not a hard requirement dropping that should meet the need

Looking at this I think you might be right and its not really needed. Maybe update the PR to get rid of it and we can see .. will also have to check other database scripts and see if there is something related going on.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
```diff
@@ -62,7 +62,7 @@ SELECT count(1) FROM query_history
     Long count(@Define("condition") String condition);
 
     @SqlQuery("""
-            SELECT CONCAT(FLOOR(created / 1000 / 60)) AS minute,
+            SELECT FLOOR(created / 1000 / 60) AS minute,
```
ebyhr, line 65: Can you test with PostgreSQL and MySQL and share the screenshot of Web UI?

mosabua:
I started CI and it seems good. We will have to wait for you to submit a CLA and CLA processing. In the meantime .. @vishalya @willmostly @Chaho12 .. any concerns?

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
Chaho12: Can't see screenshot

Chaho12 (approved):
LGTM 👍

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistoryDao.java
ebyhr: Thanks for testing, but the screenshot doesn't cover the actual situation. We should run a query before accessing the UI. 

Note that current change causes NumberFormatException if the table isn't an empty.

ebyhr (changes requested):

oneonestar:
I guess the `CONCAT()` was added to convert the data type into String.
Without `CONCAT()`, mysql and postgres behave differently when access through JDBI (can't reproduce using database CLI):
```
postgres:
minute -> {Double@9333} 3.033864E7

mysql:
minute -> {BigDecimal@9775} "30338640"
```
Change from `Long.parseLong(model.get("minute").toString());` to `(long) Float.parseFloat(model.get("minute").toString());` does solve the problem.

The change is OK but the test doesn't cover the above issue.
Could you update the test? Also you need to sign a CLA before we can merge your code.

grahamplata:
> I guess the `CONCAT()` was added to convert the data type into String. Without `CONCAT()`, mysql and postgres behave differently when access through JDBI (can't reproduce using database CLI):
> 
> ```
> postgres:
> minute -> {Double@9333} 3.033864E7
> 
> mysql:
> minute -> {BigDecimal@9775} "30338640"
> ```
> 
> Change from `Long.parseLong(model.get("minute").toString());` to `(long) Float.parseFloat(model.get("minute").toString());` does solve the problem.
> 
> The change is OK but the test doesn't cover the above issue. Could you update the test? Also you need to sign a CLA before we can merge your code.

I'll update that shortly and the CLA email was sent last week

mosabua:
CLA will be processed soon.

mosabua:
@cla-bot check

mosabua (approved):
Looks good now to me now. Thanks for working with us on iterating to the right fix @grahamplata 

Any concerns with merging this @ebyhr and @oneonestar ?

oneonestar:
Waiting for the test to be updated.

oneonestar:
> I'll update that shortly and the CLA email was sent last week

@grahamplata Any updates?

grahamplata:
> > I'll update that shortly and the CLA email was sent last week
> 
> @grahamplata Any updates?

Yep, sorry

oneonestar (approved):

oneonestar:
Thanks.

PR #287, opened 2024-03-20
Use JDBI for ResourceGroups and Remove activejdbc

oneonestar:
## Description
Close #229. Depends on #284.
* Use JDBI for ResourceGroups
* Remove activejdbc completely

The cache provided by activejdbc for ResourceGroups will be gone with this PR.
We might need to implement a cache if performance degradation is significant.

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #379, opened 2024-06-04
Fix create new resource group

oneonestar:
## Description
Create resource group was broken after #287.
`resource_group_id` should be generated by database with AUTO_INCREMENT in DDL.
Tests didn't cover create multiple resource groups so we missed that.

## Release notes

```markdown
* Fix create new resource group
```

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
@@ -50,17 +50,23 @@ public class TestResourceGroupsManager
     {
         ResourceGroupsDetail resourceGroup = new ResourceGroupsDetail();
 
-        resourceGroup.setResourceGroupId(0L);
         resourceGroup.setName("admin");
         resourceGroup.setHardConcurrencyLimit(20);
         resourceGroup.setMaxQueued(200);
         resourceGroup.setJmxExport(true);
         resourceGroup.setSoftMemoryLimit("80%");
-
-        ResourceGroupsDetail newResourceGroup = resourceGroupManager.createResourceGroup(resourceGroup,
+        ResourceGroupsDetail adminResourceGroup = resourceGroupManager.createResourceGroup(resourceGroup,
                 null);
+        assertThat(adminResourceGroup).isEqualTo(resourceGroup);
 
-        assertThat(newResourceGroup).isEqualTo(resourceGroup);
+        resourceGroup.setName("user");
+        resourceGroup.setHardConcurrencyLimit(10);
+        resourceGroup.setMaxQueued(100);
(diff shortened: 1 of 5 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -50,27 +50,35 @@ public void testCreateResourceGroup()
     {
         ResourceGroupsDetail resourceGroup = new ResourceGroupsDetail();
 
-        resourceGroup.setResourceGroupId(0L);
+        resourceGroup.setResourceGroupId(0L);   // This should be ignored. Id are generated by database (AUTO_INCREMENT) starting from 1.
```
mosabua, line 53: What do you mean .. it should be ignored? I think it is ignored right?
```suggestion
        resourceGroup.setResourceGroupId(0L);   // The value is ignored. IDs are generated by the database (AUTO_INCREMENT) starting from 1.
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -50,27 +50,35 @@ public void testCreateResourceGroup()
...
         resourceGroup.setJmxExport(true);
         resourceGroup.setSoftMemoryLimit("80%");
-
-        ResourceGroupsDetail newResourceGroup = resourceGroupManager.createResourceGroup(resourceGroup,
+        ResourceGroupsDetail adminResourceGroup = resourceGroupManager.createResourceGroup(resourceGroup,
                 null);
+        assertThat(adminResourceGroup).isEqualTo(resourceGroup);
 
-        assertThat(newResourceGroup).isEqualTo(resourceGroup);
+        resourceGroup.setResourceGroupId(0L);   // Id are generated by database (AUTO_INCREMENT). This will be 2.
```
mosabua, line 63: would it make more sense to NOT have a setter at all? Or is this something that JDBI doesnt like

mosabua: Or do we want another ID that we generate (like a UUID string or so?)

mosabua (approved):
Looks good. Just have some questions and minor updates to comments.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
oneonestar: I agree that's confusing. I'll delete these two lines.
This setter will remain for Jackson deserialization.

mosabua (approved):
Looks good. Merging this.

PR #381, opened 2024-06-06
Remove non-null restriction for QueryHistory

oneonestar:
## Description
Both `userName` and `source` in QueryHistory are nullable.

`userName` (ie. `X-Trino-User`) is optional in request. If not supplied, the session user is automatically determined via user mapping by coordinator. Trino-gateway won't be able to know the user determined by user mapping.

`source` is optional in Trino:
https://github.com/trinodb/trino/blob/edd5d5bec766e90053e7c28ff9c9dc254790fdb0/core/trino-spi/src/main/java/io/trino/spi/eventlistener/QueryContext.java#L45

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistory.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistory.java
@@ -13,6 +13,7 @@
  */
 package io.trino.gateway.ha.persistence.dao;
 
+import jakarta.annotation.Nullable;
 import org.jdbi.v3.core.mapper.reflect.ColumnName;
 
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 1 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/QueryHistory.java
```diff
@@ -30,7 +30,5 @@ public record QueryHistory(
         requireNonNull(queryId, "queryId is null");
         requireNonNull(queryText, "queryText is null");
         requireNonNull(backendUrl, "backendUrl is null");
-        requireNonNull(userName, "userName is null");
-        requireNonNull(source, "source is null");
```
ebyhr, line 34: Can we add `@Nullable` annotation to the constructor?

ebyhr (approved):

mosabua:
I added a release notes entry for this since I think it is kinda user facing..

PR #384, opened 2024-06-12
Add proxytarget header to query requests

andythsu:
## Description

Add `proxytarget` header to requests where `previousBackend` is found. Otherwise this line will return warning https://github.com/trinodb/trino-gateway/blob/d45c64f516062841339e9ce93e4675257e6e93e8/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java#L154-L174

## Release notes

```markdown
* 
```


wendigo (approved):

wendigo:
@willmostly do you want to take a look?

oneonestar:
Actually `PROXY_TARGET_HEADER` will be gone after migrate to Airlift.
The header was needed because we were hooking into different execution phases in jetty proxy. We used the header to pass information between these hooks. So that's kind of a hacky way to persist information and it'll be gone.

PR #388, opened 2024-06-17
Add custom exception when clusters are not found

andythsu:
## Description

If no adhoc clusters are found, we should throw a meaningful exception with some explanations.

## Release notes

```markdown
* 
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -328,7 +328,7 @@ Other timeout parameters are not applicable to the JDBC connection.
 
 This pulls cluster information from the `ui/api/stats` REST endpoint. This is
 supported for legacy reasons and may be deprecated in the future. It is only 
-supported for backend clusters with `web-ui.authentication.type = PASSWORD`. Set
+supported for backend clusters with `web-ui.authentication.type=FORM`. Set
 a username and password using `backendState` as with the `JDBC` option.
 
 #### NOOP


gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -224,7 +224,7 @@ private synchronized Optional<String> getBackendForRoutingGroup(String routingGr
     @Override
     public String provideAdhocBackend(String user)
     {
-        return getBackendForRoutingGroup("adhoc", user).orElseThrow();
+        return getBackendForRoutingGroup("adhoc", user).orElseThrow(() -> new ClusterNotFoundException("did not find any cluster for the adhoc routing group"));
```
mosabua, line 227: Should we make this more generic of an exception? Like RoutingException or so? Also should we pass in the magic `adhoc` from the prior code?

And update wording?

```
No cluster in routing group `adhoc` available.
```

Or similar

mosabua (changes requested):
I like the idea of improving that experience .. but we need think about how to implement  that better

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
andythsu: we could put it the method `getBackendForRoutingGroup`. That method has the `routingGroup` passed in, so we can use that in our error message. That should be cleaner

andythsu: Open to making the error more generic to be `RoutingException`

Chaho12: I think more general version, RoutingException is better :)

Issue #394, opened 2024-06-28
Pass object instead of multiple strings to routingManager?


Chaho12:
RoutingManager's provideBackendForRoutingGroup or provideAdhocBackend functions are called in [QueryIdCachingProxyHandler.java](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/handler/QueryIdCachingProxyHandler.java#L332-L341) but within the function, it sends only routing group (if there is a header named "X-Trino-Routing-Group") and user info.

But in request headers, there are much more helpful informations, such as source, catalog, clientTags etc.
![image](https://github.com/trinodb/trino-gateway/assets/25147023/6f3365d2-479c-40b1-8ab2-072058b02962)

So, how about passing an object which includes all headers?

mosabua:
I think thats a good idea .. there might also be more headers in the future so its a bit easier to extend.

oneonestar:
Make sense. We need to choose a data structure for the headers, eg.
* `Map<String, List<String>>`
* `ListMultimap<String, String>`
* `MultivaluedMap<String, String>`

Chaho12:
I think `MultivaluedMap<String, String>` sounds good as trino uses it at multiple cases, [case1](https://github.com/trinodb/trino/blob/89328441556474f5a69b7259dd7b4aa0a83a3da4/core/trino-main/src/main/java/io/trino/dispatcher/QueuedStatementResource.java#L176-L194), [case2](https://github.com/trinodb/trino/blob/89328441556474f5a69b7259dd7b4aa0a83a3da4/core/trino-main/src/main/java/io/trino/server/security/PasswordAuthenticator.java#L109-L123)

willmostly:
>  it sends only routing group (if there is a header named "X-Trino-Routing-Group") 
if you use `RuleReloadingRoutingGroupSelector`, the routing group as determined by the routing rules will be passed.

willmostly:
https://github.com/trinodb/trino-gateway/pull/398#pullrequestreview-2156954353

PR #398, opened 2024-07-01
Pass all headers as MultiValueMap

Chaho12:
## Description

Pass all headers as MultivaluedMap to routing manager.
After this pr, i'll add docs on how to add a module to customize routing logic

https://github.com/trinodb/trino-gateway/issues/394

## Release notes

```markdown
* 
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -19,8 +19,12 @@ import io.trino.gateway.ha.router.GatewayCookie;
 import io.trino.gateway.ha.router.RoutingGroupSelector;
 import io.trino.gateway.ha.router.RoutingManager;
 import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.core.MultivaluedHashMap;
+import jakarta.ws.rs.core.MultivaluedMap;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
 import java.util.List;
 import java.util.Optional;
 import java.util.regex.Pattern;
@@ -32,11 +36,11 @@ import static io.trino.gateway.ha.handler.ProxyUtils.extractQueryIdIfPresent;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.OAUTH_PATH;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.TRINO_UI_PATH;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.UI_API_STATS_PATH;
-import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.USER_HEADER;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.V1_INFO_PATH;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.V1_NODE_PATH;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.V1_QUERY_PATH;
 import static io.trino.gateway.ha.handler.QueryIdCachingProxyHandler.V1_STATEMENT_PATH;
+import static java.util.Collections.list;
 import static java.util.Objects.requireNonNull;
 
 public class RoutingTargetHandler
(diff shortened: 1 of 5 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -222,16 +225,16 @@ private synchronized Optional<String> getBackendForRoutingGroup(String routingGr
     }
 
     @Override
-    public String provideAdhocBackend(String user)
+    public String provideAdhocBackend(MultivaluedMap<String, String> headers)
     {
-        return getBackendForRoutingGroup("adhoc", user).orElseThrow();
+        return getBackendForRoutingGroup("adhoc", headers.getFirst(USER_HEADER)).orElseThrow();
```
Chaho12, line 230: UserHeader (X-Trino-User) always has one value, so getFirst is practically the same

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -81,13 +83,19 @@ public boolean isPathWhiteListed(String path)
 
     private String getBackendFromRoutingGroup(HttpServletRequest request)
     {
+        MultivaluedMap<String, String> headers = new MultivaluedHashMap<>();
+        Enumeration<String> headerNames = request.getHeaderNames();
+        while (headerNames.hasMoreElements()) {
+            String key = headerNames.nextElement();
+            headers.add(key, request.getHeader(key));
```
oneonestar, line 90: Header can be set multiple times with different values.
This only gets the first value. See `getHeaders(String name)`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
oneonestar: > UserHeader (X-Trino-User) always has one value

No. `X-Trino-User` could be null.

> getFirst is practically the same

Yes. `user` could be null, so as `getFirst()`.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStochasticRoutingManager.java
```diff
@@ -66,7 +71,9 @@ public void addMockBackends()
             haRoutingManager.updateBackEndHealth(backend, false);
         }
 
-        assertThat(haRoutingManager.provideBackendForRoutingGroup(groupName, ""))
+        headers.add(USER_HEADER, "");
```
oneonestar, line 74: Could you add another test for not setting `USER_HEADER` at all?

oneonestar (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStochasticRoutingManager.java
Chaho12: of course :)

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStochasticRoutingManager.java
```diff
@@ -58,15 +66,77 @@ public void addMockBackends()
...
+    }
+
+    @Test
+    public void testActiveBackendCount()
+    {
+        assertThat((long) backendManager.getAllActiveBackends().size()).isEqualTo(numTestBackends + 1);
+    }
+
+    @Test
+    public void testNoHealthyTestBackend()
```
oneonestar, line 101: I don't know what's the correct response for:
* missing a user header
* all backends are not healthy

I'm okay with adding these tests since they reflect the current behavior.
We might want to review them later.

oneonestar (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStochasticRoutingManager.java
Chaho12: Well this test is to test if it routes to adhoc on case
where all request group (test_group) backends are not healthy

oneonestar:
LGTP except the CI error.

Chaho12:
Fixed CI error.
It was due to `MemberName: Name 'ADHOC' must match pattern '^[a-z][a-zA-Z0-9]*$'.` so i had to change `ADHOC` -> `adhoc`

oneonestar (approved):

willmostly:
This looks like it is blurring the lines between `RoutingGroupSelector` and `RoutingManager`. The `request` object is passed to `RoutingGroupSelector.findRoutingGroup(HttpServletRequest request)`, so why do it again here? 

If the issue is that `RoutingGroupSelector` may pick a Routing Group with no healthy backends, it seems more appropriate to pass health info into the  `RoutingGroupSelector` than to add business logic into `RoutingManager`.

Chaho12:
> I think the intent here is to solve the problem of picking an appropriate backup routing group if the selected routing group has zero healthy backends.

So the reasons for this change is because RoutingManager is the one who is extended when creating custom routing module, like StochasticRoutingManager or QueryCountBasedRouter, but it lacks some important information like `sql`, `source`,  `catalog`, `clientTags`(if it is sent via trino cli)) and any other information that is essential when writing enhanced routing logic.

I am not 100% sure about the relationship between `RoutingGroupSelector` and `RoutingManager`, but I think we can see this as two different ways of handling routing.
- via routing rules file (mvel files)
- add a custom java module

@vishalya any comments on this?

Chaho12:
Close this PR because this is against current logical routing architecture where routing manager routes to given routing group, so other information passed on to routing group is unncessary (or even duplicate)

PR #403, opened 2024-07-05
Remove docs for TLS 1.0 and 1.1

oneonestar:
## Description
Remove docs for TLS 1.0 and 1.1.
TLS 1.0 and TLS 1.1 were deprecated since 2021.

Refs:
* https://datatracker.ietf.org/doc/html/rfc8996
* https://techcommunity.microsoft.com/t5/windows-it-pro-blog/tls-1-0-and-tls-1-1-soon-to-be-disabled-in-windows/ba-p/3887947

## Release notes


mosabua (approved):
Looks good. I hope nobody runs into this at this stage.

PR #412, opened 2024-07-17
Improve the time of the UI distribution graph (#369)

### LINKED ISSUE #369: Query distribution graph time is in UTC
Go to the dashboard, observe that the chart time is UTC and not in the local time. The time axis should be in local user time zone.

ytwp:
## Description
Fix #369 

## Release notes

```markdown
* 
```

### DIFF
--- a/webapp/src/components/dashboard.tsx
+++ b/webapp/src/components/dashboard.tsx
@@ -10,6 +10,7 @@ import { IconHelpCircle } from "@douyinfe/semi-icons";
 import { useNavigate } from "react-router-dom";
 import { hasPagePermission, routersMapper } from "../router";
 import { useAccessStore } from "../store";
+import { formatZonedDateTime } from "../utils/time";
 
 export function Dashboard() {
   const access = useAccessStore();
@@ -26,7 +27,7 @@ export function Dashboard() {
   const data = [
     {
       key: Locale.Dashboard.StartTime,
-      value: distributionDetail?.startTime
+      value: distributionDetail && formatZonedDateTime(distributionDetail.startTime)
     },
     {
       key: Locale.Dashboard.Backends,
--- a/webapp/src/utils/time.ts
+++ b/webapp/src/utils/time.ts
@@ -29,8 +29,7 @@ const format = (
     .format(formatString);
 };
 
-export const formatYYYYMMddHHMMSS = (timestamp: number): string => {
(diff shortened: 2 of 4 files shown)


webapp/src/utils/time.ts
```diff
@@ -39,3 +39,14 @@ export const formatYYYYMMddHHMMSS = (timestamp: number): string => {
   const seconds = String(date.getSeconds()).padStart(2, '0');
   return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`;
 }
+
+export const zonedDateFormatYYYYMMddHHMMSS = (utcDateTime: string): string => {
```
mosabua, line 43: Can we find a better name for this .. or maybe even some standard method that creates this output .. it is basically ISO date time format right?

mosabua: Also do we need a test for this?

webapp/src/components/dashboard.tsx
```diff
@@ -26,7 +27,7 @@ export function Dashboard() {
   const data = [
     {
       key: Locale.Dashboard.StartTime,
-      value: distributionDetail?.startTime
+      value: distributionDetail && zonedDateFormatYYYYMMddHHMMSS(distributionDetail.startTime)
```
mosabua, line 30: I know this fixes the linked bug .. but are there other fields where this would be needed as well .. like endtime or whatever that appear in the UI?

mosabua (changes requested):

webapp/src/utils/time.ts
willmostly, line 43: Validate that the zone offset is 0

willmostly (changes requested):
One comment and one nit, looks good overall

webapp/src/utils/time.ts
ytwp: My time zone is UTC+8. I have tested it.

ytwp: I have changed the function name.

mosabua (approved):
Looks good now.

webapp/src/utils/time.ts
mosabua: Thanks

mosabua:
Wdyt @willmostly .. ship it?

willmostly (approved):
lgtm

PR #416, opened 2024-07-23
Fix indentation for deployment annotation

lazy-eval:
## Release notes

```markdown
* 
```

### DIFF
--- a/helm/trino-gateway/templates/deployment.yaml
+++ b/helm/trino-gateway/templates/deployment.yaml
@@ -14,10 +14,12 @@ spec:
   template:
     metadata:
       annotations:
-          # Include the version of trino-gateway-configuration as an input to the deployment checksum. This causes pods to restart on helm update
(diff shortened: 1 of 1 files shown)


helm/trino-gateway/templates/deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
           # Include the version of trino-gateway-configuration as an input to the deployment checksum. This causes pods to restart on helm update
```
mosabua, line 17: Can you move the docs to also use two space indent. Also can you make the comment line shorter .. maybe 80 char width or so

mosabua (changes requested):

mosabua:
Also .. does this break the chart and should be a release blocker for version 10?

willmostly (approved):
LGTM

mosabua (approved):
Looks good now.

lazy-eval:
> Also .. does this break the chart and should be a release blocker for version 10?

As you might have already figured out this is a blocker for deployment

PR #417, opened 2024-07-23
Add option to not add X-Forwarded headers

willmostly:
## Description
In some cases it can be useful to have a Trino backend communicate directly with the client once it has been selected by the Trino Gateway's routing logic. This can be accomplished by not setting the `X-Forwarded` headers.

## Release notes

Optionally disable adding X-Forwarded-* headers to the request

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -63,7 +63,8 @@ From a users perspective Trino Gateway acts as a transparent proxy for one
 or more Trino clusters. The following Trino configuration tips should be 
 taken into account for all clusters behind the Trino Gateway.
 
-Process forwarded HTTP headers must be enabled:
+If all client and server communication is routed through Trino Gateway, 
+then process forwarded HTTP headers must be enabled:
 
 ```commandline
 http-server.process-forwarded=true
@@ -73,15 +74,21 @@ Without this setting, first requests go from the user to Trino Gateway and then
 to Trino correctly. However, the URL for subsequent next URIs for more results
 in a query provided by Trino is then using the local URL of the Trino cluster,
 and not the URL of the Trino Gateway. This circumvents the Trino Gateway for all
-these requests, and is contrary to the purpose of using the Trino Gateway. In
-scenarios, where the local URL of the Trino cluster is private to the Trino
-cluster on the DNS/network level, these following calls might not work at all
-for users.
+these requests. In scenarios, where the local URL of the Trino cluster is private 
+to the Trino cluster on the network level, these following calls do not work
+at all for users.
 
 This setting is also required for Trino to authenticate in the case TLS is 
 terminated at the Trino Gateway. Normally it refuses to authenticate plain HTTP 
 requests, but if `http-server.process-forwarded=true` it authenticates over 
 HTTP if the request includes `X-Forwarded-Proto: HTTPS`.
+
+To prevent Trino Gateway from sending `X-Forwarded-*` headers, add the following configuration:
+
+```yaml
+routing:
+  addXForwardedHeaders: false
+```
 
 Find more information in [the related Trino documentation](https://trino.io/docs/current/security/tls.html#use-a-load-balancer-to-terminate-tls-https).
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/RoutingConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/RoutingConfiguration.java
@@ -21,6 +21,8 @@ public class RoutingConfiguration
 {
     private Duration asyncTimeout = new Duration(2, MINUTES);
 
+    private boolean addXForwardedHeaders = true;
+
     public Duration getAsyncTimeout()
     {
         return asyncTimeout;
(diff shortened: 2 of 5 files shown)


mosabua:
Need to add user facing docs for this.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -104,7 +111,9 @@ public static TestConfig buildGatewayConfigAndSeedDb(int routerPort, String conf
                                 "APPLICATION_CONNECTOR_PORT", String.valueOf(30000 + (int) (Math.random() * 1000)))
                         .replace("ADMIN_CONNECTOR_PORT", String.valueOf(31000 + (int) (Math.random() * 1000)))
                         .replace("LOCALHOST_JKS", Paths.get(resource.toURI()).toFile().getAbsolutePath());
-
+        if (extraConfig.isPresent()) {
+            configStr += "\n" + extraConfig.orElseThrow();
+        }
```
oneonestar, line 116: I think adding a `test-config-without-x-forwarded.yml` might be better than this.

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -104,7 +104,6 @@ public static TestConfig buildGatewayConfigAndSeedDb(int routerPort, String conf
                                 "APPLICATION_CONNECTOR_PORT", String.valueOf(30000 + (int) (Math.random() * 1000)))
                         .replace("ADMIN_CONNECTOR_PORT", String.valueOf(31000 + (int) (Math.random() * 1000)))
                         .replace("LOCALHOST_JKS", Paths.get(resource.toURI()).toFile().getAbsolutePath());
-
```
oneonestar, line 107: nit

oneonestar (approved):

willmostly:
@oneonestar I changed the behavior slightly. In addition to not adding them ourselves, we will now actively filter out `X-Forwarded` headers that are on the incoming request already. See the change in `ProxyRequestHandler`. Are you ok with this modification?

oneonestar (approved):
LGTM.
Not sure is `addXForwardedHeaders` a good name for the behavior, but I can't come up with a better name.

willmostly:
@mosabua any comments on the docs?

docs/installation.md
```diff
@@ -63,7 +63,8 @@ From a users perspective Trino Gateway acts as a transparent proxy for one
 or more Trino clusters. The following Trino configuration tips should be 
 taken into account for all clusters behind the Trino Gateway.
 
-Process forwarded HTTP headers must be enabled:
+If all client/server communication is to be routed through the Trino Gateway,
```
mosabua, line 66:
```suggestion
If all client/server communication is routed through the Trino Gateway, 
```

docs/installation.md
```diff
@@ -73,16 +74,21 @@ Without this setting, first requests go from the user to Trino Gateway and then
 to Trino correctly. However, the URL for subsequent next URIs for more results
 in a query provided by Trino is then using the local URL of the Trino cluster,
 and not the URL of the Trino Gateway. This circumvents the Trino Gateway for all
-these requests, and is contrary to the purpose of using the Trino Gateway. In
-scenarios, where the local URL of the Trino cluster is private to the Trino
-cluster on the DNS/network level, these following calls might not work at all
-for users.
+these requests. In scenarios, where the local URL of the Trino cluster is private 
+to the Trino cluster on the DNS/network level, these following calls might not work
```
mosabua, line 78:
```suggestion
to the Trino cluster on the network level, these following calls do not work
```

mosabua:
My one question would be why this cant be a normal part of serverConfig ... I would rather not have all these separate configs and add yet one more.

willmostly:
> My one question would be why this cant be a normal part of serverConfig ... I would rather not have all these separate configs and add yet one more.

That is a good suggestion. The answer to your question is that `RoutingConfiguration` is already used in `ProxyRequestHandler` to set request timeout. 

`serverConfig` makes logical sense to me, but unfortunately it is just a map, we don't have a proper `ServerConfig` class where we could define this currently. I will see if I can figure something out here...

willmostly:
> > My one question would be why this cant be a normal part of serverConfig ... I would rather not have all these separate configs and add yet one more.
> 
> That is a good suggestion. The answer to your question is that `RoutingConfiguration` is already used in `ProxyRequestHandler` to set request timeout.
> 
> `serverConfig` makes logical sense to me, but unfortunately it is just a map, we don't have a proper `ServerConfig` class where we could define this currently. I will see if I can figure something out here...

I played around with this - mixing additional properties in with the airlift http-server properties was awkward. I could move `addXForwardedHeaders` out from `routing` and make it a top level property, wdyt @mosabua

mosabua (approved):
This looks good now. I think we will probably end up cramming more config into `routing` and we will have to think how we want this to grow in terms using config.properties as in trino and the yaml file we have. Might end up having to figure out something in airlift

PR #418, opened 2024-07-24
Add process forwarded info

mosabua:
## Description

As discussed in the sync

## Release notes

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -63,7 +63,7 @@ From a users perspective Trino Gateway acts as a transparent proxy for one
 or more Trino clusters. The following Trino configuration tips should be 
 taken into account for all clusters behind the Trino Gateway.
 
-Process forwarding must be enabled:
+Process forwarded HTTP headers must be enabled:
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -131,6 +131,18 @@ extraWhitelistPaths:
...
+## Configure behind a load balancer
+
+A possible deployment of Trino Gateway is to run multiple instances of Trino 
+Gateway behind another generic load balancer, such as a load balancer from 
+your cloud hosting provider. In this deployment you must configure the 
+`serverConfig` to include enabling process forwarded HTTP headers:
+
+```yaml
+serverConfig:
+    http-server.process-forwarded: true
```
willmostly, line 143: nit: use two spaces for indent

willmostly (approved):
Nit, otherwise lgtm

docs/installation.md
mosabua: ok .. will adjust in other docs to that as well in a follow up

PR #424, opened 2024-07-26
Configure health check timeouts

willmostly:
## Description
Add timeout parameters for INFO_API and JDBC health checks.

## Release notes

Add timeout parameters for INFO_API and JDBC health checks.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -248,6 +248,72 @@ configuration. Note that the `subPath` is not strictly necessary, and if it
 is not specified the file is mounted at `mountPath/<configMap key>`. 
 Kubernetes updates the mounted file when the ConfigMap is updated.
 
-Standard Helm oœptions such as `replicaCount`, `image`, `imagePullSecrets`, 
+Standard Helm options such as `replicaCount`, `image`, `imagePullSecrets`, 
 `service`, `ingress` and `resources` are supported. These are defined in 
 `helm/values.yaml`. 
+
+### Health Checks
+
+Trino Gateway checks the health of each backend and **deactivates it if 
+unhealthy**. A backend that fails a health check must be manually reset to 
+active. Automatic recovery is not supported.
+
+The type of health check is configured by setting
+
+```yaml
+clusterStatsConfiguration:
+  monitorType: ""
+```
+
+to one of the following values.
+
+#### INFO_API (default)
+
+By default Trino Gateway uses the `v1/info` REST endpoint. A successful check is
+defined as a 200 response with `starting: false`. Connection timeout parameters 
+can be defined through the `monitor` node, for example
+
+```yaml
+monitor:
+  connectTimeoutSeconds: 5
+  requestTimeoutSeconds: 10
+  idleTimeoutSeconds: 1
+  retries: 1
+```
+
+All timeout parameters are optional.
+
+#### JDBC
+
+This uses a JDBC connection to query `system.runtime` tables for cluster 
+information. It is required for the query count based routing strategy. This is
+recommended over `UI_API` since it does not restrict the Web UI authentication
+method of backend clusters. Configure a username and password by adding
+`backendState` to your configuration. The username and password must be valid 
+across all backends.
+
+```yaml
+backendState:
+  username: "user"
+  password: "password"
+```
+
+The request timeout can be set through
+
+```yaml
+monitor:
+  requestTimeoutSeconds: 10
+```
+
+Other timeout parameters are not applicable to the JDBC connection.
+
+#### UI_API
+
+This pulls cluster information from the `ui/api/stats` REST endpoint. This is
+supported for legacy reasons and may be deprecated in the future. It is only 
+supported for backend clusters with `web-ui.authentication.type = PASSWORD`. Set
+a username and password using `backendState` as with the `JDBC` option.
+
+#### NOOP
+
+This option disables health checks.
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
@@ -16,9 +16,10 @@ package io.trino.gateway.ha.clustermonitor;
 import io.airlift.http.client.HttpClient;
 import io.airlift.http.client.JsonResponseHandler;
 import io.airlift.http.client.Request;
+import io.airlift.http.client.UnexpectedResponseException;
+import io.airlift.log.Logger;
+import io.trino.gateway.ha.config.MonitorConfiguration;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 
@@ -26,18 +27,23 @@ import static io.airlift.http.client.HttpUriBuilder.uriBuilderFrom;
 import static io.airlift.http.client.JsonResponseHandler.createJsonResponseHandler;
 import static io.airlift.http.client.Request.Builder.prepareGet;
 import static io.airlift.json.JsonCodec.jsonCodec;
+import static java.net.HttpURLConnection.HTTP_BAD_GATEWAY;
+import static java.net.HttpURLConnection.HTTP_GATEWAY_TIMEOUT;
+import static java.net.HttpURLConnection.HTTP_UNAVAILABLE;
 import static java.util.Objects.requireNonNull;
 
 public class ClusterStatsInfoApiMonitor
         implements ClusterStatsMonitor
 {
-    private static final Logger log = LoggerFactory.getLogger(ClusterStatsInfoApiMonitor.class);
+    private static final Logger log = Logger.get(ClusterStatsInfoApiMonitor.class);
     private static final JsonResponseHandler<ServerInfo> SERVER_INFO_JSON_RESPONSE_HANDLER = createJsonResponseHandler(jsonCodec(ServerInfo.class));
     private final HttpClient client;
+    int retries;
 
-    public ClusterStatsInfoApiMonitor(HttpClient client)
+    public ClusterStatsInfoApiMonitor(HttpClient client, MonitorConfiguration monitorConfiguration)
     {
         this.client = requireNonNull(client, "client is null");
+        retries = monitorConfiguration.getRetries();
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


docs/installation.md
```diff
@@ -244,3 +244,51 @@ Kubernetes updates the mounted file when the ConfigMap is updated.
 Standard Helm oœptions such as `replicaCount`, `image`, `imagePullSecrets`, 
 `service`, `ingress` and `resources` are supported. These are defined in 
 `helm/values.yaml`. 
+
+### Health Checks
+
+Trino Gateway checks the health of each backend and deactivates it if unhealthy.
```
andythsu, line 250: I'd probably bold "deactivates it if unhealthy" because it's an important behavior and sometimes confuses people why their backends are deactivated.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -47,18 +63,42 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
                 .build();
-
         try {
             ServerInfo serverInfo = client.execute(request, SERVER_INFO_JSON_RESPONSE_HANDLER);
             return !serverInfo.isStarting();
         }
+        catch (UnexpectedResponseException e) {
+            if (shouldRetry(e.getStatusCode())) {
+                if (retriesRemaining > 0) {
+                    return isReadyStatus(baseUrl, retriesRemaining - 1);
```
andythsu, line 82: I'd log something to inform that the retries are happening. Maybe something like `log.warn("received an retriable error, retrying...")

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -47,18 +63,42 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
+                }
+            }
+            log.warn(e, "Health check failed with non-retryable response. %s", e.toString());
```
andythsu, line 85: Not only does non-retryable response get here. We will also end up in this statement if `retries == 0`. We should add something to distinguish between these two states

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
```diff
@@ -73,8 +81,16 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
+            networkTimeoutMillis.ifPresent(t -> {
+                try {
+                    conn.setNetworkTimeout(Executors.newSingleThreadExecutor(), 10); //executor is unused in TrinoConnection.setNetworkTimeout
+                }
+                catch (SQLException e) {
+                    log.error(e, "Could not set network timeout for JDBC health monitor");
+                }
+            });
+            PreparedStatement stmt = SimpleTimeLimiter.create(Executors.newSingleThreadExecutor()).callWithTimeout(
+                    () -> conn.prepareStatement(STATE_QUERY), max(10, 1000 * networkTimeoutMillis.orElse(0)) + 1, TimeUnit.SECONDS);
```
andythsu, line 93: we should add a comment here to explain the 10 seconds magic number here. Basically this is querying the `system.runtime` table and we don't expect more than 10 seconds, and hence it is "hardcoded" here and not configurable. There were some asks to make it configurable, which is something we can do, but it doesn't add much value since 10 seconds should be more than enough

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
vishalya: Why is this a warning and not an error? even though it's non retry-able response.

mosabua:
Let us know when this is rebased and ready again @willmostly

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
willmostly: This is making it configurable - I was using 10 as the default, but did so in a confusing way, and screwed up the conversion from millis to seconds to boot.

willmostly: this was reverted since the network config was superseded by https://github.com/trinodb/trino-gateway/pull/436

willmostly:
Note that the configuration options introduced here are superseded by https://github.com/trinodb/trino-gateway/pull/436. Retries are now the only thing being added

willmostly:
@andythsu and @vishalya this is ready for another look

andythsu (approved):
I'm good with this one! 🚢 it

mosabua (commented):
Please add empty lines after titles, before and after codeblocks and lists and verify that the docs actually render properly

mosabua (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -50,18 +56,48 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
     }
 
     private boolean isReadyStatus(String baseUrl)
+    {
+        return isReadyStatus(baseUrl, retries);
+    }
```
ebyhr, line 61: This method seems redundant. L52 could call `isReadyStatus(String baseUrl, int retriesRemaining)` instead.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -50,18 +56,48 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
         Request request = prepareGet()
                 .setUri(uriBuilderFrom(URI.create(baseUrl)).appendPath("/v1/info").build())
                 .build();
-
         try {
             ServerInfo serverInfo = client.execute(request, SERVER_INFO_JSON_RESPONSE_HANDLER);
             return !serverInfo.isStarting();
         }
+        catch (UnexpectedResponseException e) {
+            if (shouldRetry(e.getStatusCode())) {
```
ebyhr, line 73: Please consider replacing this retry logic with [Failsafe](https://failsafe.dev/). We widely use this library in Trino.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -50,18 +56,48 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
+    public static boolean shouldRetry(int statusCode)
+    {
+        switch (statusCode) {
+            case HTTP_BAD_GATEWAY:
+            case HTTP_UNAVAILABLE:
+            case HTTP_GATEWAY_TIMEOUT:
+                return true;
+            default:
+                return false;
+        }
```
ebyhr, line 101: We could use switch expression.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -16,28 +16,34 @@
...
 import static java.util.Objects.requireNonNull;
 
 public class ClusterStatsInfoApiMonitor
         implements ClusterStatsMonitor
 {
-    private static final Logger log = LoggerFactory.getLogger(ClusterStatsInfoApiMonitor.class);
+    private static final Logger log = Logger.get(ClusterStatsInfoApiMonitor.class);
     private static final JsonResponseHandler<ServerInfo> SERVER_INFO_JSON_RESPONSE_HANDLER = createJsonResponseHandler(jsonCodec(ServerInfo.class));
     private final HttpClient client;
+    int retries;
```
ebyhr, line 41: Add `private final`.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
```diff
@@ -43,7 +44,7 @@ public class ClusterStatsJdbcMonitor
             + "WHERE user != ? AND date_diff('hour',created,now()) <= 1 "
             + "GROUP BY state";
 
-    public ClusterStatsJdbcMonitor(BackendStateConfiguration backendStateConfiguration)
+    public ClusterStatsJdbcMonitor(BackendStateConfiguration backendStateConfiguration, MonitorConfiguration monitorConfiguration)
```
ebyhr, line 47: `MonitorConfiguration monitorConfiguration` is unused.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
```diff
@@ -73,8 +74,8 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
         }
 
         try (Connection conn = DriverManager.getConnection(jdbcUrl, properties)) {
-            PreparedStatement stmt = SimpleTimeLimiter.create(Executors.newSingleThreadExecutor())
-                    .callWithTimeout(() -> conn.prepareStatement(STATE_QUERY), 10, TimeUnit.SECONDS);
+            PreparedStatement stmt = SimpleTimeLimiter.create(Executors.newSingleThreadExecutor()).callWithTimeout(
+                    () -> conn.prepareStatement(STATE_QUERY), 10, TimeUnit.SECONDS);
```
ebyhr, line 78: This change seems unrelated to "Configure health check timeouts".

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsInfoApiMonitor.java
```diff
@@ -50,18 +56,48 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
                 .build();
-
         try {
             ServerInfo serverInfo = client.execute(request, SERVER_INFO_JSON_RESPONSE_HANDLER);
             return !serverInfo.isStarting();
         }
+        catch (UnexpectedResponseException e) {
+            if (shouldRetry(e.getStatusCode())) {
+                if (retriesRemaining > 0) {
+                    log.warn("Retrying health check on error: %s, ", e.toString());
```
ebyhr, line 75: `,` after `%s` seems redundant.

PR #425, opened 2024-07-29
Extract query ID from all kill_query procedure variations

willmostly:
## Description

Trino Gateway attempts to extract the query id from the body of queries using the [runtime.kill_query](https://trino.io/docs/current/connector/system.html#system-connector-procedures) procedure from the system catalog. This adds a test for this functionality.

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
@@ -17,7 +17,7 @@ import jakarta.validation.constraints.Max;
 
 public class RequestAnalyzerConfig
 {
-    private Integer maxBodySize = 1_000_000;
+    private int maxBodySize = 1_000_000;
 
     private boolean isClientsUseV2Format;
     private String tokenUserField = "email";
@@ -26,13 +26,13 @@ public class RequestAnalyzerConfig
 
     public RequestAnalyzerConfig() {}
 
-    public Integer getMaxBodySize()
+    public int getMaxBodySize()
     {
         return maxBodySize;
     }
 
     @Max(Integer.MAX_VALUE)
-    public void setMaxBodySize(Integer maxBodySize)
+    public void setMaxBodySize(int maxBodySize)
     {
         this.maxBodySize = maxBodySize;
     }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
@@ -15,8 +15,11 @@ package io.trino.gateway.ha.handler;
 
 import com.google.common.io.CharStreams;
 import io.airlift.log.Logger;
+import io.trino.gateway.ha.router.TrinoQueryProperties;
 import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.HttpMethod;
 
+import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.List;
 import java.util.Optional;
@@ -49,51 +52,50 @@ public final class ProxyUtils
      * capitalization.
      */
     private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
-    private static final Pattern EXTRACT_BETWEEN_SINGLE_QUOTES = Pattern.compile("'([^\\s']+)'");
 
     private ProxyUtils() {}
 
-    public static String extractQueryIdIfPresent(HttpServletRequest request, List<String> statementPaths)
+    public static Optional<String> extractQueryIdIfPresent(
+            HttpServletRequest request,
+            List<String> statementPaths,
+            boolean requestAnalyserClientsUseV2Format,
+            int requestAnalyserMaxBodySize)
     {
         String path = request.getRequestURI();
         String queryParams = request.getQueryString();
+        if (!request.getMethod().equals(HttpMethod.POST)) {
+            return extractQueryIdIfPresent(path, queryParams, statementPaths);
+        }
+        String queryText;
         try {
-            String queryText = CharStreams.toString(new InputStreamReader(request.getInputStream(), UTF_8));
-            if (!isNullOrEmpty(queryText)
-                    && queryText.toLowerCase(ENGLISH).contains("system.runtime.kill_query")) {
-                // extract and return the queryId
-                String[] parts = queryText.split(",");
-                for (String part : parts) {
-                    if (part.contains("query_id")) {
-                        Matcher matcher = EXTRACT_BETWEEN_SINGLE_QUOTES.matcher(part);
-                        if (matcher.find()) {
-                            String queryQuoted = matcher.group();
-                            if (!isNullOrEmpty(queryQuoted) && queryQuoted.length() > 0) {
-                                return queryQuoted.substring(1, queryQuoted.length() - 1);
-                            }
-                        }
-                    }
-                }
-            }
+            queryText = CharStreams.toString(new InputStreamReader(request.getInputStream(), UTF_8));
         }
-        catch (Exception e) {
-            log.error(e, "Error extracting query payload from request");
+        catch (IOException e) {
+            throw new RuntimeException("Error reading request body", e);
         }
-
-        return extractQueryIdIfPresent(path, queryParams, statementPaths);
+        if (!isNullOrEmpty(queryText) && queryText.toLowerCase(ENGLISH).contains("kill_query")) {
+            TrinoQueryProperties trinoQueryProperties = new TrinoQueryProperties(request, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
+            return trinoQueryProperties.getQueryId();
+        }
+        return Optional.empty();
     }
 
-    public static String extractQueryIdIfPresent(String path, String queryParams, List<String> statementPaths)
+    public static Optional<String> extractQueryIdIfPresent(String path, String queryParams, List<String> statementPaths)
     {
         if (path == null) {
-            return null;
+            return Optional.empty();
         }
-        String queryId = null;
         log.debug("Trying to extract query id from path [%s] or queryString [%s]", path, queryParams);
         // matchingStatementPath should match paths such as /v1/statement/executing/query_id/nonce/sequence_number,
         // and if custom paths are supplied using the statementPaths configuration, paths such as
         // /custom/statement/path/executing/query_id/nonce/sequence_number
         Optional<String> matchingStatementPath = statementPaths.stream().filter(path::startsWith).findAny();
+        if (!isNullOrEmpty(queryParams)) {
+            Matcher matcher = QUERY_ID_PARAM_PATTERN.matcher(queryParams);
+            if (matcher.matches()) {
+                return Optional.of(matcher.group(1));
+            }
+        }
         if (matchingStatementPath.isPresent() || path.startsWith(V1_QUERY_PATH)) {
             path = path.replace(matchingStatementPath.orElse(V1_QUERY_PATH), "");
             String[] tokens = path.split("/");
@@ -102,27 +104,20 @@ public final class ProxyUtils
                         || tokens[1].equals("scheduled")
                         || tokens[1].equals("executing")
                         || tokens[1].equals("partialCancel")) {
-                    queryId = tokens[2];
+                    return Optional.of(tokens[2]);
                 }
                 else {
-                    queryId = tokens[1];
+                    return Optional.of(tokens[1]);
                 }
             }
         }
         else if (path.startsWith(TRINO_UI_PATH)) {
             Matcher matcher = QUERY_ID_PATH_PATTERN.matcher(path);
             if (matcher.matches()) {
-                queryId = matcher.group(1);
+                return Optional.of(matcher.group(1));
             }
         }
-        if (!isNullOrEmpty(queryParams)) {
-            Matcher matcher = QUERY_ID_PARAM_PATTERN.matcher(queryParams);
-            if (matcher.matches()) {
-                queryId = matcher.group(1);
-            }
-        }
-        log.debug("Query id in URL [%s]", queryId);
-        return queryId;
+        return Optional.empty();
     }
 
     public static String buildUriWithNewBackend(String backendHost, HttpServletRequest request)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -47,6 +47,8 @@ public class RoutingTargetHandler
     private final RoutingGroupSelector routingGroupSelector;
     private final List<String> statementPaths;
     private final List<Pattern> extraWhitelistPaths;
+    private final boolean requestAnalyserClientsUseV2Format;
+    private final int requestAnalyserMaxBodySize;
     private final boolean cookiesEnabled;
 
     @Inject
@@ -57,8 +59,10 @@ public class RoutingTargetHandler
     {
         this.routingManager = requireNonNull(routingManager);
         this.routingGroupSelector = requireNonNull(routingGroupSelector);
-        this.statementPaths = requireNonNull(haGatewayConfiguration.getStatementPaths());
-        this.extraWhitelistPaths = requireNonNull(haGatewayConfiguration.getExtraWhitelistPaths()).stream().map(Pattern::compile).collect(toImmutableList());
+        statementPaths = requireNonNull(haGatewayConfiguration.getStatementPaths());
+        extraWhitelistPaths = requireNonNull(haGatewayConfiguration.getExtraWhitelistPaths()).stream().map(Pattern::compile).collect(toImmutableList());
+        requestAnalyserClientsUseV2Format = haGatewayConfiguration.getRequestAnalyzerConfig().isClientsUseV2Format();
+        requestAnalyserMaxBodySize = haGatewayConfiguration.getRequestAnalyzerConfig().getMaxBodySize();
         cookiesEnabled = GatewayCookieConfigurationPropertiesProvider.getInstance().isEnabled();
     }
 
@@ -96,9 +100,9 @@ public class RoutingTargetHandler
 
     private Optional<String> getPreviousBackend(HttpServletRequest request)
     {
-        String queryId = extractQueryIdIfPresent(request, statementPaths);
-        if (!isNullOrEmpty(queryId)) {
-            return Optional.of(routingManager.findBackendForQueryId(queryId));
+        Optional<String> queryId = extractQueryIdIfPresent(request, statementPaths, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
+        if (queryId.isPresent()) {
+            return queryId.map(routingManager::findBackendForQueryId);
         }
         if (cookiesEnabled && request.getCookies() != null) {
             List<GatewayCookie> cookies = Arrays.stream(request.getCookies())
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
@@ -14,21 +14,23 @@
 package io.trino.gateway.ha.router;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.airlift.compress.zstd.ZstdDecompressor;
 import io.airlift.json.JsonCodec;
 import io.airlift.log.Logger;
-import io.trino.gateway.ha.config.RequestAnalyzerConfig;
 import io.trino.sql.parser.ParsingException;
 import io.trino.sql.parser.SqlParser;
 import io.trino.sql.tree.AddColumn;
 import io.trino.sql.tree.Analyze;
+import io.trino.sql.tree.Call;
 import io.trino.sql.tree.CreateCatalog;
 import io.trino.sql.tree.CreateMaterializedView;
 import io.trino.sql.tree.CreateSchema;
@@ -40,6 +42,7 @@ import io.trino.sql.tree.DropSchema;
 import io.trino.sql.tree.DropTable;
 import io.trino.sql.tree.Execute;
 import io.trino.sql.tree.ExecuteImmediate;
+import io.trino.sql.tree.Expression;
 import io.trino.sql.tree.Identifier;
 import io.trino.sql.tree.Node;
 import io.trino.sql.tree.NodeLocation;
@@ -57,6 +60,7 @@ import io.trino.sql.tree.ShowCreate;
 import io.trino.sql.tree.ShowSchemas;
 import io.trino.sql.tree.ShowTables;
 import io.trino.sql.tree.Statement;
+import io.trino.sql.tree.StringLiteral;
 import io.trino.sql.tree.Table;
 import io.trino.sql.tree.TableFunctionInvocation;
 import jakarta.servlet.http.HttpServletRequest;
@@ -73,6 +77,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.io.BaseEncoding.base64Url;
 import static io.airlift.json.JsonCodec.jsonCodec;
 import static java.lang.Math.toIntExact;
@@ -85,6 +90,7 @@ public class TrinoQueryProperties
 {
     private final Logger log = Logger.get(TrinoQueryProperties.class);
     private final boolean isClientsUseV2Format;
+    private final int maxBodySize;
     private String body = "";
     private String queryType = "";
     private String resourceGroupQueryType = "";
@@ -96,6 +102,7 @@ public class TrinoQueryProperties
     private Set<String> catalogSchemas = ImmutableSet.of();
     private boolean isNewQuerySubmission;
     private Optional<String> errorMessage = Optional.empty();
+    private Optional<String> queryId = Optional.empty();
 
     public static final String TRINO_CATALOG_HEADER_NAME = "X-Trino-Catalog";
     public static final String TRINO_SCHEMA_HEADER_NAME = "X-Trino-Schema";
@@ -128,21 +135,24 @@ public class TrinoQueryProperties
         this.isNewQuerySubmission = isNewQuerySubmission;
         this.errorMessage = requireNonNullElse(errorMessage, Optional.empty());
         isClientsUseV2Format = false;
+        maxBodySize = -1;
     }
 
-    public TrinoQueryProperties(HttpServletRequest request, RequestAnalyzerConfig config)
+    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
     {
-        isClientsUseV2Format = config.isClientsUseV2Format();
+        requireNonNull(request, "request is null");
+        this.isClientsUseV2Format = isClientsUseV2Format;
+        this.maxBodySize = maxBodySize;
 
         defaultCatalog = Optional.ofNullable(request.getHeader(TRINO_CATALOG_HEADER_NAME));
         defaultSchema = Optional.ofNullable(request.getHeader(TRINO_SCHEMA_HEADER_NAME));
         if (request.getMethod().equals(HttpMethod.POST)) {
             isNewQuerySubmission = true;
-            processRequestBody(request, config);
+            processRequestBody(request);
         }
     }
 
-    private void processRequestBody(HttpServletRequest request, RequestAnalyzerConfig config)
+    private void processRequestBody(HttpServletRequest request)
     {
         try (BufferedReader reader = request.getReader()) {
             if (reader == null) {
@@ -153,11 +163,11 @@ public class TrinoQueryProperties
 
             Map<String, String> preparedStatements = getPreparedStatements(request);
             SqlParser parser = new SqlParser();
-            reader.mark(config.getMaxBodySize());
-            char[] buffer = new char[config.getMaxBodySize()];
-            int nChars = reader.read(buffer, 0, config.getMaxBodySize());
+            reader.mark(maxBodySize);
+            char[] buffer = new char[maxBodySize];
+            int nChars = reader.read(buffer, 0, maxBodySize);
             reader.reset();
-            if (nChars == config.getMaxBodySize()) {
+            if (nChars == maxBodySize) {
                 log.warn("Query length greater or equal to requestAnalyzerConfig.maxBodySize detected");
                 return;
                 //The body is truncated - there is a chance that it could still be syntactically valid SQL, for example if truncated on
@@ -199,7 +209,7 @@ public class TrinoQueryProperties
             ImmutableSet.Builder<String> schemaBuilder = ImmutableSet.builder();
             ImmutableSet.Builder<String> catalogSchemaBuilder = ImmutableSet.builder();
 
-            getNames(statement, tableBuilder, catalogBuilder, schemaBuilder, catalogSchemaBuilder);
+            visitNode(statement, tableBuilder, catalogBuilder, schemaBuilder, catalogSchemaBuilder);
             tables = tableBuilder.build();
             catalogBuilder.addAll(tables.stream().map(q -> q.getParts().getFirst()).iterator());
             catalogs = catalogBuilder.build();
@@ -260,7 +270,7 @@ public class TrinoQueryProperties
         return new String(preparedStatement, UTF_8);
     }
 
-    private void getNames(Node node, ImmutableSet.Builder<QualifiedName> tableBuilder,
+    private void visitNode(Node node, ImmutableSet.Builder<QualifiedName> tableBuilder,
             ImmutableSet.Builder<String> catalogBuilder,
             ImmutableSet.Builder<String> schemaBuilder,
             ImmutableSet.Builder<String> catalogSchemaBuilder)
@@ -269,6 +279,7 @@ public class TrinoQueryProperties
         switch (node) {
             case AddColumn s -> tableBuilder.add(qualifyName(s.getName()));
             case Analyze s -> tableBuilder.add(qualifyName(s.getTableName()));
+            case Call call -> queryId = extractQueryIdFromCall(call);
             case CreateCatalog s -> catalogBuilder.add(s.getCatalogName().getValue());
             case CreateMaterializedView s -> tableBuilder.add(qualifyName(s.getName()));
             case CreateSchema s -> setCatalogAndSchemaNameFromSchemaQualifiedName(Optional.of(s.getSchemaName()), catalogBuilder, schemaBuilder, catalogSchemaBuilder);
@@ -342,8 +353,20 @@ public class TrinoQueryProperties
         }
 
         for (Node child : node.getChildren()) {
-            getNames(child, tableBuilder, catalogBuilder, schemaBuilder, catalogSchemaBuilder);
+            visitNode(child, tableBuilder, catalogBuilder, schemaBuilder, catalogSchemaBuilder);
         }
+    }
+
+    private Optional<String> extractQueryIdFromCall(Call call)
+            throws RequestParsingException
+    {
+        QualifiedName callName = qualifyName(call.getName());
+        if (callName.equals(QualifiedName.of("system", "runtime", "kill_query"))) {
+            Expression argument = call.getArguments().getFirst().getValue();
+            checkArgument(argument instanceof StringLiteral, "Unable to route kill_query procedures where the first argument is not a String Literal");
+            return Optional.of(((StringLiteral) argument).getValue());
+        }
+        return Optional.empty();
     }
 
     private void setCatalogAndSchemaNameFromSchemaQualifiedName(
@@ -381,15 +404,16 @@ public class TrinoQueryProperties
         return new RequestParsingException("Name not fully qualified");
     }
 
-    private QualifiedName qualifyName(QualifiedName table)
+    private QualifiedName qualifyName(QualifiedName name)
             throws RequestParsingException
     {
-        List<String> tableParts = table.getParts();
-        return switch (tableParts.size()) {
-            case 1 -> QualifiedName.of(defaultCatalog.orElseThrow(this::unsetDefaultExceptionSupplier), defaultSchema.orElseThrow(this::unsetDefaultExceptionSupplier), tableParts.getFirst());
-            case 2 -> QualifiedName.of(defaultCatalog.orElseThrow(this::unsetDefaultExceptionSupplier), tableParts.getFirst(), tableParts.get(1));
-            case 3 -> QualifiedName.of(tableParts.getFirst(), tableParts.get(1), tableParts.get(2));
-            default -> throw new RequestParsingException("Unexpected table name: " + table.getParts());
+        List<String> nameParts = name.getParts();
+        return switch (nameParts.size()) {
+            case 1 ->
+                    QualifiedName.of(defaultCatalog.orElseThrow(this::unsetDefaultExceptionSupplier), defaultSchema.orElseThrow(this::unsetDefaultExceptionSupplier), nameParts.getFirst());
+            case 2 -> QualifiedName.of(defaultCatalog.orElseThrow(this::unsetDefaultExceptionSupplier), nameParts.getFirst(), nameParts.get(1));
+            case 3 -> QualifiedName.of(nameParts.getFirst(), nameParts.get(1), nameParts.get(2));
+            default -> throw new RequestParsingException("Unexpected qualified name: " + name.getParts());
         };
     }
 
@@ -518,6 +542,12 @@ public class TrinoQueryProperties
     public Optional<String> getErrorMessage()
     {
         return errorMessage;
+    }
+
+    @JsonIgnore
+    public Optional<String> getQueryId()
+    {
+        return queryId;
     }
 
     public static class AlternateStatementRequestBodyFormat
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
@@ -66,7 +66,6 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.net.URLDecoder;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
@@ -107,7 +106,7 @@ public class TrinoQueryProperties
             @JsonProperty("body") String body,
             @JsonProperty("queryType") String queryType,
             @JsonProperty("resourceGroupQueryType") String resourceGroupQueryType,
-            @JsonProperty("tables") String[] tables,
+            @JsonProperty("tables") List<String> tables,
             @JsonProperty("defaultCatalog") Optional<String> defaultCatalog,
             @JsonProperty("defaultSchema") Optional<String> defaultSchema,
             @JsonProperty("catalogs") Set<String> catalogs,
@@ -119,7 +118,8 @@ public class TrinoQueryProperties
         this.body = requireNonNullElse(body, "");
         this.queryType = requireNonNullElse(queryType, "");
         this.resourceGroupQueryType = resourceGroupQueryType;
-        this.tables = Arrays.stream(requireNonNullElse(tables, new String[] {})).map(this::parseIdentifierStringToQualifiedName).collect(Collectors.toSet());
+        List<String> defaultTables = ImmutableList.of();
+        this.tables = requireNonNullElse(tables, defaultTables).stream().map(this::parseIdentifierStringToQualifiedName).collect(Collectors.toSet());
         this.defaultCatalog = requireNonNullElse(defaultCatalog, Optional.empty());
         this.defaultSchema = requireNonNullElse(defaultSchema, Optional.empty());
         this.catalogs = requireNonNullElse(catalogs, ImmutableSet.of());
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
@@ -14,15 +14,27 @@
 package io.trino.gateway.ha.handler;
 
 import com.google.common.collect.ImmutableList;
+import jakarta.servlet.ReadListener;
+import jakarta.servlet.ServletInputStream;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.HttpMethod;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.mockito.Mockito;
 
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.StringReader;
 import java.util.List;
+import java.util.Optional;
 
 import static io.trino.gateway.ha.handler.ProxyUtils.extractQueryIdIfPresent;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.when;
 
 @TestInstance(Lifecycle.PER_CLASS)
 final class TestQueryIdCachingProxyHandler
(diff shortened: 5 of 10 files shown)


mosabua (approved):
Looks fine .. I wish we would not have to mock around for stuff like that but a full integration test seems worse..

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,44 @@ public void testExtractQueryIdFromUrl()
                 .isNull();
     }
 
+    @Test
+    public void testQueryIdFromKill()
+            throws IOException
+    {
+        HttpServletRequest req = Mockito.mock(HttpServletRequest.class);
```
ebyhr, line 73: Rename req to request. Please avoid abbreviations: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,44 @@ public void testExtractQueryIdFromUrl()
...
+
+        String query = "CALL system.runtime.kill_query(query_id ==> '20200416_160256_03078_6b4yt', message ==> 'If he dies, he dies')";
```
ebyhr, line 75: `==>` should be `=>`. Also, please add other cases:
* Without message argument `CALL system.runtime.kill_query(query_id => '')`
* Without name arguments, `CALL system.runtime.kill_query('...', '...')`

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,44 @@ public void testExtractQueryIdFromUrl()
...
+        ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(query.getBytes(StandardCharsets.UTF_8));
```
ebyhr, line 76: Static import `UTF_8`.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,44 @@ public void testExtractQueryIdFromUrl()
                 .isNull();
     }
 
+    @Test
+    public void testQueryIdFromKill()
```
ebyhr, line 70: Remove `public`. https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
> Test methods should be defined as package-private.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -50,7 +50,8 @@ public final class ProxyUtils
      * capitalization.
      */
     private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
-    private static final Pattern EXTRACT_BETWEEN_SINGLE_QUOTES = Pattern.compile("'([^\\s']+)'");
+    private static final Pattern KILL_QUERY_PROCEDURE_PATTERN
```
ebyhr, line 53: > Test extraction of query ID from kill_query procedure

The change isn't specific to tests. Both PR and commit title should be updated.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
willmostly: The extra test cases identified that the logic was relying on the named `query_id` argument, so I have corrected the logic in `extractQueryIdIfPresent` to handle these variations. Thanks!

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,63 @@ public void testExtractQueryIdFromUrl()
...
+                extractQueryIdIfPresent(
+                        prepareMockRequestWithBody("CALL system.runtime.kill_query(query_id => '20200416_160256_03078_6b4yt', message => 'If he dies, he dies')")))
+                .isEqualTo("20200416_160256_03078_6b4yt");
+
+        assertThat(
+                extractQueryIdIfPresent(
+                        prepareMockRequestWithBody("CALL system.runtime.kill_query('20200416_160256_03078_6b4yt', 'If he dies, he dies')")))
+                .isEqualTo("20200416_160256_03078_6b4yt");
+
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("CALL system.runtime.kill_query( query_id=>'20200416_160256_03078_6b4yt')")))
```
ebyhr, line 83:
```suggestion
        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("CALL system.runtime.kill_query(query_id=>'20200416_160256_03078_6b4yt')")))
```

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,63 @@ public void testExtractQueryIdFromUrl()
...
+                .isEqualTo("20200416_160256_03078_6b4yt");
+
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("CALL system.runtime.kill_query( '20200416_160256_03078_6b4yt')")))
```
ebyhr, line 86:
```suggestion
        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("CALL system.runtime.kill_query('20200416_160256_03078_6b4yt')")))
```

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,63 @@ public void testExtractQueryIdFromUrl()
...
+                .isEqualTo("20200416_160256_03078_6b4yt");
+    }
+
+    HttpServletRequest prepareMockRequestWithBody(String query)
```
ebyhr, line 90: This method can be `private static`.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -50,7 +50,13 @@ public final class ProxyUtils
...
+            = Pattern.compile(".*system\\.runtime\\.kill_query\\s*\\(\\s*(query_id\\s*=>)?\\s*'([^\\\\s]+)'(,\\s*(message\\s*=>\\s*)?('.*'))?\\)");
+    /**
+     * This regular expression extracts the query id from a CALL system.runtime.kill_query procedure call. It extracts the first string between quotes
+     * following the open parentheses after system.runtime.kill_query. The pattern handles the optional named arguments, and optional message argument
+     * as well as arbitrary whitespace.
+     */
```
ebyhr, line 59: The place of javadoc seems wrong.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -61,6 +66,63 @@ public void testExtractQueryIdFromUrl()
                 .isNull();
     }
 
+    @Test
+    void testQueryIdFromKill()
+            throws IOException
+    {
+        assertThat(
+                extractQueryIdIfPresent(
+                        prepareMockRequestWithBody("CALL system.runtime.kill_query(query_id => '20200416_160256_03078_6b4yt', message => 'If he dies, he dies')")))
```
ebyhr, line 75: Can we add another test case which `message` argument contains query_id text pattern?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -50,7 +50,14 @@ public final class ProxyUtils
...
     private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
-    private static final Pattern EXTRACT_BETWEEN_SINGLE_QUOTES = Pattern.compile("'([^\\s']+)'");
+
+    /**
+     * This regular expression extracts the query id from a CALL system.runtime.kill_query procedure call. It extracts the first string between quotes
+     * following the open parentheses after system.runtime.kill_query. The pattern handles the optional named arguments, and optional message argument
+     * as well as arbitrary whitespace.
+     */
+    private static final Pattern KILL_QUERY_PROCEDURE_PATTERN
+            = Pattern.compile(".*system\\.runtime\\.kill_query\\s*\\(\\s*(query_id\\s*=>)?\\s*'([^\\\\s]+?)'(,\\s*(message\\s*=>\\s*)?('.*'))?\\)");
```
ebyhr, line 60: The pattern looks wrong. Users can run the procedure without `system.runtime.` if the default catalog and schema are set.

ebyhr (changes requested):
The regular expression is incorrect in some cases.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
willmostly: the unqualified case is handled now as well, ptal. I think this should cover everything, but if not I am considering using the parser instead of this regex

mosabua:
Needs a rebase now..

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -66,6 +71,107 @@ public void testExtractQueryIdFromUrl()
                 .isNull();
     }
 
+    @Test
+    void testQueryIdFromKill()
+            throws IOException
+    {
+        assertThat(
+                extractQueryIdIfPresent(
+                        prepareMockRequestWithBody("CALL system.runtime.kill_query(query_id => '20200416_160256_03078_6b4yt', message => 'If he dies, he dies')"),
```
ebyhr, line 80: Please add a test case with uppercase procedure name `KILL_QUERY`.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -66,6 +71,107 @@ public void testExtractQueryIdFromUrl()
...
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("CALL system.runtime.kill_query('20200416_160256_03078_6b4yt')"), ImmutableList.of()))
+                .isEqualTo("20200416_160256_03078_6b4yt");
+
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("CALL kill_query('20200416_160256_03078_6b4yt')"), ImmutableList.of()))
+                .isEqualTo("20200416_160256_03078_6b4yt");
+
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("call Kill_Query('20200416_160256_03078_6b4yt')"), ImmutableList.of()))
+                .isEqualTo("20200416_160256_03078_6b4yt");
+
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("select * from postgres.query_logs.queries where sql LIKE  '%kill_query(''20200416_160256%' "),
```
ebyhr, line 124: Uppercase `select`, `from` and `where`.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -66,6 +71,107 @@ public void testExtractQueryIdFromUrl()
...
+
+        assertThat(extractQueryIdIfPresent(
+                prepareMockRequestWithBody("select * from postgres.query_logs.queries where sql LIKE  '%kill_query(''20200416_160256_03078_6b4yt' "),
+                ImmutableList.of()))
+                .isNull();
+
+        assertThat(extractQueryIdIfPresent(
+                prepareMockRequestWithBody("select * from postgres.query_logs.queries where sql LIKE 'CALL kill_query(_20200416_160256_03078_6b4yt_)' "),
+                ImmutableList.of()))
+                .isNull();
```
ebyhr, line 136: What happens if `kill_query` procedure exists in a comment? e.g.
```sql
// CALL kill_query('20200416_160256_03078_6b4yt', 'If he dies, he dies')
SELECT 1
```

willmostly:
@ebyhr I switched to using the full query parser, which should be resilient to whitespace, comments, the string literals containing the procedure name, etc

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -66,6 +75,186 @@ public void testExtractQueryIdFromUrl()
                 .isNull();
     }
 
+    @Test
+    void testQueryIdFromKill()
```
ebyhr, line 79: What's the expected behavior when the 1st argument isn't a single string literal? 
```
CALL kill_query('20200416' || '_160256_03078_6b4yt')
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,26 +90,19 @@ public static String getQueryUser(String userHeader, String authorization)
...
-                                return queryQuoted.substring(1, queryQuoted.length() - 1);
-                            }
-                        }
-                    }
+                    && queryText.toLowerCase().contains("kill_query")) {
+                TrinoQueryProperties trinoQueryProperties = new TrinoQueryProperties(request, requestAnalyzerConfig);
+                if (trinoQueryProperties.getProcedure().isPresent()
+                        && trinoQueryProperties.getProcedure().orElseThrow().getName().getParts().getLast().equalsIgnoreCase("kill_query")) {
+                    return trinoQueryProperties.getProcedureArgs().getFirst().getValue().toString()
+                            .replaceFirst("^'", "").replaceFirst("'$", "");
```
ebyhr, line 105: Why not get the value with `StringLiteral.getValue` method?

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
willmostly: IMO these should be rejected, since in general Trino Gateway cannot resolve arbitrary expressions. I'll add an explicit check

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -262,6 +267,11 @@ private void getNames(Node node, ImmutableSet.Builder<QualifiedName> tableBuilde
             ImmutableSet.Builder<String> catalogSchemaBuilder)
             throws RequestParsingException
     {
+        if (node instanceof Call) {
+            procedure = Optional.of((Call) node);
+            procedureArgs = ((Call) node).getArguments();
```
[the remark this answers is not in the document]
willmostly: i renamed the method to be more generic. The initial implementation was only concerned with extracting qualified names from SQL, but there is no reason IMO not to extract other information as well. In addition to this specific case of the `kill_query` call which contains a query id, the `Call` and `CallArguments` could be useful for routing logic. For example, perhaps the `sync_partition_metadata` procedure should always be run on a specific set of clusters.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,31 +91,28 @@ public static String getQueryUser(String userHeader, String authorization)
         return parts.get(0);
     }
 
-    public static String extractQueryIdIfPresent(HttpServletRequest request, List<String> statementPaths)
+    public static String extractQueryIdIfPresent(
+            HttpServletRequest request, List<String> statementPaths,
```
ebyhr, line 95: Wrap after `request,`

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,31 +91,28 @@ public static String getQueryUser(String userHeader, String authorization)
...
-                            if (!isNullOrEmpty(queryQuoted) && queryQuoted.length() > 0) {
-                                return queryQuoted.substring(1, queryQuoted.length() - 1);
-                            }
-                        }
+            if (!isNullOrEmpty(queryText) && queryText.toLowerCase().contains("kill_query")) {
+                TrinoQueryProperties trinoQueryProperties = new TrinoQueryProperties(request, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
+                if (trinoQueryProperties.getProcedure().isPresent()
+                        && trinoQueryProperties.getProcedure().orElseThrow().getName().getParts().getLast().equalsIgnoreCase("kill_query")) {
+                    if (!(trinoQueryProperties.getProcedureArgs().getFirst().getValue() instanceof StringLiteral)) {
+                        throw new IllegalArgumentException("Trino Gateway cannot route kill_query procedures where the first argument is not a String Literal");
```
ebyhr, line 108: No need to include software name `Trino Gateway` to the error message.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -127,21 +133,28 @@ public TrinoQueryProperties(
...
 
     public TrinoQueryProperties(HttpServletRequest request, RequestAnalyzerConfig config)
     {
-        isClientsUseV2Format = config.isClientsUseV2Format();
+        this(request, config.isClientsUseV2Format(), config.getMaxBodySize());
+    }
+
+    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
+    {
+        this.isClientsUseV2Format = isClientsUseV2Format;
```
ebyhr, line 146: Add `requireNonNull`.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
ebyhr: Use `checkArgument`.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,31 +91,28 @@ public static String getQueryUser(String userHeader, String authorization)
...
                     }
+                    return ((StringLiteral) trinoQueryProperties.getProcedureArgs().getFirst().getValue()).getValue();
                 }
             }
         }
-        catch (Exception e) {
-            log.error(e, "Error extracting query payload from request");
+        catch (IOException e) {
```
ebyhr, line 114: Please move parsing logic outside of try block.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -94,6 +98,8 @@ public class TrinoQueryProperties
     private Set<String> catalogs = ImmutableSet.of();
     private Set<String> schemas = ImmutableSet.of();
     private Set<String> catalogSchemas = ImmutableSet.of();
+    private Optional<Call> procedure = Optional.empty();
+    private List<CallArgument> procedureArgs = ImmutableList.of();
```
ebyhr, line 102: Please avoid abbreviations: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -265,6 +278,11 @@ private void getNames(Node node, ImmutableSet.Builder<QualifiedName> tableBuilde
         switch (node) {
             case AddColumn s -> tableBuilder.add(qualifyName(s.getName()));
             case Analyze s -> tableBuilder.add(qualifyName(s.getTableName()));
+            case Call s -> {
```
ebyhr, line 281: Please avoid abbreviations: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,31 +91,28 @@ public static String getQueryUser(String userHeader, String authorization)
...
-                            String queryQuoted = matcher.group();
-                            if (!isNullOrEmpty(queryQuoted) && queryQuoted.length() > 0) {
-                                return queryQuoted.substring(1, queryQuoted.length() - 1);
-                            }
-                        }
+            if (!isNullOrEmpty(queryText) && queryText.toLowerCase().contains("kill_query")) {
+                TrinoQueryProperties trinoQueryProperties = new TrinoQueryProperties(request, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
+                if (trinoQueryProperties.getProcedure().isPresent()
+                        && trinoQueryProperties.getProcedure().orElseThrow().getName().getParts().getLast().equalsIgnoreCase("kill_query")) {
+                    if (!(trinoQueryProperties.getProcedureArgs().getFirst().getValue() instanceof StringLiteral)) {
```
ebyhr, line 107: Please avoid nested code as much as possible. It's hard to read.

ebyhr (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
willmostly: this is a `boolean` primitive, so the null check is redundant

ebyhr: Sorry, I forgot to mention the variable name. We should add the check to `HttpServletRequest request`

gateway-ha/src/main/java/io/trino/gateway/ha/config/RequestAnalyzerConfig.java
```diff
@@ -17,7 +17,7 @@
 
 public class RequestAnalyzerConfig
 {
-    private Integer maxBodySize = 1_000_000;
+    private int maxBodySize = 1_000_000;
```
ebyhr, line 20: Please separate a commit and explain the motivation in a commit body.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -45,18 +46,24 @@ public class RoutingTargetHandler
...
+    private final int requestAnalyserMaxBodySize;
     private final boolean cookiesEnabled;
 
     public RoutingTargetHandler(
             RoutingManager routingManager,
             RoutingGroupSelector routingGroupSelector,
             List<String> statementPaths,
-            List<String> extraWhitelistPaths)
+            List<String> extraWhitelistPaths,
+            RequestAnalyzerConfig requestAnalyzerConfig)
```
ebyhr, line 58: I think we should bind this class and inject `RoutingManager` & `RoutingGroupSelector` instead of initializing in `HaGatewayProviderModule#getRoutingTargetHandler`.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,31 +92,28 @@ public static String getQueryUser(String userHeader, String authorization)
...
-                        }
-                    }
-                }
-            }
+            queryText = CharStreams.toString(new InputStreamReader(request.getInputStream()));
         }
-        catch (Exception e) {
-            log.error(e, "Error extracting query payload from request");
+        catch (IOException e) {
+            log.error(e, "Error reading request body");
```
ebyhr, line 108: Why do we suppress IOException here and throw an exception for non string literal argument of `kill_query` procedure? 

The `request.getInputStream()` may contain `kill_query`, but it can go to `extractQueryIdIfPresent` in case of `IOException`, right?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
willmostly: should that be a separate PR, or should I wrap it into this one?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
willmostly: If the request body contains the `kill_query` SQL, then I already know at this point that `extractQueryIdIfPresent` will not find a query id because Trino doesn't have any endpoints with query id embedded in the `path` or `queryParams` that accept a `POST`. So falling back won't help. If I don't throw on a non-string literal argument, then this request will most likely get routed to a cluster that the target query is not running on, and the user will get a potentially confusing error message back about the query not existing. So throwing on a non-string literal allows us to fail fast with a more meaningful error.  

Do you think we should throw on `IOException` as well? I was just trying to match the existing behavior, but i'm not sure that it is correct.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,34 +93,34 @@ public static String getQueryUser(String userHeader, String authorization)
...
+            throw new RuntimeException("Error reading request body", e);
```
ebyhr, line 113: Please remove the above log. Just throwing an exception here is sufficient.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
ebyhr: Follow-up PR is fine for me.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -127,21 +133,29 @@ public TrinoQueryProperties(
...
 
     public TrinoQueryProperties(HttpServletRequest request, RequestAnalyzerConfig config)
     {
-        isClientsUseV2Format = config.isClientsUseV2Format();
+        this(request, config.isClientsUseV2Format(), config.getMaxBodySize());
+    }
+
+    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
+    {
+        requireNonNull(request, "Request is null");
```
ebyhr, line 146:
```suggestion
        requireNonNull(request, "request is null");
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,34 +93,34 @@ public static String getQueryUser(String userHeader, String authorization)
...
+            List<String> statementPaths,
+            boolean requestAnalyserClientsUseV2Format,
+            int requestAnalyserMaxBodySize)
     {
         String path = request.getRequestURI();
         String queryParams = request.getQueryString();
+        if (!request.getMethod().equals(HttpMethod.POST)) {
+            return extractQueryIdIfPresent(path, queryParams, statementPaths);
+        }
+        String queryText = null;
```
ebyhr, line 107: ` = null` is redundant.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,34 +93,34 @@ public static String getQueryUser(String userHeader, String authorization)
...
         }
-        catch (Exception e) {
-            log.error(e, "Error extracting query payload from request");
+        catch (IOException e) {
+            log.error(e, "Error reading request body");
+            throw new RuntimeException("Error reading request body", e);
         }
-
-        return extractQueryIdIfPresent(path, queryParams, statementPaths);
+        if (!isNullOrEmpty(queryText) && queryText.toLowerCase().contains("kill_query")) {
```
ebyhr, line 115: Do we still need the 1st condition?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,34 +93,34 @@ public static String getQueryUser(String userHeader, String authorization)
...
+            TrinoQueryProperties trinoQueryProperties = new TrinoQueryProperties(request, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
+            if (trinoQueryProperties.getProcedure().filter(p -> p.getName().getSuffix().equals("kill_query")).isPresent()) {
+                checkArgument(trinoQueryProperties.getProcedureArguments().getFirst().getValue() instanceof StringLiteral,
+                        "Unable to route kill_query procedures where the first argument is not a String Literal");
+                return ((StringLiteral) trinoQueryProperties.getProcedureArguments().getFirst().getValue()).getValue();
+            }
+        }
+        return null;
```
ebyhr, line 123: When do we reach this condition? Is it tested?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,34 +93,34 @@ public static String getQueryUser(String userHeader, String authorization)
...
-            log.error(e, "Error extracting query payload from request");
+        catch (IOException e) {
+            log.error(e, "Error reading request body");
+            throw new RuntimeException("Error reading request body", e);
         }
-
-        return extractQueryIdIfPresent(path, queryParams, statementPaths);
+        if (!isNullOrEmpty(queryText) && queryText.toLowerCase().contains("kill_query")) {
+            TrinoQueryProperties trinoQueryProperties = new TrinoQueryProperties(request, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
+            if (trinoQueryProperties.getProcedure().filter(p -> p.getName().getSuffix().equals("kill_query")).isPresent()) {
```
ebyhr, line 117: What happens if developers create their own `kill_query` procedure in a different schema. Can we start with strict name matching like the existing implementation `system.runtime.kill_query`?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -89,34 +93,34 @@ public static String getQueryUser(String userHeader, String authorization)
...
+                checkArgument(trinoQueryProperties.getProcedureArguments().getFirst().getValue() instanceof StringLiteral,
+                        "Unable to route kill_query procedures where the first argument is not a String Literal");
+                return ((StringLiteral) trinoQueryProperties.getProcedureArguments().getFirst().getValue()).getValue();
```
ebyhr, line 120: nit: I would extract a varible:
```java
Expression argument = trinoQueryProperties.getProcedureArguments().getFirst().getValue();
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
ebyhr: Please add `@Nullable` annotation to this method.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
willmostly: https://github.com/trinodb/trino-gateway/pull/488 a separate PR seemed simpler

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
willmostly: Unfortunately `null` is used to signify that no query ID was found in either the path, query parameters or request body. Any new query that is not a `kill_query`, or a GET to a path such as `/oauth2` or `/v1/info` will reach this condition. So it is tested by the routing tests like `TestGatewayHaMultipleBackends`.

That said, I will upgrade this logic to use `Optional`, as we have done in a few other places in the code, and I'll do a follow up PR to remove the use of `null` as a special value in other places as well

ebyhr:
Could you rebase on main to resolve conflicts?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
willmostly: Just noting that this has been changed such that the IOException is not suppressed, and is treated as an error. If the gateway can't read the request body, Trino will not be able to either, so there is no reason to forward the request.

mosabua:
@ebyhr is this good to go now?

ebyhr:
@mosabua I don't think it's ready to merge. We will need a few more review rounds.

willmostly:
@ebyhr any comments on the latest revision? All previous comments have been addressed

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -95,6 +99,8 @@ public class TrinoQueryProperties
     private Set<String> catalogs = ImmutableSet.of();
     private Set<String> schemas = ImmutableSet.of();
     private Set<String> catalogSchemas = ImmutableSet.of();
+    private Optional<Call> procedure = Optional.empty();
+    private List<CallArgument> procedureArguments = ImmutableList.of();
```
ebyhr, line 103: These fields are different from other values. Please move it under `isQueryParsingSuccessful` and leave a code comment explaining they are just for internal processing.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -480,6 +499,18 @@ public boolean tablesContains(String testName)
         }
     }
 
+    public boolean procedureNameEquals(String testName)
```
ebyhr, line 502: testName -> name.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -34,32 +45,289 @@ void testExtractQueryIdFromUrl()
...
+                .hasValue("20200416_160256_03078_6b4yt");
+
+        assertThat(extractQueryIdIfPresent(prepareMockRequestWithBody("SELECT * FROM postgres.query_logs.queries WHERE sql LIKE  '%kill_query(''20200416_160256%' ", "system", "runtime"),
+                ImmutableList.of(),
+                requestAnalyzerClientsUseV2Format,
+                requestAnalyzerMaxBodySize))
+                .isEmpty();
+
+        assertThat(extractQueryIdIfPresent(
+                prepareMockRequestWithBody("select * from postgres.query_logs.queries where sql like  '%kill_query(''20200416_160256_03078_6b4yt' ", "system", "runtime"),
```
ebyhr, line 159: Uppercase keywords. Same for others.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -34,32 +45,289 @@ void testExtractQueryIdFromUrl()
...
+                .isEmpty();
+
+        assertThat(
+                extractQueryIdIfPresent(
+                        prepareMockRequestWithBody("CALL kill_query(query_id => '20200416_160256_03078_6b4yt', message => 'If he dies, he dies')", "system", "notruntime"),
+                        ImmutableList.of(), requestAnalyzerClientsUseV2Format, requestAnalyzerMaxBodySize))
+                .isEmpty();
+    }
+
+    private static HttpServletRequest prepareMockRequestWithBody(String query, String defaultCatalog)
```
ebyhr, line 270: I would rename it to `request`. Longer helper method name reduces readability of each assertion.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -34,32 +45,289 @@ void testExtractQueryIdFromUrl()
...
-                .isNull();
+                .isEmpty();
+    }
+
+    @Test
+    void testQueryIdFromKill()
+            throws IOException
+    {
+        int requestAnalyzerMaxBodySize = 1_000_000;
+        boolean requestAnalyzerClientsUseV2Format = false;
```
ebyhr, line 81: I would recommend extracting a helper method:
```java
    private static Optional<String> extractQueryId(HttpServletRequest request)
    {
        return extractQueryIdIfPresent(request, ImmutableList.of(), false, 1_000_000);
    }
```

ebyhr (commented):
Looks good except for comments.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -270,6 +284,11 @@ private void getNames(Node node, ImmutableSet.Builder<QualifiedName> tableBuilde
         switch (node) {
             case AddColumn s -> tableBuilder.add(qualifyName(s.getName()));
             case Analyze s -> tableBuilder.add(qualifyName(s.getTableName()));
+            case Call call -> {
+                procedure = Optional.of(call);
+                procedureArguments = call.getArguments();
```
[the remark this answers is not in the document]
willmostly: I guess there may be some use cases for routing procedures to specific clusters as well, such as `CALL iceberg.system.register_table`. I'll add the catalog and schema to the builders and retain the procedure information in the serialized object.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -95,6 +100,8 @@ public class TrinoQueryProperties
     private Set<String> catalogs = ImmutableSet.of();
     private Set<String> schemas = ImmutableSet.of();
     private Set<String> catalogSchemas = ImmutableSet.of();
+    private Optional<Call> procedure = Optional.empty();
+    private List<CallArgument> procedureArguments = ImmutableList.of();
```
ebyhr, line 104: Please consider changing these fields to `Optional<String>` and `List<SerializableCallArgument>`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -113,6 +120,8 @@ public TrinoQueryProperties(
             @JsonProperty("catalogs") Set<String> catalogs,
             @JsonProperty("schemas") Set<String> schemas,
             @JsonProperty("catalogSchemas") Set<String> catalogSchemas,
+            @JsonProperty("procedure") Optional<String> procedure,
+            @JsonProperty("procedureArguments") SerializableCallArgument[] procedureArguments,
```
ebyhr, line 124: Why do we use array instead of list?

ebyhr (commented):
Please squash the 3rd commit into the 1st commit.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
willmostly: This was done because of the use of `requireNonNullElse`, which infers type from its second argument. `[Immutable]List.of()` produces a `List<Object>`, where I needed a `List<SerializableCallArgument>`. It seemed convenient to create an array of the required type inline. 

I'll revert to a list and declare a default of the required type explicitly

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
```diff
@@ -62,6 +64,26 @@ void setup()
...
+
+        RequestBody requestBody = RequestBody.create("", null);
+        Request request =
+                new Request.Builder()
+                        .url("http://localhost:" + routerPort + "/v1/statement")
+                        .addHeader("X-Trino-User", "test")
+                        .addHeader("Host", "test.host.com")
+                        .post(requestBody)
+                        .build();
+        Response response = httpClient.newCall(request).execute();
```
ebyhr, line 82: Fix warning about the response without try-with-resource.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
```diff
@@ -62,6 +64,26 @@ void setup()
...
+        String responseBody = response.body().string();
```
ebyhr, line 83: Fix warning about NPE.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
```diff
@@ -62,6 +64,26 @@ void setup()
...
+        assertThat(responseBody.toLowerCase(ENGLISH)).matches(".*statement is empty");
```
ebyhr, line 84: `toLowerCase(ENGLISH)` seems redundant.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
```diff
@@ -62,6 +64,26 @@ void setup()
                 "trino1", "http://localhost:" + backendPort, "externalUrl", true, "adhoc", routerPort);
     }
 
+    @ParameterizedTest
+    @MethodSource("protocols")
+    void testEmptyBody(Protocol protocol)
```
ebyhr, line 69: How does this test relate to `kill_query` procedure change?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
willmostly: `ProxyUtils::extractQueryIdFromRequest` checked that `argument instanceof StringLiteral`, so when changing `procedureArguments` to `List<SerializableCallArgument>` I added a field to record the original type. Is there a better way to go about this?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -107,42 +113,55 @@ public TrinoQueryProperties(
...
-        this.tables = Arrays.stream(requireNonNullElse(tables, new String[] {})).map(this::parseIdentifierStringToQualifiedName).collect(Collectors.toSet());
+        List<String> defaultTables = ImmutableList.of();
+        this.tables = requireNonNullElse(tables, defaultTables).stream().map(this::parseIdentifierStringToQualifiedName).collect(Collectors.toSet());
         this.defaultCatalog = requireNonNullElse(defaultCatalog, Optional.empty());
         this.defaultSchema = requireNonNullElse(defaultSchema, Optional.empty());
         this.catalogs = requireNonNullElse(catalogs, ImmutableSet.of());
         this.schemas = requireNonNullElse(schemas, ImmutableSet.of());
         this.catalogSchemas = requireNonNullElse(catalogSchemas, ImmutableSet.of());
         this.isNewQuerySubmission = isNewQuerySubmission;
+        this.procedureArguments = requireNonNullElse(procedureArguments, ImmutableList.of());
```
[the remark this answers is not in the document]
willmostly: `procedureArguments` will be null if TrinoQueryProperties is constructed from a json that does not include this field. If the field is missing, the constructor will be called with a `null` argument, except for the `Optional` fields such as `procedure`. I will add a test case.

gateway-ha/src/test/java/io/trino/gateway/ha/TestGatewayHaSingleBackend.java
willmostly: I must have pulled something in while rebasing without noticing, this was unintentional. I will revert

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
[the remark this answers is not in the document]
ebyhr: I see you added `testJsonMissingFields` test. Who creates `{}` json?

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableCallArgument.java
```diff
@@ -0,0 +1,78 @@
...
+            return Objects.equals(this.value, that.value) && Objects.equals(this.name, that.name);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return value.hashCode() + name.hashCode();
+    }
```
ebyhr, line 77: Please generate these methods with IntelliJ instead of writing by yourself.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableExpression.java
```diff
@@ -0,0 +1,84 @@
...
+    {
+        if (this == obj) {
+            return true;
+        }
+        else if (obj != null && this.getClass() == obj.getClass()) {
+            SerializableExpression that = (SerializableExpression) obj;
+            return Objects.equals(this.value, that.value) && Objects.equals(this.originalClass, that.originalClass);
+        }
+        return false;
+    }
```
ebyhr, line 83: Please generate these methods with IntelliJ instead of writing by yourself.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableCallArgument.java
```diff
@@ -0,0 +1,78 @@
...
+// CallArgument is final, so this class just replicates it. Location is not preserved, since it would require
+// additional complexity and is not meaningful in this context
+public class SerializableCallArgument
+{
+    private final Optional<String> name;
+    private final SerializableExpression value;
+
+    @JsonCreator
+    public SerializableCallArgument(
+            @JsonProperty("name") Optional<String> name,
```
ebyhr, line 35: Add `requireNonNull`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableCallArgument.java
```diff
@@ -0,0 +1,78 @@
...
+            @JsonProperty("value") SerializableExpression value)
+    {
+        this.name = name;
+        this.value = requireNonNull(value, "value is null");
+    }
+
+    public SerializableCallArgument(CallArgument callArgument)
+    {
```
ebyhr, line 43: Add `requireNonNull`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableCallArgument.java
```diff
@@ -0,0 +1,78 @@
...
+        this.name = callArgument.getName().map(Identifier::getValue);
+        this.value = new SerializableExpression(callArgument.getValue());
```
ebyhr, line 45:
```suggestion
        this(callArgument.getName().map(Identifier::getValue), new SerializableExpression(callArgument.getValue()));
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableCallArgument.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @JsonProperty
+    public SerializableExpression getValue()
+    {
+        return value;
+    }
+
+    @JsonProperty
+    public Optional<String> getName()
```
ebyhr, line 55: Replace the order of `getValue` & `getName` for consistency with fields.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableExpression.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class SerializableExpression
+{
+    private final String value;
+    private final Class originalClass;
+
+    @JsonCreator
+    public SerializableExpression(@JsonProperty("value") String value, @JsonProperty("originalClass") String originalClass)
+    {
+        this.value = value;
```
ebyhr, line 33: Add `requireNonNull` to value & originalClass.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableExpression.java
```diff
@@ -0,0 +1,84 @@
...
+import io.trino.sql.tree.StringLiteral;
+
+import java.util.Objects;
+
+import static java.lang.Class.forName;
+
+public class SerializableExpression
+{
+    private final String value;
+    private final Class originalClass;
```
ebyhr, line 28: Fix warning of raw usage of parameterized class.

gateway-ha/src/main/java/io/trino/gateway/ha/router/SerializableExpression.java
```diff
@@ -0,0 +1,84 @@
...
+            throw new RuntimeException("Original class name provided by SerializableExpression not found", e);
+        }
+    }
+
+    public SerializableExpression(Expression expression)
+    {
+        originalClass = expression.getClass();
+        if (expression instanceof StringLiteral) {
+            // special handling for this common case so that quotes do not need to be stripped
+            value = ((StringLiteral) expression).getValue();
```
ebyhr, line 47: Use enhanced instanceof.

mosabua:
Can we get this in now @ebyhr and @willmostly so we can ideally cut 12?

ebyhr:
@mosabua No, the review hasn't yet completed.

ebyhr (changes requested):
Now I strongly disagree with the current approach because we have to fight with serialization issues of procedure arguments. The maintainability seems so bad. 

Please consider different approach. For instance, a) adding `Optional<String> queryId` to `TrinoQueryProperties` and setting the value during parse. b) extracting a helper method which retrieves query id from `kill_query` procedure without relying on `TrinoQueryProperties`. 

I remember you mentioned that `TrinoQueryProperties` class is used because some other information might be helpful in the future. This is assumption, not required for now. I don't think we want to handle in this PR.

willmostly:
> Now I strongly disagree with the current approach because we have to fight with serialization issues of procedure arguments. The maintainability seems so bad.

I agree, having access to the procedure and arguments for routing was not worth it. I took the approach of extracting the queryId from `kill_query` during parsing. It isn't serialized because if a query ID is found, routing rules will not be invoked.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -568,15 +604,15 @@ public RequestParsingException(String message)
         }
     }
 
-    public static class QualifiedNameJsonSerializer
+    public static class SetQualifiedNameJsonSerializer
```
ebyhr, line 607: Why did you rename this class?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestTrinoQueryProperties.java
```diff
@@ -56,4 +57,38 @@ void testJsonCreator()
         assertThat(deserializedTrinoQueryProperties.isQueryParsingSuccessful()).isEqualTo(trinoQueryProperties.isQueryParsingSuccessful());
         assertThat(deserializedTrinoQueryProperties.getErrorMessage()).isEqualTo(trinoQueryProperties.getErrorMessage());
     }
+
+    @Test
+    void testJsonCreatorWithEmptyProperties()
```
ebyhr, line 62: I guess code readers won't be able to understand what's empty because some fields are present. Please rename this test method.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -128,21 +136,29 @@ public TrinoQueryProperties(
...
+        maxBodySize = -1;
     }
 
     public TrinoQueryProperties(HttpServletRequest request, RequestAnalyzerConfig config)
     {
-        isClientsUseV2Format = config.isClientsUseV2Format();
+        this(request, config.isClientsUseV2Format(), config.getMaxBodySize());
+    }
+
+    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
```
ebyhr, line 147: Please deprecate another constructor `HttpServletRequest request, RequestAnalyzerConfig config`.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -102,27 +104,20 @@ public static String extractQueryIdIfPresent(String path, String queryParams, Li
                         || tokens[1].equals("scheduled")
                         || tokens[1].equals("executing")
                         || tokens[1].equals("partialCancel")) {
-                    queryId = tokens[2];
+                    return Optional.of(tokens[2]);
                 }
                 else {
-                    queryId = tokens[1];
+                    return Optional.of(tokens[1]);
```
ebyhr, line 110: Remove redundant `else`: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#additional-ide-configuration

ebyhr (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
oneonestar: https://github.com/trinodb/trino-gateway/pull/541

PR #426, opened 2024-07-30
Allow non destructive reads of request body

willmostly:
## Description

`HttpServeletRequest` only supports destructive reads of a request's body. This change ensures different components can access the request body without interfering with one another.

The body of a request is consumed by the `ServeletHandler`s that supply the `body` argument to `RouteToBackendResource.postHandler`. It is necessary to re-associate the body with the request so it can be used by downstream query id extraction logic and routing rules.

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouteToBackendResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouteToBackendResource.java
@@ -60,11 +60,12 @@ public class RouteToBackendResource
             @Context HttpServletRequest servletRequest,
             @Suspended AsyncResponse asyncResponse)
     {
-        if (servletRequest.getRequestURI().startsWith(V1_STATEMENT_PATH)) {
+        MultiReadHttpServletRequest multiReadHttpServletRequest = new MultiReadHttpServletRequest(servletRequest, body);
+        if (multiReadHttpServletRequest.getRequestURI().startsWith(V1_STATEMENT_PATH)) {
             proxyHandlerStats.recordRequest();
         }
-        String remoteUri = routingTargetHandler.getRoutingDestination(servletRequest);
-        proxyRequestHandler.postRequest(body, servletRequest, asyncResponse, URI.create(remoteUri));
+        String remoteUri = routingTargetHandler.getRoutingDestination(multiReadHttpServletRequest);
+        proxyRequestHandler.postRequest(body, multiReadHttpServletRequest, asyncResponse, URI.create(remoteUri));
     }
 
     @GET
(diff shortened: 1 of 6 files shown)


gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouteToBackendResource.java
```diff
@@ -60,11 +60,12 @@ public void postHandler(
...
     {
-        if (servletRequest.getRequestURI().startsWith(V1_STATEMENT_PATH)) {
+        MultiReadHttpServletRequest multiReadHttpServletRequest = new MultiReadHttpServletRequest(servletRequest, body);
+        if (multiReadHttpServletRequest.getRequestURI().startsWith(V1_STATEMENT_PATH)) {
             proxyHandlerStats.recordRequest();
         }
-        String remoteUri = routingTargetHandler.getRoutingDestination(servletRequest);
-        proxyRequestHandler.postRequest(body, servletRequest, asyncResponse, URI.create(remoteUri));
+        String remoteUri = routingTargetHandler.getRoutingDestination(multiReadHttpServletRequest);
+        proxyRequestHandler.postRequest(body, multiReadHttpServletRequest, asyncResponse, URI.create(remoteUri));
```
Chaho12, line 68: Why would this apply only to POST and not to others GET, DELETE?
Would it not be safer to use MultiReadHttpServletRequest instead of HttpServletRequest in all cases?
including RoutingGroup codes?

willmostly: the GETs and DELETEs won't have bodies, so `getReader` and `getInputStream` will be null either way

Chaho12: True, hmm.. how about using one type just for unifcation for all cases? even if it doesn't have body?

Chaho12 (approved):
LGTM. Btw would this change not be applied to TrinoQueryProperties too?

gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouteToBackendResource.java
willmostly: Yeah, I think that is what @oneonestar was suggesting with the plan to replace `HttpServeletRequest` with a dedicated class like `TrinoRequestContext`. This is the way to go, but would broaden the scope of this PR by quite a bit. So for now I think we can get by with `HttpServletRequest`, using the MultiRead implementation when there is a body

willmostly:
> LGTM. Btw would this change not be applied to TrinoQueryProperties too?

`TrinoQueryProperties` is downstream of `postHandler`, so it will get this change

PR #432, opened 2024-08-06
Allow OAuth-SSO also when form is available

mheers:
## Description
This PR allows the single-sign-on via OAuth even when formAuth is also available.

To have the API working with username/password we need to enable formAuth. When a user also configures OAuth the user could never login using this SSO because the form was prioritized.

## Release notes

```markdown
* Allow OAuth-SSO also when form is available
```


oneonestar:
Maybe we should use [AuthenticationConfiguration.defaultType](https://github.com/trinodb/trino-gateway/blob/e6b45cf45564ee185273ee5f229da0f1492d6af0/gateway-ha/src/main/java/io/trino/gateway/ha/config/AuthenticationConfiguration.java#L18) to decide which `loginType` to be used.

oneonestar:
@cla-bot check

oneonestar:
> Maybe we should use [AuthenticationConfiguration.defaultType](https://github.com/trinodb/trino-gateway/blob/e6b45cf45564ee185273ee5f229da0f1492d6af0/gateway-ha/src/main/java/io/trino/gateway/ha/config/AuthenticationConfiguration.java#L18) to decide which loginType to be used.

@mheers Any thoughts?

mosabua:
I think it might be best to look how this is handled in Trino and then just do the same thing.

willmostly:
@mheers are you still interested in working on this?

mosabua:
Feel free to reopen when you want to continue working on this.

PR #433, opened 2024-08-09
Show and modify routing rules from the UI

prakhar10:
## Description
1. Adding a way to display routing rules and also modify them from the UI
2. Minor change to show a user icon with a gear if user is ADMIN or a regular user icon if its a regular USER on the top right corner of the UI
3. Add config option to disable UI pages

Screenshot 2024-08-08 at 8 22 16 PM

Screenshot 2024-08-08 at 8 22 35 PM

Screenshot 2024-08-08 at 8 23 06 PM

## Release notes

```markdown
* 
```

### DIFF
--- a/docs/gateway-api.md
+++ b/docs/gateway-api.md
@@ -91,3 +91,37 @@ Will return a JSON array of active Trino cluster backends:
 curl -X POST http://localhost:8080/gateway/backend/activate/trino-2
 ```
 
+## Update Routing Rules
+
+This API can be used to programmatically update the Routing Rules.
+Rule will be updated based on the rule name.
+
+For this feature to work with multiple replicas of the Trino Gateway, you will need to provide a shared storage that supports file locking for the routing rules file. If multiple replicas are used with local storage, then rules will get out of sync when updated.
+
+```shell
+curl -X POST http://localhost:8080/webapp/updateRoutingRules \
+ -H 'Content-Type: application/json' \
+ -d '{  "name": "trino-rule",
+        "description": "updated rule description",
+        "priority": 0,
+        "actions": ["updated action"],
+        "condition": "updated condition"
+    }'
+```
+### Disable Routing Rules UI
+
+You can set the `disablePages` config to disable pages on the UI.
+
+The following pages are available:
+- `dashboard`
+- `cluster`
+- `resource-group`
+- `selector`
+- `history`
+- `routing-rules`
+
+```yaml
+uiConfiguration:
+  disablePages:
+    - 'routing-rules'
+```
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
@@ -46,6 +46,8 @@ public class HaGatewayConfiguration
 
     private RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
 
+    private UIConfiguration uiConfiguration = new UIConfiguration();
+
     // List of Modules with FQCN (Fully Qualified Class Name)
     private List<String> modules;
 
@@ -212,6 +214,16 @@ public class HaGatewayConfiguration
     public void setRequestAnalyzerConfig(RequestAnalyzerConfig requestAnalyzerConfig)
     {
         this.requestAnalyzerConfig = requestAnalyzerConfig;
+    }
+
+    public UIConfiguration getUiConfiguration()
+    {
+        return uiConfiguration;
+    }
+
+    public void setUiConfiguration(UIConfiguration uiConfiguration)
+    {
+        this.uiConfiguration = uiConfiguration;
     }
 
     public List<String> getModules()
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/UIConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/UIConfiguration.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.config;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+public class UIConfiguration
+{
+    private List<String> disablePages;
+
+    @JsonProperty
+    public List<String> getDisablePages()
+    {
+        return disablePages;
+    }
+
+    public void setDisablePages(List<String> disablePages)
+    {
+        this.disablePages = disablePages;
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.domain;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElse;
+
+/**
+ * RoutingRules
+ *
+ * @param name name of the routing rule
+ * @param description description of the routing rule. Defaults to an empty string if not provided, indicating the user intends it to be blank.
+ * @param priority priority of the routing rule. Higher number represents higher priority. If two rules have same priority then order of execution is not guaranteed.
+ * @param actions actions of the routing rule
+ * @param condition condition of the routing rule
+ */
+public record RoutingRule(
+        String name,
+        String description,
+        Integer priority,
+        List<String> actions,
+        String condition)
+{
+    public RoutingRule {
+        requireNonNull(name, "name is null");
+        description = requireNonNullElse(description, "");
+        priority = requireNonNullElse(priority, 0);
+        actions = ImmutableList.copyOf(actions);
+        requireNonNull(condition, "condition is null");
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
@@ -16,8 +16,11 @@ package io.trino.gateway.ha.resource;
 import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import io.trino.gateway.ha.config.UIConfiguration;
 import io.trino.gateway.ha.domain.Result;
+import io.trino.gateway.ha.domain.RoutingRule;
 import io.trino.gateway.ha.domain.TableData;
 import io.trino.gateway.ha.domain.request.GlobalPropertyRequest;
 import io.trino.gateway.ha.domain.request.QueryDistributionRequest;
@@ -34,8 +37,10 @@ import io.trino.gateway.ha.router.GatewayBackendManager;
 import io.trino.gateway.ha.router.HaGatewayManager;
 import io.trino.gateway.ha.router.QueryHistoryManager;
 import io.trino.gateway.ha.router.ResourceGroupsManager;
+import io.trino.gateway.ha.router.RoutingRulesManager;
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
@@ -44,6 +49,7 @@ import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
 
+import java.io.IOException;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
@@ -67,18 +73,25 @@ public class GatewayWebAppResource
     private final QueryHistoryManager queryHistoryManager;
     private final BackendStateManager backendStateManager;
     private final ResourceGroupsManager resourceGroupsManager;
+    // TODO Avoid putting mutable objects in fields
+    private final UIConfiguration uiConfiguration;
+    private final RoutingRulesManager routingRulesManager;
 
     @Inject
     public GatewayWebAppResource(
             GatewayBackendManager gatewayBackendManager,
             QueryHistoryManager queryHistoryManager,
             BackendStateManager backendStateManager,
-            ResourceGroupsManager resourceGroupsManager)
+            ResourceGroupsManager resourceGroupsManager,
+            RoutingRulesManager routingRulesManager,
+            HaGatewayConfiguration configuration)
     {
         this.gatewayBackendManager = requireNonNull(gatewayBackendManager, "gatewayBackendManager is null");
         this.queryHistoryManager = requireNonNull(queryHistoryManager, "queryHistoryManager is null");
         this.backendStateManager = requireNonNull(backendStateManager, "backendStateManager is null");
         this.resourceGroupsManager = requireNonNull(resourceGroupsManager, "resourceGroupsManager is null");
+        this.uiConfiguration = configuration.getUiConfiguration();
+        this.routingRulesManager = requireNonNull(routingRulesManager, "routingRulesManager is null");
     }
 
     @POST
@@ -424,5 +437,37 @@ public class GatewayWebAppResource
     {
         List<ResourceGroupsManager.ExactSelectorsDetail> selectorsDetailList = resourceGroupsManager.readExactMatchSourceSelector();
         return Response.ok(Result.ok(selectorsDetailList)).build();
+    }
+
+    @GET
+    @RolesAllowed("ADMIN")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/getRoutingRules")
+    public Response getRoutingRules()
+            throws IOException
+    {
+        List<RoutingRule> routingRulesList = routingRulesManager.getRoutingRules();
+        return Response.ok(Result.ok(routingRulesList)).build();
+    }
+
+    @POST
+    @RolesAllowed("ADMIN")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/updateRoutingRules")
+    public Response updateRoutingRules(RoutingRule routingRule)
+            throws IOException
+    {
+        List<RoutingRule> routingRulesList = routingRulesManager.updateRoutingRule(routingRule);
+        return Response.ok(Result.ok(routingRulesList)).build();
+    }
+
+    @GET
+    @RolesAllowed("ADMIN")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/getUIConfiguration")
+    public Response getUIConfiguration()
+    {
+        return Response.ok(Result.ok(uiConfiguration)).build();
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLParser;
+import com.google.common.collect.ImmutableList;
+import com.google.inject.Inject;
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.domain.RoutingRule;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.nio.file.StandardOpenOption.READ;
+import static java.nio.file.StandardOpenOption.WRITE;
+
+public class RoutingRulesManager
+{
+    private final String rulesConfigPath;
+
+    @Inject
+    public RoutingRulesManager(HaGatewayConfiguration configuration)
+    {
+        this.rulesConfigPath = configuration.getRoutingRules().getRulesConfigPath();
+    }
+
+    public List<RoutingRule> getRoutingRules()
+    {
+        YAMLFactory yamlFactory = new YAMLFactory();
+        ObjectMapper yamlReader = new ObjectMapper(yamlFactory);
+        ImmutableList.Builder<RoutingRule> routingRulesBuilder = ImmutableList.builder();
+        try {
+            String content = Files.readString(Path.of(rulesConfigPath), UTF_8);
+            YAMLParser parser = yamlFactory.createParser(content);
+            while (parser.nextToken() != null) {
+                RoutingRule routingRule = yamlReader.readValue(parser, RoutingRule.class);
+                routingRulesBuilder.add(routingRule);
+            }
+            return routingRulesBuilder.build();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException("Failed to read or parse routing rules configuration from path : " + rulesConfigPath, e);
+        }
+    }
+
+    public synchronized List<RoutingRule> updateRoutingRule(RoutingRule routingRule)
+    {
+        ImmutableList.Builder<RoutingRule> updatedRoutingRulesBuilder = ImmutableList.builder();
+        List<RoutingRule> currentRoutingRulesList = getRoutingRules();
+        Path path = Path.of(rulesConfigPath);
+        try (FileChannel fileChannel = FileChannel.open(path, WRITE, READ);
+                FileLock lock = fileChannel.lock()) {
+            ObjectMapper yamlWriter = new ObjectMapper(new YAMLFactory());
+            StringBuilder yamlContent = new StringBuilder();
+            for (RoutingRule rule : currentRoutingRulesList) {
+                if (rule.name().equals(routingRule.name())) {
+                    yamlContent.append(yamlWriter.writeValueAsString(routingRule));
+                    updatedRoutingRulesBuilder.add(routingRule);
+                }
+                else {
+                    yamlContent.append(yamlWriter.writeValueAsString(rule));
+                    updatedRoutingRulesBuilder.add(rule);
+                }
+            }
+            Files.writeString(path, yamlContent.toString(), UTF_8);
+            lock.release();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException("Failed to parse or update routing rules configuration form path : " + rulesConfigPath, e);
+        }
+        return updatedRoutingRulesBuilder.build();
+    }
+}
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.config.RoutingRulesConfiguration;
+import io.trino.gateway.ha.domain.RoutingRule;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.NoSuchFileException;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+final class TestRoutingRulesManager
+{
+    @Test
+    void testGetRoutingRules()
+            throws IOException
+    {
+        HaGatewayConfiguration configuration = new HaGatewayConfiguration();
+        RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_atomic.yml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        configuration.setRoutingRules(routingRulesConfiguration);
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager(configuration);
+
+        List<RoutingRule> result = routingRulesManager.getRoutingRules();
+
+        assertThat(result).hasSize(2);
+        assertThat(result.getFirst()).isEqualTo(
+                new RoutingRule(
+                        "airflow",
+                        "if query from airflow, route to etl group",
+                        null,
+                        List.of("result.put(FileBasedRoutingGroupSelector.RESULTS_ROUTING_GROUP_KEY, \"etl\")"),
+                        "request.getHeader(\"X-Trino-Source\") == \"airflow\" && (request.getHeader(\"X-Trino-Client-Tags\") == null || request.getHeader(\"X-Trino-Client-Tags\").isEmpty())"));
+        assertThat(result.get(1)).isEqualTo(
+                new RoutingRule(
+                        "airflow special",
+                        "if query from airflow with special label, route to etl-special group",
+                        null,
+                        List.of("result.put(FileBasedRoutingGroupSelector.RESULTS_ROUTING_GROUP_KEY, \"etl-special\")"),
+                        "request.getHeader(\"X-Trino-Source\") == \"airflow\" && request.getHeader(\"X-Trino-Client-Tags\") contains \"label=special\""));
+    }
+
+    @Test
+    void testRoutingRulesNoSuchFileException()
+    {
+        HaGatewayConfiguration configuration = new HaGatewayConfiguration();
+        RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_test.yaml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        configuration.setRoutingRules(routingRulesConfiguration);
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager(configuration);
+
+        assertThatThrownBy(routingRulesManager::getRoutingRules).hasRootCauseInstanceOf(NoSuchFileException.class);
+    }
+
+    @Test
+    void testUpdateRoutingRulesFile()
+            throws IOException
+    {
+        HaGatewayConfiguration configuration = new HaGatewayConfiguration();
+        RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_update.yml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        configuration.setRoutingRules(routingRulesConfiguration);
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager(configuration);
+
+        RoutingRule routingRules = new RoutingRule("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"adhoc\")"), "request.getHeader(\"X-Trino-Source\") == \"JDBC\"");
+
+        List<RoutingRule> updatedRoutingRules = routingRulesManager.updateRoutingRule(routingRules);
+        assertThat(updatedRoutingRules.getFirst().actions().getFirst()).isEqualTo("result.put(\"routingGroup\", \"adhoc\")");
+        assertThat(updatedRoutingRules.getFirst().condition()).isEqualTo("request.getHeader(\"X-Trino-Source\") == \"JDBC\"");
+
+        RoutingRule originalRoutingRules = new RoutingRule("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"etl\")"), "request.getHeader(\"X-Trino-Source\") == \"airflow\"");
+        List<RoutingRule> updateRoutingRules = routingRulesManager.updateRoutingRule(originalRoutingRules);
+
+        assertThat(updateRoutingRules).hasSize(2);
+        assertThat(updateRoutingRules.getFirst().actions().getFirst()).isEqualTo("result.put(\"routingGroup\", \"etl\")");
+        assertThat(updateRoutingRules.getFirst().condition()).isEqualTo("request.getHeader(\"X-Trino-Source\") == \"airflow\"");
+    }
+
+    @Test
+    void testUpdateRoutingRulesNoSuchFileException()
+    {
+        HaGatewayConfiguration configuration = new HaGatewayConfiguration();
+        RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_updated.yaml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        configuration.setRoutingRules(routingRulesConfiguration);
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager(configuration);
+        RoutingRule routingRules = new RoutingRule("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"adhoc\")"), "request.getHeader(\"X-Trino-Source\") == \"JDBC\"");
+
+        assertThatThrownBy(() -> routingRulesManager.updateRoutingRule(routingRules)).hasRootCauseInstanceOf(NoSuchFileException.class);
+    }
+
+    @Test
+    void testConcurrentUpdateRoutingRule()
+            throws IOException
+    {
+        HaGatewayConfiguration configuration = new HaGatewayConfiguration();
+        RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_concurrent.yml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        configuration.setRoutingRules(routingRulesConfiguration);
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager(configuration);
+
+        RoutingRule routingRule1 = new RoutingRule("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"etl\")"), "request.getHeader(\"X-Trino-Source\") == \"airflow\"");
+        RoutingRule routingRule2 = new RoutingRule("airflow", "if query from airflow, route to adhoc group", 0, List.of("result.put(\"routingGroup\", \"adhoc\")"), "request.getHeader(\"X-Trino-Source\") == \"datagrip\"");
+
+        ExecutorService executorService = Executors.newFixedThreadPool(2);
+
+        executorService.submit(() ->
+        {
+            try {
+                routingRulesManager.updateRoutingRule(routingRule1);
+            }
+            catch (UncheckedIOException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        executorService.submit(() ->
+        {
+            try {
+                routingRulesManager.updateRoutingRule(routingRule2);
+            }
+            catch (UncheckedIOException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        executorService.shutdown();
+        List<RoutingRule> updatedRoutingRules = routingRulesManager.getRoutingRules();
+        assertThat(updatedRoutingRules).hasSize(1);
+        assertThat(updatedRoutingRules.getFirst().condition()).isEqualTo("request.getHeader(\"X-Trino-Source\") == \"datagrip\"");
+        assertThat(updatedRoutingRules.getFirst().actions().getFirst()).isEqualTo("result.put(\"routingGroup\", \"adhoc\")");
+        assertThat(updatedRoutingRules.getFirst().description()).isEqualTo("if query from airflow, route to adhoc group");
+    }
+}
--- a/webapp/src/components/routing-rules.tsx
+++ b/webapp/src/components/routing-rules.tsx
@@ -1,0 +1,175 @@
+import {useEffect, useState} from "react";
+import {routingRulesApi, updateRoutingRulesApi} from "../api/webapp/routing-rules.ts";
+import {RoutingRulesData} from "../types/routing-rules";
+import {Button, Card, Form, Toast} from "@douyinfe/semi-ui";
+import {FormApi} from "@douyinfe/semi-ui/lib/es/form";
+import {Role, useAccessStore} from "../store";
+
+export function RoutingRules() {
+    const [rules, setRules] = useState<RoutingRulesData[]>([]);
+    const [editingStates, setEditingStates] = useState<boolean[]>([]);
+    const [formApis, setFormApis] = useState<(FormApi<any> | null)[]>([]);
+    const access = useAccessStore();
+
+    useEffect(() => {
+        fetchRoutingRules();
+    }, []);
+
+    const fetchRoutingRules = () => {
+        routingRulesApi()
+                .then(data => {
+                    setRules(data);
+                    setEditingStates(new Array(data.length).fill(false));
+                    setFormApis(new Array(data.length).fill(null));
+                }).catch(() => {
+            Toast.error("Failed to fetch routing rules");
+        });
+    };
+
+    const handleEdit = (index: number) => {
+        setEditingStates(prev => {
+            const newStates = [...prev];
+            newStates[index] = true;
+            return newStates;
+        });
+    };
+
+    const handleSave = async (index: number) => {
+        const formApi = formApis[index];
+        if (formApi) {
+            try {
+                const values = formApi.getValues();
+                const actionsArray = Array.isArray(values.actions)
+                        ? values.actions.map((action: string) => action.trim())
+                        : [values.actions.trim()];
+
+                const updatedRule: RoutingRulesData = {
+                    ...rules[index],
+                    ...values,
+                    actions: actionsArray
+                };
+
+                await updateRoutingRulesApi(updatedRule);
+
+                setEditingStates(prev => {
+                    const newStates = [...prev];
+                    newStates[index] = false;
+                    return newStates;
+                });
+
+                setRules(prev => {
+                    const newRules = [...prev];
+                    newRules[index] = updatedRule;
+                    return newRules;
+                });
+
+                Toast.success("Routing rule updated successfully");
+            } catch (error) {
+                Toast.error("Failed to update routing rule");
+            }
+        }
+    };
+
+    const setFormApiForIndex = (index: number) => (api: FormApi<any>) => {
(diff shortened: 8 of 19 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +436,67 @@ public Response readExactMatchSourceSelector()
         List<ResourceGroupsManager.ExactSelectorsDetail> selectorsDetailList = resourceGroupsManager.readExactMatchSourceSelector();
         return Response.ok(Result.ok(selectorsDetailList)).build();
     }
+
+    @GET
+    @RolesAllowed("USER")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/getRoutingRules")
```
Chaho12, line 443: @ytwp shouldn't we use kebab-case and not camelCase for apis. just noticed that all apis is camelCase

webapp/src/components/routing-rules.tsx
```diff
@@ -0,0 +1,177 @@
...
+    }, []);
+
+    const fetchRoutingRules = () => {
+        routingRulesApi()
+                .then(data => {
+                    setRules(data);
+                    setEditingStates(new Array(data.length).fill(false));
+                    setFormApis(new Array(data.length).fill(null));
+                }).catch((error) => {
+            console.error("Error fetching routing rules: ", error);
```
Chaho12, line 25: Remove printing to console

webapp/src/components/routing-rules.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+                setRules(prev => {
+                    const newRules = [...prev];
+                    newRules[index] = updatedRule;
+                    return newRules;
+                });
+
+                Toast.success("Routing rule updated successfully");
+            } catch (error) {
+                console.error("Error updating routing rule: ", error);
```
Chaho12, line 69: here too

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +436,67 @@ public Response readExactMatchSourceSelector()
...
+    public Response readRoutingRules()
```
Chaho12, line 444: Why does function have different name with API path?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
ytwp: The previous API was camel case, so we will continue to use this style.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: I saw some other APIs having similar convention, so i kept it like that. If you think its better to rename the API to `getRoutingRules` then i'll change that.

Chaho12: Let's have it same as path

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +436,67 @@ public Response readExactMatchSourceSelector()
...
+            return Response.ok(Result.ok(routingRulesList)).build();
+        }
+        catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @POST
+    @RolesAllowed("ADMIN")
+    @Consumes(MediaType.APPLICATION_JSON)
```
willmostly, line 466: We should document how to use this API for clients other than the web ui

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +436,67 @@ public Response readExactMatchSourceSelector()
...
+                RoutingRules routingRule = yamlReader.readValue(parser, RoutingRules.class);
+                routingRulesList.add(routingRule);
+            }
+
+            for (int i = 0; i < routingRulesList.size(); i++) {
+                if (routingRulesList.get(i).name().equals(routingRules.name())) {
+                    routingRulesList.set(i, routingRules);
+                    break;
+                }
+            }
```
willmostly, line 488: as the name of this method suggests, this updates an existing rule. It does not appear that this change adds support for deleting an existing rule or adding a new one. I guess this may be due to the UI changes required, but would it be difficult to add this functionality?

willmostly (changes requested):
This looks like a nice improvement, thank you! I reviewed the Java half, and will try to find someone for the typescript. 

My biggest concern here is not with the code, but with allowing users to supply input that will be written to disk on the server, then executed by the `jeasy` rules engine. We should take a close look at `jeasy` and `mvel` to understand what security vulnerabilities we might be introducing. At minimum we should log all of the updates so that they can be audited. @wendigo and @ebyhr wdyt?

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * RoutingRules
+ *
+ * @param name name of the routing rule
+ * @param description description of the routing rule
+ * @param priority priority of the routing rule
+ * @param actions actions of the routing rule
+ * @param condition condition of the routing rule
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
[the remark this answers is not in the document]
prakhar10: Makes sense, i'll fix that!

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +436,67 @@ public Response readExactMatchSourceSelector()
...
+        try {
+            String rulesConfigPath = configuration.getRoutingRules().getRulesConfigPath();
+            content = new String(Files.readAllBytes(Paths.get(rulesConfigPath)));
+            ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
+            YAMLParser parser = new YAMLFactory().createParser(content);
+            List<RoutingRules> routingRulesList = new ArrayList<>();
+            while (parser.nextToken() != null) {
+                RoutingRules routingRules = yamlReader.readValue(parser, RoutingRules.class);
+                routingRulesList.add(routingRules);
+            }
```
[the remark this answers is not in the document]
prakhar10: Sure, let me try.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: I don't think it would be difficult. I thought of initially going with only updates and in a separate PR i will put Add and Delete functionality. Will that be ok? Or should I add it in this PR itself?

prakhar10:
> This looks like a nice improvement, thank you! I reviewed the Java half, and will try to find someone for the typescript. 
> 
> 
> 
> My biggest concern here is not with the code, but with allowing users to supply input that will be written to disk on the server, then executed by the `jeasy` rules engine. We should take a close look at `jeasy` and `mvel` to understand what security vulnerabilities we might be introducing. At minimum we should log all of the updates so that they can be audited. @wendigo and @ebyhr wdyt?

This functionality will be accessible only for users with ADMIN privileges. I also have a functionality implemented in our internal repo where we are auditing changes being made to the Clusters from the UI. I can raise a PR for that as well and will add functionality to audit routing rules as well. I was thinking of getting this reviewed first and then from comments and suggestions I will raise the Audit logs PR.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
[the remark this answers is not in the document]
prakhar10: I tried this but it does not seem to work. Looks like if we use this typereference code it wants the yaml file in a certain format to correctly deserialize it into an object, something like - 

`- name: "airflow"
     description: "if query from airflow, route to etl group"
     priority: 0
     condition: 'request.getHeader("X-Trino-Source") == "airflow"'
     actions:
         - 'result.put("routingGroup", "etl")' `
         
But `jeasy` rules engine is not able to read this format when we try to fetch the rules. So it doesn't seem to work.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: Should I add it here - https://github.com/trinodb/trino-gateway/blob/main/docs/gateway-api.md ?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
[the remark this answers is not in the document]
willmostly: My mistake, I forgot `jeasy` concatenates multiple documents instead of having a list of rules. I think 

```
YAMLFactory yamlFactory = new YAMLFactory();
ObjectMapper yamlReader = new ObjectMapper(yamlFactory);

YAMLParser yamlParser = yamlFactory.createParser(rulesConfigPath)
List<ObjectNode> docs = yamlReader
      .readValues(yamlParser, new TypeReference<RoutingRulesRecord>() {})
      .readAll();
```
should work ([ref](https://stackoverflow.com/questions/25222327/deserialize-pojos-from-multiple-yaml-documents-in-a-single-file-in-jackson)).

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
willmostly: yes, that looks perfect

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +436,67 @@ public Response readExactMatchSourceSelector()
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @POST
+    @RolesAllowed("ADMIN")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/updateRoutingRules")
+    public Response updateRoutingRules(RoutingRules routingRules)
```
willmostly, line 469: The following sequence would result in user 1's update being lost:
* user 1 calls `updateRoutingRules` to update rule 1
* user 2 calls `updateRoutingRules` to update rule 2, which reads the rules file BEFORE user 1's update finishes writing
* user 1's call to `updateRoutingRules` completes and writes new rules out
* user 2's call to `updateRoutingRules` completes and writes new rules out

I think this issue could be avoided by adding `synchronized` to the method and making the method static or marking the class `@Singleton`

willmostly (changes requested):

willmostly (changes requested):
this is looking good, we're almost there. @mosabua and I discussed the need to support this for clustered deployments. We should also provide a way to disable this feature in case someone doesn't want to bother setting up a shared filesystem

prakhar10:
> this is looking good, we're almost there. @mosabua and I discussed the need to support this for clustered deployments. We should also provide a way to disable this feature in case someone doesn't want to bother setting up a shared filesystem

So this means provide an option or a config to show/disable the Routing Rules option in the left panel of the UI right?

mosabua:
> > this is looking good, we're almost there. @mosabua and I discussed the need to support this for clustered deployments. We should also provide a way to disable this feature in case someone doesn't want to bother setting up a shared filesystem
> 
> So this means provide an option or a config to show/disable the Routing Rules option in the left panel of the UI right?

I think for now we can just say in the docs that you must provide shared storage for that feature to work and in terms of disabling it could be just a config option option .. I think over time we will need various config options for the UI anyway

prakhar10:
@mosabua @willmostly  can you please review?

docs/gateway-api.md
```diff
@@ -91,3 +91,16 @@ Will return a JSON array of active Trino cluster backends:
 curl -X POST http://localhost:8080/gateway/backend/activate/trino-2
 ```
 
+## Update Routing Rules
```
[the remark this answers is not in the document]
xkrogen: > This emphasizes that we should really move the rules to the DB, but that will be future work.

Just as a heads up, we (LinkedIn) have a DB-based routing rule implementation internally. We're looking at finally adopting trino-gateway soon (we are still on a presto-gateway fork) and contributing this functionality will likely be part of that process.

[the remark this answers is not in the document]
mosabua: @willmostly will want to work on that with you

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,36 @@
...
+ * @param priority priority of the routing rule
+ * @param actions actions of the routing rule
+ * @param condition condition of the routing rule
+ */
+public record RoutingRules(
+        @JsonProperty("name") String name,
+        @JsonProperty("description") String description,
+        @JsonProperty("priority") Integer priority,
+        @JsonProperty("actions") List<String> actions,
+        @JsonProperty("condition") String condition)
```
ebyhr, line 34: Can we remove `@JsonProperty` annotation?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -67,18 +80,21 @@ public class GatewayWebAppResource
     private final QueryHistoryManager queryHistoryManager;
     private final BackendStateManager backendStateManager;
     private final ResourceGroupsManager resourceGroupsManager;
+    private final HaGatewayConfiguration configuration;
```
ebyhr, line 83: HaGatewayConfiguration is a mutable object. Please avoid putting it in fields.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +439,71 @@ public Response readExactMatchSourceSelector()
...
+    @RolesAllowed("USER")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/getRoutingRules")
+    public Response getRoutingRules()
+    {
+        try {
+            String rulesConfigPath = configuration.getRoutingRules().getRulesConfigPath();
+            YAMLFactory yamlFactory = new YAMLFactory();
+            ObjectMapper yamlReader = new ObjectMapper(yamlFactory);
+            YAMLParser yamlParser = yamlFactory.createParser(new String(Files.readAllBytes(Paths.get(rulesConfigPath))));
```
ebyhr, line 453: Please consider moving `ObjectMapper` to fields and initialize it in the constructor.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +439,71 @@ public Response readExactMatchSourceSelector()
...
+            List<RoutingRules> routingRulesList = yamlReader
+                    .readValues(yamlParser, new TypeReference<RoutingRules>() {})
+                    .readAll();
+            return Response.ok(Result.ok(routingRulesList)).build();
+        }
+        catch (IOException e) {
+            throw new RuntimeException(e);
```
ebyhr, line 460: Why do we rethrow as RuntimeException? Why not throw IOException directly?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -423,4 +439,71 @@ public Response readExactMatchSourceSelector()
         List<ResourceGroupsManager.ExactSelectorsDetail> selectorsDetailList = resourceGroupsManager.readExactMatchSourceSelector();
         return Response.ok(Result.ok(selectorsDetailList)).build();
     }
+
+    @GET
```
ebyhr, line 443: Can we extract a service class and add some tests?

prakhar10: Can we move these to 'RoutingManager' class instead of creating a new class?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: My bad, i will fix it.

mosabua:
I assume that @prakhar10 and @willmostly are collaborating on this currently .. ping me if review or anything else is needed.

prakhar10:
I need to make changes according to @ebyhr 's comments. So i will get that done by this weekend.

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param priority priority of the routing rule
+ * @param actions actions of the routing rule
+ * @param condition condition of the routing rule
+ */
+public record RoutingRules(
+        String name,
+        String description,
+        Integer priority,
+        List<String> actions,
+        String condition)
```
ebyhr, line 32: All fields are nullable? Please add `@Nullable` to nullable fields and a compact constructor which verifies non-null and copy `actions` with `ImmutableList.copyOf` method.

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -44,6 +44,18 @@ public class HaGatewayConfiguration
...
+
+    public UIConfiguration getUiConfiguration()
+    {
+        return uiConfiguration;
+    }
+
+    public void setUiConfiguration(UIConfiguration uiConfiguration)
+    {
+        this.uiConfiguration = uiConfiguration;
+    }
```
ebyhr, line 57: Move these getter and setter under `setRequestAnalyzerConfig` method.

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,34 @@
...
+package io.trino.gateway.ha.domain;
+
+import java.util.List;
+
+/**
+ * RoutingRules
+ *
+ * @param name name of the routing rule
+ * @param description description of the routing rule
+ * @param priority priority of the routing rule
```
ebyhr, line 23: Could you document which (small or large) number wins and what happens if we specify the same priority?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+
+            ObjectMapper yamlWriter = new ObjectMapper(new YAMLFactory());
+            StringBuilder yamlContent = new StringBuilder();
+            for (RoutingRules rule : routingRulesList) {
+                yamlContent.append(yamlWriter.writeValueAsString(rule));
+            }
+            Files.write(Paths.get(rulesConfigPath), yamlContent.toString().getBytes(UTF_8));
+        }
+        catch (IOException e) {
+            throw new IOException(e);
```
ebyhr, line 80: Please rethrow an appropriate exception.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+            YAMLParser parser = new YAMLFactory().createParser(content);
+            List<RoutingRules> routingRulesList = new ArrayList<>();
+            while (parser.nextToken() != null) {
+                RoutingRules routingRules = yamlReader.readValue(parser, RoutingRules.class);
+                routingRulesList.add(routingRules);
+            }
+            return routingRulesList;
+        }
+        catch (IOException e) {
+            throw new IOException(e);
```
ebyhr, line 48: Please rethrow an appropriate exception.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class RoutingRulesManager
+{
+    public List<RoutingRules> getRoutingRules(RoutingRulesConfiguration configuration, ObjectMapper yamlReader)
+            throws IOException
+    {
+        String content = null;
```
ebyhr, line 35: `= null` is redundant.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class RoutingRulesManager
+{
+    public List<RoutingRules> getRoutingRules(RoutingRulesConfiguration configuration, ObjectMapper yamlReader)
```
ebyhr, line 32: This method can be static.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+                routingRulesList.add(routingRules);
+            }
+            return routingRulesList;
+        }
+        catch (IOException e) {
+            throw new IOException(e);
+        }
+    }
+
+    public List<RoutingRules> updateRoutingRules(RoutingRules routingRules, RoutingRulesConfiguration configuration, ObjectMapper yamlReader)
```
ebyhr, line 52: This method can be static.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
ebyhr, line 35: No need to split into declaration and initialization.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class RoutingRulesManager
+{
+    public List<RoutingRules> getRoutingRules(RoutingRulesConfiguration configuration, ObjectMapper yamlReader)
+            throws IOException
+    {
+        String content = null;
+        try {
+            String rulesConfigPath = configuration.getRulesConfigPath();
+            content = new String(Files.readAllBytes(Paths.get(rulesConfigPath)), UTF_8);
```
ebyhr, line 38: Why not use `Files.readString` method?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+            YAMLParser parser = new YAMLFactory().createParser(content);
+            List<RoutingRules> routingRulesList = new ArrayList<>();
```
ebyhr, line 40: Use `ImmutableList.Builder`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+    {
+        String rulesConfigPath = configuration.getRulesConfigPath();
+        List<RoutingRules> routingRulesList = new ArrayList<>();
+        try {
+            String content = new String(Files.readAllBytes(Paths.get(rulesConfigPath)), UTF_8);
+            YAMLParser parser = new YAMLFactory().createParser(content);
+            while (parser.nextToken() != null) {
+                RoutingRules routingRule = yamlReader.readValue(parser, RoutingRules.class);
+                routingRulesList.add(routingRule);
+            }
```
ebyhr, line 63: Can we reuse `getRoutingRules` method?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,84 @@
...
+                    break;
+                }
+            }
+
+            ObjectMapper yamlWriter = new ObjectMapper(new YAMLFactory());
+            StringBuilder yamlContent = new StringBuilder();
+            for (RoutingRules rule : routingRulesList) {
+                yamlContent.append(yamlWriter.writeValueAsString(rule));
+            }
+            Files.write(Paths.get(rulesConfigPath), yamlContent.toString().getBytes(UTF_8));
```
ebyhr, line 77: Why not use `Files.writeString` method?

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
prakhar10: Sorry I don't think I understand why should we move this into setRequestAnalyzerConfig? I believe that serves a different purpose and this class I added so that we can control things on the UI like enabling/disabling pages.

ebyhr: I mean moving this method in this class.

prakhar10:
@ebyhr can you review now please?

prakhar10:
@mosabua can you please review?

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,44 @@
...
+public record RoutingRules(
+        String name,
+        @Nullable String description,
+        @Nullable Integer priority,
+        List<String> actions,
+        String condition)
+{
+    public RoutingRules {
+        requireNonNull(name, "name must not be null");
+        requireNonNull(condition, "condition must not be null");
```
ebyhr, line 41:
```suggestion
        requireNonNull(name, "name is null");
        requireNonNull(condition, "condition is null");
```

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,44 @@
...
+        actions = ImmutableList.copyOf(actions);
```
ebyhr, line 42: Move above `condition` check for consistency with arguments' order of the constructor.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -59,6 +69,7 @@
 import static java.util.Objects.requireNonNullElse;
 
 @Path("/webapp")
+@Singleton
```
ebyhr, line 72: Why do need this change?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public final class RoutingRulesManager
```
ebyhr, line 31: I'm not sure why this class is a static helper class. Could you consider changing to a service class?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,99 @@
...
+{
+    private RoutingRulesConfiguration routingRulesConfiguration;
+
+    private ObjectMapper yamlReader;
+
+    @BeforeEach
+    void setUp()
+    {
+        routingRulesConfiguration = new RoutingRulesConfiguration();
+        yamlReader = new ObjectMapper(new YAMLFactory());
```
ebyhr, line 40: Why do need to initialize `yamlReader` every test? Please change to constant if there is no reason.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.IOException;
+import java.nio.file.NoSuchFileException;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatException;
+
+class TestRoutingRulesManager
+{
+    private RoutingRulesConfiguration routingRulesConfiguration;
```
ebyhr, line 32: Why do need to manage `RoutingRulesConfiguration` in fields? Please inline if there is no reason.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    void testGetRoutingRules()
+            throws IOException
+    {
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_atomic.yml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+
+        List<RoutingRules> result = RoutingRulesManager.getRoutingRules(routingRulesConfiguration, yamlReader);
+
+        assertThat(2).isEqualTo(result.size());
```
ebyhr, line 52: The order of actual and expected is wrong. Same for others.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,99 @@
...
+        assertThat("airflow special").isEqualTo(result.get(1).name());
+    }
+
+    @Test
+    void testRoutingRulesNoSuchFileException()
+    {
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_test.yaml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+
+        assertThatException().isThrownBy(() -> {
```
ebyhr, line 67: Replace with expression lambda.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Test
+    void testUpdateRoutingRulesNoSuchFileException()
+    {
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_updated.yaml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        RoutingRules routingRules = new RoutingRules("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"adhoc\")"), "request.getHeader(\"X-Trino-Source\") == \"JDBC\"");
+
+        assertThatException().isThrownBy(() -> {
```
ebyhr, line 95: Replace with expression lambda.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,99 @@
...
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_update.yml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        RoutingRules routingRules = new RoutingRules("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"adhoc\")"), "request.getHeader(\"X-Trino-Source\") == \"JDBC\"");
+
+        List<RoutingRules> updatedRoutingRules = RoutingRulesManager.updateRoutingRules(routingRules, routingRulesConfiguration, yamlReader);
+        assertThat("result.put(\"routingGroup\", \"adhoc\")").isEqualTo(updatedRoutingRules.get(0).actions().get(0));
+        assertThat("request.getHeader(\"X-Trino-Source\") == \"JDBC\"").isEqualTo(updatedRoutingRules.get(0).condition());
+
+        RoutingRules originalRoutingRules = new RoutingRules("airflow", "if query from airflow, route to etl group", 0, List.of("result.put(\"routingGroup\", \"etl\")"), "request.getHeader(\"X-Trino-Source\") == \"airflow\"");
+        RoutingRulesManager.updateRoutingRules(originalRoutingRules, routingRulesConfiguration, yamlReader);
```
ebyhr, line 85: Could you verify the updated result?

ebyhr (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
prakhar10: makes sense, i'll update it.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
prakhar10: By inline you mean initialize in the test method itself right?

`RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
String rulesConfigPath = "src/test/resources/rules/routing_rules_atomic.yml";
routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);`

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: Sorry not needed, forgot to remove it. I'll remove 👍

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
prakhar10: Screenshot 2024-10-28 at 7 22 30 PM

So I was trying to implement these comments and then I got an error saying I need to make the class final as well to have all static methods. Maybe I misunderstood what you meant?

gateway-ha/src/main/java/io/trino/gateway/ha/config/UIConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+    {
+        this.disablePages = disablePages;
+    }
+
+    @Override
+    public String toString()
+    {
+        return "UIConfiguration{" +
+                "disablePages=" + disablePages +
+                '}';
```
willmostly, line 40: use `String.format`

gateway-ha/src/main/java/io/trino/gateway/ha/config/UIConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+        return disablePages;
+    }
+
+    public void setDisablePages(List<String> disablePages)
+    {
+        this.disablePages = disablePages;
+    }
+
+    @Override
+    public String toString()
```
willmostly, line 36: this is unused, do we need it?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,80 @@
...
+                    break;
+                }
+            }
+            ObjectMapper yamlWriter = new ObjectMapper(new YAMLFactory());
+            StringBuilder yamlContent = new StringBuilder();
+            for (RoutingRules rule : routingRulesList) {
+                yamlContent.append(yamlWriter.writeValueAsString(rule));
+                routingRulesBuilder.add(rule);
+            }
+            Files.writeString(Paths.get(rulesConfigPath), yamlContent.toString(), UTF_8);
```
willmostly, line 73: multiple writers could overwrite one another here. You will need to use filesystem locking since there could be multiple trino gateway instances simultaneously attempting an update.

willmostly (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
prakhar10: Makes sense, i'll check on this

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRules.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * RoutingRules
+ *
+ * @param name name of the routing rule
+ * @param description description of the routing rule
+ * @param priority priority of the routing rule. Higher number represents higher priority. If two rules have same priority then order of execution is not guaranteed.
+ * @param actions actions of the routing rule
+ * @param condition condition of the routing rule
+ */
+public record RoutingRules(
```
willmostly, line 32: this is a single rule
```suggestion
public record RoutingRule(
```

willmostly (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public final class RoutingRulesManager
+{
+    private RoutingRulesManager() {}
+
+    public static List<RoutingRules> getRoutingRules(RoutingRulesConfiguration configuration, ObjectMapper yamlReader)
```
willmostly, line 35: Iirc the Jackson yaml processor is not thread safe. Creating it locally would bypass these issues

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
prakhar10: @ebyhr any inputs?

ebyhr: Please ignore the previous comment and change to a service class.

prakhar10:
@ebyhr  can you take a look now?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,96 @@
...
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager();
+
+        List<RoutingRule> result = routingRulesManager.getRoutingRules(routingRulesConfiguration);
+
+        assertThat(result.size()).isEqualTo(2);
+        assertThat(result.getFirst().name()).isEqualTo("airflow");
+        assertThat(result.getFirst().description()).isEqualTo("if query from airflow, route to etl group");
+        assertThat(result.getFirst().condition()).isEqualTo("request.getHeader(\"X-Trino-Source\") == \"airflow\" && (request.getHeader(\"X-Trino-Client-Tags\") == null || request.getHeader(\"X-Trino-Client-Tags\").isEmpty())");
+        assertThat(result.getFirst().actions().getFirst()).isEqualTo("result.put(\"routingGroup\", \"etl\")");
+        assertThat(result.getFirst().actions().size()).isEqualTo(1);
```
ebyhr, line 45: Why not verify `priority` field? In other words, why not use `isEqualTo`? 
```java
        assertThat(result.getFirst())
                .isEqualTo(new RoutingRule(...));
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    public List<RoutingRule> updateRoutingRules(RoutingRule routingRules, RoutingRulesConfiguration configuration)
+            throws IOException
+    {
+        ImmutableList.Builder<RoutingRule> routingRulesBuilder = ImmutableList.builder();
+        String rulesConfigPath = configuration.getRulesConfigPath();
+        try {
+            List<RoutingRule> routingRulesList = getRoutingRules(configuration);
```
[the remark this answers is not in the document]
prakhar10: i'll update the names

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,79 @@
...
+{
+    public List<RoutingRule> getRoutingRules(RoutingRulesConfiguration configuration)
+            throws IOException
+    {
+        ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
+        String rulesConfigPath = configuration.getRulesConfigPath();
+        try {
+            String content = Files.readString(Paths.get(rulesConfigPath), UTF_8);
+            YAMLParser parser = new YAMLFactory().createParser(content);
+            List<RoutingRule> routingRulesList = new ArrayList<>();
```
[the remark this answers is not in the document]
prakhar10: I'm using the result of getRoutingRules() method here - https://github.com/trinodb/trino-gateway/blob/1fb283d402788a406fdc1c9d7d7a8f0717950222/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java#L59

If I make it immutable then we wont be able to update it. Or should I convert it back into mutable list?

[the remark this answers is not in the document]
ebyhr: Please change the caller logic not to depend on mutable list.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,96 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class RoutingRulesManager
+{
+    private final RoutingRulesConfiguration routingRulesConfiguration;
```
ebyhr, line 35: Please don't put mutable objects in fields.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,98 @@
...
+        RoutingRulesConfiguration routingRulesConfiguration = new RoutingRulesConfiguration();
+        String rulesConfigPath = "src/test/resources/rules/routing_rules_atomic.yml";
+        routingRulesConfiguration.setRulesConfigPath(rulesConfigPath);
+        configuration.setRoutingRules(routingRulesConfiguration);
+        RoutingRulesManager routingRulesManager = new RoutingRulesManager(configuration);
+
+        List<RoutingRule> result = routingRulesManager.getRoutingRules();
+
+        assertThat(result).hasSize(2);
+        assertThat(result.getFirst()).isEqualTo(new RoutingRule("airflow", "if query from airflow, route to etl group", null, Collections.singletonList("result.put(\"routingGroup\", \"etl\")"), "request.getHeader(\"X-Trino-Source\") == \"airflow\" && (request.getHeader(\"X-Trino-Client-Tags\") == null || request.getHeader(\"X-Trino-Client-Tags\").isEmpty())"));
```
ebyhr, line 45: Put arguments on separate lines. This line is too long.

ebyhr: Replace Collections.singletonList with List.of.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
prakhar10: Screenshot 2024-11-21 at 12 58 29 PM

I was following this comment @ebyhr

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
willmostly: We do, since multiple gateway instances may try to access the same file

prakhar10:
@willmostly @ebyhr can you review now please?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -424,4 +436,36 @@ public Response readExactMatchSourceSelector()
         List<ResourceGroupsManager.ExactSelectorsDetail> selectorsDetailList = resourceGroupsManager.readExactMatchSourceSelector();
         return Response.ok(Result.ok(selectorsDetailList)).build();
     }
+
+    @GET
+    @RolesAllowed("USER")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/getRoutingRules")
+    public Response getRoutingRules()
```
willmostly, line 444: Should these methods return the bare type instead of wrapping them in a `Response`, similar to the Trino UI? E.g. https://github.com/trinodb/trino/blob/bb80ff8be6073bc970501825e2e7f304b3b9d643/core/trino-main/src/main/java/io/trino/server/ui/ClusterResource.java#L51

docs/gateway-api.md
```diff
@@ -91,3 +91,36 @@ Will return a JSON array of active Trino cluster backends:
...
+
+This API can be used to programmatically update the Routing Rules.
+Rule will be updated based on the rule name.
+
+For this feature to work with multiple replicas of the Trino Gateway, you will need to provide a shared storage for the routing rules file. If multiple replicas are used with local storage, then rules will get out of sync when updated.
```
willmostly, line 99:
```suggestion
For this feature to work with multiple replicas of the Trino Gateway, you will need to provide a shared storage that supports file locking for the routing rules file. If multiple replicas are used with local storage, then rules will get out of sync when updated.
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,100 @@
...
+        catch (IOException e) {
+            throw new IOException("Failed to read or parse routing rules configuration from path : " + rulesConfigPath, e);
+        }
+    }
+
+    public synchronized List<RoutingRule> updateRoutingRule(RoutingRule routingRule)
+            throws IOException
+    {
+        ImmutableList.Builder<RoutingRule> updatedRoutingRulesBuilder = ImmutableList.builder();
+        List<RoutingRule> currentRoutingRulesList = new ArrayList<>();
```
willmostly, line 68:
```suggestion
        List<RoutingRule> currentRoutingRulesList = getRoutingRules();
```
also remove the the logic below to read rules

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,100 @@
...
+                    break;
+                }
+            }
+            ObjectMapper yamlWriter = new ObjectMapper(new YAMLFactory());
+            StringBuilder yamlContent = new StringBuilder();
+            for (RoutingRule rule : currentRoutingRulesList) {
+                yamlContent.append(yamlWriter.writeValueAsString(rule));
+                updatedRoutingRulesBuilder.add(rule);
+            }
+            try (FileChannel fileChannel = FileChannel.open(Paths.get(rulesConfigPath), StandardOpenOption.WRITE, StandardOpenOption.READ);
```
willmostly, line 89: Lock the file before it is read. Otherwise you allow a sequence like
1. Gateway 1 reads rules
2. Gateway 2 reads rules
3. Gateway 1 locks rules, writes rules, releases lock.
4. Gateway 2 locks rules, writes rules, releases lock. 

Which results in the changes by Gateway 1 being lost.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
```diff
@@ -0,0 +1,149 @@
...
+import java.io.IOException;
+import java.nio.file.NoSuchFileException;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+final class TestRoutingRulesManager
```
willmostly, line 30: Please add tests that use the REST API

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
```diff
@@ -0,0 +1,44 @@
...
+ * @param name name of the routing rule
+ * @param description description of the routing rule
+ * @param priority priority of the routing rule. Higher number represents higher priority. If two rules have same priority then order of execution is not guaranteed.
+ * @param actions actions of the routing rule
+ * @param condition condition of the routing rule
+ */
+public record RoutingRule(
+        String name,
+        @Nullable String description,
+        @Nullable Integer priority,
```
willmostly, line 35: priority should not be nullable. It may be defaulted to `0`.

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
```diff
@@ -0,0 +1,44 @@
...
+        List<String> actions,
+        String condition)
+{
+    public RoutingRule {
+        requireNonNull(name, "name is null");
+        actions = ImmutableList.copyOf(actions);
+        requireNonNull(condition, "condition is null");
+    }
```
willmostly, line 43: Set description to `""` if it is null

willmostly (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
prakhar10: @willmostly  sorry I didn't understand this, can you provide an examples or please help me understand?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: I tried to keep it similar to other methods and also here - https://github.com/trinodb/trino-gateway/blob/64dcbd578c6f4b1fac574f2ae09bd85ba2dedaaf/webapp/src/api/base.ts#L6

we are checking the response status and then extracting the json body, so to not make any changes to this code I kept it as is.

docs/gateway-api.md
```diff
@@ -91,3 +91,36 @@ Will return a JSON array of active Trino cluster backends:
...
+
+```shell
+curl -X POST http://localhost:8080/webapp/updateRoutingRules \
```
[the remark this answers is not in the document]
prakhar10: Let me check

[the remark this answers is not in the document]
prakhar10: It does not work without it, I will add it to the documentation.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingRulesManager.java
prakhar10: Discussed with Will, working on this.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
willmostly: Ok in that case

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
prakhar10: Works great, thank you Will!

willmostly (approved):
Thanks for all the work @prakhar10 !

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new IOException("Failed to read or parse routing rules configuration from path : " + rulesConfigPath, e);
+        }
+    }
+
+    public synchronized List<RoutingRule> updateRoutingRule(RoutingRule routingRule)
+            throws IOException
+    {
+        ImmutableList.Builder<RoutingRule> updatedRoutingRulesBuilder = ImmutableList.builder();
+        List<RoutingRule> currentRoutingRulesList = getRoutingRules();
+        try (FileChannel fileChannel = FileChannel.open(Path.of(rulesConfigPath), StandardOpenOption.WRITE, StandardOpenOption.READ);
```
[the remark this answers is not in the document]
prakhar10: @ebyhr the `rulesConfigPath` is `final` so I don't think it can be modified.

or do you mean what if other threads modify the contents of the file on that path?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingRulesManager.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public List<RoutingRule> getRoutingRules()
+            throws IOException
+    {
+        ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
+        ImmutableList.Builder<RoutingRule> routingRulesBuilder = ImmutableList.builder();
+        try {
+            String content = Files.readString(Path.of(rulesConfigPath), UTF_8);
+            YAMLParser parser = new YAMLFactory().createParser(content);
```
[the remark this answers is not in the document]
prakhar10: yes right, I'll fix this.

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
willmostly: IMO the semantics of an empty value are more appropriate than `null`. The routing rule comes from a web form or an API call. If a description is not supplied, then I would say that the user intends it to be blank, not that the description is unknown to them.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -425,4 +438,36 @@ public Response readExactMatchSourceSelector()
         List<ResourceGroupsManager.ExactSelectorsDetail> selectorsDetailList = resourceGroupsManager.readExactMatchSourceSelector();
         return Response.ok(Result.ok(selectorsDetailList)).build();
     }
+
+    @GET
+    @RolesAllowed("USER")
```
willmostly, line 443: lets make this `ADMIN` to avoid users attempting to see what the rules are so that they can craft a query that goes to a particular cluster

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -425,4 +438,36 @@ public Response readExactMatchSourceSelector()
...
+    @Path("/updateRoutingRules")
+    public Response updateRoutingRules(RoutingRule routingRule)
+            throws IOException
+    {
+        List<RoutingRule> routingRulesList = routingRulesManager.updateRoutingRule(routingRule);
+        return Response.ok(Result.ok(routingRulesList)).build();
+    }
+
+    @GET
+    @RolesAllowed("USER")
```
willmostly, line 466: lets make this `ADMIN` as well. I don't see a reason for a normal user to need this, so PoLP would suggest `ADMIN`

willmostly (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: done

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: done

willmostly:
@mosabua I recommend we just use the PR title as the release note:
```
Show and modify routing rules from the UI
```

Any cleanup can be done in a follow up

Issue #437, opened 2024-08-14
`appVersion` in Chart.yaml should be updated with each release


oneonestar:
`appVersion` in Chart.yaml should be updated with each release

> Shouldn't the Chart.yaml be updated with each release?

_Originally posted by @wccropper in https://github.com/trinodb/trino-gateway/issues/435#issuecomment-2289091742_

Ref:
https://github.com/trinodb/trino-gateway/blob/112c500321805de962f6d9874ec5145c70dc8631/helm/trino-gateway/Chart.yaml#L24
https://github.com/trinodb/trino-gateway/blob/112c500321805de962f6d9874ec5145c70dc8631/helm/trino-gateway/values.yaml#L6-L7

PR #438, opened 2024-08-16
Update Helm chart version to 10

oneonestar:
## Description
Update Helm chart version to 10.
Related to #437.

`version` should follow semantic versioning (X.Y.Z).
I'm thinking about:
```
X follows trino-gateway release
Y for helm chart breaking updates
Z for minor bug fixes
```

wdyt @willmostly 

## Release notes

### DIFF
--- a/helm/trino-gateway/Chart.yaml
+++ b/helm/trino-gateway/Chart.yaml
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "8"
+version: "10.0.0"
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "8"
+appVersion: "10"


helm/trino-gateway/Chart.yaml
```diff
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "8"
+version: "10.0.0"
```
mosabua, line 18: In our prior syncs we decided to keep Trino Gateway and helm chart versions in sync for now. So at this stage it should be just "10".

mosabua: But we can also discuss in the dev sync on Wednesday

mosabua (changes requested):

mosabua:
Also note that 10 is a perfectly fine semantic version number as well

oneonestar:
> Also note that 10 is a perfectly fine semantic version number as well

I don't think so.

https://helm.sh/docs/topics/charts/#charts-and-versioning
> Every chart must have a version number. A version must follow the [SemVer 2](https://semver.org/spec/v2.0.0.html) standard.

https://semver.org/
```
<version core> ::= <major> "." <minor> "." <patch>
```

mosabua (approved):
Lets roll with this.

Issue #370, opened 2024-05-29
Username in the filter box get reloaded


vishalya:
Go to the query history page, change the user name to filter by a different name, go to dashboard and comeback to the history page. It will load the origin name.

PR #439, opened 2024-08-19
Fix username reload in history page

prakhar10:
## Description
Fixes https://github.com/trinodb/trino-gateway/issues/370

## Release notes

```markdown
* 
```

### DIFF
--- a/webapp/src/components/history.tsx
+++ b/webapp/src/components/history.tsx
@@ -19,8 +19,8 @@ export function History() {
   const [page, setPage] = useState(1);
   const [size] = useState(15);
   const [form, setForm] = useState(() => {
-    const savedForm = sessionStorage.getItem('historyForm');
-    return savedForm ? JSON.parse(savedForm) : { user: access.userName };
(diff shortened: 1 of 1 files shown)


webapp/src/components/history.tsx
```diff
@@ -41,6 +42,10 @@ export function History() {
     list(1);
   }, [form]);
 
+  useEffect(() => {
+    sessionStorage.setItem('historyForm', JSON.stringify(form));
```
vishalya, line 46: why not just save the username? does the form save all the history data too!

prakhar10: No, its just saving the search fields `RoutedTo`, `User` and `QueryId`. I thught what if in the future we need to save the other fields as well just like username when someone goes away from the page and come back.

Currently other fields are not using the `initValue` attribute so we are not setting any default values, which makes sense for `RoutedTo` and `QueryId`.

I think for now like you suggested we can just save the username.

mosabua (approved):
Looks good. Wdyt @colebow and @vishalya

PR #448, opened 2024-08-27
Add commonLabels to helm chart

alaturqua:
## Description
This PR adds commonLabels to all resources. For some kubernetes instances, it is required to have custom labels.

## Release notes

```markdown
Add commonLabels to helm chart 
```

### DIFF
--- a/helm/trino-gateway/values.yaml
+++ b/helm/trino-gateway/values.yaml
@@ -54,6 +54,7 @@ config:
 
 service:
   type: ClusterIP
+  port: 8080
 
 ingress:
   enabled: false
@@ -109,6 +110,10 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+# -- Labels that get applied to every resource's metadata
+commonLabels: {}
+  # example: "some label"
 
 podAnnotations: {}
 
--- a/helm/trino-gateway/templates/_helpers.tpl
+++ b/helm/trino-gateway/templates/_helpers.tpl
@@ -40,6 +40,9 @@ helm.sh/chart: {{ include "trino-gateway.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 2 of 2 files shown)


helm/trino-gateway/values.yaml
```diff
@@ -110,6 +111,11 @@ tolerations: []
 
 affinity: {}
 
+commonLabels: {}
+  # io.example/cost-center: "000000"
+  # io.example/team: "trino"
```
mosabua, line 116: Do the labels have to have a namespace or can it also be

example: "some value"

willmostly (approved):
This looks good. @alaturqua could you squash your commits? Please capitalize your commit message as per the guidance here: https://cbea.ms/git-commit/

helm/trino-gateway/values.yaml
mosabua: If yes .. we should add such a simpler example. Also

helm/trino-gateway/values.yaml
```diff
@@ -110,6 +111,11 @@ tolerations: []
 
 affinity: {}
 
+commonLabels: {}
```
mosabua, line 114: Add docs like in https://github.com/trinodb/charts/blob/ac81a3cc14b766d3c2b2ed70e85863f67dc36a2a/charts/trino/values.yaml#L799

That way we can produce docs later like in the Trino chart..

mosabua (changes requested):
Looks good to me, just some needs and probably want a look from @willmostly

helm/trino-gateway/values.yaml
alaturqua: It does need to have namespace. That was just the way we were using commonLabels. I can change them and add comment for docs.

helm/trino-gateway/values.yaml
alaturqua: added

alaturqua:
@willmostly 
Squashed my commits and added docs. Please review again.

mosabua (approved):
Looks good now. Thanks @alaturqua

PR #455, opened 2024-08-29
Correct OAuth cookie logout behavior

willmostly:
## Description

The cookie used for maintaining stickiness during an OAuth2 handshake forwards requests to a path in the `deleteCookiePath` list such as `/logout` to a random backend. The requests should be forwarded to the same backend to allow any server side sessions to be cleaned up. 

## Release notes

```markdown
* Ensure OAuth logout requests are forwarded to the Trino cluster that initiated the handshake.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
@@ -177,10 +177,6 @@ public class GatewayCookie
 
     public boolean matchesRoutingPath(String path)
     {
-        if (matchesDeletePath(path)) {
-            return false;
-        }
-
         return unsignedGatewayCookie.getRoutingPaths().stream().anyMatch(path::startsWith);
     }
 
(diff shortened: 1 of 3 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
```diff
@@ -177,10 +177,6 @@ public static GatewayCookie fromCookie(Cookie cookie)
 
     public boolean matchesRoutingPath(String path)
     {
-        if (matchesDeletePath(path)) {
```
mosabua, line 180: Do you know why this was there in the first place then?

mosabua (approved):
Looks good to go apart from maybe understanding why this was not the case before and checking if we need to doc some more config for this.

gateway-ha/src/main/java/io/trino/gateway/ha/router/GatewayCookie.java
willmostly: I believe that this logic was there because the original model of the flow was basically 
* identify the first request in a "session" and create a cookie to track it
* if the client makes another request that initiates a session, delete the old cookie and create a new session

This logic was carried over from that model, which doesn't properly handle logout.

PR #456, opened 2024-08-30
Set X-Forwarded-Host when client using HTTP/2

### LINKED ISSUE #451: Support HTTP/2 requests
When clients send HTTP/2 requests (without `Host` headers), X-Forwarded headers will be set except `X-Forwarded-Host`, which causes making improper nextUris in a coordinator, resulting in connection failures from next client requests.

oneonestar:
## Description
Set X-Forwarded-Host when client using HTTP/2.
Fix #451 

image
OkHttp convert `Host` header into HTTP/2 `:authority` header, thus the test is valid.

Ref:
https://github.com/square/okhttp/blob/28075d78310e895b4673cd5433a69b80d42433c2/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2ExchangeCodec.kt#L139-L143
https://github.com/square/okhttp/blob/28075d78310e895b4673cd5433a69b80d42433c2/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2ExchangeCodec.kt#L172-L175

## Release notes

```markdown
* Fix X-Forwarded-Host not being set when client using HTTP/2
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java
@@ -45,7 +45,6 @@ import java.util.HashMap;
 import java.util.concurrent.ExecutorService;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.common.net.HttpHeaders.HOST;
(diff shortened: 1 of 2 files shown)


wendigo (approved):

mosabua (approved):
Nice. Lets ship this!

PR #457, opened 2024-08-30
Update dependencies to align with Trino

wendigo:


wendigo:
@willmostly does anyone work on 22? :)

PR #463, opened 2024-09-05
Improve mkocs setup info

mosabua:
## Description

Result from testing with Will.

## Release notes


willmostly (approved):
LGTM

PR #469, opened 2024-09-11
Fixed analyze request

### LINKED ISSUE #468: analyzeRequest not explained correctly in documentation
The documentation states that [isAnalyzeRequest](https://trinodb.github.io/trino-gateway/routing-rules/#trinoqueryproperties) should be set to make `trinoQueryProperties` and `trinoRequestUser` available. However, the property that should be set is `analyzeRequest` as given in the test case:
https://github.com/trinodb/trino-gateway/blob/b549b249854d2f466db997ecaf67904539649933/gateway-ha/src/test/resources/test-config-with-routing-template.yml#L27

athultr1997:
## Description
Fixes #468

Objectmapper expects the boolean fields without the `is` prefix. Hence, instead of `isAnalyzeRequest`, `analyzeRequest` should be used for configuration.

## Release notes

### DIFF
--- a/docs/routing-rules.md
+++ b/docs/routing-rules.md
@@ -48,7 +48,7 @@ You can use an external service for processing your routing by setting the
 `rulesType` to `EXTERNAL` and configuring the `rulesExternalConfiguration`.
 
 Trino Gateway then sends all headers as a map in the body of a POST request to the external service.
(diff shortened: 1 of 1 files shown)


ebyhr (approved):

ebyhr:
@athultr1997 Thanks for fixing this. Please avoid past tense in the commit message from next time. https://cbea.ms/git-commit/ is our commit message guideline.

PR #470, opened 2024-09-13
Use blacklistHeaders instead of blackListHeaders

Chaho12:
## Description

This fixes typo at [Guide](https://github.com/trinodb/trino-gateway/blob/c03df851d642757e2db6f7f6081a28cb5a549d7b/docs/routing-rules.md?plain=1#L34) where `blacklistHeaders` has `L` capitalized in code.
 Currently, conf has to be set using `L`, which isn't correct, so changed code to use black`l`istHeaders instead.

## Release notes

```markdown
* Use blacklistHeaders instead of blackListHeaders
```


gateway-ha/src/main/java/io/trino/gateway/ha/config/RulesExternalConfiguration.java
```diff
@@ -18,7 +18,7 @@
 public class RulesExternalConfiguration
 {
     private String urlPath;
-    private List<String> blackListHeaders;
+    private List<String> excludeHeaderList;
```
ebyhr, line 21: How about renaming to `excludeHeaders` (removing List)?

mosabua: good idea

mosabua:
Agreed .. this is a good change @Chaho12

gateway-ha/src/main/java/io/trino/gateway/ha/config/RulesExternalConfiguration.java
Chaho12: yup changed it. 
image
image

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -61,10 +61,10 @@ public class ExternalRoutingGroupSelector
     @VisibleForTesting
     ExternalRoutingGroupSelector(RulesExternalConfiguration rulesExternalConfiguration, RequestAnalyzerConfig requestAnalyzerConfig)
     {
-        Set<String> defaultBlacklistHeaders = ImmutableSet.of("Content-Length");
-        this.blacklistHeaders = ImmutableSet.<String>builder()
-                .addAll(defaultBlacklistHeaders)
-                .addAll(rulesExternalConfiguration.getBlackListHeaders())
+        Set<String> defaultExcludeHeaders = ImmutableSet.of("Content-Length");
```
ebyhr, line 64: This variable seems redundant. We could simply call `.add("Content-Length")`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -142,8 +142,8 @@ private Multimap<String, String> getValidHeaders(HttpServletRequest servletReque
         Multimap<String, String> headers = ArrayListMultimap.create();
         for (String name : list(servletRequest.getHeaderNames())) {
             for (String value : list(servletRequest.getHeaders(name))) {
-                // Add all headers to ListMultimap except those in blacklist
-                if (!blacklistHeaders.contains(name)) {
+                // Add all headers to ListMultimap except those in excludeHeaders
```
ebyhr, line 145: This comment seems unhelpful as it's obvious when reading the code.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
```diff
@@ -84,7 +84,7 @@ static Stream<RulesExternalConfiguration> provideRoutingRuleExternalConfig()
     {
         RulesExternalConfiguration restConfig = new RulesExternalConfiguration();
         restConfig.setUrlPath("http://localhost:8080/api/public/gateway_rules");
-        restConfig.setBlackListHeaders(new ArrayList<>(List.of("Authorization")));
+        restConfig.setExcludeHeaders(new ArrayList<>(List.of("Authorization")));
```
ebyhr, line 87: Why are we calling `new ArrayList` here? Also, why does this test class use `@ParameterizedTest` just for single case?

ebyhr:
> * Use excludeHeaders instead of blackListHeaders

Thanks for writing a release note entry. I would recommend mentioning where the `excludeHeaders` is used, e.g. "Use Replace `blackListHeaders` with `excludeHeaders` in the routing rule configuration"

ebyhr (approved):
Left come comments which can be addressed in a follow-up PR.

mosabua:
please rebase and address the few small comments @Chaho12 .. then we can merge

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
Chaho12: > Why are we calling new ArrayList here? 

True a list is sufficent.

> Also, why does this test class use @ParameterizedTest just for single case?

what do you mean by a single case?

ebyhr: I meant this method returns a single `RulesExternalConfiguration`.

Chaho12: I am not that familiar with Junit 5 so please bear with me.

Well, I wanted to reduce code duplication for configs, as it could be used several times. 
And i thought this way of using parameterized test makes it concise to test.

But now that you mention, since this is not complex and returns as a single object,
i could simply create this object in initialize() function, make it where it is `@BeforeAll`. right?

ebyhr: RulesExternalConfiguration is a mutable object. Please don't put it in the fields. 
We can simply create a helper method and use it in each test.

ebyhr:
> Use excludeHeaderList instead of blackListHeaders #470

Please update the PR titlte.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
ebyhr: I pushed the small change.

PR #474, opened 2024-09-18
Throw exception when no clusters are found for adhoc routing group

andythsu:
## Description

Adhoc routing group is the fallback routing group if no routing group is provided. If there are no clusters available for this group, we should log and throw an error

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
@@ -224,7 +224,7 @@ public class QueryCountBasedRouter
     @Override
     public String provideAdhocBackend(String user)
     {
-        return getBackendForRoutingGroup("adhoc", user).orElseThrow();
+        return getBackendForRoutingGroup("adhoc", user).orElseThrow(() -> new RouterException("did not find any cluster for the adhoc routing group"));
     }
 
     @Override
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
@@ -17,9 +17,8 @@ package io.trino.gateway.ha.router;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
+import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -32,7 +31,7 @@ import java.util.stream.Collectors;
 public class QueryCountBasedRouter
         extends StochasticRoutingManager
 {
-    private static final Logger log = LoggerFactory.getLogger(QueryCountBasedRouter.class);
+    private static final Logger log = Logger.get(QueryCountBasedRouter.class);
     @GuardedBy("this")
     private List<LocalStats> clusterStats;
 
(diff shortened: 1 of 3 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -224,7 +224,11 @@ private synchronized Optional<String> getBackendForRoutingGroup(String routingGr
     @Override
     public String provideAdhocBackend(String user)
     {
-        return getBackendForRoutingGroup("adhoc", user).orElseThrow();
+        return getBackendForRoutingGroup("adhoc", user).orElseThrow(() -> {
+            RouterException adhocRoutingGroupException = new RouterException("did not find any cluster for the adhoc routing group");
+            log.error(adhocRoutingGroupException.getMessage(), adhocRoutingGroupException);
```
willmostly, line 229: Update the logger to `io.airlift.log.Logger` and
```suggestion
            log.error(adhocRoutingGroupException, adhocRoutingGroupException.getMessage())
```
This is the last class using the slf4j logger, so you can remove that dependency after the change

willmostly (changes requested):
Looks good, will approve after the change to the logger

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -224,7 +224,11 @@ private synchronized Optional<String> getBackendForRoutingGroup(String routingGr
     @Override
     public String provideAdhocBackend(String user)
     {
-        return getBackendForRoutingGroup("adhoc", user).orElseThrow();
+        return getBackendForRoutingGroup("adhoc", user).orElseThrow(() -> {
+            RouterException adhocRoutingGroupException = new RouterException("did not find any cluster for the adhoc routing group");
+            log.error(adhocRoutingGroupException, adhocRoutingGroupException.getMessage());
+            return adhocRoutingGroupException;
```
ebyhr, line 230: Logging an error and throwing an exception is weird. I would recommend removing logging.

ebyhr (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
andythsu: sure, I was logging and throwing at the same time in case other systems rely on the format of the log

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -32,7 +31,8 @@
 public class QueryCountBasedRouter
         extends StochasticRoutingManager
 {
-    private static final Logger log = LoggerFactory.getLogger(QueryCountBasedRouter.class);
+    private static final Logger log = Logger.get(HaGatewayManager.class);
```
ebyhr, line 34: The argument should be `QueryCountBasedRouter.class`

andythsu: oops good catch

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -224,11 +224,7 @@ private synchronized Optional<String> getBackendForRoutingGroup(String routingGr
     @Override
     public String provideAdhocBackend(String user)
     {
-        return getBackendForRoutingGroup("adhoc", user).orElseThrow(() -> {
-            RouterException adhocRoutingGroupException = new RouterException("did not find any cluster for the adhoc routing group");
-            log.error(adhocRoutingGroupException.getMessage(), adhocRoutingGroupException);
-            return adhocRoutingGroupException;
-        });
+        return getBackendForRoutingGroup("adhoc", user).orElseThrow(() -> new RouterException("did not find any cluster for the adhoc routing group"));
```
ebyhr, line 227: The commit boundary is wrong. This change should be included in the 1st commit.

ebyhr (approved):

Issue #478, opened 2024-09-25
Fix IndexOutOfBoundsException from TrinoQueryProperties


JeongDaeKim:
IndexOutOfBoundsException occurs for some statements without catalogs defined explicitly, such as `create schema {schema_name}`, when request analyzer enabled.

```
2024-09-23T16:00:00.456+0900    ERROR   http-worker-101 io.trino.gateway.ha.router.RuleReloadingRoutingGroupSelector    Error opening rules configuration file, using routing group header as default.
java.lang.IndexOutOfBoundsException: index (1) must be less than size (1)
        at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1374)
        at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1356)
        at com.google.common.collect.SingletonImmutableList.get(SingletonImmutableList.java:46)
        at io.trino.gateway.ha.router.TrinoQueryProperties.setCatalogAndSchemaNameFromSchemaQualifiedName(TrinoQueryProperties.java:331)
        at io.trino.gateway.ha.router.TrinoQueryProperties.getNames(TrinoQueryProperties.java:297)
        at io.trino.gateway.ha.router.TrinoQueryProperties.processRequestBody(TrinoQueryProperties.java:163)
        at io.trino.gateway.ha.router.TrinoQueryProperties.<init>(TrinoQueryProperties.java:103)
        at io.trino.gateway.ha.router.RuleReloadingRoutingGroupSelector.findRoutingGroup(RuleReloadingRoutingGroupSelector.java:98)
        at io.trino.gateway.ha.handler.RoutingTargetHandler.getBackendFromRoutingGroup(RoutingTargetHandler.java:84)
        at io.trino.gateway.ha.handler.RoutingTargetHandler.lambda$getRoutingDestination$0(RoutingTargetHandler.java:64)
        at java.base/java.util.Optional.orElseGet(Optional.java:364)
        at io.trino.gateway.ha.handler.RoutingTargetHandler.getRoutingDestination(RoutingTargetHandler.java:64)
        at io.trino.gateway.proxyserver.RouteToBackendResource.postHandler(RouteToBackendResource.java:67)
```

PR #479, opened 2024-09-25
Fix IndexOutOfBoundsException caused by switch-case statements with no breaks

JeongDaeKim:
## Description

Fix https://github.com/trinodb/trino-gateway/issues/478

## Release notes

```markdown
* 
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
@@ -357,16 +357,17 @@ public class TrinoQueryProperties
         else {
             QualifiedName schema = schemaOptional.orElseThrow();
             switch (schema.getParts().size()) {
-                case 1:
+                case 1 -> {
                     schemaBuilder.add(schema.getParts().getFirst());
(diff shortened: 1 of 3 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -357,16 +357,17 @@ private void setCatalogAndSchemaNameFromSchemaQualifiedName(
...
-                case 2:
+                }
+                case 2 -> {
                     schemaBuilder.add(schema.getParts().get(1));
                     catalogBuilder.add(schema.getParts().getFirst());
                     catalogSchemaBuilder.add(format("%s.%s", schema.getParts().getFirst(), schema.getParts().getLast()));
-                default:
-                    log.error("Schema has >2 parts: " + schema);
+                }
+                default -> log.error("Schema has >2 parts: " + schema);
```
ebyhr, line 370: optional: This line should use a logger template instead of string concat. Could you fix it?

ebyhr:
> Fix IOOBE caused by switch-case statements with no breaks.

Please avoid abbreviations (IOOBE) and remove a following dot. 
* https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations
* https://cbea.ms/git-commit/

PR #481, opened 2024-09-25
Upgrade UI

mosabua:
## Description

- Node from v18.9.1 to latest LTS version v20.17.0
- pnpm from 8.6.12 to latest version 9.11.0
- Including all transitive dependencies

## Release notes

(x) No release notes entry needed.

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -489,8 +489,8 @@
                             <goal>install-node-and-pnpm</goal>
                         </goals>
                         <configuration>
-                            <nodeVersion>v18.9.1</nodeVersion>
-                            <pnpmVersion>8.6.12</pnpmVersion>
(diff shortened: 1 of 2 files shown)


ytwp (approved):

mosabua:
Did you do any further testing on this @ytwp  .. cause I did not..

ebyhr (approved):

ebyhr:

Why do we need to mention this change in release notes?

ytwp:
> Did you do any further testing on this @ytwp .. cause I did not..

I have tested it, and Node v20.17.0 works well.

mosabua:
> > (x) Release notes are required, with the following suggested text:
> 
> Why do we need to mention this change in release notes?

I was thinking it might introduce some incompatibility with extensions or so but I think we are all good.

PR #482, opened 2024-09-25
Update MySQL JDBC driver to 9.0.0

mosabua:
## Release notes

```markdown
* tbd .. probably affects what MySQL version we support
```

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -24,7 +24,7 @@
         <dep.jeasy.version>4.1.0</dep.jeasy.version>
         <dep.mockito.version>5.13.0</dep.mockito.version>
         <dep.okhttp3.version>4.12.0</dep.okhttp3.version>
-        <dep.mysqlconnector.version>8.3.0</dep.mysqlconnector.version>
+        <dep.mysqlconnector.version>9.0.0</dep.mysqlconnector.version>
         <dep.trino.version>446</dep.trino.version>
     </properties>
 


gateway-ha/pom.xml
```diff
@@ -24,7 +24,7 @@
         <dep.jeasy.version>4.1.0</dep.jeasy.version>
         <dep.mockito.version>5.13.0</dep.mockito.version>
         <dep.okhttp3.version>4.12.0</dep.okhttp3.version>
-        <dep.mysqlconnector.version>8.3.0</dep.mysqlconnector.version>
+        <dep.mysqlconnector.version>9.0.0</dep.mysqlconnector.version>
```
ebyhr, line 27: Trino user reported 8.4.0 caused performance regression. https://github.com/trinodb/trino/issues/21945
I'm not sure whether this version is safe to use.

mosabua: Hmm.. good question.. I guess we would need to try with 9.0.0 in Trino as well and decide from there.

ebyhr: https://dev.mysql.com/doc/relnotes/mysql/9.0/en/news-9-0-0.html
> This release is no longer available for download. It was removed due to a critical issue that could stop the server from restarting following the creation of a very large number of tables (8001, or more). Please upgrade to MySQL 9.0.1 instead.

I don't think we should use this version.

mosabua:
> @mosabua we should update the image used by the runner, regardless of 8.3 vs 9 for the driver version. It looks like none of the tests use a MySQL image 😬
> 
> https://github.com/trinodb/trino-gateway/blob/c0f11de3d1a58b7ff98dcb4d239fb5f93a671e54/gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java#L53

Sounds good .. I will try to attack this at some stage soon.

PR #486, opened 2024-09-26
Use and require Java 23

oneonestar:
## Description
Use and require Java 23.

## Release notes

```markdown
* Require Java 23 for build and runtime.
```


willmostly:
@oneonestar is this ready to review?

mosabua:
I think this might also need an upgrade of airlift and airbase -- please check with @wendigo

mosabua:
Also .. we probably want a separate PR in parallel that runs with 24-ea similar to how we run it in Trino

oneonestar:
> is this ready to review?

@willmostly Yea.

> I think this might also need an upgrade of airlift and airbase -- please check with @wendigo

Sure #529

Since Java 22 is already EOL, we should roll this out soon.

willmostly (approved):
LGTM

oneonestar:
Let's ship it together with Trino. 🚀
https://github.com/trinodb/trino/pull/23783

PR #488, opened 2024-09-26
Inject RoutingTargetHandler

willmostly:
## Description

Inject `RoutingTargetHandler` instead of initializing it in `HaGatewayProviderModule`

## Release notes


gateway-ha/src/main/java/io/trino/gateway/baseapp/BaseApp.java
```diff
@@ -169,4 +171,9 @@ private static void registerProxyResources(Binder binder)
         httpClientBinder(binder).bindHttpClient("proxy", ForProxy.class);
         httpClientBinder(binder).bindHttpClient("monitor", ForMonitor.class);
     }
+
+    private static void registerRoutingHandler(Binder binder)
```
ebyhr, line 175: Inline this method.

ebyhr (approved):

PR #489, opened 2024-09-27
Update docs to add encoding header to be excluded

Chaho12:
## Description

When decoding JSON response using `io.airlift.http.client#createJsonResponseHandler` at [external routing group selector code](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java#L98), 
JSON response could be compressed in GZIP (or whatever format) depending on the network configuration.

This is not expecected behaviour by Airlift as it read bytes.
- https://github.com/airlift/airlift/blob/master/http-client/src/main/java/io/airlift/http/client/JsonResponseHandler.java#L68-L91

So, I suggest adding `'Accept-Encoding'` to docs so that user can know they might need to exclude it when making API request to server. 

## Release notes

```markdown
* 
```

### DIFF
--- a/docs/routing-rules.md
+++ b/docs/routing-rules.md
@@ -33,11 +33,13 @@ routingRules:
         urlPath: https://router.example.com/gateway-rules # replace with your own API path
         excludeHeaders:
             - 'Authorization'
+            - 'Accept-Encoding'
 ```
 
 * Redirect URLs are not supported
 * Optionally add headers to the `excludeHeaders` list to exclude requests with corresponding header values
   from being sent in the POST request.
+* Check headers to exclude when making API requests, specifics depend on the network configuration.
 
 If there is error parsing the routing rules configuration file, an error is logged,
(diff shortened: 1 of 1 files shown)


docs/routing-rules.md
```diff
@@ -33,11 +33,13 @@ routingRules:
         urlPath: https://router.example.com/gateway-rules # replace with your own API path
         excludeHeaders:
             - 'Authorization'
+            - 'Accept-Encoding'
 ```
 
 * Redirect URLs are not supported
 * Optionally add headers to the `excludeHeaders` list to exclude requests with corresponding header values
   from being sent in the POST request.
+* Check headers that need to be excluded when making API requests, depending on network configuration.
```
mosabua, line 42:
```suggestion
* Check headers to exclude when making API requests, specifics depend on the network configuration.
```

ebyhr:
@mosabua Gentle reminder.

mosabua (approved):
This is good to go.

PR #497, opened 2024-09-30
Add parsing error message field to TrinoQueryProperties

Chaho12:
## Description

Add a field (private Optional<String> ParseErrorMessage) to TrinoQueryProperties class that stores `e.getMessage()` when there is error.
https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java#L213-L224

We have server connected with gateway that runs explain before actually running the query.
But not all queries need to be sent with explain.

For example there are cases where queries with parsing failure (no need to run explain cause it is highly likely to fail anyway due to syntax error)
- we need to tell user on WHY query has failed, but running another sql parser or explain is waste of resource
- since gateway already does query parsing, why not use that failure info.

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
@@ -96,7 +96,7 @@ public class TrinoQueryProperties
     private Set<String> schemas = ImmutableSet.of();
     private Set<String> catalogSchemas = ImmutableSet.of();
     private boolean isNewQuerySubmission;
-    private boolean isQueryParsingSuccessful;
+    private Optional<String> errorMessage = Optional.empty();
 
     public static final String TRINO_CATALOG_HEADER_NAME = "X-Trino-Catalog";
     public static final String TRINO_SCHEMA_HEADER_NAME = "X-Trino-Schema";
@@ -114,7 +114,7 @@ public class TrinoQueryProperties
             @JsonProperty("schemas") Set<String> schemas,
             @JsonProperty("catalogSchemas") Set<String> catalogSchemas,
             @JsonProperty("isNewQuerySubmission") boolean isNewQuerySubmission,
-            @JsonProperty("isQueryParsingSuccessful") boolean isQueryParsingSuccessful)
+            @JsonProperty("errorMessage") Optional<String> errorMessage)
     {
         this.body = requireNonNullElse(body, "");
         this.queryType = requireNonNullElse(queryType, "");
@@ -126,7 +126,7 @@ public class TrinoQueryProperties
         this.schemas = requireNonNullElse(schemas, ImmutableSet.of());
         this.catalogSchemas = requireNonNullElse(catalogSchemas, ImmutableSet.of());
         this.isNewQuerySubmission = isNewQuerySubmission;
-        this.isQueryParsingSuccessful = isQueryParsingSuccessful;
+        this.errorMessage = requireNonNullElse(errorMessage, Optional.empty());
         isClientsUseV2Format = false;
     }
 
@@ -208,19 +208,18 @@ public class TrinoQueryProperties
             catalogSchemaBuilder.addAll(
                     tables.stream().map(qualifiedName -> format("%s.%s", qualifiedName.getParts().getFirst(), qualifiedName.getParts().get(1))).iterator());
             catalogSchemas = catalogSchemaBuilder.build();
-            isQueryParsingSuccessful = true;
         }
         catch (IOException e) {
             log.warn("Error extracting request body for rules processing: %s", e.getMessage());
-            isQueryParsingSuccessful = false;
+            errorMessage = Optional.of(e.getMessage());
         }
         catch (ParsingException e) {
             log.info("Could not parse request body as SQL: %s; Message: %s", body, e.getMessage());
-            isQueryParsingSuccessful = false;
+            errorMessage = Optional.of(e.getMessage());
         }
         catch (RequestParsingException e) {
             log.warn(e, "Error parsing request for rules");
-            isQueryParsingSuccessful = false;
+            errorMessage = Optional.of(e.getMessage());
         }
     }
 
@@ -291,7 +290,7 @@ public class TrinoQueryProperties
                         targetSchema = QualifiedName.of(defaultCatalog.orElseThrow(), s.getTarget().getValue());
                     }
                     else {
-                        isQueryParsingSuccessful = false;
+                        errorMessage = Optional.of("defaultCatalog is not present");
                         return;
                     }
                 }
@@ -379,7 +378,6 @@ public class TrinoQueryProperties
 
     private RequestParsingException unsetDefaultExceptionSupplier()
     {
-        isQueryParsingSuccessful = false;
         return new RequestParsingException("Name not fully qualified");
     }
 
@@ -513,7 +511,13 @@ public class TrinoQueryProperties
     @JsonProperty("isQueryParsingSuccessful")
     public boolean isQueryParsingSuccessful()
     {
-        return isQueryParsingSuccessful;
+        return errorMessage.isEmpty();
+    }
+
+    @JsonProperty
+    public Optional<String> getErrorMessage()
+    {
+        return errorMessage;
     }
 
     public static class AlternateStatementRequestBodyFormat
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestTrinoQueryProperties.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestTrinoQueryProperties.java
@@ -38,7 +38,7 @@ final class TestTrinoQueryProperties
                 ImmutableSet.of("s"),
                 ImmutableSet.of("c.s"),
                 true,
-                true);
+                Optional.empty());
 
         String trinoQueryPropertiesJson = codec.toJson(trinoQueryProperties);
         TrinoQueryProperties deserializedTrinoQueryProperties = codec.fromJson(trinoQueryPropertiesJson);
@@ -54,5 +54,6 @@ final class TestTrinoQueryProperties
         assertThat(deserializedTrinoQueryProperties.getCatalogSchemas()).isEqualTo(trinoQueryProperties.getCatalogSchemas());
         assertThat(deserializedTrinoQueryProperties.isNewQuerySubmission()).isEqualTo(trinoQueryProperties.isNewQuerySubmission());
         assertThat(deserializedTrinoQueryProperties.isQueryParsingSuccessful()).isEqualTo(trinoQueryProperties.isQueryParsingSuccessful());
+        assertThat(deserializedTrinoQueryProperties.getErrorMessage()).isEqualTo(trinoQueryProperties.getErrorMessage());
     }
 }
--- a/docs/routing-rules.md
+++ b/docs/routing-rules.md
@@ -195,9 +195,10 @@ added to the list of tables in all contexts, including  statements such as
 A routing rule can call the following methods on the `trinoQueryProperties`
 object:
 
+* `String errorMessage()`: the error message only if there was any error while
+  creating `trinoQueryProperties` object. 
 * `boolean isNewQuerySubmission()`: is the request a POST to the `v1/statement`
-  query endpoint.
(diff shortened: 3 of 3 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -519,6 +528,12 @@ public boolean isQueryParsingSuccessful()
         return isQueryParsingSuccessful;
     }
 
+    @JsonProperty
+    public Optional<String> getParseErrorMessage()
```
ebyhr, line 532: This method is used only from tests. What's the motivation of this change?

ebyhr (commented):
> * Add parsing error message field to TrinoQueryProperties

Why do we need to mention this change in a release note?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
Chaho12: We have server connected with gateway that runs explain before actually running the query. 
But not all queries need to be sent with explain.

For example there are cases where queries with parsing failure (no need to run explain cause it is highly likely to fail anyway due to syntax error)
  - we need to tell user on WHY query has failed, but running another sql parser or explain is waste of resource
  - since gateway already does query parsing, why not use that failure info.

Chaho12:
> Why do we need to mention this change in a release note?

Thought there might be some people like me who would need the error. 
Not necessary though, I could just update the docs instead.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -114,7 +115,8 @@ public TrinoQueryProperties(
             @JsonProperty("schemas") Set<String> schemas,
             @JsonProperty("catalogSchemas") Set<String> catalogSchemas,
             @JsonProperty("isNewQuerySubmission") boolean isNewQuerySubmission,
-            @JsonProperty("isQueryParsingSuccessful") boolean isQueryParsingSuccessful)
+            @JsonProperty("isQueryParsingSuccessful") boolean isQueryParsingSuccessful,
+            @JsonProperty("parseErrorMessage") Optional<String> parseErrorMessage)
```
ebyhr, line 119: Why do we need to include "parse" in the field name? Should we add another variable if we find different error cause?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -212,14 +215,17 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
         }
         catch (IOException e) {
             log.warn("Error extracting request body for rules processing: %s", e.getMessage());
+            parseErrorMessage = Optional.ofNullable(e.getMessage());
```
ebyhr, line 218: This isn't a parse error because IOException happens before parsing. The variable should be renamed or we shouldn't set the variable here.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -379,8 +386,10 @@ private void setCatalogAndSchemaNameFromSchemaQualifiedName(
 
     private RequestParsingException unsetDefaultExceptionSupplier()
     {
+        String errorString = "Name not fully qualified";
         isQueryParsingSuccessful = false;
-        return new RequestParsingException("Name not fully qualified");
+        parseErrorMessage = Optional.of(errorString);
+        return new RequestParsingException(errorString);
```
ebyhr, line 392: `parseErrorMessage` is set twice here and catch block of `processRequestBody`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -212,14 +215,17 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
         }
         catch (IOException e) {
             log.warn("Error extracting request body for rules processing: %s", e.getMessage());
+            parseErrorMessage = Optional.ofNullable(e.getMessage());
             isQueryParsingSuccessful = false;
         }
         catch (ParsingException e) {
             log.info("Could not parse request body as SQL: %s; Message: %s", body, e.getMessage());
+            parseErrorMessage = Optional.ofNullable(e.getMessage());
```
ebyhr, line 223: Is `e.getMessage()` really nullable?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -212,14 +215,17 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
...
             isQueryParsingSuccessful = false;
         }
         catch (RequestParsingException e) {
             log.warn(e, "Error parsing request for rules");
+            parseErrorMessage = Optional.ofNullable(e.getMessage());
```
ebyhr, line 228: Is `e.getMessage()` really nullable?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -97,6 +97,7 @@ public class TrinoQueryProperties
     private Set<String> catalogSchemas = ImmutableSet.of();
     private boolean isNewQuerySubmission;
     private boolean isQueryParsingSuccessful;
+    private Optional<String> parseErrorMessage;
```
ebyhr, line 100: Why do we need both isQueryParsingSuccessful and parseErrorMessage variables? Is it possible to remove isQueryParsingSuccessful and return the value by checking the size of parseErrorMessage on the fly?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
ebyhr: > since gateway already does query parsing, why not use that failure info.

I don't think relying on Trino Gateway's parser is safe. For instance, the behavior might be different from the backend Trino because of version mismatch.

Chaho12: Sure updated description.

> I don't think relying on Trino Gateway's parser is safe. For instance, the behavior might be different from the backend Trino because of version mismatch.

That is true. Gateway does not guarantee that, but I think we should leave this up to the users to decide on how much they rely on the result. 
Btw in our case, we just use `isQueryParsingSuccessful` to see if we will run explain to fetch IO Plan or not, as it is part of our routing rules. (There are several steps on deciding which routing group to route, but one heavy step of running explain can be skipped). 
This failure info is a supplementary info where we tell users that query parsing has failed so please do check your sql 
- As of now, we use sql parser libraries to see why parsing has failed, which is also not that accurate.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
Chaho12: So if there isn't an errorMessage than it is considered successful? 
Hmm.. I prefer having a separate variable. @willmostly any thoughts?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
Chaho12: Making it a more general field as `errorMessage` and storing any sort of errors here sounds good.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
Chaho12: Unless we have a logical bug, not really. 
Since parseErrorMessage is type `Optional<String>`, i couldn't directly save it as e.getMessage().
Any better ideas?

ebyhr: Why not use `Optional.of()` method?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
ebyhr: When does isQueryParsingSuccessful become true when parseErrorMessage is present?
When does isQueryParsingSuccessful become false when parseErrorMessage is empty? 

Having two variables is troublesome as it's easy to miss updating another one. Why do you prefer having a separate variable?

Chaho12: Nowadays am used to using Golang as it always has error field to check before doing any further actions. 
- https://go.dev/blog/error-handling-and-go

Quite convenient should i say. 
But am ok with just having error message and removing boolean field.

mosabua:
Looks like @ebyhr and @Chaho12 are closing this out together shortly. Let me know if you want a review or other help from me.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
Chaho12: Any ideas on handling this? because when i do this
```java
    @JsonProperty("isQueryParsingSuccessful")
    public boolean isQueryParsingSuccessful()
    {
        return errorMessage.isEmpty();
    }
```
I get this error. As of now, isQueryParsingSuccessful is assigned but never accessed. 
There is no need to access the field in code, but i don't see any other better ways than simply adding @SuppressWarnings("unused")
```
gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java:99:21
java: [UnusedVariable] The field 'isQueryParsingSuccessful' is never read.
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestTrinoQueryProperties.java
```diff
@@ -53,6 +54,7 @@ void testJsonCreator()
         assertThat(deserializedTrinoQueryProperties.getCatalogs()).isEqualTo(trinoQueryProperties.getCatalogs());
         assertThat(deserializedTrinoQueryProperties.getCatalogSchemas()).isEqualTo(trinoQueryProperties.getCatalogSchemas());
         assertThat(deserializedTrinoQueryProperties.isNewQuerySubmission()).isEqualTo(trinoQueryProperties.isNewQuerySubmission());
-        assertThat(deserializedTrinoQueryProperties.isQueryParsingSuccessful()).isEqualTo(trinoQueryProperties.isQueryParsingSuccessful());
+        assertThat(deserializedTrinoQueryProperties.isQueryParsingSuccessful()).isEqualTo(trinoQueryProperties.isQueryParsingSuccessful());;
```
ebyhr, line 57:
```suggestion
        assertThat(deserializedTrinoQueryProperties.isQueryParsingSuccessful()).isEqualTo(trinoQueryProperties.isQueryParsingSuccessful());
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -114,7 +116,8 @@ public TrinoQueryProperties(
             @JsonProperty("schemas") Set<String> schemas,
             @JsonProperty("catalogSchemas") Set<String> catalogSchemas,
             @JsonProperty("isNewQuerySubmission") boolean isNewQuerySubmission,
-            @JsonProperty("isQueryParsingSuccessful") boolean isQueryParsingSuccessful)
+            @JsonProperty("isQueryParsingSuccessful") boolean isQueryParsingSuccessful,
```
ebyhr, line 119: Why do we still need this field? I thought just having the getter is sufficient.

Chaho12: @ebyhr PTAL if I understood correctly, when you mean just having getter.

PR #500, opened 2024-10-02
Enable some errorprone rules and remove redundant mock usage

ebyhr:
## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
@@ -98,7 +98,6 @@ public class LbFormAuthManager
      *
      * @param idToken the access token provided back by the authorization server.
      * @return a map with the token claims
-     * @throws Exception is thrown if the access token is invalid
      */
     public Optional<Map<String, Claim>> getClaimsFromIdToken(String idToken)
     {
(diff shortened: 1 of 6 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -98,7 +98,6 @@ public Result<?> processRESTLogin(RestLoginRequest loginForm)
      *
      * @param idToken the access token provided back by the authorization server.
      * @return a map with the token claims
-     * @throws Exception is thrown if the access token is invalid
```
ebyhr, line 101: This method returns `Optional.empty()` if the access token is invalid.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbOAuthManager.java
```diff
@@ -163,7 +163,6 @@ public Response getAuthorizationCodeResponse()
      *
      * @param idToken the access token provided back by the authorization server.
      * @return a map with the token claims
-     * @throws Exception is thrown if the access token is invalid
```
ebyhr, line 166: This method returns `Optional.empty()` if the access token is invalid.

wendigo (approved):

PR #501, opened 2024-10-02
Add a cache for GatewayBackend to HaGatewayManager

choiwaiyiu:
## Description
Currently, Trino Gateway stops routing queries when the database is unavailable.

This PR adds a cache of `GatewayBackend` to the `HaGatewayManager`, ensuring that queries can still be routed even when the database is unavailable.

The cache is read-only, so update operations to `GatewayBackend` will still fail if the database is unavailable.

## Release notes


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -23,78 +22,84 @@
...
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
 public class HaGatewayManager
         implements GatewayBackendManager
 {
-    private static final Logger log = Logger.get(HaGatewayManager.class);
-
     private final GatewayBackendDao dao;
+    private final AtomicReference<List<GatewayBackend>> cache = new AtomicReference<>();
```
xkrogen, line 34: We should use something like a Guava/Caffeine cache here that has more proper cache management including cache expiration times. Remember that the same database will be shared by multiple GW instances. This code seems to assume that it can keep the cache in sync by simply invalidating the cache when update operations are made, but updates may be made out-of-band by other GW instances.

xkrogen (commented):
I am supportive of adding caching at this layer, but it's important to note that this alone won't be sufficient for the GW to continue routing traffic when the DB is unavailable, since writes happen in-the-loop of new query submission, e.g. here:
https://github.com/trinodb/trino-gateway/blob/110cd7cad9ecdac5326b13305d5ba61e03c3044a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java#L275

We need a more holistic strategy around how to handle DB unavailability which may include treating some writes as best-effort and skipping them in case of DB unavailability.

cc @surajkn

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
choiwaiyiu: @xkrogen 
Thank you for your review. @oneonestar and I are currently working on addressing the issues in this PR based on your feedback and advice :)

choiwaiyiu:
@xkrogen 
Thanks for the review, and sorry for the late reply! 
I’ve worked on fixing the cache inconsistency issue by changing how we use the cache. 
As for the implementation details, `Supplier` might not be strictly necessary, but I added it to help with performance.

choiwaiyiu:
cc: @oneonestar

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -23,59 +23,63 @@
...
+                .filter(backend -> backend.routingGroup().equals(routingGroup))
+                .collect(toImmutableList());
         return upcast(proxyBackendList);
     }
 
     @Override
     public Optional<ProxyBackendConfiguration> getBackendByName(String name)
     {
-        List<GatewayBackend> proxyBackendList = dao.findByName(name);
+        List<GatewayBackend> proxyBackendList = fetchAllBackends().stream()
```
xkrogen, line 80: So with this change, instead of letting the SQL backend do the filtering (by name / active status / routing group), we always fetch the full set of backends, and do the filtering in-process.

I think this is fine. We expect the number of backends to be small, no more than in the tens, so it shouldn't be a substantial perf hit. But let's call it out clearly in the PR description.

This also means we can remove all the (newly) unused methods in `GatewayBackendDao`: https://github.com/trinodb/trino-gateway/blob/d9998b16e8f847ba596955fb1c9629b92cdbfbd8/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/GatewayBackendDao.java#L26-L48

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -130,4 +134,16 @@ private static List<ProxyBackendConfiguration> upcast(List<GatewayBackend> gatew
         }
         return proxyBackendConfigurations;
     }
+
+    private List<GatewayBackend> fetchAllBackends()
+    {
+        try {
+            List<GatewayBackend> backends = gatewayBackendSupplier.get();
+            cache.set(backends);
```
xkrogen, line 142: Trying to understand the logic here since we are kind of layering two caches on top of each other -- the memoized supplier and the `AtomicReference`. I guess you are making it such that:

- We use the memoized supplier to reduce DB load / increase performance for frequent re-lookups (within 500ms)
- We use the AtomicReference, which never expires, to provide a _stale_ view of the data when the DB is inaccessible

This makes sense to me, but it's a bit confusing to track how these two interact. Why not just use a Guava Cache? It gives you much more configurability/flexibility so we can have all of this logic in a single place (using `refreshAfterWrite()`, which does exactly what we want here) and gives us more hooks to add monitoring:
```java
    private static final Logger LOG = Logger.get(HaGatewayManager.class);
    private static final Object ALL_BACKEND_CACHE_KEY = new Object();

    private final GatewayBackendDao dao;
    private final LoadingCache<Object, List<GatewayBackend>> backendCache;

    private final CounterStat backendLookupSuccesses = new CounterStat();
    private final CounterStat backendLookupFailures = new CounterStat();


    public HaGatewayManager(Jdbi jdbi)
    {
        dao = requireNonNull(jdbi, "jdbi is null").onDemand(GatewayBackendDao.class);
        backendCache = CacheBuilder
                .newBuilder()
                .initialCapacity(1)
                .refreshAfterWrite(Duration.ofSeconds(1))
                .build(CacheLoader.asyncReloading(
                        CacheLoader.from(this::loadAllBackends),
                        MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor())));
    }

    private List<GatewayBackend> loadAllBackends()
    {
        try {
            var backends = dao.findAll();
            backendLookupSuccesses.update(1);
            return backends;
        } catch (Exception e) {
            backendLookupFailures.update(1);
            LOG.warn(e, "Failed to load backends");
            throw e;
        }
    }

    // probably should be named "getAllBackends()" since it may or may not actually do any fetch
    private List<GatewayBackend> fetchAllBackends()
    {
        return backendCache.getUnchecked(ALL_BACKEND_CACHE_KEY);
    }
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -26,31 +27,32 @@
...
     private final GatewayBackendDao dao;
+    private final Supplier<List<GatewayBackend>> gatewayBackendSupplier;
     private final AtomicReference<List<GatewayBackend>> cache = new AtomicReference<>();
 
     public HaGatewayManager(Jdbi jdbi)
     {
         dao = requireNonNull(jdbi, "jdbi is null").onDemand(GatewayBackendDao.class);
-        cache.set(ImmutableList.of());
-        fetchAllBackendsToCache();
+        gatewayBackendSupplier = Suppliers.memoizeWithExpiration(dao::findAll, 500, MILLISECONDS);
```
xkrogen, line 42: Should probably let the cache TTL be configurable since there is a perf vs staleness tradeoff here

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
xkrogen, line 1: Let's add some testing for this as well. We want to ensure that, even if the DB goes down, we can still submit new queries.

We can add a new `TestGatewayHaDataStoreFailure` where we set up a gateway, populate some backends, submit a query to populate the cache, then simulate a database failure. After this, we should try to submit new queries, and ensure everything works as expected.

xkrogen:
Sorry for delay in my response, got swamped with some semesterly planning exercises. Business cycles and bureaucracy, ah the fun :)

mosabua:
@oneonestar and @choiwaiyiu .. I assume you are still working on this .. let us know if you need anything

ebyhr:
Closing as a stale PR. Please feel free to reopen if you continue the work.

PR #505, opened 2024-10-02
Make startup command configurable

willmostly:
## Description

Make the startup command configurable

This allows users to set java options such as `-Djavax.net.ssl.truststore=/path/to/truststore`, run short scripts, and otherwise customize startup. 

## Release notes

- ( ) This is not user-visible or is docs only, and no release notes are 
  required.
- ( ) Release notes are required. Please propose a release note for me.
- ( x) Release notes are required, with the following suggested text:
### Helm Chart
Make the startup command configurable

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -248,10 +248,22 @@ helm install tg --values values-override.yaml helm/trino-gateway
 Secrets for `authenticationSecret` and `backendState` can be provisioned
 similarly. Alternatively,  you can directly define the `config.backEndState` 
 node in `values-override.yaml` and leave `backendStateSecret` undefined. 
-However, a [Secret](https://kubernetes.
-io/docs/concepts/configuration/secret/)
-is recommended to protect the  database credentials required for this 
+However, a [Secret](https://kubernetes.io/docs/concepts/configuration/secret/)
+is recommended to protect the database credentials required for this 
 configuration.
+
+By default, the Trino Gateway process is started with the following command:
+
+```shell
+java -XX:MinRAMPercentage=80.0 -XX:MaxRAMPercentage=80.0 -jar /usr/lib/trino/gateway-ha-jar-with-dependencies.jar /etc/gateway/config.yaml
+```
+
+You can customize details with the `command` node. It accepts a list, that must
+begin with an executable such as `java` or `bash` that is available on the PATH.
+The following list elements are provided as arguments to the executable. It is
+not typically necessary to modify this node. You can use it to change of JVM
+startup parameters to control memory settings and other aspects, or to use other
+configuration file names.
 
 #### Additional options
 
--- a/helm/trino-gateway/templates/deployment.yaml
+++ b/helm/trino-gateway/templates/deployment.yaml
@@ -43,12 +43,7 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
-              - "java"
-              - "-XX:MinRAMPercentage=80.0"
-              - "-XX:MaxRAMPercentage=80.0"
-              - "-jar"
-              - "/usr/lib/trino/gateway-ha-jar-with-dependencies.jar"
-              - "/etc/gateway/config.yaml"
+            {{- toYaml .Values.command | nindent 12}}
           ports:
             - name: request
               containerPort: {{ index .Values "config" "serverConfig" "http-server.http.port" }}
--- a/helm/trino-gateway/values.yaml
+++ b/helm/trino-gateway/values.yaml
@@ -52,7 +52,7 @@ config:
     managedApps:
         - io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor
 
-# Startup command for Trino Gateway process. Add additional java options here if required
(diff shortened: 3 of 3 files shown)


helm/trino-gateway/templates/deployment.yaml
```diff
@@ -43,12 +43,8 @@ spec:
...
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
-              - "java"
-              - "-XX:MinRAMPercentage=80.0"
-              - "-XX:MaxRAMPercentage=80.0"
-              - "-jar"
-              - "/usr/lib/trino/gateway-ha-jar-with-dependencies.jar"
-              - "/etc/gateway/config.yaml"
+              - java
+              {{- toYaml append .Values.javaOptions list "-XX:MinRAMPercentage=80.0" "-XX:MaxRAMPercentage=80.0" "-jar" "/usr/lib/trino/gateway-ha-jar-with-dependencies.jar" "/etc/gateway/config.yaml"}}
```
mosabua, line 47: How about moving the options all into the values file

mosabua: In this setup you cant really override those values right? Or at least they would show up twice..

willmostly: that could be nice and simple, i'll update this

helm/trino-gateway/values.yaml
```diff
@@ -52,6 +52,15 @@ config:
     managedApps:
         - io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor
 
+# Startup command for Trino Gateway process. Add additional java options here if required
+command:
+    - "java"
```
mosabua, line 57: two space indent like in below or four like above.. maybe we should make the whole file consistent

mosabua:
If we go with this approach where the whole command is configurable.. which is fine by me .. we should update commit message, PR name, release notes suggestion, and also add relevant info to the docs.

mosabua (changes requested):
See earlier comment of mine...

helm/trino-gateway/values.yaml
willmostly: Lets go with 2 - I'll make the indentation consistent in the rest of the file in a follow up PR

docs/installation.md
```diff
@@ -253,6 +253,15 @@ io/docs/concepts/configuration/secret/)
 is recommended to protect the  database credentials required for this 
 configuration.
 
+By default, the Trino Gateway process is started with the command
```
mosabua, line 256:
```suggestion
By default, the Trino Gateway process is started with the following command:

```

docs/installation.md
```diff
@@ -253,6 +253,15 @@ io/docs/concepts/configuration/secret/)
 is recommended to protect the  database credentials required for this 
 configuration.
 
+By default, the Trino Gateway process is started with the command 
+```shell
+java -XX:MinRAMPercentage=80.0 -XX:MaxRAMPercentage=80.0 -jar /usr/lib/trino/gateway-ha-jar-with-dependencies.jar /etc/gateway/config.yaml
+```
+This can be customized through the `command` node. This node accepts a list, which
```
mosabua, line 260:
```suggestion

You can customize details with the `command` node. It accepts a list, that 
```

docs/installation.md
```diff
@@ -253,6 +253,15 @@ io/docs/concepts/configuration/secret/)
 is recommended to protect the  database credentials required for this 
 configuration.
 
+By default, the Trino Gateway process is started with the command 
+```shell
+java -XX:MinRAMPercentage=80.0 -XX:MaxRAMPercentage=80.0 -jar /usr/lib/trino/gateway-ha-jar-with-dependencies.jar /etc/gateway/config.yaml
+```
+This can be customized through the `command` node. This node accepts a list, which 
+should begin with an executable such as `java` or `bash`. The following list
```
mosabua, line 261:
```suggestion
must begin with an executable such as `java` or `bash` that is available on the PATH. The following list 
```

docs/installation.md
```diff
@@ -253,6 +253,15 @@ io/docs/concepts/configuration/secret/)
...
 configuration.
 
+By default, the Trino Gateway process is started with the command 
+```shell
+java -XX:MinRAMPercentage=80.0 -XX:MaxRAMPercentage=80.0 -jar /usr/lib/trino/gateway-ha-jar-with-dependencies.jar /etc/gateway/config.yaml
+```
+This can be customized through the `command` node. This node accepts a list, which 
+should begin with an executable such as `java` or `bash`. The following list 
+elements are provided as arguments to the executable. It is not typically necessary
+to modify this node.
```
mosabua, line 263:
```suggestion
to modify this node. You can use it to change of JVM startup parameters to control memory settings and other aspects, or to use other configuration file names.
```

helm/trino-gateway/values.yaml
```diff
@@ -52,6 +52,15 @@ config:
     managedApps:
         - io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor
 
+# Startup command for Trino Gateway process. Add additional java options here if required
```
mosabua, line 55:
```suggestion
# Startup command for Trino Gateway process. Add additional Java options and other modifications as desired.
```

mosabua (changes requested):

mosabua (approved):
Looks good now. Thank you @willmostly

PR #513, opened 2024-10-03
Enable syntax highlight in documentation

ebyhr:
## Description

https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#configuration

Before this change:
Screenshot 2024-10-03 at 11 54 48

After this change:
Screenshot 2024-10-03 at 11 54 07

## Release notes


mosabua (approved):
Nice .. we probably have to check for the right hints in a lot of the snippets but this is a great first step.

PR #518, opened 2024-10-04
Move docker readme to docs

mosabua:
## Description

This is a simple lift and shift with minor adjustments only for me. Ultimately we need to break up the installation section much better.

Details about follow up work in https://github.com/trinodb/trino-gateway/issues/519

## Release notes

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -2,14 +2,21 @@
 
 The Docker image of Trino Gateway is designed for the following use cases:
 
-* Manual usage in front of Trino clusters by mounting in a configuration file
-* Automated usage with an orchestration system like Kubernetes to simplify
-  deployment
+* Manual usage in front of Trino clusters with `docker` or `docker-compose` 
+  commands for development and testing.
+* Automated usage with [Helm chart for deployment of the container on
+  Kubernetes](installation.md#helm).
+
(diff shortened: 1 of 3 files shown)


mkdocs.yml
```diff
@@ -94,5 +94,6 @@ nav:
       - Migration to Airlift: migration-to-airlift.md
   - Contribute:
     - Code: development.md
+    - Docker: docker.md
```
mosabua, line 97: Maybe it should go in the installation section already now.. or in the follow up work as trimmed down separate page.

docs/docker.md
```diff
@@ -2,14 +2,21 @@
 
 The Docker image of Trino Gateway is designed for the following use cases:
 
-* Manual usage in front of Trino clusters by mounting in a configuration file
-* Automated usage with an orchestration system like Kubernetes to simplify
-  deployment
+* Manual usage in front of Trino clusters by mounting in a configuration file.
```
willmostly, line 5: "mounting in a configuration file" - I think this means mounting a configuration file to the container, since you can't mount an image in a file :). The wording may be confusing to some

willmostly (changes requested):

willmostly (approved):
LGTM

PR #521, opened 2024-10-11
Shorten release note permalink

ebyhr:
## Description

`https://trinodb.github.io/trino-gateway/release-notes/#trino-gateway-11-12-sep-2024`
will be
`https://trinodb.github.io/trino-gateway/release-notes/#11`

## Release notes


mosabua (approved):
Thats a cool trick .. Do you know if that works in wiki markdown pages as well .. would be good for our contributor call meetings and such.

I think I am going to test that.

PR #407, opened 2024-07-10
Small improvements

wendigo:
## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
* 
```

PR #533, opened 2024-10-24
Code Refactoring

oneonestar:
## Description
Continue the work from #407

## Release notes


gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ServerInfo.java
```diff
@@ -13,27 +13,11 @@
...
-import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import static java.util.Objects.requireNonNull;
-
 // based on https://github.com/trinodb/trino/blob/439/client/trino-client/src/main/java/io/trino/client/ServerInfo.java
 // without unused fields
-public class ServerInfo
+public record ServerInfo(
+        @JsonProperty("starting") boolean starting)
```
ebyhr, line 21: Do we still need `@JsonProperty`?

oneonestar: I thought that is the coding style for Trino to write all the `@JsonProperty` explicitly.

ebyhr: We usually don't use `@JsonProperty` for record in Trino.

ebyhr (approved):

PR #540, opened 2024-10-30
Remove easy-rules

### LINKED ISSUE #527: Remove easyrules
Currently Trino Gateway uses easyrules https://github.com/j-easy/easy-rules to allow custom routing rule definition.

While this is very flexible and powerful the project has gone stale and unmaintained.

In addition there is a severe security issue https://nvd.nist.gov/vuln/detail/CVE-2023-50571
even though that relies on a insecure class being loaded and used as part of rule validation. Since dynamic loading of such classes is not part of Trino Gateway this most likely does not apply.

We should remove easy-rules usage and find alternatives. We also discussed declaring rules in a scripting language (or even plain java). In any case .. we will have to figure out security aspects around all that. 

Options might be:

* https://commons.apache.org/proper/commons-jexl/
* https://github.com/mvel/mvel (we use it already indirectly, same security issue applies)

willmostly:
## Description

Remove easy-rules and use mvel directly for rule evaluation This PR closes #527. The existing rule format is retained, with the exception that `easy-rules` style composite rules have been removed and replaced with a `state` map that allows passing information between rule executions.

Trino Gateway currently leverages easy-rules to evaluate routing rules against an http request. This PR 
* removes easy-rules due to #527 
* Defines a rule as the combination of a condition, set of actions, and a priority
* replicate existing easy-rules functionality: routing rules with local file based storage and the MVEL engine for evaluation

The `MVELRoutingRule` and `FileBasedRoutingGroupSelector` replicate the existing easy-rules functionality, minus composite rules. Easy-rules uses mvel as the execution engine, so rule condition and action syntax does not need to be migrated. This PR drops support for [composite rules](https://trinodb.github.io/trino-gateway/routing-rules/#rule-priority), which may be a breaking change for some users. 

An additional `state` object is made available to allow passing evaluation results from lower priority rules to higher priority rules. This should compensate for the removal of the `easy-rules` composite rules. For usage examples, see tests and docs.

Database storage for MVEL rules will be added as a follow up.

## Release notes

Remove easy-rules library. **Breaking change**: composite rules are deprecated


mosabua:
Have not looked but removing one layer of indirection that just adds security issues and a bunch of syntax sugar seems like a good idea.

mosabua:
Can you ask in the maintainers group about any potential suggestion or preference for a rules engine

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -67,7 +67,6 @@ static Stream<String> provideRoutingRuleConfigFiles()
         String rulesDir = "src/test/resources/rules/";
         return Stream.of(
                 rulesDir + "routing_rules_atomic.yml",
-                rulesDir + "routing_rules_composite.yml",
```
willmostly, line 70: **This PR removes support for composite rules**

oneonestar:
I agree to remove the unmaintained `easy-rules` and the support for composite rule.
However, I think the abstraction for multi-engine support is overkill at this moment.
It's difficult to make a good abstraction when we only have one case to support.
We can add an abstraction later when we decided to add another engine.

vishalya:
Do the rest of the [examples and rules ](https://github.com/trinodb/trino-gateway/blob/main/docs/routing-rules.md#if-statements-mvel-flow-control )are taken care by your 'MVEL adapter'?

willmostly:
@vishalya it is tested with all the rules in https://github.com/willmostly/trino-gateway/tree/will/mvel-rules/gateway-ha/src/test/resources/rules. I think your issue is that `trinoRequestUser.getUser()` returns an `Optional`, so that condition will always be false. You can use `trinoRequestUser.userExistsAndEquals(\"will\")` instead.

gateway-ha/src/main/java/io/trino/gateway/ha/router/MVELFileRoutingGroupSelector.java
```diff
@@ -0,0 +1,80 @@
...
+
+        setRules(readRulesFromPath(this.rulesPath));
+    }
+
+    @Override
+    void reloadRules(long lastUpdatedTimeMillis)
+    {
+        try {
+            BasicFileAttributes attr = Files.readAttributes(this.rulesPath, BasicFileAttributes.class);
+            if (attr.lastModifiedTime().toMillis() > lastUpdatedTimeMillis) {
```
vishalya, line 49: nit: I prefer to return earlier with  `attr.lastModifiedTime().toMillis() <= lastUpdatedTimeMillis` condition.
Also, I see the exact same condition one more time, could please explain the logic here.

willmostly:
@oneonestar ptal, I've refactored to remove abstractions for multi-engine support, with the exception of an interface to define the form of a rule.

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingGroupSelector.java
```diff
@@ -0,0 +1,132 @@
...
+                // Prevent re-entry in case another thread passes the first check while rules are being updated
+                if (attr.lastModifiedTime().toMillis() > lastUpdatedTimeMillis) {
+                    List<RoutingRule> ruleList = readRulesFromPath(this.rulesPath);
+                    setRules(ruleList);
+                }
+            }
+        }
+        catch (IOException e) {
+            throw new RuntimeException("Could not access rules file", e);
+        }
```
oneonestar, line 112: All the checks and sync things can be avoided by using `memoizeWithExpiration`
https://github.com/trinodb/trino/blob/a9d83df685be215b6719f631d7b6c871ea540d1d/lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/security/FileBasedSystemAccessControlModule.java#L68-L77

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingGroupSelector.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    public List<RoutingRule> readRulesFromPath(Path rulesPath)
+    {
+        ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
```
oneonestar, line 117: Move to class level `private static final` variable.

gateway-ha/src/main/java/io/trino/gateway/ha/router/MVELRoutingRule.java
```diff
@@ -0,0 +1,122 @@
...
+                condition instanceof String stringCondition ? compileExpression(stringCondition, parserContext) : condition,
+                "condition is null");
+        this.actions = requireNonNull(actions.stream().map(action -> {
+            if (action instanceof String stringAction) {
+                return compileExpression(stringAction, parserContext);
+            }
+            else {
+                return action;
+            }
+        }).collect(toImmutableList()), "actions is null");
```
oneonestar, line 64: Could you improve the readability of this?

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingGroupSelector.java
willmostly: `memoizeWithExpiration` will change the semantics from the existing behavior in [RuleReloadingRoutingGroupSelector](https://github.com/trinodb/trino-gateway/blob/a54cb142193560da0459fd9da4a714955104b4aa/gateway-ha/src/main/java/io/trino/gateway/ha/router/RuleReloadingRoutingGroupSelector.java#L70), which this class replaces. It will also require making a change to [testByRoutingRulesEngineFileChange](https://github.com/trinodb/trino-gateway/blob/e82114083f1a4244ec960cce7fa771656af4270b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java#L249) to wait until the refresh period expires.

I don't think that the existing behavior of file updates being recognized immediately is a critical feature, but I am trying to minimize changes in behavior and avoiding changes to test logic if possible. Do you think the reduction in complexity is worth it?

oneonestar: Writing concurrent code correctly and efficiently is difficult. Both read and write to `rules` must be `synchronized`. Adding `synchronized` to `findRoutingGroup` could cause performance problems. I think using a lib here would be better.

andythsu:
Unrelated to this PR, but just want to raise a discussion:

What would be an ideal way to verify these routing rules work? Ideally we want to test
- when a request matches a particular rule, does it get routed to the right cluster
- verify that the request has the right headers (?)

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingGroupSelector.java
willmostly: @mosabua we will have to document this change in behavior

oneonestar:
@andythsu It's a good to have feature. You could open an issue to keep track of it.

gateway-ha/src/test/resources/rules/routing_rules_state.yml
```diff
@@ -0,0 +1,31 @@
+---
+name: "initialize state"
+description: "Add a set to the state map to track rules that have evaluated to true"
+priority: 0
+condition: "true"
+actions:
+  - |
```
oneonestar, line 7: Is `- |` required, or is that some kind of coding style?

willmostly: The `-` indicates this is an array element, the `|` indicates the beginning of a block scalar: https://github.com/yaml/yaml-grammar/blob/75083c2aaabd0d7453d72de895666b21021826c1/yaml-spec-1.2.txt#L1689

In this instance I'm using it because it allows me to avoid escaping all of the quotes in lines like
```
state.get("triggeredRules").contains("airflow") && request.getHeader("X-Trino-Client-Tags") contains "label=special"
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingGroupSelector.java
```diff
@@ -0,0 +1,107 @@
...
+
+import static com.google.common.base.Suppliers.memoizeWithExpiration;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.sort;
+
+public class FileBasedRoutingGroupSelector
+        implements RoutingGroupSelector
+{
+    private static final Logger log = Logger.get(FileBasedRoutingGroupSelector.class);
+    public static final String RESULTS_ROUTING_GROUP_KEY = "routingGroup";
```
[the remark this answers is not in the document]
willmostly: The intent is for this to be used in routing rules so that users do not need to know this special string `"routingGroup"`. I updated the rule files in the tests to demonstrate

oneonestar (approved):

PR #473, opened 2024-09-18
Add release notes for Trino Gateway 12 and related changes 

mosabua:
## Description

Assemble the release notes for Trino Gateway 12 release and adjust rest of docs to version 12

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

## Verification for each pull request

Format: PR/issue number, ✅ / ❌ rn ✅ / ❌ docs
✅ rn - release note added and verified, or assessed to be not necessary, set to ❌ rn before completion
✅ docs - need for docs assessed and merged, or assessed to be not necessary, set to ❌ docs before completion

Any dates missing in the list just had no merged PRs.

## 12 Sep 2024
 
* #463 ✅ rn ✅ docs

## 15 Sep 2024

* #469 ✅ rn ✅ docs

## 18 Sep 2024

* #471 ✅ rn ✅ docs

## 20 Sep 2024

* #474 ✅ rn ✅ docs

## 23 Sep 2024

* #470 ✅ rn ✅ docs

## 24 Sep 2024

* #477 ✅ rn ✅ docs
* #479 ✅ rn ✅ docs

## 27 Sep 2024

* #466 ✅ rn ✅ docs
* #488 ✅ rn ✅ docs
* #481 ✅ rn ✅ docs

## 29 Sep 2024

* #494 ✅ rn ✅ docs
* #496 ✅ rn ✅ docs
* #492 ✅ rn ✅ docs
* #495 ✅ rn ✅ docs
* #493 ✅ rn ✅ docs

## 1 Oct 2024

* #483 ✅ rn ✅ docs
* #489 ✅ rn ✅ docs
* #499 ✅ rn ✅ docs
* #500 ✅ rn ✅ docs

## 2 Oct 2024

* #490 ✅ rn ✅ docs
* #503 ✅ rn ✅ docs
* #506 ✅ rn ✅ docs
* #507 ✅ rn ✅ docs
* #508 ✅ rn ✅ docs
* #509 ✅ rn ✅ docs
* #511 ✅ rn ✅ docs
* #512 ✅ rn ✅ docs

## 3 Oct 2024

* #510 ✅ rn ✅ docs
* #513 ✅ rn ✅ docs
* #514 ✅ rn ✅ docs
* #515 ✅ rn ✅ docs
* #517 ✅ rn ✅ docs

## 11 Oct 2024

* #360 ✅ rn ✅ docs

## 15 Oct 2024 

* #522 ✅ rn ✅ docs
* #521 ✅ rn ✅ docs
[...]

PR #545, opened 2024-11-08
Fix source artifact links

mosabua:
## Description

See title

Follow up to #473 

## Release notes


vishalya (approved):
Checked the few links, they are working!

PR #549, opened 2024-11-11
Remove GitHub button in UI

Lukashmoser:
## Description
Removed GitHub button in UI as a link to the src code.
A link to the src code within an application feels unnecessary, also the website also has a link to the src code already.

More info at https://trino.io/development/process#release-note -->
## Release notes

```markdown
* Remove GitHub button from header in UI
```


mosabua (approved):
Agreed .. seems wrong for a user to link to the source code

We could link to the docs .. but imho they are not good enough for that yet. Also we could add a link in the footer .. but for now .. just removing it seems the best option.

mosabua:
Wdyt @vishalya @willmostly @oneonestar

Chaho12 (approved):
I agree.Github/docs is meant for developers or engineers like us to refer.
It is not meant for querying users to see (no need and no useful information for them)
Rather Trino session docs would be more useful for them.

ebyhr (approved):

ebyhr:
> * Remove GitHub button from header in UI

@Lukashmoser I don't think we need a release note entry.

PR #550, opened 2024-11-11
Upgrade Trino version in test runner

Lukashmoser:
## Description
Upgrade Trino version in test runner to the same version as the dependency in the pom.

Will allow us to test for http/2 and spooling protocol

## Release notes


mosabua (approved):
Looks good. Low impact.

Also note @willmostly that the quickstart guide uses trinodb:trino .. so latest which is a moving target but probably also fine for a quickstart. 

Do we potentially want to drop version completely here as well? Seems a bit too loose for a test runner to me, compared to a quick start for users.

ebyhr (approved):

PR #553, opened 2024-11-19
Support environment variables in Helm config and remove lookup (WIP)

willmostly:
## Description
Support interpolation of environment variables from secrets in Helm configuration. Remove usage of Helm lookup function in generating configuration.

The existing Helm chart uses Helm's [lookup](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function) to combine Kubernetes secrets with non-sensitive configuration. This caused incompatibilities with some frameworks such as ArgoCD.

The lookup function is replaced with an initContainer which merges secrets with the configuration provided directly in the chart. Adding an initContainer also allows providing sensitive values through environment variables using the `envFrom` feature. 

## Release notes

Support environment variable references in configuration and improve compatibility with ArgoCD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,8 @@ jobs:
           platforms: amd64,arm64,ppc64le
       - name: Build and Test Docker Image
         run: docker/build.sh
+      - name: Build and Test Init Image
+        run: helm/init-container-image/build.sh
       - name: Set up Helm
         uses: azure/setup-helm@v4.2.0
         with:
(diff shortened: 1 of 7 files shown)


mosabua:
So .. I am confused now .. should this not be a PR in the charts repo and in this repo we update the docs to point over there? And then we also update the release process to cut a charts release for the Trino Gateway part each time we release Trino Gateway?

willmostly:
@mosabua old habits die hard... I will move this PR over

PR #558, opened 2024-11-29
Adjust to the moved Helm chart

mosabua:
## Description

As discussed in dev sync.

Related to https://github.com/trinodb/charts/pull/268

## Release notes


docs/installation.md
```diff
@@ -204,8 +204,16 @@ java -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80 \
 ### Helm <a name="helm"></a>
```
willmostly, line 204: This section is linked from docker.md - perhaps it should just link directly to the chart repo - wdyt?

willmostly (approved):
one comment but I think this is good to merge

docs/installation.md
mosabua: No .. I want to keep this since I want to actually get this onto a separate page and much improved.. so for now .. lets leave as is

Issue #516, opened 2024-10-03
Wrap too long code snippet in documentation


ebyhr:
https://trinodb.github.io/trino-gateway/quickstart/#add-trino-backends

![Screenshot 2024-10-04 at 8 03 38](https://github.com/user-attachments/assets/1a26e88f-3a74-409c-b8cc-2fcd7266af52)

Benji0187:
Is anyone working on this? If you assign it to me, I can pick it up.

mosabua:
Go right ahead .. send a pull request and link to this issue.

PR #565, opened 2024-12-05
Refactoring code snippets in quickstart.md

sanspareilsmyn:
## Description
Refactoring code snippets in quickstart.md. This PR closes https://github.com/trinodb/trino-gateway/issues/516. 

The refactoring of the script enhances readability and maintainability by organizing code into functions and adding clear comments.

## Release notes

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -22,49 +22,66 @@ It  copies the following, necessary files to current directory:
 ```shell
 #!/usr/bin/env sh
 
+# Set the version for the Gateway Jar
 VERSION=12
 
-# Copy necessary files to current directory
+# Function to copy necessary files to the current directory
+copy_files() {
+    # Check and get the Gateway Jar
+    if [[ -f "gateway-ha.jar" ]]; then
+        echo "Found gateway-ha.jar file in current directory."
+    else
+        echo "Fetching gateway-ha.jar version $VERSION from Maven Central repository."
+        curl -O "https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha/${VERSION}/gateway-ha-${VERSION}-jar-with-dependencies.jar"
+        mv "gateway-ha-${VERSION}-jar-with-dependencies.jar" ./gateway-ha.jar
+    fi
 
-# Check and get the Gateway Jar
-if [[ -f "gateway-ha.jar" ]]; then
-    echo "Found gateway-har.jar file in current directory."
-else
-    echo "Failed to find gateway-ha.jar in current directory. Fetching version $VERSION from Maven Central repository."
-    curl https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha/${VERSION}/gateway-ha-${VERSION}-jar-with-dependencies.jar -o ./gateway-ha.jar
-fi
+    # Check and get the Config.yaml
+    if [[ -f "quickstart-config.yaml" ]]; then
+        echo "Found quickstart-config.yaml file in current directory."
+    else
+        cp ../docs/quickstart-config.yaml ./quickstart-config.yaml
+    fi
 
-# Check and get the Config.yaml
-if [[ -f "quickstart-config.yaml" ]]; then
-    echo "Found quickstart-config.yaml file in current directory."
-else
-    cp ../docs/quickstart-config.yaml ./quickstart-config.yaml
-fi
+    # Check and get the postgres.sql
+    if [[ -f "gateway-ha-persistence-postgres.sql" ]]; then
+        echo "Found gateway-ha-persistence-postgres.sql file in current directory."
+    else
+        cp ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql ./gateway-ha-persistence-postgres.sql
+    fi
+}
 
-# Check and get the postgres.sql
-if [[ -f "gateway-ha-persistence-postgres.sql" ]]; then
-    echo "Found gateway-ha-persistence-postgres.sql file in current directory."
-else
-    cp ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql ./gateway-ha-persistence-postgres.sql
-fi
+# Function to check if PostgreSQL database is running and start it if not
+start_postgres_db() {
+    if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
+        echo "PostgreSQL database container 'local-postgres' is already running. Only starting Trino Gateway."
+    else
+        echo "Starting PostgreSQL database container 'local-postgres'."
+        
+        # Set the password for PostgreSQL
+        export PGPASSWORD=mysecretpassword
+        
+        # Run PostgreSQL container with necessary configurations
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -22,49 +22,66 @@ It  copies the following, necessary files to current directory:
...
-
-#Start Trino Gateway server.
+# Function to copy necessary files to the current directory
+copy_files() {
+    # Check and get the Gateway Jar
+    if [[ -f "gateway-ha.jar" ]]; then
+        echo "Found gateway-ha.jar file in current directory."
+    else
+        echo "Fetching gateway-ha.jar version $VERSION from Maven Central repository."
+        curl -O "https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha/${VERSION}/gateway-ha-${VERSION}-jar-with-dependencies.jar"
```
mosabua, line 35: Can we somehow shorten this? Maybe define a baseurl as `https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha/` and use it

docs/quickstart.md
```diff
@@ -22,49 +22,66 @@ It  copies the following, necessary files to current directory:
...
+        echo "Found quickstart-config.yaml file in current directory."
+    else
+        cp ../docs/quickstart-config.yaml ./quickstart-config.yaml
+    fi
+
+    # Check and get the postgres.sql
+    if [[ -f "gateway-ha-persistence-postgres.sql" ]]; then
+        echo "Found gateway-ha-persistence-postgres.sql file in current directory."
+    else
+        cp ../gateway-ha/src/main/resources/gateway-ha-persistence-postgres.sql ./gateway-ha-persistence-postgres.sql
```
mosabua, line 50: shorten

docs/quickstart.md
```diff
@@ -22,49 +22,66 @@ It  copies the following, necessary files to current directory:
...
+    fi
+}
+
+# Function to check if PostgreSQL database is running and start it if not
+start_postgres_db() {
+    if docker ps --format '{{.Names}}' | grep -q '^local-postgres$'; then
+        echo "PostgreSQL database container 'local-postgres' is already running. Only starting Trino Gateway."
```
mosabua, line 57: shorten

docs/quickstart.md
```diff
@@ -22,49 +22,66 @@ It  copies the following, necessary files to current directory:
...
+    else
+        echo "Starting PostgreSQL database container 'local-postgres'."
+        
+        # Set the password for PostgreSQL
+        export PGPASSWORD=mysecretpassword
+        
+        # Run PostgreSQL container with necessary configurations
+        docker run -v "$(pwd)"/gateway-ha-persistence-postgres.sql:/tmp/gateway-ha-persistence-postgres.sql \
```
mosabua, line 65: shorten

sanspareilsmyn:
@mosabua Thanks for the reivew. I've just applied your feedbacks!

docs/quickstart.md
```diff
@@ -77,28 +72,37 @@ kill -5 $(jps | grep gateway-ha.jar | cut -d' ' -f1)
...
+              \"name\": \"$1\",
+              \"proxyTo\": \"http://localhost:808$2\",
+              \"active\": true,
+              \"routingGroup\": \"adhoc\"
+            }"
+}
+
+for i in 1 2; do
+    add_backend "trino$i" "$i"
+done                                                                                   }'
```
mosabua, line 105: End of line is wrong

mosabua (changes requested):

docs/quickstart.md
```diff
@@ -23,49 +23,44 @@ It  copies the following, necessary files to current directory:
...
-fi
-
-
-#Start Trino Gateway server.
-java -Xmx1g -jar ./gateway-ha.jar ./quickstart-config.yaml
+BASE_URL="https://repo1.maven.org/maven2/io/trino/gateway/gateway-ha"
+POSTGRES_SQL="gateway-ha-persistence-postgres.sql"
+JAR_FILE="gateway-ha-$VERSION-jar-with-dependencies.jar"
+GATEWAY_JAR="gateway-ha.jar"
+CONFIG_YAML="quickstart-config.yaml"
```
mosabua, line 30: The script doesnt actually work .. options:

* remove the log levels line in the config file `log.levels-file: ..`
* add a command here that adds an empty log properties file in the right folder
* add a command here that adds and empty log.properties file in the current folder and update config for current folder

mosabua: Personally I think we dont need the log levels in the quickstart .. what do you think @willmostly ?

mosabua:
Also with this whole clean up .. could we add one more config for the folder of the Trino Gateway source code .. the default could be `.` or so 

Then the script could be used in any folder .. we would just have to document to change that variable to an absolute path to the checkout .. or even a relative one from the script directory.

docs/quickstart.md
willmostly: sorry @mosabua and @sanspareilsmyn, i missed this comment. IMO simpler is better for this example, so removing the `log.levels-file` line is the best option

mosabua:
I squash commits, address the invalid line, removed the log config, and rebased. However I ran out of time to fix it up.. currently the script fails due to the bootstrap changes .. have to investigate or @sanspareilsmyn can pull the current version and push further improvements.

mosabua:
Thoughts @willmostly or @oneonestar ?

mosabua:
So this currently breaks because the quickstart script is a mixture of using the config file from the current checkout and the files from an actual released version. Wdyt @willmostly .. should we try to change it so it uses only the current checkout consistently? The problem then is that people have to build Trino Gateway .. but then on the other hand we dont have to set version, its good for testing any specific version and building can be easy if you just use mvnw clean install with skipping tests.. 

Thoughts from others like @vishalya @oneonestar or @Chaho12 ?

mosabua:
Also @sanspareilsmyn .. I messed up the commits.. can you slack me your email and full name so I can use it on co-authored-by in the commit .. or otherwise .. pull my changes down and push on top of it with your user..

Chaho12:
There is this part so that if there is built jar there, it would use that jar instead so we can test any version.

```
    if [[ ! -f "$GATEWAY_JAR" ]]; then
        echo "Fetching $GATEWAY_JAR version $VERSION"
        curl -O "$BASE_URL/$VERSION/$JAR_FILE"
        mv "$JAR_FILE" "$GATEWAY_JAR"
    fi
```

Regarding the conf mismatch with released version, we could add msg that saids sth like "Use preferred release tag for testing" 

Or maybe maintain separate configuration files for each version? For example, `config_v13.yaml`, `config_LTS.yaml`. And for every release, we update LTS one, make copy (config_v14.yaml) and remove last config(config_v13)? But i don't think this is necessary because user can just checkout tag.

oneonestar:
This quickstart script predates gateway docker image.
IMO, the objective of this script is for new comer to run and try out the latest released gateway with minimal effort.
It should use gateway docker image and the corresponding config should be embedded into the docker image.

We could provide another guide or script for developer who want's to setup a dev environment.

mosabua (approved):
Looks good.. I will send follow up PR for the docs.

Issue #564, opened 2024-12-04
Trailing / in backend ProxyTo URL causes routing error


willmostly:
A trailing slash in the proxyTo URL, e.g. 
```
{"name": "my-proxy-target", "routingGroup":"adhoc", "ProxyTo": "https://my-proxy-target.io/", "ExternalUrl": "https://my-proxy-target.io/"}
```
 will produce a URL like https://my-proxy-target.io//v1/statement. This will in turn produce an obscure error message such as 

```
[...]

PR #570, opened 2024-12-13
Remove trailing slash in proxyTo and externalUrl in create and update backend methods

prakhar10:
## Description
Remove trailing slash in proxyTo and externalUrl in create and update backend methods. Fixes https://github.com/trinodb/trino-gateway/issues/564

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
@@ -94,19 +94,23 @@ public class HaGatewayManager
     @Override
     public ProxyBackendConfiguration addBackend(ProxyBackendConfiguration backend)
     {
-        dao.create(backend.getName(), backend.getRoutingGroup(), backend.getProxyTo(), backend.getExternalUrl(), backend.isActive());
+        String backendProxyTo = removeTrailingSlash(backend.getProxyTo());
+        String backendExternalUrl = removeTrailingSlash(backend.getExternalUrl());
+        dao.create(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
         return backend;
     }
 
     @Override
     public ProxyBackendConfiguration updateBackend(ProxyBackendConfiguration backend)
     {
+        String backendProxyTo = removeTrailingSlash(backend.getProxyTo());
+        String backendExternalUrl = removeTrailingSlash(backend.getExternalUrl());
         GatewayBackend model = dao.findFirstByName(backend.getName());
         if (model == null) {
-            dao.create(backend.getName(), backend.getRoutingGroup(), backend.getProxyTo(), backend.getExternalUrl(), backend.isActive());
+            dao.create(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
         }
         else {
-            dao.update(backend.getName(), backend.getRoutingGroup(), backend.getProxyTo(), backend.getExternalUrl(), backend.isActive());
+            dao.update(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
         }
         return backend;
     }
@@ -129,5 +133,10 @@ public class HaGatewayManager
             proxyBackendConfigurations.add(backendConfig);
         }
         return proxyBackendConfigurations;
+    }
+
+    public static String removeTrailingSlash(String url)
+    {
+        return url.replaceAll("/$", "");
     }
 }
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
@@ -83,4 +83,30 @@ final class TestHaGatewayManager
                 .extracting(ProxyBackendConfiguration::getRoutingGroup)
                 .containsExactly("adhoc");
     }
+
+    @Test
+    void testRemoveTrailingSlashInUrl()
+    {
+        ProxyBackendConfiguration etl = new ProxyBackendConfiguration();
+        etl.setActive(false);
+        etl.setRoutingGroup("etl");
+        etl.setName("new-etl1");
+        etl.setProxyTo("https://etl1.trino.gateway.io:443/");
+        etl.setExternalUrl("https://etl1.trino.gateway.io:443/");
+        haGatewayManager.addBackend(etl);
+
+        assertThat(haGatewayManager.getBackendByName("new-etl1").map(ProxyBackendConfiguration::getProxyTo).orElseThrow()).isEqualTo("https://etl1.trino.gateway.io:443");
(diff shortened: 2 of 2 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -94,19 +94,23 @@ public void activateBackend(String backendName)
     @Override
     public ProxyBackendConfiguration addBackend(ProxyBackendConfiguration backend)
     {
-        dao.create(backend.getName(), backend.getRoutingGroup(), backend.getProxyTo(), backend.getExternalUrl(), backend.isActive());
+        String backendProxyTo = backend.getProxyTo().replaceAll("/$", "");
```
ebyhr, line 97: Please add a helper method to remove trailing slash.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -82,5 +82,16 @@ void testGatewayManager()
         assertThat(haGatewayManager.getAllBackends())
                 .extracting(ProxyBackendConfiguration::getRoutingGroup)
                 .containsExactly("adhoc");
+
+        ProxyBackendConfiguration etl = new ProxyBackendConfiguration();
```
ebyhr, line 86: Please extract a test method.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
willmostly: I think the same class should be fine

willmostly (commented):
Looks good % @ebyhr's comment

prakhar10:
@willmostly @ebyhr can you please review.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -83,4 +83,30 @@ void testGatewayManager()
...
+
+    @Test
+    void testRemoveTrailingSlashesInUrl()
+    {
+        ProxyBackendConfiguration etl = new ProxyBackendConfiguration();
+        etl.setActive(false);
+        etl.setRoutingGroup("etl");
+        etl.setName("new-etl1");
+        etl.setProxyTo("https://etl1.trino.gateway.io:443/");
+        etl.setExternalUrl("https://etl1.trino.gateway.io:443//");
```
ebyhr, line 95: Why do we want to remove more than one `/`?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -130,4 +134,9 @@ private static List<ProxyBackendConfiguration> upcast(List<GatewayBackend> gatew
         }
         return proxyBackendConfigurations;
     }
+
+    private static String removeTrailingSlashes(String url)
+    {
+        return url.replaceAll("/+$", "");
```
ebyhr, line 140: We should remove only one `/` in my opinion.

prakhar10: Ok, I was thinking by chance if someone adds multiple slashes by mistake we should be able to handle that. Should I revert @ebyhr ?

ebyhr: Yes.

ebyhr:

prakhar10:
@ebyhr  can you check now please?

ebyhr:
> Revert handling multiple trailing slashes in url

Please remove the commit body.

willmostly (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -94,19 +94,23 @@ public void activateBackend(String backendName)
     @Override
     public ProxyBackendConfiguration addBackend(ProxyBackendConfiguration backend)
     {
-        dao.create(backend.getName(), backend.getRoutingGroup(), backend.getProxyTo(), backend.getExternalUrl(), backend.isActive());
+        String backendProxyTo = removeTrailingSlash(backend.getProxyTo());
+        String backendExternalUrl = removeTrailingSlash(backend.getExternalUrl());
```
ebyhr, line 98: These getters are called from other places. Please check all usages. For instance, The following place has the same issue because `UI_LOGIN_PATH` is `/ui/login`. 

https://github.com/trinodb/trino-gateway/blob/40337f6444362478beba6d555c1425f3490b3cc1/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java#L132

prakhar10: So the way I was thinking about this is to remove the slash while saving a new backend or updating a new backend so that later on any method that pulls this will always get the correct url from the database instead of trimming it in every part of the code.

prakhar10: @ebyhr  any thoughts on this?

mosabua:
Is anything here still to be done besides a merge @willmostly @ebyhr ?

ebyhr (changes requested):
(This PR isn't ready for merge)

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
prakhar10: > So the way I was thinking about this is to remove the slash while saving a new backend or updating a new backend so that later on any method that pulls this will always get the correct url from the database instead of trimming it in every part of the code.

@ebyhr does this make sense?

ebyhr: > while saving a new backend or updating a new backend

How does it work in existing environments?

prakhar10: yeah I think if a cluster that is already added and has a trailing slash then that would still be an issue and they'll need to manually fix it. Ok I'll probably extract the helper method into a common Util class and then use it across other places in the codebase, does that sound ok @ebyhr ?

prakhar10: Please can you review now.

willmostly: > if a cluster that is already added and has a trailing slash then that would still be an issue and they'll need to manually fix it.

This  should be fine - clusters with URLs containing trailing slashes are unusable, so they should be very rare in existing environments

prakhar10: Should I revert the change from other files then?

mosabua:
@ebyhr @willmostly .. what are we doing here? Can we get this merged or are there still open concerns?

willmostly:
@mosabua suggested release note: 
```
* fix issue with backend URLs containing trailing slash (#564)
```

PR #573, opened 2024-12-14
Ignore version output file for git

mosabua:
## Description

The docker build creates that file and we should ignore it so accidental commit of it can be avoided.

na

## Release notes


Chaho12 (approved):
LGTM 👍

PR #575, opened 2024-12-16
Use flyway to perform database migrations on startup

posulliv:
## Description

Add support for running mirations automatically with flyway on server startup. Since users may not want this, I added a config option to opt out of having the gateway process perform migrations. Migrations will only be performed by the gateway process if this config option is enabled.

This change will allow us to add backend DB support for Oracle in the future.

## Release notes

( X) Release notes are required. Please propose a release note for me.

```markdown
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -38,15 +38,36 @@ distribution is installed.
 
 ### Backend database
 
-Trino Gateway requires a MySQL or PostgreSQL database.
+Trino Gateway requires a MySQL or PostgreSQL database. Database initialization
+is performed automatically when the Trino Gateway process starts. Migrations
+are performed using `Flyway`.
 
-Use the following scripts in the `gateway-ha/src/main/resources/` folder to
-initialize the database:
-
-* `gateway-ha-persistence-mysql.sql` for MySQL
-* `gateway-ha-persistence-postgres.sql` for PostgreSQL
+The migration files can viewed in the `gateway-ha/src/main/resources/` folder.
+Each database type supported has its own sub-folder.
 
 The files are also included in the JAR file.
+
+If you do not want migrations to be performed automatically on startup, then
+you can set `runMigrationsEnabled` to `false` in the data store configuration.
+For example:
+
+```yaml
+dataStore:
+  jdbcUrl: jdbc:postgresql://postgres:5432/trino_gateway_db
+  user: USER
+  password: PASSWORD
+  driver: org.postgresql.Driver
+  queryHistoryHoursRetention: 24
+  runMigrationsEnabled: false
+```
+
+`Flyway` uses a transactional lock in databases that support it such as 
+[PostgreSQL](https://documentation.red-gate.com/fd/postgresql-database-235241807.html#).
+In the scenario where multiple Trino Gateway instances are running and sharing
+the same backend database, the first Trino Gateway instance to start will get
+the lock and run the database migrations with `Flyway`. Other Trino Gateway
+instances might fail during startup while migrations are running but once migrations
+are completed they will start as expected.
 
 ### Trino clusters
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
@@ -20,14 +20,16 @@ public class DataStoreConfiguration
     private String password;
     private String driver;
     private Integer queryHistoryHoursRetention = 4;
+    private boolean runMigrationsEnabled = true;
 
-    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, Integer queryHistoryHoursRetention)
+    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled)
     {
         this.jdbcUrl = jdbcUrl;
         this.user = user;
         this.password = password;
         this.driver = driver;
         this.queryHistoryHoursRetention = queryHistoryHoursRetention;
+        this.runMigrationsEnabled = runMigrationsEnabled;
     }
 
     public DataStoreConfiguration() {}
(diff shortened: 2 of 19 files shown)


mosabua:
Awesome.. we just talked about this recently in the dev sync - fyi @willmostly

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -20,14 +20,16 @@ public class DataStoreConfiguration
     private String password;
     private String driver;
     private Integer queryHistoryHoursRetention = 4;
+    private boolean runMigrationsEnabled;
```
willmostly, line 23: This could be defaulted to `true`. Most users should allow the gateway to manage their DB schema, and it makes things easier for new users as well.

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/BaseTestDatabaseMigrations.java
```diff
@@ -0,0 +1,102 @@
...
+    protected abstract String getDriver();
+
+    @AfterAll
+    public final void close()
+    {
+        container.close();
+    }
+
+    @Test
+    public void testMigrationWithEmptyDatabase()
```
willmostly, line 51: Can we add a test case for a non-empty database as well?

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/BaseTestDatabaseMigrations.java
```diff
@@ -0,0 +1,102 @@
...
+    protected void dropAllTables()
+    {
+        String gatewayBackendTable = "DROP TABLE IF EXISTS gateway_backend";
+        String queryHistoryTable = "DROP TABLE IF EXISTS query_history";
+        String propertiesTable = "DROP TABLE IF EXISTS resource_groups_global_properties";
+        String resourceGroupsTable = "DROP TABLE IF EXISTS resource_groups";
+        String selectorsTable = "DROP TABLE IF EXISTS selectors";
+        String exactMatchTable = "DROP TABLE IF EXISTS exact_match_source_selectors";
+        String flywayHistoryTable = "DROP TABLE IF EXISTS flyway_schema_history";
+        Handle jdbiHandle = jdbi.open();
```
willmostly, line 92: Can we add a check that the schema is now empty? A query against `information_schema` should work for both pg and mysql. This will ensure that if someone adds a table to the schema, they remember to update this method

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
willmostly: Please document this setting

willmostly (commented):
Just a few minor comments - this looks good to me overall and should be a nice improvement!

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
willmostly: I think https://github.com/trinodb/trino-gateway/blob/f4b78f48ca5e882575a4f82640bb70940195c79b/docs/installation.md?plain=1#L39 is probably the best place for it now. You can revise the section that refers to initializing the DB manually and let users know that the Gateway will do it for them, unless they set `runMigrationsEnabled: false`, in which case they will need to manually run the schema SQLs

mosabua:
Don't have time to review but here are some thoughts:

- Can we check database permissions before scripts are run
- I agree that we want to have this migration on by default, but be able to deactivate it
- Must be part of the app start so it works for any deployment
- If migration fails or some problem is found the app startup should probably fail hard so the app is not used at all in some weird state
- have to figure out how this works in a cluster where we have multiple instance, maybe first instance gets a lock and then all other instances are prevented from starting (or fail to start) until the migration is done

posulliv:
thanks @mosabua Replying to your comments:

> Can we check database permissions before scripts are run

flyway will fail if database permissions are incorrect and the gateway process will not start up. trino has the same behavior for database backed resource groups.

> I agree that we want to have this migration on by default, but be able to deactivate it

Got it. This is implemented in the PR now with an option to disable it.

> Must be part of the app start so it works for any deployment

This is in the PR. Migrations occur on startup unless they are explitly disabled in the data store config.

> If migration fails or some problem is found the app startup should probably fail hard so the app is not used at all in some weird state

Yep, this is behavior as implemented in this PR. If a migration fails the process will not start and the logs will contain the migration error.

> have to figure out how this works in a cluster where we have multiple instance,

Are you saying we need to handle this in this PR? This state indicating one gateway process is performed migrations would need to be maintained somewhere. Seems like backing DB is the only only place we could store it when there are multiple gateway processes?

mosabua:
Awesome.. thanks for answering @posulliv 

The only thing I think we need to figure out is how to deal with multiple nodes. Can you find out how flyway deals with this. Feels like it should support that out of the box somehow.

Maybe the first node that starts up locks the db for the upgrade and any additional nodes connecting just fail or have to wait until migration is completed. We just need to avoid having multiple nodes connecting and all of them starting to migrate the database and ending up with a mess

posulliv:
Yeah @willmostly suggested flyway might take care of this as well. It does in fact use locks - https://documentation.red-gate.com/fd/postgresql-database-235241807.html#

> By default Flyway uses a transactional lock with PostgreSQL

So the behavior should be the first gateway instance to start will get the lock and run the migrations. Other instances might fail during startup but if this is a k8s deployment, once they are automatically restarted they should be good as migrations will be completed.

mosabua:
Sweet.. lets just add that to the docs then.

posulliv:
@mosabua ok cool. I updated the docs in latest push.

docs/installation.md
```diff
@@ -38,16 +38,37 @@ distribution is installed.
...
+dataStore:
+  jdbcUrl: jdbc:postgresql://postgres:5432/trino_gateway_db
+  user: USER
+  password: PASSWORD
+  driver: org.postgresql.Driver
+  queryHistoryHoursRetention: 24
+  runMigrationsEnabled: false
+```
+
+`flyway` uses a transactional lock in databases that support it such as
```
willmostly, line 64: Nit: Capitalize  `flyway`

willmostly (approved):
Looks good!

willmostly:
@mosabua ptal

Issue #585, opened 2025-01-08
UI theme change looses label texts in charts


mosabua:
Steps to reproduce

* Access UI and observe labels on charts
* Switch through to other theme (from light to dark or dark to light)
* Note that the label texts on the charts disappear (white text on white background, or black text on black background) 

Somehow the theme and css change does not propagate to the chart library - thats the bug.

However note that this only fails when using the button to change theme. If you refresh the page (browser refresh) the labels are correct.

PR #593, opened 2025-01-18
Fix chart legend font color issue after theme change

prakhar10:
## Description

Fix chart legend font color issue after changing the UI theme. Fixes https://github.com/trinodb/trino-gateway/issues/585

## Release notes

### DIFF
--- a/webapp/src/components/dashboard.tsx
+++ b/webapp/src/components/dashboard.tsx
@@ -117,10 +117,28 @@ export function Dashboard() {
   );
 }
 
+function updateLegendColor() {
+    const [color, setColor] = useState(getCSSVar('--semi-color-text-0'));
+
+    useEffect(() => {
+        const handleThemeChange = () => {
+            setColor(getCSSVar('--semi-color-text-0'));
+        };
+
+        const observer = new MutationObserver(handleThemeChange);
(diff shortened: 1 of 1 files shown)


mosabua (approved):
This looks good although I really dislike the fact that this is even necessary. It feels brittle for any further changes or wider need to change color on chart. However .. I think we probably still want to merge this.

webapp/src/components/dashboard.tsx
```diff
@@ -150,7 +168,7 @@ function LineChart(props: {
     const option = {
       legend: {
         textStyle: {
-          color: getCSSVar('--semi-color-text-0')
+            color: legendColor
```
ebyhr, line 171: Revert a change for indentation.

prakhar10:
> This looks good although I really dislike the fact that this is even necessary. It feels brittle for any further changes or wider need to change color on chart. However .. I think we probably still want to merge this.

I agree. This seemed a bit weird, for some reason its not triggering a refresh to the legend color when the theme is updated, that's why I had to it this way.

mosabua:
Haha .. you beat me to it @ebyhr .. i was just about to switch gears from the Trino release notes..

PR #594, opened 2025-01-20
Remove unused code

oneonestar:
## Description
Remove unused code.

Forgot to remove the dead code when  refactoring from while loop to ScheduledExecutorService in #533.

## Release notes


ebyhr (approved):

mosabua (approved):
Haha .. nice catch. Maybe we should do a search for Thread.sleep overall.

PR #599, opened 2025-01-22
Remove "Optional[]" from getResourceGroupQueryType string

willmostly:
TrinoQueryProperties.getResourceGroupQueryType() called `toString` on an `Optional`. This unintended behavior produced output such as `"Optional[DATA_DEFINITION]"`, complicating rule writing for end users.

## Release notes

```markdown
* Fix TrinoQueryProperties.getResourceGroupQueryType()
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/StatementUtils.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/StatementUtils.java
@@ -14,6 +14,7 @@
 package io.trino.gateway.ha.router;
 
 import com.google.common.collect.ImmutableList;
+import io.airlift.log.Logger;
 import io.trino.sql.tree.AddColumn;
 import io.trino.sql.tree.Analyze;
 import io.trino.sql.tree.Call;
@@ -89,7 +90,6 @@ import io.trino.sql.tree.Update;
 import io.trino.sql.tree.Use;
 
 import java.util.Map;
-import java.util.Optional;
 
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static io.trino.gateway.ha.router.QueryType.ALTER_TABLE_EXECUTE;
@@ -108,6 +108,8 @@ import static java.util.function.Function.identity;
 //modified version of io.trino.util.StatementUtils
 public final class StatementUtils
 {
+    private static final Logger log = Logger.get(StatementUtils.class);
+
     private StatementUtils() {}
 
     private static final Map<Class<? extends Statement>, StatementTypeInfo<? extends Statement>> STATEMENT_QUERY_TYPES = ImmutableList.<StatementTypeInfo<?>>builder()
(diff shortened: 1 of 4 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/StatementUtils.java
```diff
@@ -191,13 +193,17 @@ private StatementUtils() {}
...
-            return getQueryType(((ExplainAnalyze) statement).getStatement());
+            return getResourceGroupQueryType(((ExplainAnalyze) statement).getStatement());
+        }
+        StatementTypeInfo<? extends Statement> statementTypeInfo = STATEMENT_QUERY_TYPES.get(statement.getClass());
+        if (statementTypeInfo != null) {
+            return statementTypeInfo.getQueryType().toString();
         }
-        return Optional.ofNullable(STATEMENT_QUERY_TYPES.get(statement.getClass()))
-                .map(StatementTypeInfo::getQueryType);
+        log.warn("Encountered unknown statement type: %s", statement.getClass());
```
ebyhr, line 205: We usually log messages when encountered something. I would remove "Encountered" from the message. 
```suggestion
        log.warn("Unsupported statement type: %s", statement.getClass());
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/StatementUtils.java
```diff
@@ -191,13 +193,17 @@ private StatementUtils() {}
             .build().stream()
             .collect(toImmutableMap(StatementTypeInfo::getStatementType, identity()));
 
-    public static Optional<QueryType> getQueryType(Statement statement)
+    public static String getResourceGroupQueryType(Statement statement)
```
ebyhr, line 196: Why did you rename this method?

willmostly: `TrinoQueryProperties` already uses the name `queryType` for the statement class name:
```
queryType = statement.getClass().getSimpleName();
resourceGroupQueryType = StatementUtils.getQueryType(statement).toString();
```
I changed the method name in this PR because I found myself confused while reading the code. I could put it in a separate commit if that is preferable. 

I would actually rather change `queryType` in  `TrinoQueryProperties` to `statementType`, but this would be a breaking change for those using `TrinoQueryProperties.getQueryType()` in their routing rules.

ebyhr: I don't request a separate commit, but explaining such a context in  a PR description or leaving a review comment will be helpful.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelector.java
```diff
@@ -178,6 +178,21 @@ void testTrinoQueryPropertiesQueryType()
...
+    void testTrinoQueryPropertiesResourceGroupQueryType()
+            throws IOException
+    {
+        RoutingGroupSelector routingGroupSelector =
+                RoutingGroupSelector.byRoutingRulesEngine("src/test/resources/rules/routing_rules_trino_query_properties.yml", requestAnalyzerConfig);
+        String query = "CREATE TABLE cat.schem.foo (c1 int)";
+        Reader reader = new StringReader(query);
+        BufferedReader bufferedReader = new BufferedReader(reader);
+        HttpServletRequest mockRequest = prepareMockRequest();
+        when(mockRequest.getReader()).thenReturn(bufferedReader);
```
ebyhr, line 191: We could inline some variables:
```java
        when(mockRequest.getReader())
                .thenReturn(new BufferedReader(new StringReader("CREATE TABLE cat.schem.foo (c1 int)")));
```

I would suggest adding a helper method in follow-up.

ebyhr (approved):

PR #601, opened 2025-01-27
Display cluster health status on cluster page

prakhar10:
## Description

Display cluster health status on the cluster page

Screenshot 2025-01-28 at 12 27 56 AM

Screenshot 2025-01-28 at 8 35 39 AM

Screenshot 2025-01-29 at 9 30 16 AM

Screenshot 2025-01-29 at 9 31 48 AM

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoStatus.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoStatus.java
@@ -18,7 +18,7 @@ package io.trino.gateway.ha.clustermonitor;
  * We should use PENDING when Trino clusters are still spinning up
  * HEALTHY is when health checks report clusters as up
  * UNHEALTHY is when health checks report clusters as down
- * UNKNOWN is when the health checks are not able to determine the status
+ * UNKNOWN is the initial default state and also used when the health checks are not able to determine the status
  */
 public enum TrinoStatus
 {
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoStatus.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoStatus.java
@@ -18,10 +18,12 @@ package io.trino.gateway.ha.clustermonitor;
  * We should use PENDING when Trino clusters are still spinning up
  * HEALTHY is when health checks report clusters as up
  * UNHEALTHY is when health checks report clusters as down
+ * UNKNOWN is when the health checks are not able to determine the status
  */
 public enum TrinoStatus
 {
     PENDING,
     HEALTHY,
-    UNHEALTHY
+    UNHEALTHY,
+    UNKNOWN
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
@@ -99,7 +99,7 @@ public class GatewayWebAppResource
             backendResponse.setProxyTo(b.getProxyTo());
             backendResponse.setActive(b.isActive());
             if (Objects.isNull(backendState.trinoStatus())) {
-                backendResponse.setStatus(TrinoStatus.PENDING.toString());
+                backendResponse.setStatus(TrinoStatus.UNKNOWN.toString());
             }
             else {
                 backendResponse.setStatus(backendState.trinoStatus().toString());
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
@@ -16,7 +16,6 @@ package io.trino.gateway.ha.resource;
 import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
-import io.trino.gateway.ha.clustermonitor.TrinoStatus;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.domain.Result;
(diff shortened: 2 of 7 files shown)


willmostly:
Nice usability improvement!

ebyhr:
Could you add another screenshot with dark theme to PR description?

prakhar10:
> Could you add another screenshot with dark theme to PR description?

Thanks @ebyhr , done.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -97,6 +98,12 @@ public Response getAllBackends()
             backendResponse.setName(b.getName());
             backendResponse.setProxyTo(b.getProxyTo());
             backendResponse.setActive(b.isActive());
+            if (Objects.isNull(backendState.trinoStatus())) {
+                backendResponse.setStatus(TrinoStatus.PENDING.toString());
```
mosabua, line 102: Is that really valid? I kinda think it is not and it would be better to display "null" or "unknown" or something

webapp/src/components/cluster.tsx
```diff
@@ -63,6 +64,24 @@ export function Cluster() {
...
+              statusColor = 'green';
+              break;
+          case 'UNHEALTHY':
+              statusColor = 'red';
+              break;
+          case 'PENDING':
+              statusColor = 'yellow';
+              break;
+          default:
+              statusColor = 'white';
```
mosabua, line 80: When is that used .. and if so .. that would not work with light theme right? .. so maybe get rid of it?

mosabua (commented):
Nice and simple PR really .. just some minor questions and things to sort out.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
prakhar10: When I added a new backend, it failed with a NPE while trying to display the status immediately so I added PENDING as a default status, since I guess that is the status that we are using to indicate a starting cluster.

UNKNOWN crossed my mind as well, if we think that's better then sure I can update this to UNKNOWN

webapp/src/components/cluster.tsx
prakhar10: I kept this as a safety net so that if at all something changes in the status names, the UI does not break.

If we use UNKNOWN then this is how it would look 

Screenshot 2025-01-29 at 8 53 57 AM

mosabua: I think this is much better and more correct.

prakhar10: ok sure I'll update it to UNKNOWN

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
mosabua: Lets go with adding unknown .. probably add it to the TrinoStatus class as the default or so and then use it here as well

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -97,6 +98,12 @@ public Response getAllBackends()
             backendResponse.setName(b.getName());
             backendResponse.setProxyTo(b.getProxyTo());
             backendResponse.setActive(b.isActive());
+            if (Objects.isNull(backendState.trinoStatus())) {
+                backendResponse.setStatus(TrinoStatus.UNKNOWN.toString());
+            }
+            else {
+                backendResponse.setStatus(backendState.trinoStatus().toString());
+            }
```
ebyhr, line 106: Can we set `UNKNOWN` by default in `ClusterStats.Builder`?

prakhar10: Something like this ?

`        private TrinoStatus trinoStatus = TrinoStatus.UNKNOWN;
`

ebyhr: Yes.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/TrinoStatus.java
```diff
@@ -18,10 +18,12 @@
  * We should use PENDING when Trino clusters are still spinning up
  * HEALTHY is when health checks report clusters as up
  * UNHEALTHY is when health checks report clusters as down
+ * UNKNOWN is when the health checks are not able to determine the status
```
mosabua, line 21:
```suggestion
 * UNKNOWN is the intial default state and also used the health checks are not able to determine the status
```

webapp/src/components/cluster.tsx
```diff
@@ -63,6 +64,24 @@ export function Cluster() {
...
+          case 'HEALTHY':
+              statusColor = 'green';
+              break;
+          case 'UNHEALTHY':
+              statusColor = 'red';
+              break;
+          case 'PENDING':
+              statusColor = 'yellow';
+              break;
+          default:
```
mosabua, line 79: Add case UNKNOWN before explicitly and maybe javadoc that we also default to unknown .. of check if it is even possible to get to that case then .. maybe not

mosabua: To clarify after you asked me on slack

something like (pseudo code)

```
case `UNKNOWN`:
  statusColor = 'white';
  break;
default:
  // should not happen really but in case .. defaulting to same as UNKNOWN setup
  statusColor = 'white';
  break;
```

mosabua (approved):
I think this looks good now. Any further comments or suggestions before we merge @ebyhr  @willmostly ?

mosabua:
If we really need another adjustment we can do a follow up PR. Merging.

PR #602, opened 2025-01-27
Fix compilation error

willmostly:
## Description

#540 included changes to test method parameters. A new test using this method was merged after the last CI run of #540, so when #540 was merged the build broke.

## Release notes

```markdown
```


mosabua (approved):
Build passed - code looks ok and good to go.

Just a general thought for this and probably all other tests .. the path used should maybe be the class path rather than the file path .. so for this test I think it would be rules/routing_rules.. 

But I am not sure if we have a Trino test dev guideline or convention for this. Maybe worth looking up, fixing and a chat with @martint

PR #609, opened 2025-02-03
Conditionally skip Oracle tests

willmostly:
The Oracle test container runs very slowly on non x86 architectures, which causes tests to fail. These tests will be skipped unless they are running in Github Actions, on an x86_64, or explicitly enabled by setting the env var TG_RUN_ORACLE_TESTS=true.

## Release notes

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestDatabaseMigrationsOracle.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestDatabaseMigrationsOracle.java
@@ -15,10 +15,10 @@ package io.trino.gateway.ha.persistence;
 
 import com.google.common.collect.ImmutableList;
 import org.jdbi.v3.core.Handle;
-import org.testcontainers.containers.OracleContainer;
 
 import java.util.List;
 
+import static io.trino.gateway.ha.HaGatewayTestUtils.getOracleContainer;
 import static java.lang.String.format;
 
 final class TestDatabaseMigrationsOracle
@@ -26,7 +26,7 @@ final class TestDatabaseMigrationsOracle
 {
     public TestDatabaseMigrationsOracle()
     {
-        super(new OracleContainer("gvenzl/oracle-xe:18.4.0-slim"), "TEST");
+        super(getOracleContainer(), "TEST");
     }
 
     @Override
--- a/docs/development.md
+++ b/docs/development.md
@@ -30,6 +30,10 @@ compilation and testing are specified in `.mvn/jvm.config`.
 Edit the configuration file `gateway-ha-config.yml` in the `gateway-ha` folder
 and update the mysql db information.
 
+Note that tests using Oracle are disabled by default on non-x86_64 CPU architectures.
+To enable them, set the environment variable `TG_RUN_ORACLE_TESTS=true`. These tests
+will always be run in GitHub CI.
+
 ```shell
 cd gateway-ha/target/
 java -jar gateway-ha-{{VERSION}}-jar-with-dependencies.jar ../gateway-ha-config.yml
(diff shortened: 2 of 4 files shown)


willmostly:
Manually verified that the Oracle tests run in CI:
```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 40.74 s -- in io.trino.gateway.ha.router.TestQueryHistoryManagerOracle
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 25.67 s -- in io.trino.gateway.ha.persistence.TestDatabaseMigrationsOracle
```
but are skipped on my M1 Macbook.

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestDatabaseMigrationsOracle.java
```diff
@@ -20,13 +20,23 @@
...
     public TestDatabaseMigrationsOracle()
     {
-        super(new OracleContainer("gvenzl/oracle-xe:18.4.0-slim"), "TEST");
+        super(getOracleContainer(), "TEST");
+    }
+
+    private static OracleContainer getOracleContainer()
+    {
+        // reference: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
+        assumeTrue("true".equals(System.getenv("GITHUB_ACTIONS"))
```
mosabua, line 36: I think we should extract this into a shared method .. truly ugly otherwise

mosabua (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestDatabaseMigrationsOracle.java
mosabua: Also .. we should add this info to the development docs

ebyhr (changes requested):
Please find a different docker image instead.

mosabua:
> Please find a different docker image instead.

Oracle or anyone else at this stage does NOT publish an image that is suitable for ARM. The same problem with the container image exists in Trino itself and we think this is a good approach to allow local testing and development on ARM-based machines .. currently without this approach you have to skip the tests

ebyhr:
That's not true https://blogs.oracle.com/database/post/announcing-oracle-database-23ai-free-container-images-for-armbased-apple-macbook-computers. I would recommend looking into this image's license.

willmostly:
Thanks @ebyhr, I was not aware of that image. The [license](https://www.oracle.com/downloads/licenses/oracle-free-license.html) looks harmless to me, but this isn't my area of expertise. @mosabua thoughts?

willmostly:
Closing in favor of moving to https://blogs.oracle.com/database/post/announcing-oracle-database-23ai-free-container-images-for-armbased-apple-macbook-computers

willmostly:
Closing in favor of moving to https://blogs.oracle.com/database/post/announcing-oracle-database-23ai-free-container-images-for-armbased-apple-macbook-computers

mosabua:
That image is only the latest version of Oracle and from the tutorial it requires an account with the Oracle container registry with a whole lot of restrictions.

I dont know if we can use it with GitHub actions either... we can look into it but for now I would think we should merge this PR after cleaning it up

mosabua:
There is lots more info at https://container-registry.oracle.com/ .. which we have to look at for more info. At first glance I saw no docs for accessing older versions

mosabua:
Also .. maybe @wendigo or someone else knows why we use the container we use here .. I will ask

mosabua (approved):
Looks good now. Thanks @willmostly

ebyhr:
@mosabua I was not convinced with this change. I'm not sure why you merged this PR even I did requested changes.

mosabua:
@ebyhr we discussed this in the dev sync this morning and agreed to merge it for now. It is currently a pain to build locally and it will also potentially cause an issue for the release build that @martint will want to run as soon as we get things ready. As such we decided to merge this and then potentially follow up with adjustments. I still have to get around to write the meeting notes .. hopefully after I get the 470 release notes completed.

ebyhr:
@mosabua I don't think this change is required for "build". Does Martin run all tests during the release process?

mosabua:
They run by default in the normal build. Not sure about release build .. but typically they do run unless we have that deactivate in airbase / the release profile.

Issue #598, opened 2025-01-22
Simplify configuration file name


mosabua:
Discussed in https://github.com/trinodb/trino-gateway/wiki/Contributor-meetings#trino-gateway-dev-sync-22-jan-2025 

Team agreed on `config.yaml`. Task here is rename and files and examples in the code base and the documentation.

PR #612, opened 2025-02-07
Simplify config filename to config.yaml

oneonestar:
## Description
Simplify config filename to `config.yaml`.
Relates to #598 

Docker image still uses `gateway-ha-config.yml`, but changing that requires updates in charts.
That's also a breaking change since user might mount their config file into the image.
I'll do it in another PR.

## Release notes


mosabua (approved):
Looks good.

mosabua (approved):

PR #621, opened 2025-02-18
Update METRICS monitor to configure custom metrics for running and qu…

ricardocardante:
## Description

This PR updates the `METRICS` monitor to allow to configure the metrics that Trino Gateway should rely on to keep track of the number of running and queued queries in a given cluster. Previously the values were configured to strictly track `trino_execution_name_QueryManager_RunningQueries` and `trino_execution_name_QueryManager_QueuedQueries`, but in some scenarios such as in the Starburst Galaxy metrics definition, those metrics don't exist with such naming. As such, while the former are still the default metric names, it is now possible to configure through the `monitor` spec in the YAML configuration.

In addition, all references to the `monitorConfiguration` config node in docs were updated to use `monitor`, as the former does not exist.

## Release notes

```markdown
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -394,6 +394,24 @@ It retrieves the number of running and queued queries for use with
 the `QueryCountBasedRouter` (either `METRICS` or `JDBC` must be enabled if
 `QueryCountBasedRouter` is used).
 
+By default, it uses the `trino_execution_name_QueryManager_RunningQueries` and
+`trino_execution_name_QueryManager_QueuedQueries` to track the number of running
+and queued queries respectively, however these metrics can be configured as follows:
+
+```yaml
+monitor:
+    runningQueriesMetricName: io_starburst_galaxy_name_GalaxyMetrics_RunningQueries
+    queuedQueriesMetricName: io_starburst_galaxy_name_GalaxyMetrics_QueuedQueries
+```
+
+Similarly, by default the monitor pulls the metrics using the `/metrics` endpoint, but it
+can be updated to use another one:
+
+```yaml
+monitor:
+    metricsEndpoint: /v1/metrics
+```
+
 This monitor allows customizing health definitions by comparing metrics to fixed
 values. This is configured through two maps: `metricMinimumValues` and 
 `metricMaximumValues`. The keys of these maps are the metric names, and the values
@@ -401,7 +419,7 @@ are the minimum or maximum values (inclusive) that are considered healthy. By de
 the only metric populated is:
 
 ```yaml
-monitorConfiguration:
+monitor:
     metricMinimumValues:
         trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount: 1
 ```
@@ -412,7 +430,7 @@ worker count to 10 and disqualify clusters that have been experiencing frequent 
 Collections, set
 
 ```yaml
-monitorConfiguration:
+monitor:
     metricMinimumValues:
         trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount: 10
     metricMaximumValues:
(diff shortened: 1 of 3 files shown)


mariafialho (approved):

docs/installation.md
```diff
@@ -394,14 +394,32 @@ It retrieves the number of running and queued queries for use with
...
+
 This monitor allows customizing health definitions by comparing metrics to fixed
 values. This is configured through two maps: `metricMinimumValues` and 
 `metricMaximumValues`. The keys of these maps are the metric names, and the values
 are the minimum or maximum values (inclusive) that are considered healthy. By default,
 the only metric populated is:
 
 ```yaml
-monitorConfiguration:
+monitor:
```
oneonestar, line 422: The change from `monitorConfiguration` -> `monitor` is a fix for the doc. Could you split the fix into another commit?

gateway-ha/src/main/java/io/trino/gateway/ha/config/MonitorConfiguration.java
```diff
@@ -33,7 +33,10 @@ public class MonitorConfiguration
 
     private String metricsEndpoint = "/metrics";
 
-    // Require 1 node for health by default. This configuration only applies to the ClusterStatsMetricsMonitor
```
oneonestar, line 36: Why remove this comment?

docs/installation.md
```diff
@@ -428,21 +446,23 @@ method of backend clusters. Configure a username and password by adding
...
 
+Trino Gateway allows to define the `explicitPrepare` property for the JDBC monitor.
+This property was introduced in Trino 431, and uses a single query for prepared
+statements when set to `false`, instead of a `PREPARE/EXECUTE` pair.
+If you are using Trino 431 or later, you can improve latency for prepared statements
+by setting:
+
+```yaml
+monitor:
+   explicitPrepare: false
```
oneonestar, line 463: IIUC, the description was wrong and remained so after the update.

In Trino JDBC driver (>=431), `explicitPrepare` defaults to `true`. It uses `PREPARE + EXECUTE`.
Setting the property `explicitPrepare` to `false` uses `EXECUTE IMMEDIATE`, which has better performance. 

The correct description should be:
Trino Gateway uses `explicitPrepare=false` by default. This property was introduced
in Trino 431, and uses a single query for prepared statements, instead of a 
`PREPARE/EXECUTE` pair. If you are using the JDBC health check option with older 
versions of Trino, set
```yaml
monitor:
   explicitPrepare: true    # <<<This should be true
```

@willmostly Could you please confirm this?

oneonestar (changes requested):
Please keep logical changes together in a commit.
We should have 2 commits, one for fixing the doc and one for adding custom metrics.
You can ref to https://trino.io/development/process.html#commit

gateway-ha/src/main/java/io/trino/gateway/ha/config/MonitorConfiguration.java
ricardocardante: Initially thought in removing this default min metric, but I did the rollback and overlooked the comment. It is now restored, thanks ✅

docs/installation.md
ricardocardante: Done ✅ https://github.com/trinodb/trino-gateway/pull/626

docs/installation.md
ricardocardante: @oneonestar 

> In Trino JDBC driver (>=431), explicitPrepare defaults to true. It uses PREPARE + EXECUTE.
Setting the property explicitPrepare to false uses EXECUTE IMMEDIATE, which has better performance.

Agreed 💯 

> Trino Gateway uses explicitPrepare=false by default.

I believe that `explicitPrepare` is left unset if not defined in the configuration which thereby defaults to `true` to ensure maximum compatibility.

https://github.com/trinodb/trino-gateway/blob/b6ee885b0ed80d11e556db7bd8ea14c108566676/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java#L56-L60

oneonestar: If not defined in the configuration, by default `explicitPrepare` is `false`.
https://github.com/trinodb/trino-gateway/blob/b6ee885b0ed80d11e556db7bd8ea14c108566676/gateway-ha/src/main/java/io/trino/gateway/ha/config/MonitorConfiguration.java#L32

`monitorConfiguration.isExplicitPrepare()` is `false`.
`!monitorConfiguration.isExplicitPrepare()` is `true`.
`properties.setProperty("explicitPrepare", "false");` is being set.

willmostly (approved):
LGTM, pending CLA approval. Please squash your commits

martint:
@cla-bot check

PR #622, opened 2025-02-19
Remove version property from docker-compose

### LINKED ISSUE #420: Docker compose warning
During release build the following warning appears.

'WARN[0000] /Users/abc/projects/trino/trino-gateway/docker/docker-compose.yml: `version` is obsolete

from the command 

docker/release-docker.sh

oneonestar:
## Description
Fix #420
Remove version property from docker-compose.

See https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-and-name-top-level-elements

## Release notes

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.4"
 services:
   trino-gateway:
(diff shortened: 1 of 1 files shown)


mosabua (approved):
Thanks for picking that up .. I keep forgetting myself

PR #630, opened 2025-02-20
Fix bug preventing the "Disable Health Checks" function from working.

### LINKED ISSUE #362: Disable health check for active backend
Hi, 

I wanted to understand, is there a way to disable health check for identifying active backend and treat all the backend as healthy or pass a custom health check url. 

Regards

### LINKED ISSUE #624: Impossible to use NOOP health check
it is impossible to use NOOP ClusterStatsMonitor without specification BackendStateConfiguration
 
bug in https://github.com/trinodb/trino-gateway/blob/bd2f5135b938a3add217e4dc2b5bc42ba3852ccf/gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java#L226-L228
when BackendStateConfiguration == null trino gw should check monitor type anyway, bit it silently use ClusterStatsInfoApiMonitor

### LINKED ISSUE #625: Trino gateway not works when used NOOP health check
When clusterStatsConfiguration.monitorType = "NOOP" there is no way to make query: occur exception: "HTTP ERROR 500 jakarta.servlet.ServletException: java.lang.IllegalStateException: Number of active backends found zero". This happens because all backends are in an UNKNOWN state.

I expect that when the health check is disabled, Trino should function as if all backends are alive.

oneonestar:
## Description
Fix #362
Fix #624 
Fix #625 

## Release notes

```markdown
* Fix bugs preventing the "Disable Health Checks" function from working.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/NoopClusterStatsMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/NoopClusterStatsMonitor.java
@@ -21,6 +21,11 @@ public class NoopClusterStatsMonitor
     @Override
     public ClusterStats monitor(ProxyBackendConfiguration backend)
     {
-        return ClusterStats.builder(backend.getName()).build();
+        return ClusterStats.builder(backend.getName())
+                .trinoStatus(TrinoStatus.HEALTHY)
+                .proxyTo(backend.getProxyTo())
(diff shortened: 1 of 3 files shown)


willmostly (changes requested):
looks good, will approve once the check for null credentials is added

willmostly (approved):
LGTM

oneonestar:
Thanks for the review

PR #634, opened 2025-02-26
Add release automation

mosabua:
## Description

Lifted and adapted from airlift. Doesnt do anything about the docker container and helm chart yet but that could also be separate steps anyway.

I think we probably have to provision various secrets and also the trino-bot email maybe.

And I have no idea how we would test this.

Do we need a CHANGES file?

## Release notes

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,24 @@
+changelog:
+  exclude:
+    labels:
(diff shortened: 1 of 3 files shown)


mosabua:
We are going to have to switch to Njord and the new Central portal .. so for now I am going to abandon this.

PR #607, opened 2025-02-03
Show RoutingGroup info in query history

andythsu:
## Description

It is useful to show what routing group a request is being routed to because if two routing groups have the same `routedTo` url, it will show the wrong information on the history page

Old:
image

New:
image

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
(X) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #639, opened 2025-03-07
Feature/jyoo/fix duplicate routed to

Chaho12:
## Description

While reviewing Andy’s PR, I noticed some areas that could benefit from improvement.
I couldn't find a better way to handle duplicate proxyTo's in [current logic](https://github.com/trinodb/trino-gateway/blob/c580816aaa45de23986f55f0caaf0e64ba111a8c/webapp/src/components/history.tsx#L31-L36). It uses a map so duplicate entry would overwrite data, causing incorrect matching in query history page.

- https://github.com/trinodb/trino-gateway/pull/607#issuecomment-2702293547

## Release notes

```markdown
* Rename RoutedTo to Name field in Query History page. ({issue}`issuenumber`)
```


docs/quickstart.md
```diff
@@ -84,8 +84,13 @@ JAVA_OPTS="-Dhttp-server.process-forwarded=true"
...
 # Start Trino servers
 for i in 1 2; do
-    docker run --name trino$i -d -p 808$i:8080 \
-        -e JAVA_TOOL_OPTIONS="$JAVA_OPTS" $TRINO_IMAGE
+    PORT=808$i
+    if ! lsof -i:$PORT > /dev/null; then
+        docker run --name trino$i -d -p $PORT:8080 \
+            -e JAVA_TOOL_OPTIONS="$JAVA_OPTS" $TRINO_IMAGE
+    else
+        echo "Port $PORT is already in use. Skipping trino$i."
```
andythsu, line 92: not sure if we wanna skip orchestrating trino here or just let it fail... I personally think letting it fail is better. Or we could programmatically find an open port...

andythsu:
I think the only way to deal with multiple clusters having the same `proxyTo` url is to save the name of the cluster to `query_history` table. It's currently not in there so it's hard to tell. I think `query_history` table should at least have the same schema as `gateway_backend`...

andythsu:
Also, it looks like `ProxyBackendConfiguration` class is a dataclass while `GatewayBackend` is the schema matching the db. The ` proxyTo` and `backend_url` is the same thing, as in this line https://github.com/trinodb/trino-gateway/blob/59c9621c882a8da350b40b0a3c1c1795d0c4842b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java#L130

We should probably change the lingo here, or at least make the variables the same to avoid confusion

webapp/src/components/history.tsx
```diff
@@ -115,7 +115,7 @@ export function History() {
           onPageChange: list,
         }}>
           <Column title="QueryId" dataIndex="queryId" key="queryId" render={linkQueryRender} />
-          <Column title="RoutedTo" dataIndex="backendUrl" key="backendUrlName" render={(text: string) => <Text>{backendMapping[text]}</Text>} />
+          <Column title="Name" dataIndex="backendUrl" key="backendUrlName" render={(text: string) => <Text>{backendMapping[text]}</Text>} />
```
andythsu, line 118: 👍

Chaho12:
> I think the only way to deal with multiple clusters having the same `proxyTo` url is to save the name of the cluster to `query_history` table. It's currently not in there so it's hard to tell. I think `query_history` table should at least have the same schema as `gateway_backend`...

I agree on saving in history, and not make FE decide which is what. 
Btw, proxyTo(or name) was not there until we moved to the new UI. I think this mismatch happened since then.
image

andythsu:
> > I think the only way to deal with multiple clusters having the same `proxyTo` url is to save the name of the cluster to `query_history` table. It's currently not in there so it's hard to tell. I think `query_history` table should at least have the same schema as `gateway_backend`...
> 
> I agree on saving in history, and not make FE decide which is what. Btw, proxyTo(or name) was not there until we moved to the new UI. I think this mismatch happened since then. image

this UI makes me nostalgic

docs/quickstart.md
Chaho12: I thought it would be better off to re-run the script to add backends to gateway if they are running in background. Failing means that i need to stop/kill current trino containers everytime my db container is gone OR manually add them.

mosabua:
Needs a rebase ..

docs/quickstart.md
andythsu: Should we add a Warning in front of the message so it's clear that something is wrong with the port.

Chaho12: Maybe like this? sure
```suggestion
        echo "Warning: Port $PORT is already in use. Skipping trino$i."
```

ebyhr:
@vishalya The commit title "Rename RoutedTo to Name field in Query History page" and the actual changes don't match. We shouldn't have squashed commits in this case. 

Please reach out to @mosabua or other maintainers if you don't understand the review process of `trinodb` organization.

mosabua:
Added as topic for the dev sync this week.

PR #640, opened 2025-03-11
Remove unnecessary db init script from quickstart guide

mosabua:
## Description

Since flyway initializes the db now. As follow up cleanup we should get rid of the sql scripts and their use in TrinoGatewayRunner as well .. but I have to get to that another day.

## Release notes


Chaho12 (approved):
LGTM 👍

PR #654, opened 2025-03-31
Add Header Modification Support to External Routing Service

### LINKED ISSUE #646: Feature: Allow External Routing Service to Set HTTP Headers
Currently, the Trino Gateway's external routing rules engine allows routing decisions based on information sent to an external service, and the service can return a routingGroup and a list of errors.

This issue proposes enhancing the external routing service response to support a mechanism for the service to specify HTTP headers that the Trino Gateway would be able to set on the incoming request before forwarding it to a Trino backend.

**Use Case**:
A primary use case for this feature is to enable more sophisticated routing logic that can also enforce policies through HTTP headers. For example:
* Query Runtime Limiting: An external routing service could determine a query's characteristics and, based on those, set specific session properties via the X-Trino-Session header to enforce runtime limits on the Trino backend. This would allow dynamic policy enforcement at the routing layer.
* Dynamic Session Property Configuration: The routing service could set other session properties based on user, source, or other request attributes, providing a more flexible way to configure query execution context.

**Proposed Response Format**:
I propose extending the JSON response format from the external routing service to include an optional setHeaders field, which would be a JSON object where keys are header names and values are header values. For example:
```
{
  "setHeaders": {
    "X-Trino-Session": "query_max_execution_time=5m",
    "X-Trino-Client-Tags": "limited_runtime"
  },
  "routingGroup": "test-group",
  "errors": []
}
```

**Considerations**:
• Security implications of allowing an external service to set arbitrary headers. The Trino Gateway would need to carefully handle and potentially restrict which headers can be set.
• The interaction with existing header-based routing (e.g., X-Trino-Routing-Group).
• The order of header processing and potential conflicts.

**Benefits**:
This enhancement would provide a more powerful and flexible way to integrate external routing logic with the Trino Gateway, enabling advanced policy enforcement and dynamic configuration based on routing decisions.

RoeyoOgen:
# Add Header Modification Support to External Routing Service

## Description
This PR
Enhances the external routing service to support header modifications in requests. It allows the external routing service to set custom headers that will be applied to the original request before it's forwarded to the target backend.

closes #646 

## Changes
- Added `setHeaders` field to `RoutingGroupExternalResponse` record to support header modifications
- Updated `ExternalRoutingGroupSelector` to handle header modifications from the external service
- Modified `RoutingTargetHandler` to apply the modified headers to the request
- Added comprehensive test coverage for header modification scenarios

## Example Usage
The external routing service can now return headers that should be set in the request:

```json
{
    "routingGroup": "test-group",
    "errors": [],
    "setHeaders": {
        "x-trino-client-tags": ["etl"],
        "x-trino-session": "query_max_memory=50GB,optimize_metadata_queries=false"
    }
}
```

## Testing
Added test cases to verify:
- Header exclusion functionality
- Different types of header values
- Multiple header values
- Error handling with headers
- Edge cases like empty routing groups

## Notes
- Headers are set as request attributes and then applied by the `RoutingTargetHandler`
- The external service can modify any headers except those in the exclude list
- Null header values are ignored

() This is not user-visible or is docs only, and no release notes are required.
( X) Release notes are required. Please propose a release note for me.

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -93,8 +93,10 @@ public class RoutingTargetHandler
     {
         RoutingSelectorResponse routingDestination = routingGroupSelector.findRoutingDestination(request);
         String user = request.getHeader(USER_HEADER);
-        // This falls back on adhoc routing group if there is no cluster found for the routing group.
-        String routingGroup = routingDestination.routingGroup() != null ? routingDestination.routingGroup() : "adhoc";
+        // This falls back on adhoc routing group if there is no cluster found (or value is empty) for the routing group.
+        String routingGroup = (routingDestination.routingGroup() != null && !routingDestination.routingGroup().isEmpty())
+                ? routingDestination.routingGroup()
+                : "adhoc";
         String clusterHost = routingManager.provideClusterForRoutingGroup(routingGroup, user);
         // Apply headers from RoutingDestination if there are any
         HttpServletRequest modifiedRequest = request;
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.ImmutableList.toImmutableList;
@@ -83,12 +84,12 @@ public class RoutingTargetHandler
             return new RoutingTargetResponse(
                     new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request)),
                     request);
-        }).orElse(getClusterFromRoutingGroup(request));
+        }).orElse(getRoutingTargetResponse(request));
         logRewrite(routingTargetResponse.routingDestination().clusterHost(), request);
         return routingTargetResponse;
     }
 
-    private RoutingTargetResponse getClusterFromRoutingGroup(HttpServletRequest request)
+    private RoutingTargetResponse getRoutingTargetResponse(HttpServletRequest request)
     {
         RoutingSelectorResponse routingDestination = routingGroupSelector.findRoutingDestination(request);
         String user = request.getHeader(USER_HEADER);
@@ -152,13 +153,10 @@ public class RoutingTargetHandler
         @Override
         public Enumeration<String> getHeaderNames()
         {
-            List<String> names = Collections.list(super.getHeaderNames());
-            for (String name : customHeaders.keySet()) {
-                if (!names.contains(name)) {
-                    names.add(name);
-                }
-            }
-            return Collections.enumeration(names);
+            return Collections.enumeration(
+                    Stream.concat(Collections.list(super.getHeaderNames()).stream(), customHeaders.keySet().stream())
+                            .distinct()
+                            .toList());
         }
     }
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -17,13 +17,20 @@ import com.google.inject.Inject;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.GatewayCookieConfigurationPropertiesProvider;
 import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.handler.schema.RoutingDestination;
+import io.trino.gateway.ha.handler.schema.RoutingTargetResponse;
 import io.trino.gateway.ha.router.GatewayCookie;
 import io.trino.gateway.ha.router.RoutingGroupSelector;
 import io.trino.gateway.ha.router.RoutingManager;
+import io.trino.gateway.ha.router.schema.RoutingSelectorResponse;
 import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletRequestWrapper;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Pattern;
 
@@ -66,27 +73,36 @@ public class RoutingTargetHandler
         cookiesEnabled = GatewayCookieConfigurationPropertiesProvider.getInstance().isEnabled();
     }
 
-    public RoutingDestination getRoutingDestination(HttpServletRequest request)
+    public RoutingTargetResponse resolveRouting(HttpServletRequest request)
     {
         Optional<String> queryId = extractQueryIdIfPresent(request, statementPaths, requestAnalyserClientsUseV2Format, requestAnalyserMaxBodySize);
         Optional<String> previousCluster = getPreviousCluster(queryId, request);
-        RoutingDestination routingDestination = previousCluster.map(cluster -> {
+        RoutingTargetResponse routingTargetResponse = previousCluster.map(cluster -> {
             String routingGroup = queryId.map(routingManager::findRoutingGroupForQueryId)
                     .orElse("adhoc");
-            return new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request));
+            return new RoutingTargetResponse(
+                    new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request)),
+                    request);
         }).orElse(getClusterFromRoutingGroup(request));
-        logRewrite(routingDestination.clusterHost(), request);
-        return routingDestination;
+        logRewrite(routingTargetResponse.routingDestination().clusterHost(), request);
+        return routingTargetResponse;
     }
 
-    private RoutingDestination getClusterFromRoutingGroup(HttpServletRequest request)
+    private RoutingTargetResponse getClusterFromRoutingGroup(HttpServletRequest request)
     {
-        Optional<String> routingGroup = routingGroupSelector.findRoutingGroup(request);
+        RoutingSelectorResponse routingDestination = routingGroupSelector.findRoutingDestination(request);
         String user = request.getHeader(USER_HEADER);
         // This falls back on adhoc routing group if there is no cluster found for the routing group.
-        String group = routingGroup.orElse("adhoc");
-        String clusterHost = routingManager.provideClusterForRoutingGroup(group, user);
-        return new RoutingDestination(group, clusterHost, buildUriWithNewCluster(clusterHost, request));
+        String routingGroup = routingDestination.routingGroup() != null ? routingDestination.routingGroup() : "adhoc";
+        String clusterHost = routingManager.provideClusterForRoutingGroup(routingGroup, user);
+        // Apply headers from RoutingDestination if there are any
+        HttpServletRequest modifiedRequest = request;
+        if (!routingDestination.externalHeaders().isEmpty()) {
+            modifiedRequest = new HeaderModifyingRequestWrapper(request, routingDestination.externalHeaders());
+        }
+        return new RoutingTargetResponse(
+                new RoutingDestination(routingGroup, clusterHost, buildUriWithNewCluster(clusterHost, request)),
+                modifiedRequest);
     }
 
     public boolean isPathWhiteListed(String path)
@@ -99,6 +115,51 @@ public class RoutingTargetHandler
                 || path.startsWith(UI_API_STATS_PATH)
                 || path.startsWith(OAUTH_PATH)
                 || extraWhitelistPaths.stream().anyMatch(pattern -> pattern.matcher(path).matches());
+    }
+
+    /**
+     * A wrapper for HttpServletRequest that allows modifying multiple headers.
+     */
+    private static class HeaderModifyingRequestWrapper
+            extends HttpServletRequestWrapper
+    {
+        private final Map<String, String> customHeaders;
+
+        public HeaderModifyingRequestWrapper(HttpServletRequest request, Map<String, String> customHeaders)
+        {
+            super(request);
+            this.customHeaders = customHeaders;
+        }
+
+        @Override
+        public String getHeader(String name)
+        {
+            if (customHeaders.containsKey(name)) {
+                return customHeaders.get(name);
+            }
+            return super.getHeader(name);
+        }
+
+        @Override
+        public Enumeration<String> getHeaders(String name)
+        {
+            if (customHeaders.containsKey(name)) {
+                return Collections.enumeration(Collections.singletonList(customHeaders.get(name)));
+            }
+            return super.getHeaders(name);
+        }
+
+        @Override
+        public Enumeration<String> getHeaderNames()
+        {
+            List<String> names = Collections.list(super.getHeaderNames());
+            for (String name : customHeaders.keySet()) {
+                if (!names.contains(name)) {
+                    names.add(name);
+                }
+            }
+            return Collections.enumeration(names);
+        }
     }
 
     private Optional<String> getPreviousCluster(Optional<String> queryId, HttpServletRequest request)
@@ -119,7 +180,6 @@ public class RoutingTargetHandler
                 return Optional.of(cookies.getFirst().getBackend());
             }
         }
-
         return Optional.empty();
     }
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
@@ -25,8 +25,8 @@ import io.airlift.json.JsonCodec;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.RequestAnalyzerConfig;
 import io.trino.gateway.ha.config.RulesExternalConfiguration;
+import io.trino.gateway.ha.router.schema.ExternalRouterResponse;
 import io.trino.gateway.ha.router.schema.RoutingGroupExternalBody;
-import io.trino.gateway.ha.router.schema.RoutingGroupExternalResponse;
 import io.trino.gateway.ha.router.schema.RoutingSelectorResponse;
 import jakarta.servlet.http.HttpServletRequest;
 
@@ -36,7 +36,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import static com.google.common.net.HttpHeaders.CONTENT_TYPE;
 import static com.google.common.net.MediaType.JSON_UTF_8;
@@ -57,8 +56,8 @@ public class ExternalRoutingGroupSelector
     private final RequestAnalyzerConfig requestAnalyzerConfig;
     private final TrinoRequestUser.TrinoRequestUserProvider trinoRequestUserProvider;
     private static final JsonCodec<RoutingGroupExternalBody> ROUTING_GROUP_EXTERNAL_BODY_JSON_CODEC = jsonCodec(RoutingGroupExternalBody.class);
-    private static final JsonResponseHandler<RoutingGroupExternalResponse> ROUTING_GROUP_EXTERNAL_RESPONSE_JSON_RESPONSE_HANDLER =
-            createJsonResponseHandler(jsonCodec(RoutingGroupExternalResponse.class));
+    private static final JsonResponseHandler<ExternalRouterResponse> ROUTING_GROUP_EXTERNAL_RESPONSE_JSON_RESPONSE_HANDLER =
+            createJsonResponseHandler(jsonCodec(ExternalRouterResponse.class));
 
     @VisibleForTesting
     ExternalRoutingGroupSelector(HttpClient httpClient, RulesExternalConfiguration rulesExternalConfiguration, RequestAnalyzerConfig requestAnalyzerConfig)
@@ -95,7 +94,7 @@ public class ExternalRoutingGroupSelector
                     .build();
 
             // Execute the request and get the response
-            RoutingGroupExternalResponse response = httpClient.execute(request, ROUTING_GROUP_EXTERNAL_RESPONSE_JSON_RESPONSE_HANDLER);
+            ExternalRouterResponse response = httpClient.execute(request, ROUTING_GROUP_EXTERNAL_RESPONSE_JSON_RESPONSE_HANDLER);
 
             // Check the response and return the routing group
             if (response == null) {
@@ -110,15 +109,12 @@ public class ExternalRoutingGroupSelector
             if (response.externalHeaders() != null) {
                 response.externalHeaders().forEach((key, value) -> {
                     if (!excludeHeaders.contains(key) && value != null) {
-                        filteredHeaders.put(key, value.toString());
+                        filteredHeaders.put(key, value);
                     }
                 });
                 // Log the headers that will be applied
                 if (!filteredHeaders.isEmpty()) {
-                    String headersStr = filteredHeaders.entrySet().stream()
-                            .map(entry -> entry.getKey() + "=" + entry.getValue())
-                            .collect(Collectors.joining(", "));
-                    log.info("External routing service modified headers to: %s", headersStr);
+                    log.info("External routing service modified headers to: %s", filteredHeaders);
                 }
             }
             return new RoutingSelectorResponse(response.routingGroup(), filteredHeaders);
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
@@ -27,12 +27,16 @@ import io.trino.gateway.ha.config.RequestAnalyzerConfig;
 import io.trino.gateway.ha.config.RulesExternalConfiguration;
 import io.trino.gateway.ha.router.schema.RoutingGroupExternalBody;
 import io.trino.gateway.ha.router.schema.RoutingGroupExternalResponse;
+import io.trino.gateway.ha.router.schema.RoutingSelectorResponse;
 import jakarta.servlet.http.HttpServletRequest;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static com.google.common.net.HttpHeaders.CONTENT_TYPE;
 import static com.google.common.net.MediaType.JSON_UTF_8;
@@ -78,7 +82,7 @@ public class ExternalRoutingGroupSelector
     }
 
     @Override
-    public Optional<String> findRoutingGroup(HttpServletRequest servletRequest)
+    public RoutingSelectorResponse findRoutingDestination(HttpServletRequest servletRequest)
     {
         try {
             RoutingGroupExternalBody requestBody = createRequestBody(servletRequest);
@@ -100,13 +104,30 @@ public class ExternalRoutingGroupSelector
             else if (response.errors() != null && !response.errors().isEmpty()) {
                 throw new RuntimeException("Response with error: " + String.join(", ", response.errors()));
             }
-            return Optional.ofNullable(response.routingGroup());
+
+            // Filter out excluded headers and null values
+            Map<String, String> filteredHeaders = new HashMap<>();
+            if (response.externalHeaders() != null) {
+                response.externalHeaders().forEach((key, value) -> {
+                    if (!excludeHeaders.contains(key) && value != null) {
+                        filteredHeaders.put(key, value.toString());
+                    }
+                });
+                // Log the headers that will be applied
+                if (!filteredHeaders.isEmpty()) {
+                    String headersStr = filteredHeaders.entrySet().stream()
+                            .map(entry -> entry.getKey() + "=" + entry.getValue())
+                            .collect(Collectors.joining(", "));
+                    log.info("External routing service modified headers to: %s", headersStr);
+                }
+            }
+            return new RoutingSelectorResponse(response.routingGroup(), filteredHeaders);
         }
         catch (Exception e) {
             log.error(e, "Error occurred while retrieving routing group "
                     + "from external routing rules processing at " + uri);
         }
-        return Optional.ofNullable(servletRequest.getHeader(ROUTING_GROUP_HEADER));
+        return new RoutingSelectorResponse(servletRequest.getHeader(ROUTING_GROUP_HEADER));
     }
 
     private RoutingGroupExternalBody createRequestBody(HttpServletRequest request)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/ExternalRouterResponse.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/ExternalRouterResponse.java
@@ -13,6 +13,7 @@
  */
 package io.trino.gateway.ha.router.schema;
 
+import com.google.common.collect.ImmutableMap;
 import jakarta.annotation.Nullable;
 
 import java.util.List;
@@ -28,4 +29,9 @@ public record ExternalRouterResponse(
         @Nullable String routingGroup,
         List<String> errors,
         Map<String, String> externalHeaders)
-        implements RoutingGroupResponse {}
+        implements RoutingGroupResponse
+{
+    public ExternalRouterResponse {
+        externalHeaders = ImmutableMap.copyOf(externalHeaders);
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/ExternalRouterResponse.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/ExternalRouterResponse.java
@@ -15,19 +15,17 @@ package io.trino.gateway.ha.router.schema;
 
 import jakarta.annotation.Nullable;
 
-import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 
 /**
- * Response from the routing service that includes:
- * - routingGroup: The target routing group for the request (Optional)
- * - externalHeaders: Headers that can be set in the request (Currently can only be set in ExternalRoutingGroupSelector)
+ * Response from the external routing service that includes:
+ * - routingGroup: The target routing group for the request (optional)
+ * - errors: Any errors that occurred during routing
+ * - externalHeaders: Headers that can be set in the request
  */
-public record RoutingSelectorResponse(@Nullable String routingGroup, Map<String, String> externalHeaders)
-        implements RoutingGroupResponse
-{
-    public RoutingSelectorResponse(String routingGroup)
-    {
-        this(routingGroup, Collections.emptyMap());
-    }
-}
+public record ExternalRouterResponse(
+        @Nullable String routingGroup,
+        List<String> errors,
+        Map<String, String> externalHeaders)
+        implements RoutingGroupResponse {}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupResponse.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupResponse.java
@@ -15,16 +15,20 @@ package io.trino.gateway.ha.router.schema;
 
 import jakarta.annotation.Nullable;
 
-import java.util.List;
 import java.util.Map;
 
 /**
- * Response from the external routing service that includes:
- * - routingGroup: The target routing group for the request (optional)
- * - errors: Any errors that occurred during routing
- * - externalHeaders: Headers that can be set in the request
+ Interface representing the response from a routing group selector.
+ This interface defines the contract for responses that determine how requests should be routed within
+    the Trino Gateway system.
+
+ Implementations of this interface are used to:
+    * Specify the target routing group for a request
+    * Provide additional headers that should be added to the request
  */
-public record RoutingGroupExternalResponse(
-        @Nullable String routingGroup,
-        List<String> errors,
-        Map<String, Object> externalHeaders){}
+public interface RoutingGroupResponse
+{
+    @Nullable String routingGroup();
+
+    Map<String, String> externalHeaders();
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
@@ -15,16 +15,20 @@ package io.trino.gateway.ha.router.schema;
 
 import jakarta.annotation.Nullable;
 
-import java.util.List;
+import java.util.Collections;
 import java.util.Map;
 
 /**
- * Response from the external routing service that includes:
- * - routingGroup: The target routing group for the request (optional)
- * - errors: Any errors that occurred during routing
- * - externalHeaders: Headers that can be set in the request
+ * Response from the routing service that includes:
+ * - routingGroup: The target routing group for the request (Optional)
+ * - externalHeaders: Headers that can be set in the request (Currently can only be set in ExternalRoutingGroupSelector)
  */
-public record RoutingGroupExternalResponse(
+public record RoutingSelectorResponse(
         @Nullable String routingGroup,
-        List<String> errors,
-        Map<String, Object> externalHeaders){}
+        Map<String, String> externalHeaders)
+{
+    public RoutingSelectorResponse(String routingGroup)
+    {
+        this(routingGroup, Collections.emptyMap());
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
@@ -13,9 +13,9 @@
  */
 package io.trino.gateway.ha.router.schema;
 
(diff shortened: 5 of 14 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,6 +100,18 @@ public String findRoutingGroup(HttpServletRequest servletRequest)
...
+            if (response.setHeaders() != null) {
+                response.setHeaders().forEach((key, value) -> {
+                    if (!excludeHeaders.contains(key) && value != null) {
+                        // Store the header in a request attribute
+                        // This will be used by the RoutingTargetHandler to set the actual header
+                        servletRequest.setAttribute("MODIFIED_HEADER_" + key, value);
+                    }
+                });
+            }
+
```
andythsu, line 114: instead of modifying the `setvletRequest`, maybe we can change the return type to `Optional<RoutingGroupExternalResponse>`? I have a [PR open](https://github.com/trinodb/trino-gateway/pull/607/files#diff-09f965fbf93971267d5b5aca0b85fc75a90ed23e45f90d06af0ef8f1adac6185R74) that is doing similar things, except I'm trying to recourd `RoutingGroup` but you are trying to get extra headers

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -446,6 +449,11 @@ public Response readExactMatchSourceSelector()
     public Response getRoutingRules()
             throws IOException
     {
+        if (configuration.getRoutingRules().isRulesEngineEnabled() && configuration.getRoutingRules().getRulesType() == RulesType.EXTERNAL) {
+            return Response.status(Response.Status.NO_CONTENT)
+                    .entity(Result.fail("Routing rules are managed by an external service"))
```
andythsu, line 454: instead of failing, do we want to "enforce" external service to expose an endpoint where we can query routing rules? I think it might be useful

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -66,13 +70,28 @@ public RoutingTargetHandler(
...
 
-    public String getRoutingDestination(HttpServletRequest request)
+    public RoutingResult resolveRouting(HttpServletRequest request)
     {
         Optional<String> previousBackend = getPreviousBackend(request);
         String clusterHost = previousBackend.orElseGet(() -> getBackendFromRoutingGroup(request));
         logRewrite(clusterHost, request);
 
-        return buildUriWithNewBackend(clusterHost, request);
+        // Apply headers from ExternalRoutingGroupSelector
```
andythsu, line 79: This will fail for subsequent calls to the coordinator if `getPreviousBackend` resolves. So only the first request will have headers modified. I think we want to apply modified headers to all requests to coordinator right?

Chaho12: As Andy mentioned, it's more complicated because requests for the same query (same query ID) are routed to the same backend before going through the external routing service. See the [code](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L69-L76).
Following this, have you tested querying a query and confirmed that all requests have the headers applied?

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupExternalResponse.java
```diff
@@ -14,5 +14,15 @@
...
+/**
+ * Response from the external routing service that includes:
+ * - routingGroup: The target routing group for the request
+ * - errors: Any errors that occurred during routing
+ * - setHeaders: Headers that should be set in the request
+ */
+public record RoutingGroupExternalResponse(
+        String routingGroup,
+        List<String> errors,
+        Map<String, Object> setHeaders) {}
```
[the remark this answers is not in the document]
RoeyoOgen: Why misleading? these are headers set by the external service. 'extra' might actually be confusing as these can override existing ones and not just add extra.

How about "transformedHeaders" or "enrichedHeaders"?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
RoeyoOgen: If I understood correctly, I don't think this is possible, as who says the external service has a set of rules which it can expose? 
I mean the rules can be really complex code, how will you be able to query them?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
RoeyoOgen: I haven't, will test and fix if needed. 
(missed this as In my specific case I'd want all queries to go to the external service and not use the cache)

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
```diff
@@ -76,6 +77,7 @@ public class GatewayWebAppResource
     // TODO Avoid putting mutable objects in fields
     private final UIConfiguration uiConfiguration;
     private final RoutingRulesManager routingRulesManager;
+    private final HaGatewayConfiguration configuration;
```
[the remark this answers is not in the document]
RoeyoOgen: done in 
my bad , this should be addressed in [https://github.com/trinodb/trino-gateway/pull/660](https://github.com/trinodb/trino-gateway/pull/url)

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
andythsu: I think it's just better to keep the original request without modifying it, if possible. @Chaho12 @vishalya what do you think?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
andythsu: I'm guessing headers won't be applied to subsequent calls to coordinator...

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupExternalResponse.java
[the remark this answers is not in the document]
andythsu: what about `externalHeaders`?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
RoeyoOgen: weird, 
this whole file shouldn't be here (when i split the CR per Jaeho's request i missed this)

the correct link is: #660  
i can add this comment there and resolve here...

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
RoeyoOgen: I can defenitly do this by addding them to the cache (like the routing group is saved) do we actually need them though?

attaching an image which @xkrogen made:
```mermaid
sequenceDiagram
    User->>+Client: "SELECT a FROM tbl"

    Client->>+GW: POST /v1/statement "SELECT a FROM tbl"
    GW->>+Backend: POST /v1/statement "SELECT a FROM tbl"
    Backend->>-GW: {"id": "123", "state": "QUEUED", "nextUri": "/v1/statement/queued/123/xxx/1"}
    GW->>Database: INSERT {"queryID": "123", pageID: "POST", content: "..."}
    GW->>-Client: {"id": "123", "state": "QUEUED", "nextUri": "/v1/statement/queued/123/xxx/1"}

    Client->>+GW: GET /v1/statement/queued/123/xxx/1
    GW->>+Backend: GET /v1/statement/queued/123/xxx/1
    Backend->>-GW: {"id": "123", "state": "RUNNING", "nextUri": "/v1/statement/executing/123/xxx/0"}
    GW->>Database: INSERT {"queryID": "123", pageID: "queued/1", content: "..."}
    GW->>-Client: {"id": "123", "state": "RUNNING", "nextUri": "/v1/statement/executing/123/xxx/0"}

    Client->>+GW: GET /v1/statement/executing/123/xxx/0
    GW->>+Backend: GET /v1/statement/executing/123/xxx/0
    Backend->>-GW: {"id": "123", "state": "RUNNING", "nextUri": "/v1/statement/executing/123/xxx/1", "data": ["a", "b"]}
    GW->>Database: INSERT {"queryID": "123", pageID: "executing/0", content: "..."}
    GW->>-Client: {"id": "123", "state": "RUNNING", "nextUri": "/v1/statement/executing/123/xxx/1", "data": ["a", "b"]}

    Client->>+GW: GET /v1/statement/executing/123/xxx/1
    GW->>+Backend: GET /v1/statement/executing/123/xxx/1
    Backend->>-GW: {"id": "123", "state": "FINISHED", "nextUri": "/v1/statement/executing/123/xxx/2", "data": ["c", "d"]}
    GW->>Database: INSERT {"queryID": "123", pageID: "executing/1", content: "..."}
    GW->>-Client: {"id": "123", "state": "FINISHED", "nextUri": "/v1/statement/executing/123/xxx/2", "data": ["c", "d"]}

    Client->>+GW: GET /v1/statement/executing/123/xxx/2
    GW->>+Backend: GET /v1/statement/executing/123/xxx/2
    Backend->>-GW: {"id": "123", "state": "FINISHED"}
    GW->>Database: INSERT {"queryID": "123", pageID: "executing/2", content: "..."}
    GW->>Database: INSERT {"queryID": "123", completed: TRUE}
    GW->>-Client: {"id": "123", "state": "FINISHED"}

    Client->>-User: ["a", "b", "c", "d"]


    User->>+Client: "SELECT a FROM tbl"

    Client->>+GW: POST /v1/statement "SELECT a FROM tbl"
    GW->>+Database: SELECT {"queryID": "123", pageID: "POST"}
    Database->>-GW: content: {...}
    GW->>-Client: {"id": "123", "state": "QUEUED", "nextUri": "/v1/statement/queued/123/xxx/1"}

    Client->>+GW: GET /v1/statement/queued/123/xxx/1
    GW->>+Database: SELECT {"queryID": "123", pageID: "queued/1"}
    Database->>-GW: content: {...}
    GW->>-Client: {"id": "123", "state": "RUNNING", "nextUri": "/v1/statement/executing/123/xxx/0"}

    Client->>+GW: GET /v1/statement/executing/123/xxx/0
    GW->>+Database: SELECT {"queryID": "123", pageID: "executing/0"}
    Database->>-GW: content: {...}
    GW->>-Client: {"id": "123", "state": "RUNNING", "nextUri": "/v1/statement/executing/123/xxx/1", "data": ["a", "b"]}

    Client->>+GW: GET /v1/statement/executing/123/xxx/1
    GW->>+Database: SELECT {"queryID": "123", pageID: "executing/1"}
    Database->>-GW: content: {...}
    GW->>-Client: {"id": "123", "state": "FINISHED", "nextUri": "/v1/statement/executing/123/xxx/2", "data": ["c", "d"]}

    Client->>+GW: GET /v1/statement/executing/123/xxx/2
    GW->>+Database: SELECT {"queryID": "123", pageID: "executing/2"}
    Database->>-GW: content: {...}
    GW->>-Client: {"id": "123", "state": "FINISHED"}

    Client->>-User: ["a", "b", "c", "d"]
```

if the header modification is only for the actual query **POST** command which is forwarded to the relevant trino cluster. so why would we care if the rest of the GET commands dont include the headers? they will have no effect on the allready running query...?

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupExternalResponse.java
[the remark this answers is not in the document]
RoeyoOgen: done, changed setHeaders to externalHeaders

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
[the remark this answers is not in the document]
RoeyoOgen: removed GatewayWebAppResource changes

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
RoeyoOgen: removed GatewayWebAppResource changes

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
andythsu: Are you planning to set the headers for Trino? If so, then that's fine. I was assuming the headers were for all requests and hence the confusion. If the headers are meant for Trino only, then I think it's important enough to mention somewhere in the doc that the headers will be sent only once to the coordinator initially

RoeyoOgen: Yeah, for my case it's just for trino. If anyone needs anything else I'll be happy to add 😀 

I'll add to the docs.

andythsu: Thanks! I'm fine with this PR as long as we have a decision for https://github.com/trinodb/trino-gateway/pull/654#discussion_r2025095799

RoeyoOgen: Talked to @vishalya he'll take a look in the next two days

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,6 +100,18 @@ public String findRoutingGroup(HttpServletRequest servletRequest)
             else if (response.errors() != null && !response.errors().isEmpty()) {
                 throw new RuntimeException("Response with error: " + String.join(", ", response.errors()));
             }
+
+            // Apply headers from response if not in excludeHeaders
+            if (response.externalHeaders() != null) {
```
vishalya, line 105: Why not add a copy of the externalHeaders() to the servletRequest? so that could be used effectively in the class `HttpServletRequestWrapper` where you can override the values. Separating out individual header doesn't offer any advantage.

gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouteToBackendResource.java
```diff
@@ -65,26 +66,26 @@ public void postHandler(
...
     }
 
     @GET
     public void getHandler(
             @Context HttpServletRequest servletRequest,
             @Suspended AsyncResponse asyncResponse)
     {
-        String remoteUri = routingTargetHandler.getRoutingDestination(servletRequest);
-        proxyRequestHandler.getRequest(servletRequest, asyncResponse, URI.create(remoteUri));
+        RoutingResult result = routingTargetHandler.resolveRouting(servletRequest);
```
vishalya, line 78: Can we keep the original method with it's signature and logic as is? It's not suppose to modify a request, instead add a new method in this `RouteToBackendResource` class which would provide the modified headers. You can move the HeaderModifyingRequestWrapper and it's related logic in this class. `routingTargetHandler` should just do the routing.
RouteToBackendResouce class sits a bit further down from where we send the request to the backend trino service which would be a better place have the logic to modify the request.

vishalya (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
RoeyoOgen: Changed to be more similar to #607 @andythsu 
and added a log @Chaho12

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -66,13 +70,28 @@ public RoutingTargetHandler(
...
+        HttpServletRequest modifiedRequest = request;
+        Enumeration<String> attributeNames = request.getAttributeNames();
+        while (attributeNames.hasMoreElements()) {
+            String attributeName = attributeNames.nextElement();
+            if (attributeName.startsWith("MODIFIED_HEADER_")) {
+                String headerName = attributeName.substring("MODIFIED_HEADER_".length());
+                Object headerValue = request.getAttribute(attributeName);
+                if (headerValue != null) {
+                    // Create a wrapper request that includes the modified header
+                    modifiedRequest = new HeaderModifyingRequestWrapper(modifiedRequest, headerName, headerValue);
```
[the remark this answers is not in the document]
RoeyoOgen: changed after @andythsu 's comment above

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
RoeyoOgen: @andythsu suggested to return a model instead, like will suggested to him in #607 

so I did like he suggested....

gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouteToBackendResource.java
RoeyoOgen: same as above, 

@andythsu suggested to return a model instead, like will suggested to him in #607 

so I did like he suggested....

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,13 +104,30 @@ public String findRoutingGroup(HttpServletRequest servletRequest)
...
+                    if (!excludeHeaders.contains(key) && value != null) {
+                        filteredHeaders.put(key, value.toString());
+                    }
+                });
+                // Log the headers that will be applied
+                if (!filteredHeaders.isEmpty()) {
+                    String headersStr = filteredHeaders.entrySet().stream()
+                            .map(entry -> entry.getKey() + "=" + entry.getValue())
+                            .collect(Collectors.joining(", "));
+                    log.info("External routing service modified headers to: %s", headersStr);
```
Chaho12, line 121: Why not simply put line 121 after line 113?

RoeyoOgen: This will make a separate log for each change instead of one containing all of them.

andythsu:
Could you merge main? It'll probably be more straightforward

Chaho12:
I think you can squash commits too

RoeyoOgen:
@Chaho12 @andythsu @vishalya 
pushed changes after pulling #607 
I'd appreciate if you can resolve comments and approve the logic. 
After that I'll add tests and we can get this merged :)

vishalya (approved):
Approved!

RoeyoOgen:
@Chaho12 tagging you again :)

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,13 +104,30 @@ public Optional<String> findRoutingGroup(HttpServletRequest servletRequest)
...
+
+            // Filter out excluded headers and null values
+            Map<String, String> filteredHeaders = new HashMap<>();
+            if (response.externalHeaders() != null) {
+                response.externalHeaders().forEach((key, value) -> {
+                    if (!excludeHeaders.contains(key) && value != null) {
+                        filteredHeaders.put(key, value.toString());
+                    }
+                });
+                // Log the headers that will be applied
```
andythsu, line 116: I think you can just do 
`log.info("External routing service modified headers to: %s", filteredHeaders.toString());` 
ref: https://github.com/openjdk/jdk/blob/86f9b3f52a0675be4dd8096da0c65d6bda442f7b/src/java.base/share/classes/java/util/AbstractMap.java#L516-L533

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupExternalResponse.java
```diff
@@ -13,6 +13,18 @@
...
+/**
+ * Response from the external routing service that includes:
+ * - routingGroup: The target routing group for the request (optional)
+ * - errors: Any errors that occurred during routing
+ * - externalHeaders: Headers that can be set in the request
+ */
+public record RoutingGroupExternalResponse(
+        @Nullable String routingGroup,
+        List<String> errors,
+        Map<String, Object> externalHeaders){}
```
andythsu, line 30: Is there why the value will be `Object`? I assume we are only supporting `string:string` pair

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,13 +104,30 @@ public Optional<String> findRoutingGroup(HttpServletRequest servletRequest)
...
                 throw new RuntimeException("Response with error: " + String.join(", ", response.errors()));
             }
-            return Optional.ofNullable(response.routingGroup());
+
+            // Filter out excluded headers and null values
+            Map<String, String> filteredHeaders = new HashMap<>();
+            if (response.externalHeaders() != null) {
+                response.externalHeaders().forEach((key, value) -> {
+                    if (!excludeHeaders.contains(key) && value != null) {
+                        filteredHeaders.put(key, value.toString());
```
andythsu, line 113: following up on previous comment, if value has type of `Object`, then `toString()` may have bad side effects depending on the actual implementation of `value`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupExternalResponse.java
andythsu: According to a blog post from DigitalOcean, header's value should always be a string https://docs.digitalocean.com/products/functions/reference/http-headers/#other-headers

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -66,27 +73,36 @@ public RoutingTargetHandler(
...
         Optional<String> previousCluster = getPreviousCluster(queryId, request);
-        RoutingDestination routingDestination = previousCluster.map(cluster -> {
+        RoutingTargetResponse routingTargetResponse = previousCluster.map(cluster -> {
             String routingGroup = queryId.map(routingManager::findRoutingGroupForQueryId)
                     .orElse("adhoc");
-            return new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request));
+            return new RoutingTargetResponse(
+                    new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request)),
+                    request);
         }).orElse(getClusterFromRoutingGroup(request));
```
andythsu, line 86: let's change this name as it does more things now. Maybe something like `getRoutingTargetResponse`?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -101,6 +117,51 @@ public boolean isPathWhiteListed(String path)
...
+        public Enumeration<String> getHeaders(String name)
+        {
+            if (customHeaders.containsKey(name)) {
+                return Collections.enumeration(Collections.singletonList(customHeaders.get(name)));
+            }
+            return super.getHeaders(name);
+        }
+
+        @Override
+        public Enumeration<String> getHeaderNames()
```
andythsu, line 153: nit:
```
return Collections.enumeration(
    Stream.concat(Collections.list(super.getHeaderNames()).stream(), customHeaders.keySet().stream())
          .distinct()
          .toList()
);
```
faster performance :)

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,13 +104,30 @@ public Optional<String> findRoutingGroup(HttpServletRequest servletRequest)
...
+                });
+                // Log the headers that will be applied
+                if (!filteredHeaders.isEmpty()) {
+                    String headersStr = filteredHeaders.entrySet().stream()
+                            .map(entry -> entry.getKey() + "=" + entry.getValue())
+                            .collect(Collectors.joining(", "));
+                    log.info("External routing service modified headers to: %s", headersStr);
+                }
+            }
+            return new RoutingSelectorResponse(response.routingGroup(), filteredHeaders);
```
andythsu, line 124: I noticed you are only adding `filteredHeaders`, thereby letting the external service manage the entire header going to trino cluster. Is this the intended behavior? I thought we are "merging" original headers and external headers.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
RoeyoOgen: done :)

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupExternalResponse.java
RoeyoOgen: Just my confusion with fastapi reponse model (forgot it modifies to JSON in background). changed to str.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
RoeyoOgen: removed.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
RoeyoOgen: done

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
RoeyoOgen: changed, cheers!

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
RoeyoOgen: Nope, works fine :)

1. In `ExternalRoutingGroupSelector.java`, it:
   - Gets original headers from the request
   - Filters out excluded headers
   - Sends request to external service
   - Returns only the filtered headers from external service response

2. In `RoutingTargetHandler.java`, when processing the response:

```java
HttpServletRequest modifiedRequest = request;
if (!routingDestination.externalHeaders().isEmpty()) {
    modifiedRequest = new HeaderModifyingRequestWrapper(request, routingDestination.externalHeaders());
}
```
Thereafter HeaderModifyingRequest is sent to the cluster where the behavior (of HeaderModifyingRequestWrapper) is:
- If a header exists in customHeaders (from external service), it takes precedence over the original header
- If a header doesn't exist in customHeaders, the original header is used

The final set of headers includes both original headers and custom headers, with custom headers overriding original ones when there's a conflict

So bottom line, the HeaderModifyingRequestWrapper does merge both sets of headers, with external headers taking precedence over original headers when there's a conflict.

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Map;
+
+/**
+ * Response from the routing service that includes:
+ * - routingGroup: The target routing group for the request (Optional)
+ * - externalHeaders: Headers that can be set in the request (Currently can only be set in ExternalRoutingGroupSelector)
+ */
+public record RoutingSelectorResponse(@Nullable String routingGroup, Map<String, String> externalHeaders)
+        implements RoutingGroupResponse
+{
```
ebyhr, line 28: Add a compact constructor with `ImmutableMap.copyOf`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
```diff
@@ -0,0 +1,33 @@
...
+    public RoutingSelectorResponse(String routingGroup)
+    {
+        this(routingGroup, Collections.emptyMap());
```
ebyhr, line 31: Use `ImmutableMap.of()` instead.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -119,7 +178,6 @@ private Optional<String> getPreviousCluster(Optional<String> queryId, HttpServle
                 return Optional.of(cookies.getFirst().getBackend());
             }
         }
-
```
ebyhr, line 122: Revert unrelated change.

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,13 +103,27 @@ public Optional<String> findRoutingGroup(HttpServletRequest servletRequest)
...
+            Map<String, String> filteredHeaders = new HashMap<>();
+            if (response.externalHeaders() != null) {
+                response.externalHeaders().forEach((key, value) -> {
+                    if (!excludeHeaders.contains(key) && value != null) {
+                        filteredHeaders.put(key, value);
+                    }
+                });
+                // Log the headers that will be applied
+                if (!filteredHeaders.isEmpty()) {
+                    log.info("External routing service modified headers to: %s", filteredHeaders);
```
ebyhr, line 117: Why does this log level need to be info instead of debug?

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingGroupResponse.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ Interface representing the response from a routing group selector.
+ This interface defines the contract for responses that determine how requests should be routed within
+    the Trino Gateway system.
+
+ Implementations of this interface are used to:
+    * Specify the target routing group for a request
+    * Provide additional headers that should be added to the request
+ */
+public interface RoutingGroupResponse
```
ebyhr, line 29: Add permits for ExternalRouterResponse and RoutingSelectorResponse

andythsu:
Overall I think it LGTM. Could we add some test cases?

RoeyoOgen:
> Overall I think it LGTM. Could we add some test cases?

doing this tomorrow morning and will ping you!

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
RoeyoOgen: done

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
RoeyoOgen: no changes done here

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
RoeyoOgen: why not? is it not info?

RoeyoOgen:
@andythsu pining you

andythsu (approved):
Left some comments to be addressed. Otherwise LGTM

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
```diff
@@ -196,6 +198,122 @@ void testExcludeHeader()
...
+        // Execute
+        RoutingSelectorResponse routingSelectorResponse = selector.findRoutingDestination(mockRequest);
+
+        // Verify
+        assertThat(routingSelectorResponse.routingGroup()).isNull();
+        assertThat(routingSelectorResponse.externalHeaders().get(headerKey)).isNull();
+    }
+
+    @Test
+    void testHeaderModificationWithEmptyRoutingGroup()
```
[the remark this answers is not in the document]
RoeyoOgen: I think there is:
this comes to test `assertThat(routingSelectorResponse.routingGroup()).isEmpty()` as it may be valid for the router not to return a routing group.

then as you said resolveRouting will put adhock.

also added tests for RoutingTargetHandler -- TestRoutingTargetHandler() which you can check out :)

gateway-ha/src/main/java/io/trino/gateway/ha/router/ExternalRoutingGroupSelector.java
```diff
@@ -100,13 +103,27 @@ public Optional<String> findRoutingGroup(HttpServletRequest servletRequest)
             else if (response.errors() != null && !response.errors().isEmpty()) {
                 throw new RuntimeException("Response with error: " + String.join(", ", response.errors()));
```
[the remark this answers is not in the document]
RoeyoOgen: at the end of the function this is caught and logged, and I kept the same behaviour of `return new RoutingSelectorResponse(servletRequest.getHeader(ROUTING_GROUP_HEADER))`:


```
catch (Exception e) {
            log.error(e, "Error occurred while retrieving routing group "
                    + "from external routing rules processing at " + uri);
        }
        return new RoutingSelectorResponse(servletRequest.getHeader(ROUTING_GROUP_HEADER));
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
```diff
@@ -196,6 +198,122 @@ void testExcludeHeader()
...
+                ImmutableList.of("Error occurred"),
+                ImmutableMap.of(headerKey, headerValue));
+
+        when(httpClient.execute(any(), any())).thenReturn(mockResponse);
+
+        // Execute
+        RoutingSelectorResponse routingSelectorResponse = selector.findRoutingDestination(mockRequest);
+
+        // Verify
+        assertThat(routingSelectorResponse.routingGroup()).isNull();
```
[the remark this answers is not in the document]
RoeyoOgen: it isn't see above :)

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
andythsu: I think Yuya meant to say revert the line break change here

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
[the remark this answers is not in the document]
andythsu: I see it now!

andythsu (approved):
LGTM!

Chaho12 (approved):
LGTM. Just don't forget to squash commits + messages

RoeyoOgen:
@vishalya @ebyhr LGTM :)

ebyhr:
Please follow the commit message guideline: https://trino.io/development/process.html#pull-request-and-commit-guidelines-

ebyhr:
@vishalya Please make sure the commit granularity before merging PR.

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/ExternalRouterResponse.java
```diff
@@ -0,0 +1,37 @@
...
+ * - externalHeaders: Headers that can be set in the request
+ */
+public record ExternalRouterResponse(
+        @Nullable String routingGroup,
+        List<String> errors,
+        Map<String, String> externalHeaders)
+        implements RoutingGroupResponse
+{
+    public ExternalRouterResponse {
+        externalHeaders = ImmutableMap.copyOf(externalHeaders);
```
ebyhr, line 35: Copy for `errors` is missing.

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import jakarta.annotation.Nullable;
+
+import java.util.Map;
+
+/**
+ * Response from the routing service that includes:
+ * - routingGroup: The target routing group for the request (Optional)
+ * - externalHeaders: Headers that can be set in the request (Currently can only be set in ExternalRoutingGroupSelector)
```
ebyhr, line 24: Use `@param`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/RoutingSelectorResponse.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public record RoutingSelectorResponse(@Nullable String routingGroup, Map<String, String> externalHeaders)
+        implements RoutingGroupResponse
+{
+    public RoutingSelectorResponse {
```
ebyhr, line 29: Move { to new line.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestRoutingTargetHandler.java
```diff
@@ -0,0 +1,246 @@
...
+class TestRoutingTargetHandler
+{
+    private RoutingManager routingManager;
+    private HttpClient httpClient;
+    private HttpServletRequest request;
+
+    private RoutingTargetHandler handler;
+    private HaGatewayConfiguration config;
+
+    static HaGatewayConfiguration provideGatewayConfiguration()
```
ebyhr, line 57: Helper method should exist after the caller.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestRoutingTargetHandler.java
```diff
@@ -0,0 +1,246 @@
...
+import static io.trino.gateway.ha.handler.HttpUtils.USER_HEADER;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class TestRoutingTargetHandler
+{
+    private RoutingManager routingManager;
```
ebyhr, line 50: Change to a local variable.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestRoutingTargetHandler.java
```diff
@@ -0,0 +1,246 @@
...
+    private HttpClient httpClient;
+    private HttpServletRequest request;
+
+    private RoutingTargetHandler handler;
+    private HaGatewayConfiguration config;
```
ebyhr, line 55: Change to a local variable.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestRoutingTargetHandler.java
```diff
@@ -0,0 +1,246 @@
...
+        routingManager = Mockito.mock(RoutingManager.class);
+        request = prepareMockRequest();
+
+        // Initialize the handler with the configuration
+        handler = new RoutingTargetHandler(routingManager, RoutingGroupSelector.byRoutingExternal(httpClient, config.getRoutingRules().getRulesExternalConfiguration(), config.getRequestAnalyzerConfig()), config);
+    }
+
+    @Test
+    void testBasicHeaderModification()
+            throws Exception
```
ebyhr, line 115: Remove a redundant throws.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestRoutingTargetHandler.java
```diff
@@ -0,0 +1,246 @@
...
+import java.util.Map;
+
+import static io.trino.gateway.ha.handler.HttpUtils.USER_HEADER;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class TestRoutingTargetHandler
```
ebyhr, line 48: https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
> Test classes should be defined as package-private and final.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingGroupSelectorExternal.java
```diff
@@ -196,6 +198,122 @@ void testExcludeHeader()
...
+                ImmutableMap.of(headerKey, headerValue));
+
+        when(httpClient.execute(any(), any())).thenReturn(mockResponse);
+
+        // Execute
+        RoutingSelectorResponse routingSelectorResponse = selector.findRoutingDestination(mockRequest);
+
+        // Verify
+        assertThat(routingSelectorResponse.routingGroup()).isEmpty();
+        assertThat(routingSelectorResponse.externalHeaders().get(headerKey)).isNotNull().isEqualTo(headerValue);
```
ebyhr, line 314: Use `containsEntry()` method instead.

Issue #649, opened 2025-03-19
Proposal to improve telemetry of Trino Gateway by tracking cluster activation status


amybubu:
### Problem ###
Currently, there is no built-in metric within the Trino Gateway that consistently tracks the activation status of Trino clusters behind it. This means we have no historical data of what state the clusters are/have been in for continuous periods of time. Without this visibility, prolonged outages may go unnoticed, leading to delayed mitigation and reduced availability. An activation status metric on the Gateway would provide crucial insight into cluster behavior, enhancing both auditing and large-scale cluster management.

### Motivation ###

As a backend level metric, this data would address a gap in the open-source Trino Gateway. The OSS metrics currently emitted by airlift primarily covers JVM and request level information. An activation status metric would introduce higher visibility into our backend clusters through the gateway, without needing to make an API call or look at the database table directly.  

![Image](https://github.com/user-attachments/assets/196d3bc4-e535-43db-9a2f-787f7b3e5f19)

At LinkedIn, we introduced an activation status metric into our Trino Gateway service and it’s become an essential tool for maintaining availability at scale. By integrating this metric into our alerting system, we receive timely notifications whenever a cluster remains deactivated for an unexpected amount of time. This has proved very beneficial to us, as we operate Trino at a large scale, serving thousands of weekly active users, executing over 5.5 million queries per week. Serving this amount of traffic requires operating 10+ clusters behind Trino Gateway. Additionally, LinkedIn fleet management infrastructure automatically deactivates and reactivates our clusters during deployment and maintenance procedures to maintain high availability, but that means we are not always aware when a cluster is being taken down. The alert has helped bring our attention to any failed automatic procedures that have unintentionally left clusters in a deactivated state, reducing time of mitigation from potentially days to just hours.

The metric's data has come in handy when investigating failing clusters, uncovering bugs, and enabling faster rollbacks. It has even simply reminded us to manually bring back up clusters when we forgot to re-activate after mitigation. No matter if it’s human error or an infrastructure glitch, this metric has helped us catch the issue early on, which is crucial as such capacity shortage means more queries have to be queued on the remaining clusters, slowing down user experience. We’ve also combined this metric with other alerts to improve operational excellence, and added logs that capture activation/deactivation calls, providing a clear paper trail for historical analysis and troubleshooting. 
[...]

PR #672, opened 2025-04-15
Improve logging for backend cluster (de)activation

amybubu:
## Description
As apart of efforts to improve telemetry as described in Issue https://github.com/trinodb/trino-gateway/issues/649, this PR adds logging for whenever a backend cluster is activated/deactivated. 

## Testing
- built and ran locally
- confirmed logs appeared when calling de(activate) via curl call or changing deactivation status via the UI, e.g.:
```
2025-04-03T17:09:03.588-0700	INFO	http-worker-87	io.trino.gateway.ha.router.HaGatewayManager	Backend cluster trino-3 activation status changed to active=false (previous status: active=true).
```
- confirm error thrown if try to activate/deactivate non-existent backend
```
java.lang.IllegalStateException: No backend found with name: trino-4, could not activate
```

## Release notes

```markdown
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
@@ -86,7 +86,7 @@ public class GatewayResource
             this.gatewayBackendManager.activateBackend(name);
         }
         catch (Exception e) {
-            log.error(e);
+            log.error(e, e.getMessage());
             return throwError(e);
         }
         return Response.ok().build();
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
+import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
 public class HaGatewayManager
@@ -82,13 +83,30 @@ public class HaGatewayManager
     @Override
     public void deactivateBackend(String backendName)
     {
-        dao.deactivate(backendName);
+        updateClusterActivationStatus(backendName, false, () -> dao.deactivate(backendName));
     }
 
     @Override
     public void activateBackend(String backendName)
     {
-        dao.activate(backendName);
+        updateClusterActivationStatus(backendName, true, () -> dao.activate(backendName));
+    }
+
+    private void updateClusterActivationStatus(String clusterName, boolean newStatus, Runnable changeActiveStatus)
+    {
+        GatewayBackend model = dao.findFirstByName(clusterName);
+        checkState(model != null, "No cluster found with name: %s, could not (de)activate", clusterName);
+
+        boolean previousStatus = model.active();
+        changeActiveStatus.run();
+        logActivationStatusChange(clusterName, newStatus, previousStatus);
+    }
+
+    private static void logActivationStatusChange(String clusterName, boolean newStatus, boolean previousStatus)
+    {
+        if (previousStatus != newStatus) {
+            log.info("Backend cluster %s activation status set to active=%s (previous status: active=%s).", clusterName, previousStatus, newStatus);
+        }
     }
 
     @Override
@@ -111,6 +129,7 @@ public class HaGatewayManager
         }
         else {
             dao.update(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
+            logActivationStatusChange(backend.getName(), backend.isActive(), model.active());
         }
         return backend;
     }


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,25 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
     @Override
     public void deactivateBackend(String backendName)
     {
-        dao.deactivate(backendName);
+        GatewayBackend model = dao.findFirstByName(backendName);
+        if (model != null) {
```
xkrogen, line 87: We should throw some error if `model == null` to indicate that the operation was unsuccessful

Side note, as `activateBackend`/`deactivateBackend` get more complex, should we unify them with some helper method?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,25 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+            Boolean prevStatus = model.active();
+            dao.deactivate(backendName);
+            log.info("Backend cluster %s has been deactivated (previous status: active=%s).",
+                    backendName, prevStatus);
```
xkrogen, line 91: For `updateBackend` we only log if there was an actual change, but here we log regardless. Let's be consistent across all methods, either always logging or only logging if there is a change in state.

xkrogen (commented):
Thanks @amybubu !  Ensuring we have good logging around this has been very helpful to track down the sequence of events for cluster activation/deactivation

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
Peiyingy: +1 on using helper method as `activateBackend` and `deactivateBackend` follow the similar pattern

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,30 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+    private void updateBackendActivationStatus(String backendName, Runnable action)
+    {
+        GatewayBackend model = dao.findFirstByName(backendName);
+        if (model == null) {
+            throw new IllegalStateException("No backend found with name: " + backendName + ", could not (de)activate");
+        }
+
+        Boolean prevStatus = model.active();
+        action.run();
+        Boolean currStatus = dao.findFirstByName(backendName).active();
```
xkrogen, line 104: IMO we shouldn't do another lookup here -- remember that this is a database call and thus expensive. We already know the state. Probably we should just accept a boolean `Runnable` and then use that value here?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
xkrogen, line 1: I see from https://github.com/trinodb/trino-gateway/pull/676 that there has been a decision to move away from the "backend" terminology, I guess we should follow the new "cluster" terminology here? cc @andythsu

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,30 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+
+        if (!Objects.equals(prevStatus, currStatus)) {
```
andythsu, line 106: IMO this check is redundant as the function is "updating" the status. If prevStatus and currStatus is the same for some reason, then it should be db level failure, in any case we should throw an exception

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -110,7 +128,13 @@ public ProxyBackendConfiguration updateBackend(ProxyBackendConfiguration backend
             dao.create(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
         }
         else {
+            Boolean prevStatus = model.active();
             dao.update(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
+            Boolean currStatus = backend.isActive();
+            if (!Objects.equals(prevStatus, currStatus)) {
```
andythsu, line 134: logic is duplicate here

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
andythsu: yes. Please use "cluster" to refer to Trino cluster.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
amybubu: I can remove checking if prevStatus != currStatus in the activate/deactivateBackned methods since it's redundant, but I think we need to keep this logic here because changes made to activation status via the UI go through this method rather than the activate/deactivateBackend methods, and since not all usages of updateBackend are for activation status changes, we need to check if a change occurred to log it.

amybubu: moved logging logic to helper function

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
amybubu: Rather than a db level failure, what if someone just calls /activate on an already active backend? This check prevents us from unnecessarily logging in this case. Also, it keeps it consistent with logging in updateBackend which does require checking if prevStatus != currStatus, because changes to activation status made via the UI use updateBackend, but not all calls to updateBackend method change the activation status, so we need to check if a change has been made to log it.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -110,7 +131,10 @@ public ProxyBackendConfiguration updateBackend(ProxyBackendConfiguration backend
             dao.create(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
         }
         else {
+            Boolean prevStatus = model.active();
             dao.update(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
+            Boolean currStatus = backend.isActive();
+            logActivationStatusChange(backend.getName(), currStatus, prevStatus);
```
xkrogen, line 137: Is it possible that `model.active()` or `backend.isActive()` could return null? If so you will get NPE on L137 since `logActivationStatusChange` accepts `boolean` (primitive type) so unboxing will occur here. So if null is possible, then `logActivationStatusChange` has to accept `Boolean` (boxed type). If it can't be null, all good.

Btw the local variables don't seem necessary:
```suggestion
            dao.update(backend.getName(), backend.getRoutingGroup(), backendProxyTo, backendExternalUrl, backend.isActive());
            logActivationStatusChange(backend.getName(), backend.isActive(), model.active());
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+        }
+
+        Boolean prevStatus = model.active();
+        action.run();
+        logActivationStatusChange(clusterName, newStatus, prevStatus);
+    }
+
+    private void logActivationStatusChange(String clusterName, boolean newStatus, boolean prevStatus)
+    {
+        if (!Objects.equals(prevStatus, newStatus)) {
```
xkrogen, line 109: Since `newStatus`/`prevStatus` are primitive type (`boolean`), you can just do `if (prevStatus != newStatus)`. We only need `Objects.equals()` if we use the boxed type (`Boolean`)

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
amybubu: Looks like `.active` and `.isActive` return primitive type `boolean` so I don't think they can be null . We should be good! Thanks for pointing that out

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +82,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+
+    private void updateBackendActivationStatus(String clusterName, boolean newStatus, Runnable action)
+    {
+        GatewayBackend model = dao.findFirstByName(clusterName);
+        if (model == null) {
+            throw new IllegalStateException("No cluster found with name: " + clusterName + ", could not (de)activate");
+        }
+
+        boolean prevStatus = model.active();
+        action.run();
```
andythsu, line 102: If the update fails for some reason, this will fail silently, making the next log line confusing. Should we only log when the update succeeds?

amybubu: Do you mean the log line from `logActivationStatusChange` will be confusing? Since it only logs if the activation status changes though, doesn't that mean we would already only log if the activation status was correctly updated?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +82,39 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
     {
-        dao.activate(backendName);
+        updateClusterActivationStatus(backendName, true, () -> dao.activate(backendName));
+    }
+
+    private void updateClusterActivationStatus(String clusterName, boolean newStatus, Runnable action)
+    {
+        GatewayBackend model = dao.findFirstByName(clusterName);
+        if (model == null) {
+            throw new IllegalStateException("No cluster found with name: " + clusterName + ", could not (de)activate");
```
andythsu, line 98: nit:
use `String.format()`

andythsu:
LGTM!

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +82,39 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
     {
-        dao.activate(backendName);
+        updateClusterActivationStatus(backendName, true, () -> dao.activate(backendName));
+    }
+
+    private void updateClusterActivationStatus(String clusterName, boolean newStatus, Runnable action)
+    {
+        GatewayBackend model = dao.findFirstByName(clusterName);
+        if (model == null) {
+            throw new IllegalStateException(String.format("No cluster found with name: %s, could not (de)activate", clusterName));
```
andythsu, line 98: static import `String` please :)

ebyhr (commented):
Please squash commits into one and update the PR & commit title. 

The current PR title looks misleading. It adds not only logging but also "dao.findFirstByName" call.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,39 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+            throw new IllegalStateException(format("No cluster found with name: %s, could not (de)activate", clusterName));
+        }
+
+        boolean prevStatus = model.active();
+        try {
+            action.run();
+            logActivationStatusChange(clusterName, newStatus, prevStatus);
+        }
+        catch (Exception e) {
+            log.error("Failed to update activation status for cluster %s: %s", clusterName, e.getMessage());
```
ebyhr, line 108: GatewayResource also logs the error message. I don't think we should log the similar message twice.

ebyhr (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,38 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+        }
+
+        boolean prevStatus = model.active();
+        try {
+            action.run();
+            logActivationStatusChange(clusterName, newStatus, prevStatus);
+        }
+        catch (Exception e) {
+            throw e;
+        }
```
ebyhr, line 109: Why do we catch this exception and rethrow it?

amybubu: my bad, didn't clean up whole thing

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
     }
 
     @Override
     public void activateBackend(String backendName)
     {
-        dao.activate(backendName);
+        updateClusterActivationStatus(backendName, true, () -> dao.activate(backendName));
+    }
+
+    private void updateClusterActivationStatus(String clusterName, boolean newStatus, Runnable action)
```
ebyhr, line 95: The "action" parameter name is too general. Please consider renaming to `changeActiveStatus` or something.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+    {
+        GatewayBackend model = dao.findFirstByName(clusterName);
+        if (model == null) {
+            throw new IllegalStateException(format("No cluster found with name: %s, could not (de)activate", clusterName));
+        }
+
+        boolean prevStatus = model.active();
```
ebyhr, line 102: Please avoid abbreviations https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations or inline this variable.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
ebyhr, line 95: This method can be static.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+        action.run();
+        logActivationStatusChange(clusterName, newStatus, prevStatus);
+    }
+
+    private void logActivationStatusChange(String clusterName, boolean newStatus, boolean prevStatus)
```
ebyhr, line 107: This method can be static.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
+    {
+        if (prevStatus != newStatus) {
+            log.info("Backend cluster %s activation status set to active=%s (previous status: active=%s).",
+                    clusterName, newStatus, prevStatus);
```
ebyhr, line 111: Unwarp.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -82,13 +83,33 @@ public Optional<ProxyBackendConfiguration> getBackendByName(String name)
...
-        dao.activate(backendName);
+        updateClusterActivationStatus(backendName, true, () -> dao.activate(backendName));
+    }
+
+    private void updateClusterActivationStatus(String clusterName, boolean newStatus, Runnable action)
+    {
+        GatewayBackend model = dao.findFirstByName(clusterName);
+        if (model == null) {
+            throw new IllegalStateException(format("No cluster found with name: %s, could not (de)activate", clusterName));
+        }
```
ebyhr, line 100:
```java
checkState(model != null, "No cluster found with name: %s, could not (de)activate", clusterName);
```

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
```diff
@@ -86,7 +86,7 @@ public Response activateBackend(@PathParam("name") String name)
             this.gatewayBackendManager.activateBackend(name);
         }
         catch (Exception e) {
-            log.error(e);
+            log.error(e, e.getMessage());
```
ebyhr, line 89: This change is unrelated to "Modify HaGatewayManager to support informative logging when (de)activating backend clusters". Please separate a commit.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
amybubu: I don't think this can be static because it uses dao which is not static.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
ebyhr: Reminder.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import java.util.Optional;
 
+import static com.google.common.base.Preconditions.checkState;
```
ebyhr, line 27: The commit title is too long. Please follow https://trino.io/development/process#pull-request-and-commit-guidelines-

ebyhr, line 27:
```
Improve logging for backend cluster (de)activation
address nit
```

The commit body should have been removed.

ebyhr:
@Chaho12 @oneonestar @mosabua Please review commit messages from next time.

PR #676, opened 2025-04-17
Update operation.md

andythsu:
## Description

A few meetings ago we've decided to use **cluster** to refer to Trino clusters instead of backend. I'd like to start with changing the jargons in all docs, then refactor the code base.

## Release notes


xkrogen:
Just want to add +1 on this, "backend" has been confusing to me :)

docs/operation.md
```diff
@@ -9,12 +9,12 @@ Trino Gateway records history of recent queries and displays links to check quer
 details page in respective trino cluster.
 ![trino.gateway.io](./assets/trinogateway_query_history.png)
 
-## Admin UI - add and modify backend information
+## Cluster UI - add and modify cluster information
 
-The admin page is used to configure the gateway to multiple backends.
-Existing backend information can also be modified using the same.
+The cluster page is used to configure the gateway to multiple Trino clusters.
```
mosabua, line 14:
```suggestion
The cluster page is used to configure Trino Gateway for multiple Trino clusters.
```

docs/operation.md
```diff
@@ -25,29 +25,32 @@ Query ID.
...
 
-1. Set the backend to deactivate state, this prevents any new incoming queries
-   from getting assigned to the backend.
-2. Poll the trino backend coorinator URL until the queued query count and the
-   running query count both hit 0.
-3. Terminate the trino Coordinator & Worker Java process.
+1. Deactivate the cluster by turning off the 'Active' switch. This ensures that no 
+   new incoming queries are routed to the cluster.
+2. Poll the Trino cluster coordinator URL until the queued query count and the
+   running query count are both 0.
```
mosabua, line 31:
```suggestion
   running query count are both zero.
```

mosabua (approved):
Just a couple of suggestions to apply. Thanks so much for picking this up.. there is more to do on that front in the code and such as well .. and I think we do have an issue filed for it.

docs/operation.md
```diff
@@ -25,29 +25,32 @@ Query ID.
...
 
 ```yaml
 modules:
   - io.trino.gateway.ha.module.QueryCountBasedRouterProvider
 ```
-- The router works on the stats it receives from the clusters about the load i.e number queries queued and running on a cluster at regular intervals which can be configured like below. The default interval is 1 min
+- The router operates based on the stats it receives from the clusters, such as 
+the number of queued and running queries, which are reported at regular 
+intervals. This interval can be configured by setting `taskDelaySeconds` under
+`monitor` section in the config file. The default interval is 1 min
```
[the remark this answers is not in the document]
andythsu: default is `60`. I can send a PR to make it use Airlift duration

mosabua (approved):
Looks good now.

PR #681, opened 2025-04-28
[Improvement] JDBC Connection Configuration with Flexible SSL and Property Injection

felicity3786:
## Description
Summary: Today, the Trino Gateway only supports very basic JDBC connection initialization, assuming standard username/password settings. It lacks support for important real-world features like client-side certificates (mTLS), custom truststore setups, or fine-grained SSL configurations when connecting to databases (e.g., MySQL).
I would like to help contributing a more flexible, extensible JDBC connection management system.

Right now, current gateway code hardcodes simple username/password based JDBC setup.
It is difficult to configure SSL/TLS for databases that require client certificates (e.g., `SslMode=VERIFY_CA` for `MySQL`). Gateway users with stricter security requirements (like us at **LinkedIn**) must patch or fork the code. And it is not cleanly extensible for new future other database types. At **LinkedIn**, We already implemented the following internally and would like to contribute it upstream:

- Introduce a `JdbcPropertiesProvider `interface for generating connection properties.
- Add `BasicJdbcPropertiesProvider` (default simple username/password).
- Add `MySqlJdbcPropertiesProvider` to handle MySQL-specific SSL properties:
- Handles `clientCertificateKeyStoreUrl, clientCertificateKeyStorePassword`, etc.
- Supports different `SslMode` settings (`DISABLED`, `VERIFY_CA`, etc.).
- Introduce a `JdbcPropertiesProviderFactory` to pick the right provider automatically based on configuration.
- Refactor `JdbcConnectionManager `to use these properties instead of hardcoded username/password.
- Add Airlift-compliant `@Singleton` / Guice bindings for better dependency injection.

**Benefits:**
The changes will help 
- Secure and flexible database connections (support MySQL client cert auth, etc.).
- Easily extensible for Oracle, or other databases in the future.
- More separation of concerns: connection properties logic decoupled from connection management.
- Retains full backward compatibility (H2, MySQL username/password still work out of the box).

## Release notes

```markdown
```

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -77,6 +77,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.mysql</groupId>
+            <artifactId>mysql-connector-j</artifactId>
+            <version>9.2.0</version>
+        </dependency>
+
+        <dependency>
             <groupId>com.nimbusds</groupId>
             <artifactId>nimbus-jose-jwt</artifactId>
             <version>10.0.2</version>
@@ -268,13 +274,6 @@
         <dependency>
             <groupId>org.weakref</groupId>
             <artifactId>jmxutils</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>com.mysql</groupId>
-            <artifactId>mysql-connector-j</artifactId>
-            <version>9.2.0</version>
-            <scope>runtime</scope>
         </dependency>
 
         <dependency>
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
@@ -22,6 +22,10 @@ public class DataStoreConfiguration
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
 
+    // TODO: Refactor to decouple DataStoreConfiguration from a specific
+    //  database implementation after adopting the Airlift configuration framework (https://github.com/trinodb/trino-gateway/issues/378)
+    private MysqlConfiguration mysqlConfiguration = new MysqlConfiguration();
+
     public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled)
     {
         this.jdbcUrl = jdbcUrl;
@@ -92,5 +96,15 @@ public class DataStoreConfiguration
     public void setRunMigrationsEnabled(boolean runMigrationsEnabled)
     {
         this.runMigrationsEnabled = runMigrationsEnabled;
+    }
+
+    public MysqlConfiguration getMysqlConfiguration()
+    {
+        return mysqlConfiguration;
+    }
+
+    public void setMysqlConfiguration(MysqlConfiguration mysqlConfig)
+    {
+        this.mysqlConfiguration = mysqlConfig;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
@@ -43,6 +43,7 @@ public class HaGatewayConfiguration
     private List<String> statementPaths = ImmutableList.of(V1_STATEMENT_PATH);
     private boolean includeClusterHostInResponse;
     private ProxyResponseConfiguration proxyResponseConfiguration = new ProxyResponseConfiguration();
+    private boolean oracleBackend;
 
     private RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
 
@@ -296,5 +297,15 @@ public class HaGatewayConfiguration
         {
             super(message);
         }
+    }
+
+    public boolean isOracleBackend()
+    {
+        return oracleBackend;
+    }
+
+    public void setOracleBackend(boolean oracleBackend)
+    {
+        this.oracleBackend = oracleBackend;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/MysqlConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/MysqlConfiguration.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.config;
+
+import com.mysql.cj.conf.PropertyDefinitions.SslMode;
+
+/**
+ * Configuration for MySQL SSL (client cert and truststore settings).
+ */
+public class MysqlConfiguration
+{
+    private SslMode sslMode = SslMode.DISABLED;
+    private String clientCertificateKeyStoreUrl;
+    private String clientCertificateKeyStorePassword;
+    private String clientCertificateKeyStoreType;
+    private String trustCertificateKeyStoreUrl;
+    private String trustCertificateKeyStorePassword;
+
+    public SslMode getSslMode()
+    {
+        return sslMode;
+    }
+
+    public MysqlConfiguration setSslMode(SslMode sslMode)
+    {
+        this.sslMode = sslMode;
+        return this;
+    }
+
+    public String getClientCertificateKeyStoreUrl()
+    {
+        return clientCertificateKeyStoreUrl;
+    }
+
+    public MysqlConfiguration setClientCertificateKeyStoreUrl(String url)
+    {
+        this.clientCertificateKeyStoreUrl = url;
+        return this;
+    }
+
+    public String getClientCertificateKeyStorePassword()
+    {
+        return clientCertificateKeyStorePassword;
+    }
+
+    public MysqlConfiguration setClientCertificateKeyStorePassword(String password)
+    {
+        this.clientCertificateKeyStorePassword = password;
+        return this;
+    }
+
+    public String getClientCertificateKeyStoreType()
+    {
+        return clientCertificateKeyStoreType;
+    }
+
+    public MysqlConfiguration setClientCertificateKeyStoreType(String type)
+    {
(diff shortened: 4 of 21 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -297,4 +298,14 @@ public HaGatewayConfigurationException(String message)
...
+
+    public boolean isOracleBackend()
+    {
+        return oracleBackend;
+    }
+
+    public void setOracleBackend(boolean oracleBackend)
+    {
+        this.oracleBackend = oracleBackend;
+    }
```
xkrogen, line 310: IIUC, with this change, now users will have to manually configure `oracleBackend = true` in addition to setting their JDBC URL to `jdbc:oracle://...`, is that right?

I have two ideas here.

First, in the interest of maintaining current behavior and reducing the burden on users, I would say we can just add a new method on `DataStoreConfiguration` like:
```java
public boolean isOracleBackend() {
  return jdbcUrl.startsWith("jdbc:oracle:");
}
```

But, thinking on this more, I think we can do better. I am thinking we should create an enum for all of the supported backends:
```java
enum DataStoreBackend {
  ORACLE,
  MYSQL,
  POSTGRES;
}
```

Then in `DataStoreConfiguration` we can default this to being set by the JDBC URL, but allow users to override if needed:
```java
private DataStoreBackend dataStoreBackendType;

...

public DataStoreBackend getDataStoreBackendType() {
  if (dataStoreBackendType != null) {
    return dataStoreBackendType;
  }
  if (jdbcUrl.startsWith("jdbc:oracle") {
    return ORACLE;
  } else if (...) {
    ..
  }
}
```

And then we can pass around this nice enum instead of a boolean. I suspect that over time we're going to need more DB-specific modifications throughout the system. WDYT?

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/BasicJdbcPropertiesProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.persistence;
+
+import io.trino.gateway.ha.config.DataStoreConfiguration;
+
+import java.util.Properties;
+
+public class BasicJdbcPropertiesProvider
```
xkrogen, line 20: Maybe add Javadoc explaining that this is the fallback / default behavior in case no DB-specific prop provider is available, and that this should always be the last one in the list?

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
felicity3786: Yep I think the enum DataStoreBackend is better, updated in the new commit. Thanks!

xkrogen:
Current diff LGTM, cc @willmostly

gateway-ha/src/main/java/io/trino/gateway/ha/config/MysqlConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+ * Configuration for MySQL SSL (client cert and truststore settings).
+ */
+public class MysqlConfiguration
+{
+    private SslMode sslMode = SslMode.DISABLED;
+    private String clientCertificateKeyStoreUrl;
+    private String clientCertificateKeyStorePassword;
+    private String clientCertificateKeyStoreType;
+    private String trustCertificateKeyStoreUrl;
+    private String trustCertificateKeyStorePassword;
```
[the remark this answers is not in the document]
felicity3786: The naming is from the mysql server: https://github.com/mysql/mysql-connector-j/blob/20ef8ee9eb4294a03858acccea0ddad3525f1ff9/src/main/core-api/java/com/mysql/cj/conf/PropertyKey.java#L82-L84 so the driver can easily pick it up without translating it. I felt it might be easier keeping it this way, I can add some comments for the trino concept mapping though, WDYT?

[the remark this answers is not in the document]
xkrogen: +1 with @Felicity-3786 -- we previously discussed that if we had these as generic, Gateway-level configs, we should use the names consistent with Trino like @ebyhr said. But currently these are MySQL-specific configs, so we should be consistent with the MySQL naming convention.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -101,43 +104,55 @@ public void setRunMigrationsEnabled(boolean runMigrationsEnabled)
...
-                return DataStoreBackend.H2;
+        try {
+            Enumeration<Driver> drivers = DriverManager.getDrivers();
+            while (drivers.hasMoreElements()) {
+                Driver driver = drivers.nextElement();
+                if (driver.acceptsURL(jdbcUrl)) {
+                    String driverClass = driver.getClass().getName();
+                    if (driverClass.contains("oracle")) {
+                        return DataStoreType.ORACLE;
+                    }
```
xkrogen, line 134: This fuzzy string match against class name doesn't seem right to me. We have all of the driver classes on the classpath, can we do an actual match against the class name? Like adding a field to `DataStoreType` which tracks the expected instance of `Class<?>` for that `DataStoreType`.  Then this iteration can become something like:
```
while (drivers.hasMoreElements()) {
                Driver driver = drivers.nextElement();
                if (driver.acceptsURL(jdbcUrl)) {
                    for (DataStoreType dataStoreType : DataStoreType.values() {
                      if (dataStoreType.driverClass() == driver.getClass()) {
                        return dataStoreType;
                      }
                    }
                    break;
                 }
}
```

gateway-ha/src/main/java/io/trino/gateway/ha/config/MySqlConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+        this.sslMode = sslMode;
+        return this;
+    }
+
+    public String getClientCertificateKeyStoreUrl()
+    {
+        return clientCertificateKeyStoreUrl;
+    }
+
+    public MySqlConfiguration setClientCertificateKeyStoreUrl(String url)
```
ebyhr, line 46: Please rename the parameter. Same for others.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcPropertiesProviderFactory.java
```diff
@@ -30,13 +29,13 @@ public class JdbcPropertiesProviderFactory
     @Inject
     public JdbcPropertiesProviderFactory(List<JdbcPropertiesProvider> orderedProviders)
     {
-        this.providers = requireNonNull(orderedProviders, "providers is null");
+        this.providers = List.copyOf(orderedProviders);
```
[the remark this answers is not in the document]
ebyhr: requireNonNull is redundant when we use ImmutableList.copyOf.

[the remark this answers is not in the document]
xkrogen: Why is that @ebyhr ? `copyOf()` will throw NPE if given a null value:
https://github.com/google/guava/blob/6bbfc86eb2cf64600e1e30e112cad227f5e31f4b/guava/src/com/google/common/collect/ImmutableList.java#L258-L264

The Trino repo has hundreds of usages of this pattern: https://github.com/search?q=repo%3Atrinodb%2Ftrino%20%22ImmutableList.copyOf(requireNonNull(%22&type=code

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcPropertiesProviderFactory.java
```diff
@@ -19,40 +19,39 @@
...
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
 
 import java.util.Properties;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-class TestJdbcPropertiesProviderFactory
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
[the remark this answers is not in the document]
felicity3786: Hmm I guess we don't really need it given I have remove all the @BeforeAll, removed it

gateway-ha/src/main/java/io/trino/gateway/ha/config/MySqlConfiguration.java
felicity3786: Hi @ebyhr, @xkrogen and I both think it's would make more sense to keep the mysql naming convention as we discussed in https://github.com/trinodb/trino-gateway/pull/681#discussion_r2076642607, wdyt?

mosabua:
Needs a rebase and adjustment to recently merged PR about JDBC connection properties

mosabua:
Please let us know when you rebased and updated this PR @Felicity-3786

felicity3786:
> Please let us know when you rebased and updated this PR @Felicity-3786

Thanks @mosabua for reminding! Rebased, mind taking another look?

mosabua:
@Felicity-3786 could you rebase and confirm all comments from @ebyhr are addressed? 

Also @xkrogen .. anything to review/add here? If you are good .. please approve.

gateway-ha/pom.xml
```diff
@@ -78,6 +78,18 @@
             <classifier>classes</classifier>
         </dependency>
 
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <version>1.4.192</version>
+        </dependency>
```
xkrogen, line 92: We should keep H2 off the production classpath since it is intended to be test-only.

felicity3786: hmm but we use it in https://github.com/trinodb/trino-gateway/blob/ab73f04462983ac115ea7ca7f7589c9a375aead9/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreType.java#L22 so it needs to be in compile path?

felicity3786:
> @Felicity-3786 could you rebase and confirm all comments from @ebyhr are addressed?
> 
> Also @xkrogen .. anything to review/add here? If you are good .. please approve.

Thanks for reminder, resolved the conflict and yes all the previous comments are addressed.

gateway-ha/pom.xml
xkrogen: We'll have to find a way to keep it available for testing but not on the compile classpath. Preferably there should be no references to H2 whatsoever in production (code or classpath). Maybe there needs to be a way to register a new `DataStoreType`, and in the test code we can register H2?

mosabua: There is a PR to remove H2 .. maybe we should merge that one first or so

felicity3786: Sure let me know if I need to rebase after that one went in.

PR #683, opened 2025-05-09
Let Guice manage dependencies in router modules

andythsu:
## Description

Manual initialization is preventing us from using dependency injection. We should use Guice to manage the dependencies.

## Release notes


xkrogen (commented):
+1 from me. cc @felicity3786 this is relevant to what we were discussing in the context of another PR recently

ebyhr (approved):

Issue #81, opened 2023-10-27
[Bug] The query history UI should use externalUrl instead of backendUrl


lambrospetrou:
Hello folks, we hit a bug using the new `trino-gateway` when opening query details from the gateway's UI displaying query history.

The `trino-gateway` uses `backendUrl` instead of `externalUrl` in the links it shows in the UI ([see code](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/resources/template/query-history-view.ftl#L64)) which means when opened from our internal users, they do not resolve properly since they are using the internal endpoints that the gateway uses to communicate with the clusters.

The previous `presto-gateway` was using the gateway's URL for this ([see code](https://github.com/lyft/presto-gateway/blob/master/gateway-ha/src/main/resources/template/query-history-view.ftl#L63)), which works so far in our case, and then it routes the UI calls to fetch the query details to the right coordinator cluster using the query mapping table/cache. At some point there is a request `http://<gateway-url>/ui/api/query/<query_id>`, which is matched by the request handler and gets sent to the right coordinator that has the query details.

## Solution

I suggest one of the following:
1. Use the `externalUrl` of the backend cluster for the links in the query history, and then we will need to set those to something resolvable by the expected users of the admin UI.
2. Revert to using the gateway's URL and let the routing take care of finding the right backend cluster.

lambrospetrou:
I can provide a PR to fix this as long as we decide on which of the two solutions you would like to use. Maybe @mosabua can chime in with a preference or route to the right folks?

wewark:
Hello, any ETA for this? It seems like even setting `externalUrl` to the gateway URL has no effect. It is probably using the backend URL anyway https://github.com/trinodb/trino-gateway/blob/49c5c82dbc18a50b9104d001917dc3139088a0da/webapp/src/components/history.tsx#L57C30-L57C47

oneonestar:
This issue is a bit complicated. Currently, we only store the `backend_url` in the query history table. Both `backend_url` and `external_url` can change at any time. This can lead to problems when the URLs are changed and the old `backend_url` in the query history no longer matches with the current setting.

Storing the foreign key `gateway_backend.name` in the query history table could solve the issue. We can get the latest `backend_url` and `external_url` for the cluster. For audit and debugging purposes, `backend_url` should also be kept for reference.

```
CREATE TABLE IF NOT EXISTS gateway_backend (
name VARCHAR(256) PRIMARY KEY,
[...]

PR #693, opened 2025-06-02
Use external url for query routing history

raj-manvar:
## Description

Use external url field for query routing history page

The backendUrl used by Trino gateway may not be accessible for users, instead the external url field is used to indicate the hostname that's accessible by Trino users.  

For example, when hosted on Kubernetes, it's common to use [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) for inter apps communication like between Gateway and Trino clusters, in such cases the hostname used by Gateway would be something like `service-name.namespace.svc.cluster.local` which is only accessible from within the Kubernetes cluster. 

https://github.com/trinodb/trino-gateway/issues/81 is relevant issue with other discussion points.

## Release notes

```markdown
```

### DIFF
--- a/docs/routing-logic.md
+++ b/docs/routing-logic.md
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
+## Routing URLs
+
+Each Trino cluster configured with the Trino Gateway includes both a `proxyTo` 
+URL and an `externalURL`. The `proxyTo` URL is used internally by the Trino 
+Gateway to route requests based on routing rules and query identifiers, whereas 
+the `externalURL` serves as a UI-accessible or publicly reachable address for 
+the Trino cluster, and is commonly used to access [Trino web UI](https://trino.io/docs/current/admin/web-interface.html)  
+
+For example, in a Kubernetes environment, the `proxyTo` URL might be 
+`trino-backend-service.trino-namespace.svc.cluster.local:8083` for communication 
+between the Trino Gateway and Trino clusters, and the external URL for the same 
+backend cluster might be `trino.domain.com`.
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -17,6 +17,7 @@ import com.google.inject.Inject;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.GatewayCookieConfigurationPropertiesProvider;
 import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.handler.schema.RoutingDestination;
 import io.trino.gateway.ha.handler.schema.RoutingTargetResponse;
 import io.trino.gateway.ha.router.GatewayCookie;
@@ -81,8 +82,10 @@ public class RoutingTargetHandler
         RoutingTargetResponse routingTargetResponse = previousCluster.map(cluster -> {
             String routingGroup = queryId.map(routingManager::findRoutingGroupForQueryId)
                     .orElse("adhoc");
+            String externalUrl = queryId.map(routingManager::findExternalUrlForQueryId)
+                    .orElse(cluster);
             return new RoutingTargetResponse(
-                    new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request)),
+                    new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request), externalUrl),
                     request);
         }).orElse(getRoutingTargetResponse(request));
         logRewrite(routingTargetResponse.routingDestination().clusterHost(), request);
@@ -97,14 +100,16 @@ public class RoutingTargetHandler
         String routingGroup = (routingDestination.routingGroup() != null && !routingDestination.routingGroup().isEmpty())
                 ? routingDestination.routingGroup()
                 : "adhoc";
-        String clusterHost = routingManager.provideClusterForRoutingGroup(routingGroup, user);
+        ProxyBackendConfiguration backendConfiguration = routingManager.provideBackendConfiguration(routingGroup, user);
+        String clusterHost = backendConfiguration.getProxyTo();
+        String externalUrl = backendConfiguration.getExternalUrl();
         // Apply headers from RoutingDestination if there are any
         HttpServletRequest modifiedRequest = request;
         if (!routingDestination.externalHeaders().isEmpty()) {
             modifiedRequest = new HeaderModifyingRequestWrapper(request, routingDestination.externalHeaders());
         }
         return new RoutingTargetResponse(
-                new RoutingDestination(routingGroup, clusterHost, buildUriWithNewCluster(clusterHost, request)),
+                new RoutingDestination(routingGroup, clusterHost, buildUriWithNewCluster(clusterHost, request), externalUrl),
                 modifiedRequest);
     }
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryHistoryManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryHistoryManager.java
@@ -68,7 +68,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public String getQueryId()
         {
-            return this.queryId;
+            return queryId;
         }
 
         public void setQueryId(String queryId)
@@ -79,7 +79,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public String getQueryText()
         {
-            return this.queryText;
+            return queryText;
         }
 
         public void setQueryText(String queryText)
@@ -90,7 +90,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public String getUser()
         {
-            return this.user;
+            return user;
         }
 
         public void setUser(String user)
@@ -101,7 +101,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public String getSource()
         {
-            return this.source;
+            return source;
         }
 
         public void setSource(String source)
@@ -112,7 +112,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public String getBackendUrl()
         {
-            return this.backendUrl;
+            return backendUrl;
         }
 
         public void setBackendUrl(String backendUrl)
@@ -123,7 +123,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public long getCaptureTime()
         {
-            return this.captureTime;
+            return captureTime;
         }
 
         public void setCaptureTime(long captureTime)
@@ -134,7 +134,7 @@ public interface QueryHistoryManager
         @JsonProperty
         public String getRoutingGroup()
         {
-            return this.routingGroup;
+            return routingGroup;
         }
 
         public void setRoutingGroup(String routingGroup)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryHistoryManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryHistoryManager.java
@@ -34,6 +34,8 @@ public interface QueryHistoryManager
 
     String getRoutingGroupForQueryId(String queryId);
 
+    String getExternalUrlForQueryId(String queryId);
+
     TableData<QueryDetail> findQueryHistory(QueryHistoryRequest query);
 
     List<DistributionResponse.LineChart> findDistribution(Long ts);
@@ -48,6 +50,7 @@ public interface QueryHistoryManager
         private String backendUrl;
         private long captureTime;
         private String routingGroup;
+        private String externalUrl;
 
         public QueryDetail() {}
 
@@ -139,6 +142,17 @@ public interface QueryHistoryManager
             this.routingGroup = routingGroup;
         }
 
+        @JsonProperty
+        public String getExternalUrl()
+        {
+            return externalUrl;
+        }
+
+        public void setExternalUrl(String externalUrl)
+        {
+            this.externalUrl = externalUrl;
+        }
+
         @Override
         public boolean equals(Object o)
         {
@@ -155,13 +169,14 @@ public interface QueryHistoryManager
                     Objects.equals(user, that.user) &&
                     Objects.equals(source, that.source) &&
                     Objects.equals(backendUrl, that.backendUrl) &&
-                    Objects.equals(routingGroup, that.routingGroup);
+                    Objects.equals(routingGroup, that.routingGroup) &&
+                    Objects.equals(externalUrl, that.externalUrl);
         }
 
         @Override
         public int hashCode()
         {
-            return Objects.hash(queryId, queryText, user, source, backendUrl, captureTime, routingGroup);
+            return Objects.hash(queryId, queryText, user, source, backendUrl, captureTime, routingGroup, externalUrl);
         }
 
         @Override
@@ -175,6 +190,7 @@ public interface QueryHistoryManager
                     .add("backendUrl", backendUrl)
                     .add("captureTime", captureTime)
                     .add("routingGroup", routingGroup)
+                    .add("externalUrl", externalUrl)
                     .toString();
         }
     }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
@@ -61,7 +61,7 @@ public abstract class RoutingManager
                         .maximumSize(10000)
                         .expireAfterAccess(30, TimeUnit.MINUTES)
                         .build(
-                                new CacheLoader<String, String>()
+                                new CacheLoader<>()
                                 {
                                     @Override
                                     public String load(String queryId)
@@ -74,7 +74,7 @@ public abstract class RoutingManager
                         .maximumSize(10000)
                         .expireAfterAccess(30, TimeUnit.MINUTES)
                         .build(
-                                new CacheLoader<String, String>()
+                                new CacheLoader<>()
                                 {
                                     @Override
                                     public String load(String queryId)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
@@ -20,6 +20,7 @@ import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
 import io.trino.gateway.ha.clustermonitor.TrinoStatus;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import jakarta.annotation.Nullable;
 import jakarta.ws.rs.HttpMethod;
 
 import java.net.HttpURLConnection;
@@ -48,6 +49,7 @@ public abstract class RoutingManager
     private final GatewayBackendManager gatewayBackendManager;
     private final ConcurrentHashMap<String, TrinoStatus> backendToStatus;
     private final LoadingCache<String, String> queryIdRoutingGroupCache;
+    private final LoadingCache<String, String> queryIdExternalUrlCache;
     private final QueryHistoryManager queryHistoryManager;
 
     public RoutingManager(GatewayBackendManager gatewayBackendManager, QueryHistoryManager queryHistoryManager)
@@ -80,6 +82,19 @@ public abstract class RoutingManager
                                         return findRoutingGroupForUnknownQueryId(queryId);
                                     }
                                 });
+        queryIdExternalUrlCache =
+                CacheBuilder.newBuilder()
+                        .maximumSize(10000)
+                        .expireAfterAccess(30, TimeUnit.MINUTES)
+                        .build(
+                                new CacheLoader<>()
+                                {
+                                    @Override
+                                    public String load(String queryId)
+                                    {
+                                        return findExternalUrlForUnknownQueryId(queryId);
+                                    }
+                                });
 
         this.backendToStatus = new ConcurrentHashMap<>();
     }
@@ -94,6 +109,11 @@ public abstract class RoutingManager
         queryIdBackendCache.put(queryId, backend);
     }
 
+    public void setExternalUrlForQueryId(String queryId, String externalUrl)
+    {
+        queryIdExternalUrlCache.put(queryId, externalUrl);
+    }
+
     public void setRoutingGroupForQueryId(String queryId, String routingGroup)
     {
         queryIdRoutingGroupCache.put(queryId, routingGroup);
@@ -102,7 +122,7 @@ public abstract class RoutingManager
     /**
      * Performs routing to an adhoc backend.
      */
-    public String provideAdhocCluster(String user)
+    private ProxyBackendConfiguration provideAdhocBackendConfiguration()
     {
         List<ProxyBackendConfiguration> backends = this.gatewayBackendManager.getActiveAdhocBackends();
         backends.removeIf(backend -> isBackendNotHealthy(backend.getName()));
@@ -110,29 +130,40 @@ public abstract class RoutingManager
             throw new IllegalStateException("Number of active backends found zero");
         }
         int backendId = Math.abs(RANDOM.nextInt()) % backends.size();
-        return backends.get(backendId).getProxyTo();
+        return backends.get(backendId);
+    }
+
+    public String provideAdhocCluster(String user)
+    {
+        return provideAdhocBackendConfiguration().getProxyTo();
     }
 
     /**
      * Performs routing to a given cluster group. This falls back to an adhoc backend, if no scheduled
      * backend is found.
      */
-    public String provideClusterForRoutingGroup(String routingGroup, String user)
+    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
     {
         List<ProxyBackendConfiguration> backends =
                 gatewayBackendManager.getActiveBackends(routingGroup);
         backends.removeIf(backend -> isBackendNotHealthy(backend.getName()));
         if (backends.isEmpty()) {
-            return provideAdhocCluster(user);
+            return provideAdhocBackendConfiguration();
         }
         int backendId = Math.abs(RANDOM.nextInt()) % backends.size();
-        return backends.get(backendId).getProxyTo();
+        return backends.get(backendId);
+    }
+
+    public String provideClusterForRoutingGroup(String routingGroup, String user)
+    {
+        return provideBackendConfiguration(routingGroup, user).getProxyTo();
     }
 
     /**
      * Performs cache look up, if a backend not found, it checks with all backends and tries to find
      * out which backend has info about given query id.
      */
+    @Nullable
     public String findBackendForQueryId(String queryId)
     {
         String backendAddress = null;
@@ -140,15 +171,29 @@ public abstract class RoutingManager
             backendAddress = queryIdBackendCache.get(queryId);
         }
         catch (ExecutionException e) {
-            log.error("Exception while loading queryId from cache %s", e.getLocalizedMessage());
+            log.warn("Exception while loading queryId from cache %s", e.getLocalizedMessage());
         }
         return backendAddress;
+    }
+
+    @Nullable
+    public String findExternalUrlForQueryId(String queryId)
+    {
+        String externalUrl = null;
+        try {
+            externalUrl = queryIdExternalUrlCache.get(queryId);
+        }
+        catch (ExecutionException e) {
+            log.warn("Exception while loading queryId from cache %s", e.getLocalizedMessage());
+        }
+        return externalUrl;
     }
 
     /**
      * Looks up the routing group associated with the queryId in the cache.
      * If it's not in the cache, look up in query history
      */
+    @Nullable
     public String findRoutingGroupForQueryId(String queryId)
     {
         String routingGroup = null;
@@ -156,7 +201,7 @@ public abstract class RoutingManager
             routingGroup = queryIdRoutingGroupCache.get(queryId);
         }
         catch (ExecutionException e) {
-            log.error("Exception while loading queryId from routing group cache %s", e.getLocalizedMessage());
+            log.warn("Exception while loading queryId from routing group cache %s", e.getLocalizedMessage());
         }
         return routingGroup;
     }
@@ -225,6 +270,16 @@ public abstract class RoutingManager
         String routingGroup = queryHistoryManager.getRoutingGroupForQueryId(queryId);
         setRoutingGroupForQueryId(queryId, routingGroup);
         return routingGroup;
+    }
+
+    /**
+     * Attempts to look up the external url associated with the query id from query history table
+     */
+    protected String findExternalUrlForUnknownQueryId(String queryId)
+    {
+        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
+        setExternalUrlForQueryId(queryId, externalUrl);
+        return externalUrl;
     }
 
     // Predicate helper function to remove the backends from the list
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseExternalUrlQueryHistoryTest.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseExternalUrlQueryHistoryTest.java
@@ -1,0 +1,202 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import io.trino.gateway.ha.config.DataStoreConfiguration;
+import io.trino.gateway.ha.persistence.FlywayMigration;
+import io.trino.gateway.ha.persistence.JdbcConnectionManager;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+
+import java.util.List;
+import java.util.Optional;
+
+import static io.trino.gateway.ha.TestingJdbcConnectionManager.createTestingJdbcConnectionManager;
+import static java.util.Objects.requireNonNull;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(Lifecycle.PER_CLASS)
+abstract class BaseExternalUrlQueryHistoryTest
+{
+    private final JdbcDatabaseContainer<?> container;
+    private final QueryHistoryManager queryHistoryManager;
+
+    protected BaseExternalUrlQueryHistoryTest(JdbcDatabaseContainer<?> container)
+    {
+        this.container = requireNonNull(container, "container is null");
+        this.container.start();
+
+        DataStoreConfiguration config = new DataStoreConfiguration(
+                container.getJdbcUrl(),
+                container.getUsername(),
+                container.getPassword(),
+                container.getDriverClassName(),
+                4,
+                true);
+        FlywayMigration.migrate(config);
+        JdbcConnectionManager jdbcConnectionManager = createTestingJdbcConnectionManager(container, config);
+        queryHistoryManager = new HaQueryHistoryManager(jdbcConnectionManager.getJdbi(), container.getJdbcUrl().startsWith("jdbc:oracle"));
+    }
+
+    @AfterAll
+    public final void close()
+    {
+        container.close();
+    }
+
+    @Test
+    void testSubmitQueryWithExternalUrl()
+    {
+        // Clear existing data
+        List<QueryHistoryManager.QueryDetail> queryDetails = queryHistoryManager.fetchQueryHistory(Optional.empty());
+        assertThat(queryDetails).isEmpty();
+
+        // Create and submit query with external URL
+        QueryHistoryManager.QueryDetail queryDetail = new QueryHistoryManager.QueryDetail();
+        queryDetail.setQueryId("test-query-123");
+        queryDetail.setQueryText("SELECT * FROM test_table");
+        queryDetail.setBackendUrl("http://localhost:8080");
+        queryDetail.setUser("test-user");
+        queryDetail.setSource("sqlWorkbench");
+        queryDetail.setRoutingGroup("adhoc");
+        queryDetail.setExternalUrl("https://external-gateway.example.com");
+        queryDetail.setCaptureTime(System.currentTimeMillis());
+
+        queryHistoryManager.submitQueryDetail(queryDetail);
+
+        // Verify the query was stored with external URL
+        queryDetails = queryHistoryManager.fetchQueryHistory(Optional.empty());
+        assertThat(queryDetails).hasSize(1);
+
+        QueryHistoryManager.QueryDetail retrievedQuery = queryDetails.getFirst();
+        assertThat(retrievedQuery.getQueryId()).isEqualTo("test-query-123");
+        assertThat(retrievedQuery.getExternalUrl()).isEqualTo("https://external-gateway.example.com");
+    }
+
+    @Test
+    void testGetExternalUrlByQueryId()
+    {
+        // Create and submit query with external URL
+        QueryHistoryManager.QueryDetail queryDetail = new QueryHistoryManager.QueryDetail();
+        queryDetail.setQueryId("external-url-test-456");
+        queryDetail.setQueryText("SELECT count(*) FROM users");
+        queryDetail.setBackendUrl("http://backend:8080");
+        queryDetail.setUser("admin");
+        queryDetail.setSource("trino-cli");
+        queryDetail.setRoutingGroup("analytics");
+        queryDetail.setExternalUrl("https://analytics-gateway.company.com");
+        queryDetail.setCaptureTime(System.currentTimeMillis());
+
+        queryHistoryManager.submitQueryDetail(queryDetail);
+
+        // Test retrieving external URL by query ID
+        String externalUrl = queryHistoryManager.getExternalUrlForQueryId("external-url-test-456");
+        assertThat(externalUrl).isEqualTo("https://analytics-gateway.company.com");
+    }
+
+    @Test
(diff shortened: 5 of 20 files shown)


webapp/src/components/history.tsx
```diff
@@ -123,7 +123,7 @@ export function History() {
           <Column title="QueryId" dataIndex="queryId" key="queryId" render={linkQueryRender} />
           <Column title="RoutingGroup" dataIndex="routingGroup" key="routingGroup" render={routingGroupRender} />
           <Column title="RoutedTo" dataIndex="backendUrl" key="backendUrlName" render={(text: string) => <Text>{backendMapping[text]}</Text>} />
-          <Column title="RoutedTo" dataIndex="backendUrl" key="backendUrl" render={linkRender} />
+          <Column title="RoutedTo" dataIndex="externalUrl" key="externalUrl" render={linkRender} />
```
raj-manvar, line 126: Looking for some feedback here, should we keep the RoutedTo for backward compatibility and add a new column for externalUrl ? Or maybe just change to use externalUrl only for QueryId ? Since I believe going to Trino Query details page would be main intent to get the queryId

xkrogen:
Big +1 on this, we actually just finished an internal implementation of this only a couple of weeks ago that we have been planning to open source. cc @felicity3786 can you help collaborate on this?

webapp/src/components/history.tsx
felicity3786: I am in favor of also updating the RoutedTo column, presenting an inaccessible url on the UI could be confusing. Using externalUrl user can jump to cluster from gw.

felicity3786: Also maybe we should update the filter dropdown menu to use externalUrl as well? in L#97
```
{backendData?.map(b => (
                  <Form.Select.Option key={b.externalUrl} value={b.externalUrl}>
                    <Tag color={'blue'} style={{ marginRight: '5px' }}>{backendMapping[b.externalUrl]}</Tag>
                    <Text>{b.externalUrl}</Text>
```

raj-manvar: Thanks! yes, updated the dropdown menu to use externalUrl as well.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -227,6 +268,16 @@ protected String findRoutingGroupForUnknownQueryId(String queryId)
         return routingGroup;
     }
 
+    /**
+     * Attempts to look up the external url associated with the query id from query history table
+     */
+    protected String findExternalUrlForUnknownQueryId(String queryId)
+    {
+        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
+        setRoutingGroupForQueryId(queryId, externalUrl);
```
felicity3786: Do you meant to call `setExternalUrlForQueryId(queryId, externalUrl)` ?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -97,14 +100,16 @@ private RoutingTargetResponse getRoutingTargetResponse(HttpServletRequest reques
...
+        String clusterHost = backendConfiguration.getProxyTo();
+        String externalUrl = backendConfiguration.getExternalUrl();
         // Apply headers from RoutingDestination if there are any
         HttpServletRequest modifiedRequest = request;
         if (!routingDestination.externalHeaders().isEmpty()) {
             modifiedRequest = new HeaderModifyingRequestWrapper(request, routingDestination.externalHeaders());
         }
         return new RoutingTargetResponse(
-                new RoutingDestination(routingGroup, clusterHost, buildUriWithNewCluster(clusterHost, request)),
+                new RoutingDestination(routingGroup, clusterHost, buildUriWithNewCluster(clusterHost, request), externalUrl),
```
felicity3786, line 112: Could `externalUrl` ever be `null` here? We can add the same fallback to `clusterHost` to be safe.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -227,6 +268,16 @@ protected String findRoutingGroupForUnknownQueryId(String queryId)
         return routingGroup;
     }
 
+    /**
+     * Attempts to look up the external url associated with the query id from query history table
+     */
+    protected String findExternalUrlForUnknownQueryId(String queryId)
+    {
+        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
```
felicity3786, line 280: Same here, do we need to handle when the database has no `external_url `for that `queryId`, maybe pick a sensible default `backendUrl` and then populate the cache?

felicity3786 (commented):
Overall LGTM! 
Internally our implementation is a more "normalized" approache, we keep `query_history `without `external_url`, and join the `gateway_backend` to get it at the read time. It would leads to some potential mismatch if the `gateway_backend` table got updated, but no schema migration needed.
I feel in this PR, it's handled cleaner and simpler. Although there will be risk of "stale" data like if admins rename external_url, but given query_history will be purged periodically I don't think it would be a very big concern.
cc @xkrogen

xkrogen:
> Although there will be risk of "stale" data like if admins rename external_url, but given query_history will be purged periodically I don't think it would be a very big concern.

Agree that this _could_ be a concern, but it applies even with the `backend_url` (i.e. even without this change, if we change the `backend_url` for a backend, the old history won't update). To actually tackle this, we would have to begin registering backends under some ID, and then store the backend's ID in the history instead of the URLs, and join with the backend table to reconstruct. So I think adding `external_url` is still fine, consistent with our current approach of the history table storing point-in-time info as of when the query was executed. If/when practical issues are observed that would encourage us to normalize this data, moving out both `backend_url`/`external_url` should be no more work than just moving `backend_url`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Thanks yes! Sorry I had made a copy pasta error here. Updated to setExternalUrlForQueryId

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
raj-manvar: The UI form which asks for setting the backend has the externalUrl as a compulsory field.   <br> 
image

So if the backend is added via UI, it would likely be not null <br> 
The DB schema from https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/resources/mysql/V1__create_schema.sql#L5 does allow the externalUrl to have Null value though, so someone could insert into the DB directly and can have externalUrl to be null.  The clusterHost can be null too if inserted into the DB directly but that would fail all routing itself I believe so it would be caught early. 

I think if the expectation is that the Routing backend will be added using the UI form then it won't be null, but if we want to handle cases when the externalUrl could still be null via DB inserts we can handle this case.  What do you think ?

webapp/src/components/history.tsx
Chaho12: FYI. RoutedTo is duplicate for backendUrl and will be replaced.
https://github.com/trinodb/trino-gateway/pull/639/files#diff-7faaf9ded840c56e9b317edbaf8aeb74eb813b342bbc63051cf8e4f197f77d2e

Chaho12 (commented):
Can you update the docs on explaning difference between RoutedTo and ExternalURL? Adding purpose of why 2 url exists would be great. 
As you have noticed, external URL was not used at all before.

webapp/src/components/history.tsx
raj-manvar: Thanks for the heads up!  The changes from pr #639 seems on different line so hopefully there's no conflict, but I'll rebase with main branch if it's merged first and there's conflict.

raj-manvar:
Added documentation on two URLs, purposes and example for the same.  cc: @Chaho12

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -80,6 +81,19 @@ public String load(String queryId)
                                         return findRoutingGroupForUnknownQueryId(queryId);
                                     }
                                 });
+        queryIdExternalUrlCache =
```
Peiyingy, line 85: I agree we need an ExternalUrl cache, but I’m not sure if having `queryIdExternalUrlCache` is the most efficient approach. Perhaps we could use either:
- Single cache: map each `queryId → (backendUrl, externalUrl)` tuple 
- Two caches: keep `queryId → backendUrl` and `backendUrl → externalUrl`

I'm wondering what do you think?

docs/routing-logic.md
```diff
@@ -75,3 +75,15 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
+## Routing URLs
+
+Each Trino backend added in Trino Gateway supports proxyTo URL and external URL.
```
Chaho12:
```suggestion
Each Trino backend to the Trino Gateway supports both a proxyTo URL and external URL.
```

docs/routing-logic.md
```diff
@@ -75,3 +75,15 @@ oauth2GatewayCookieConfiguration:
...
+Proxy URL is used by Gateway internally to route the requests based on routing rules and query identifiers,
```
Chaho12:
```suggestion
While the proxy URL is used internally by the Gateway to route requests based on routing rules and query identifiers, 
```

docs/routing-logic.md
```diff
@@ -75,3 +75,15 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
+## Routing URLs
+
+Each Trino backend added in Trino Gateway supports proxyTo URL and external URL.
+Proxy URL is used by Gateway internally to route the requests based on routing rules and query identifiers, 
+while external URL is used to indicate user accessible URL for the Trino backend clusters,
```
Chaho12:
```suggestion
the external URL represents a user-accessible address for the Trino backend cluster,
```

docs/routing-logic.md
```diff
@@ -75,3 +75,15 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
+## Routing URLs
+
+Each Trino backend added in Trino Gateway supports proxyTo URL and external URL.
+Proxy URL is used by Gateway internally to route the requests based on routing rules and query identifiers, 
+while external URL is used to indicate user accessible URL for the Trino backend clusters, 
+commonly used to access [Trino web UI](https://trino.io/docs/current/admin/web-interface.html)
```
Chaho12:
```suggestion
and is commonly used to access [Trino web UI](https://trino.io/docs/current/admin/web-interface.html)  
```

docs/routing-logic.md
```diff
@@ -75,3 +75,15 @@ oauth2GatewayCookieConfiguration:
...
+
+Each Trino backend added in Trino Gateway supports proxyTo URL and external URL.
+Proxy URL is used by Gateway internally to route the requests based on routing rules and query identifiers, 
+while external URL is used to indicate user accessible URL for the Trino backend clusters, 
+commonly used to access [Trino web UI](https://trino.io/docs/current/admin/web-interface.html)  
+
+For example, the proxyTo URL could be `trino-backend-service.trino-namespace.svc.cluster.local:8083` when using [Kubernets Service](
+) to communicate between Gateway and Trino backend, 
+while external URL for the same backend cluster could be `trino.domain.com`. 
+For cases where there is no user accessible URL for backend clusters available, external URL can be set to Trino Gateway itself or proxyTo URL.
```
Chaho12:
```suggestion
For example, when using a Kubernetes Service, the proxyTo URL might be `trino-backend-service.trino-namespace.svc.cluster.local:8083`
for communication between the Gateway and the Trino backend.
Meanwhile, the external URL for the same backend cluster could be `trino.domain.com`. 
If there is no user-accessible URL available for a backend cluster,
the external URL can be set to the Trino Gateway address itself or to the proxyTo URL.
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -81,8 +82,10 @@ public RoutingTargetResponse resolveRouting(HttpServletRequest request)
         RoutingTargetResponse routingTargetResponse = previousCluster.map(cluster -> {
             String routingGroup = queryId.map(routingManager::findRoutingGroupForQueryId)
                     .orElse("adhoc");
+            String externalUrl = queryId.map(routingManager::findExternalUrlForQueryId)
+                    .orElse(cluster);
```
Chaho12, line 86: You know that if externalUrl is null, it returns proxyTo instead at [code](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/config/ProxyBackendConfiguration.java#L52-L58) right? 
Can you explain why would it not be proxyTo if it fails to find?

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -81,8 +82,10 @@ public RoutingTargetResponse resolveRouting(HttpServletRequest request)
...
             return new RoutingTargetResponse(
-                    new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request)),
+                    new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request), externalUrl),
```
Chaho12, line 88: Adding external URL to RoutingDestination is meant only for logging at queryHistoryManger right?
And this is needed to show it at History page?

If that's the only use, then it may be more than needed to include it in every RoutingDestination, as this object is passed through multiple layers. Passing it through the entire routing stack, rather than fetching it only where needed, introduces extra coupling and data flow.

@vishalya Any thoughts on this?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
Chaho12: I would say single cache.

Queryid is unique, and external url is tightly coupled with query(id) so single cache is most efficient, in both atomicity, simplicity.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
raj-manvar: Actually, the externalUrl will default to proxyUrl if it's null [here](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/config/ProxyBackendConfiguration.java#L52-L58) so externalUrl can't be null herel

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
raj-manvar: Ah thanks, I had missed that! Yes it shouldn't be null here in that case. <br>  Correspondingly the routingGroup also shouldn't be null here, because it's default set to "adhoc" [here](https://github.com/trinodb/trino-gateway/blob/1f39b982b51d4de914524c881d3095673c76604a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L102). <br> 
Can you help confirm the same for routingGroup, I can update both to `queryId.map(routingManager::findRoutingGroupForQueryId).get()` and `queryId.map(routingManager::findExternalUrlForQueryId).get()`  correspondingly

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Sorry had missed this, but since externalUrl is default to proxyUrl [here](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/config/ProxyBackendConfiguration.java#L52-L58) this shouldn't be null here as well since the externalUrl won't be null in the query_history sql table as well

docs/routing-logic.md
raj-manvar: Sorry for the grammar mistakes, I should've put the content in some online grammar correction tool / AI beforehand.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
Chaho12: - if queryId is empty calling `.get()` on an empty Optional will throw a `NoSuchElementException`. 
- If queryId is present, then `findRoutingGroupForQueryId` or `findExternalUrlForQueryId` will be called.
If either of those methods returns null (which may happen as user could override this function), then `.map(...)` will convert the Optional into `Optional.ofNullable(null)`, which is `Optional.empty()`, so calling `.get()` will again throw a `NoSuchElementException`.

Unfortunatley, we can't simply update like that.

I followed the logic and noticed that routingGroup cannot be null. Must be a value. 

@andythsu can you check this logic?

raj-manvar: > If either of those methods returns null (which may happen as user could override this function)

Sorry I don't follow this part.  Those methods shouldn't return null as well if queryId is present. Since when the queryId was first received, routingGroup / externalUrl are inserted into the database [here](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java#L194). After that subsequent calls to findRoutingGroupForQueryId would return the values from the database.

raj-manvar: @Chaho12  Was thinking if it's okay to keep the default to proxyTo just to be safe.  Defaulting to proxyTo url is mostly harmless and will throw an exception incase the value is null for any reason. wdyt ?

Chaho12: Sorry for confusion. What i meant was that it should have default orElse value because routingGroup cannot be null. 
Yes proxyTo sounds ok

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
+## Routing URLs
+
+Each Trino backend to the Trino Gateway supports both a proxyTo URL and external URL.
```
andythsu:
```suggestion
Each Trino cluster configured with the Trino Gateway includes both a `proxyTo` URL and an `externalURL`
```

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
...
+While the proxy URL is used internally by the Gateway to route the requests based on routing rules and query identifiers,
```
andythsu:
```suggestion
The `proxyTo` URL is used internally by the Trino Gateway to route requests based on routing rules and query identifiers, 
```

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
+## Routing URLs
+
+Each Trino backend to the Trino Gateway supports both a proxyTo URL and external URL.
+While the proxy URL is used internally by the Gateway to route the requests based on routing rules and query identifiers, 
+the external URL represents a user accessible address for the Trino backend cluster,
```
andythsu:
```suggestion
whereas the `externalURL` serves as a UI-accessible or publicly reachable address for the Trino cluster, 
```

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
...
 ```
+
+## Routing URLs
+
+Each Trino backend to the Trino Gateway supports both a proxyTo URL and external URL.
+While the proxy URL is used internally by the Gateway to route the requests based on routing rules and query identifiers, 
+the external URL represents a user accessible address for the Trino backend cluster, 
+and is commonly used to access [Trino web UI](https://trino.io/docs/current/admin/web-interface.html)  
+
+For example, when using a Kubernetes Service, the proxyTo URL might be `trino-backend-service.trino-namespace.svc.cluster.local:8083`
```
andythsu:
```suggestion
For example, in a Kubernetes environment, the `proxyTo` URL might be `trino-backend-service.trino-namespace.svc.cluster.local:8083`
```

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
...
+for communication between the Gateway and the Trino backend.
```
andythsu:
```suggestion
for communication between the Trino Gateway and Trino clusters, and the external URL for the same backend cluster might be `trino.domain.com`.
```

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
...
+Meanwhile, the external URL for the same backend cluster could be `trino.domain.com`.
```
andythsu:
```suggestion
```

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
...
+If there is no user-accessible URL available for a backend cluster,
```
andythsu:
```suggestion
```
I don't think this line is necessary. I think this is more like implementation details.

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
...
+the external URL can be set to the Trino Gateway address itself or to the proxyTo URL.
```
andythsu:
```suggestion
```

andythsu (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -102,31 +121,41 @@ public void setRoutingGroupForQueryId(String queryId, String routingGroup)
     /**
      * Performs routing to an adhoc backend.
      */
-    public String provideAdhocCluster(String user)
+    public ProxyBackendConfiguration provideAdhocBackendConfiguration(String user)
```
vishalya: can this method be `private`?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -102,31 +121,41 @@ public void setRoutingGroupForQueryId(String queryId, String routingGroup)
...
+    {
+        return provideAdhocBackendConfiguration(user).getProxyTo();
     }
 
     /**
      * Performs routing to a given cluster group. This falls back to an adhoc backend, if no scheduled
      * backend is found.
      */
-    public String provideClusterForRoutingGroup(String routingGroup, String user)
+    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
```
vishalya, line 145: can this method be `private`?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -80,6 +81,19 @@ public String load(String queryId)
...
+                        .expireAfterAccess(30, TimeUnit.MINUTES)
+                        .build(
+                                new CacheLoader<String, String>()
+                                {
+                                    @Override
+                                    public String load(String queryId)
+                                    {
+                                        return findExternalUrlForUnknownQueryId(queryId);
+                                    }
+                                });
```
vishalya, line 97: I am with @Peiyingy, but it could be a different PR to have a single cache.

vishalya (changes requested):
Thanks for filling the gap @raj-manvar. Could you also please add some unit tests in [TestHaGatewayManager]( https://github.com/trinodb/trino-gateway/blob/a9b10e70b2f1d858d4d19d9e34627ba59d440ff2/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java#L39)
- default value and setting explicit values
- towards cache hits and misses

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Made this method private.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: No, this method is called here https://github.com/trinodb/trino-gateway/blob/0a18f61e0074e619b9b3e0840c3b9ecd445e76a4/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L103

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Thanks, created https://github.com/trinodb/trino-gateway/issues/702 for this. I'll try to pick this issue after this

docs/routing-logic.md
```diff
@@ -75,3 +75,13 @@ oauth2GatewayCookieConfiguration:
...
+
+## Routing URLs
+
+Each Trino cluster configured with the Trino Gateway includes both a `proxyTo` URL and an `externalURL`
+The `proxyTo` URL is used internally by the Trino Gateway to route requests based on routing rules and query identifiers,  
+whereas the `externalURL` serves as a UI-accessible or publicly reachable address for the Trino cluster, 
+and is commonly used to access [Trino web UI](https://trino.io/docs/current/admin/web-interface.html)  
+
+For example, in a Kubernetes environment, the `proxyTo` URL might be `trino-backend-service.trino-namespace.svc.cluster.local:8083`
+for communication between the Trino Gateway and Trino clusters, and the external URL for the same backend cluster might be `trino.domain.com`.
```
ebyhr: Wrap at 80 characters.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryHistoryManager.java
```diff
@@ -139,6 +142,17 @@ public void setRoutingGroup(String routingGroup)
             this.routingGroup = routingGroup;
         }
 
+        @JsonProperty
+        public String getExternalUrl()
+        {
+            return this.externalUrl;
```
ebyhr: Remove redundant `this.`

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -80,6 +81,19 @@ public String load(String queryId)
                                         return findRoutingGroupForUnknownQueryId(queryId);
                                     }
                                 });
+        queryIdExternalUrlCache =
+                CacheBuilder.newBuilder()
+                        .maximumSize(10000)
+                        .expireAfterAccess(30, TimeUnit.MINUTES)
+                        .build(
+                                new CacheLoader<String, String>()
```
ebyhr: Remove redundant `String, String`

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -145,6 +174,18 @@ public String findBackendForQueryId(String queryId)
         return backendAddress;
     }
 
+    public String findExternalUrlForQueryId(String queryId)
```
ebyhr, line 180: Add `@Nullable` annotation.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -145,6 +174,18 @@ public String findBackendForQueryId(String queryId)
...
+    {
+        String externalUrl = null;
+        try {
+            externalUrl = queryIdExternalUrlCache.get(queryId);
+        }
+        catch (ExecutionException e) {
+            log.error("Exception while loading queryId from cache %s", e.getLocalizedMessage());
```
ebyhr: Change to warn.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistory.java
```diff
@@ -0,0 +1,202 @@
...
+import org.testcontainers.containers.JdbcDatabaseContainer;
+
+import java.util.List;
+import java.util.Optional;
+
+import static io.trino.gateway.ha.TestingJdbcConnectionManager.createTestingJdbcConnectionManager;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(Lifecycle.PER_CLASS)
+abstract class TestExternalUrlQueryHistory
```
ebyhr: BaseExternalUrlQueryHistoryTest

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistory.java
```diff
@@ -0,0 +1,202 @@
...
+{
+    protected final JdbcDatabaseContainer<?> container = startContainer();
```
ebyhr: private

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistory.java
```diff
@@ -0,0 +1,202 @@
...
+        queryDetail.setExternalUrl("https://external-gateway.example.com");
+        queryDetail.setCaptureTime(System.currentTimeMillis());
+
+        queryHistoryManager.submitQueryDetail(queryDetail);
+
+        // Verify the query was stored with external URL
+        queryDetails = queryHistoryManager.fetchQueryHistory(Optional.empty());
+        assertThat(queryDetails).hasSize(1);
+
+        QueryHistoryManager.QueryDetail retrievedQuery = queryDetails.get(0);
```
ebyhr: `getFirst()`

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistory.java
```diff
@@ -0,0 +1,202 @@
...
+import static io.trino.gateway.ha.TestingJdbcConnectionManager.createTestingJdbcConnectionManager;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(Lifecycle.PER_CLASS)
+abstract class TestExternalUrlQueryHistory
+{
+    protected final JdbcDatabaseContainer<?> container = startContainer();
+    private QueryHistoryManager queryHistoryManager;
+
+    protected abstract JdbcDatabaseContainer<?> startContainer();
```
ebyhr: Please add a constructor with `JdbcDatabaseContainer<?>` instead of this method.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Thanks, Created https://github.com/trinodb/trino-gateway/issues/702 for this for future.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Even tho the return value is initialized with null, we don't expect the final return value to be null here.  Also Guava's LoadingCache doesn't allow null keys or values.

webapp/src/components/history.tsx
raj-manvar: Resolved this conflict with main branch.

vishalya (approved):
Approved.
Does it good to the rest of reviewers before I merge it?

ebyhr:
> Does it good to the rest of reviewers before I merge it?

Yes, please don't merge this PR yet.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
ebyhr: I don't understand what you mean.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistoryMySql.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class TestExternalUrlQueryHistoryMySql
+        extends BaseExternalUrlQueryHistoryTest
+{
+    public TestExternalUrlQueryHistoryMySql()
+    {
+        super(createMySqlContainer());
+    }
+
+    private static JdbcDatabaseContainer<?> createMySqlContainer()
```
ebyhr: Remove redundant `createMySqlContainer`, `createOracleContainer`, `createPostgresqlContainer` methods. We can start the container in `BaseExternalUrlQueryHistoryTest`.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -82,6 +84,16 @@ void testGatewayManager()
...
                 .containsExactly("adhoc");
+
+        // Test default externalUrl to proxyUrl
+        ProxyBackendConfiguration adhoc2 = new ProxyBackendConfiguration();
+        adhoc2.setActive(true);
+        adhoc2.setRoutingGroup("adhoc2");
+        adhoc2.setName("adhoc2");
+        adhoc2.setProxyTo("adhoc2.trino.gateway.io");
+        haGatewayManager.addBackend(adhoc2);
+        final ProxyBackendConfiguration backendConfigurationAdhoc2 = haGatewayManager.getActiveBackends("adhoc2").getFirst();
```
ebyhr: Remove `final`.

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseExternalUrlQueryHistoryTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+@TestInstance(Lifecycle.PER_CLASS)
+abstract class BaseExternalUrlQueryHistoryTest
+{
+    private final JdbcDatabaseContainer<?> container;
+    private QueryHistoryManager queryHistoryManager;
+
+    protected BaseExternalUrlQueryHistoryTest(JdbcDatabaseContainer<?> container)
+    {
+        this.container = container;
```
ebyhr: Add `requireNonNull`.

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseExternalUrlQueryHistoryTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @BeforeAll
+    void setUp()
```
ebyhr: Merge this method into the constructor.

ebyhr (changes requested):
Please squash commits and separate commits into "Use external url for query routing history" and other style changes.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
raj-manvar: Can you clarify why should we add `@Nullable` annotation here ?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistoryMySql.java
raj-manvar: How would we do that in a clean way ? :thinking: Each of the tests in BaseExternalUrlQueryHistoryTest runs for each of the Mysql, Oracle and Postgresql containers. <br> There would be lot of code duplications if we keep merge in a single BaseExternalUrlQueryHistoryTest iiuc. Can you please give an example code snippet on how you envision each of the test in BaseExternalUrlQueryHistoryTest to run for all mysql, postgres and oracle containers

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseExternalUrlQueryHistoryTest.java
raj-manvar: Sorry, I don't follow purpose of this. Maybe discussion at https://github.com/trinodb/trino-gateway/pull/693#discussion_r2162790570 would help clarify.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistoryMySql.java
ebyhr: I pushed a new commit. Please take a look at that.

ebyhr:
Please rebase on main to resolve conflicts.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
ebyhr: > we don't expect the final return value to be null here.

The method returns null in case `ExecutionException` happens.

raj-manvar: Ah Thanks okay. I updated this and other similar methods of this class with `@Nullable` annotation. <br> Taking a looking the relevant code that's calling these methods, all seem to be handling the case when return value is null.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistoryMySql.java
raj-manvar: Thanks a lot!  Sorry I had misunderstood this as removing the entire TestExternalUrlQueryHistoryMySql classes instead. Thanks again.

raj-manvar:
Thanks @ebyhr for the help & feedback! I have rebased with main branch and also squashed all commits into a single commit

docs/routing-logic.md
```diff
@@ -75,3 +75,16 @@ oauth2GatewayCookieConfiguration:
     - "/custom/logout"
   lifetime: "5m"
 ```
+
```
ebyhr, line 78: @mosabua Do you want to review this doc?

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryHistoryManager.java
```diff
@@ -65,7 +68,7 @@ public int compareTo(QueryDetail o)
         @JsonProperty
         public String getQueryId()
         {
-            return this.queryId;
+            return queryId;
```
ebyhr, line 71: This change is unrelated to "use external url for query routing history". Please separate a commit.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -59,7 +61,7 @@ public RoutingManager(GatewayBackendManager gatewayBackendManager, QueryHistoryM
                         .maximumSize(10000)
                         .expireAfterAccess(30, TimeUnit.MINUTES)
                         .build(
-                                new CacheLoader<String, String>()
+                                new CacheLoader<>()
```
ebyhr, line 64: This change is unrelated to "use external url for query routing history". Please separate a commit.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalUrlQueryHistoryPostgresql.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import org.testcontainers.containers.PostgreSQLContainer;
+
+public class TestExternalUrlQueryHistoryPostgresql
```
ebyhr:
```suggestion
public class TestExternalUrlQueryHistoryPostgreSql
```

ebyhr (commented):
> use external url for query routing history

Please capitalize the commit title 
https://trino.io/development/process.html#pull-request-and-commit-guidelines-

raj-manvar:
> > use external url for query routing history
> 
> Please capitalize the commit title https://trino.io/development/process.html#pull-request-and-commit-guidelines-

Thanks! I'll be careful for future commits. Updated current MR commit messages.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -87,7 +87,7 @@ public String load(String queryId)
...
```
ebyhr, line 90: This change at L90 should go to the 1st commit.

webapp/src/components/history.tsx
Chaho12, line 1: do we now need `mapping[backend.proxyTo] = backend.name;`? i think we can now remove that part in this code
```
        for (let index = 0; index < data.length; index++) {
          const backend = data[index] as BackendData;
          mapping[backend.name] = backend.proxyTo;
          mapping[backend.proxyTo] = backend.name;
          mapping[backend.externalUrl] = backend.name;
        }
```

raj-manvar: I think it's still used in getting the backend Name from backendUrl at https://github.com/trinodb/trino-gateway/blob/main/webapp/src/components/history.tsx#L125 <br>
It's also getting used at https://github.com/trinodb/trino-gateway/blob/main/webapp/src/components/history.tsx#L99, but we can replace this usage with just `b.name` ?
We could remove the remaining two mapping ( backend.name -> proxyTo and externalUrl -> name ), should we do it as part of this pull request though, as it's not strictly related to this change ? or create a separate PR ?

Chaho12: A separate PR sounds better :) Thx for checking logic.

raj-manvar: Thanks, I created https://github.com/trinodb/trino-gateway/issues/711 so that we don't miss this.

raj-manvar:
> I think you misunderstood ebyhr comment on capitalizing commit message. We only capitalize the first letter of the subject line. e.g) `Use external URL for query routing history`, `Remove inferred types and this references`

Ah thanks okay.  I updated the commit messages accordingly.

raj-manvar:
@ebyhr Does this good look to merge now ?

docs/routing-logic.md
mosabua: Docs changes look good.

mosabua:
I checked docs now .. all good there. @andythsu @ebyhr @vishalya @Chaho12 .. is this good to go? Please update your review notes.

andythsu (approved):
🚀 !

raj-manvar:
@ebyhr @vishalya  can you please help merge this.

Chaho12:
Will rebase and merge in few hours if you don't have any further comments @ebyhr :)

Chaho12:
Just for safety, I checked that rebase has no conflict :)  so will rebase and merge

image

PR #695, opened 2025-06-05
Use Airlift Duration in MonitorConfiguration

andythsu:
## Description

Use Airlift Duration in MonitorConfiguration

## Release notes

```markdown
* Fix monitor.taskDelaySeconds to use Airlift Duration
```

### DIFF
--- a/docs/operation.md
+++ b/docs/operation.md
@@ -49,11 +49,11 @@ modules:
 ```
 - The router operates based on the stats it receives from the clusters, such as 
 the number of queued and running queries. These values are retrieved at regular 
-intervals. This interval can be configured by setting `taskDelaySeconds` under
-`monitor` section in the config file. The default interval is 60
+intervals. This interval can be configured by setting `taskDelay` under
+`monitor` section in the config file. The default interval is 1 minute
 ```yaml
 monitor:
-  taskDelaySeconds: 10
+  taskDelay: 1m
 ```
 
 ## Monitoring <a name="monitoring"></a>
--- a/docs/routers.md
+++ b/docs/routers.md
@@ -50,7 +50,7 @@ Use the following steps to create a new router:
 - The router listens to the list of `ClusterStats` via the`updateBackEndStats`
   method.
 - This method is called on regular intervals defined in the config 
-  parameter `monitor=>taskDelaySeconds`.
+  parameter `monitor=>taskDelay`.
 - Each element in the list corresponds to each backend cluster.
 - Only the stats from the healthy cluster are reported, unhealthy clusters are
   not included in the list. If you have three cluster backends and one is
@@ -70,9 +70,6 @@ backendState:
 
 clusterStatsConfiguration:
   monitorType: UI_API
-
-monitor:
-  taskDelaySeconds: 10
 
 modules:
   - io.trino.gateway.ha.module.QueryCountBasedRouterProvider
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
@@ -15,6 +15,7 @@ package io.trino.gateway.ha.clustermonitor;
 
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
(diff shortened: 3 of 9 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/config/MonitorConfiguration.java
```diff
@@ -15,15 +15,14 @@
...
-import io.trino.gateway.ha.clustermonitor.ActiveClusterMonitor;
 
 import java.util.Map;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class MonitorConfiguration
 {
-    private int taskDelaySeconds = ActiveClusterMonitor.MONITOR_TASK_DELAY_SECONDS;
+    private Duration taskDelay = new Duration(60, SECONDS);
```
ebyhr:
```java
private Duration taskDelay = new Duration(1, MINUTES);
```

docs/operation.md
```diff
@@ -49,11 +49,11 @@ modules:
 ```
 - The router operates based on the stats it receives from the clusters, such as 
 the number of queued and running queries. These values are retrieved at regular 
-intervals. This interval can be configured by setting `taskDelaySeconds` under
-`monitor` section in the config file. The default interval is 60
+intervals. This interval can be configured by setting `taskDelay` under
+`monitor` section in the config file. The default interval is 60 seconds
```
ebyhr: 60 seconds -> 1 minute

docs/operation.md
```diff
@@ -49,11 +49,11 @@ modules:
...
 ```yaml
 monitor:
-  taskDelaySeconds: 10
+  taskDelay: "60s"
```
ebyhr: 60s -> 1m

ebyhr:

docs/routers.md
```diff
@@ -72,7 +72,7 @@ clusterStatsConfiguration:
   monitorType: UI_API
 
 monitor:
-  taskDelaySeconds: 10
+  taskDelay: "10s"
```
Chaho12: if default is 1 min, whow about setting it to 1 min too in doc?

docs/operation.md
```diff
@@ -49,11 +49,11 @@ modules:
...
 - The router operates based on the stats it receives from the clusters, such as 
 the number of queued and running queries. These values are retrieved at regular 
-intervals. This interval can be configured by setting `taskDelaySeconds` under
-`monitor` section in the config file. The default interval is 60
+intervals. This interval can be configured by setting `taskDelay` under
+`monitor` section in the config file. The default interval is 1 minute
 ```yaml
 monitor:
-  taskDelaySeconds: 10
+  taskDelay: "1m"
```
Chaho12: do you need the double quotes? i thought no quotes needed as it is consired as string anyway (no special characters) and it is better style to not use double quotes for simple strings like this.

> Seems to me that the best approach would be to not use quotes unless you have to, and then to use single quotes unless you specifically want to process escape codes.

https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml

Chaho12 (approved):
LGTM 👍

docs/routers.md
andythsu: sure thing

andythsu: actually I will remove `monitor` section from here because it's irrelevant in routers doc

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ActiveClusterMonitor.java
```diff
@@ -89,7 +88,7 @@ public void start()
             catch (Exception e) {
                 log.error(e, "Error performing backend monitor tasks");
             }
-        }, 0, taskDelaySeconds, TimeUnit.SECONDS);
+        }, 0, (long) taskDelay.getValue(), taskDelay.getUnit());
```
Chaho12, line 91:
```suggestion
        }, 0, taskDelay.toMillis(), TimeUnit.MILLISECONDS);
```
If taskDelay is 1.5m, wouldn't `(long) taskDelay.getValue()` truncate the faction (turning 1.5 into 1)? causing a loss of precision?

PR #707, opened 2025-06-25
Replace H2 use for testing with PostgreSQL

Akanksha-kedia:
…ronment-Specific Configuration

## Description

- Removed H2 database dependency from pom.xml

- Enhanced DataStoreConfiguration to support environment-specific database configurations

- Updated FlywayMigration to properly handle different database types

- Modified config.yaml to include environment-based database settings

- Implemented environment detection in HaGatewayLauncher

- Migrated all tests to use PostgreSQL with testcontainers instead of H2

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
@@ -27,18 +27,6 @@ public interface ExactMatchSourceSelectorsDao
             """)
     List<ExactMatchSourceSelectors> findAll();
 
-    @SqlQuery("""
-            SELECT * FROM exact_match_source_selectors
-            WHERE
-                resource_group_id = :resourceGroupId
-                AND update_time   = :updateTime
-                AND source        = :source
-                AND environment   = :environment
-                AND query_type    = :queryType
-            LIMIT 1
-            """)
-    ExactMatchSourceSelectors findFirst(@BindBean ResourceGroupsManager.ExactSelectorsDetail exactSelectors);
-
     @SqlUpdate("""
             INSERT INTO exact_match_source_selectors
             (resource_group_id, update_time, source, environment, query_type)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaResourceGroupsManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaResourceGroupsManager.java
@@ -256,17 +256,6 @@ public class HaResourceGroupsManager
                 .collect(toImmutableList());
     }
 
-    /**
-     * Gets exact match source selector from db.
-     */
-    @Override
-    public ExactSelectorsDetail getExactMatchSourceSelector(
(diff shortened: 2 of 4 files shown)


Akanksha-kedia:
@mosabua please review

gateway-ha/pom.xml
```diff
@@ -321,14 +321,6 @@
             <scope>runtime</scope>
         </dependency>
 
-        <!-- Test deps -->
-        <dependency>
-            <groupId>com.h2database</groupId>
-            <artifactId>h2</artifactId>
-            <version>1.4.192</version>
-            <scope>test</scope>
```
xkrogen: Can you describe more about the motivation behind this change? Why do we want to use Postgres instead of H2?

gateway-ha/config.yaml
```diff
@@ -12,6 +12,20 @@ dataStore:
...
+      password: P0stG&es
+    testing:
+      jdbcUrl: jdbc:postgresql://localhost:5432/trino_gateway_test
+      user: trino_gateway_test_admin
+      password: P0stG&es
+    production:
+      jdbcUrl: jdbc:postgresql://db.example.com:5432/trino_gateway_prod
+      user: trino_gateway_prod_admin
+      password: ${ENV:PROD_DB_PASSWORD}
+      queryHistoryHoursRetention: 168  # 7 days for production
```
xkrogen: I don't think listing out configurations for all environments in the test same place is the correct strategy. Any number of configs, whether for the dataStore or otherwise, may need to be adjusted across environments. This is the whole reason we support the environment variable replacement. Or, in many cases, you will already have some kind of configuration management system that allows you to control per-environment configs.

Assuming you want to use environment variables to customize the environment, instead of listing out all possible environments, this should just look something like:
```yaml
dataStore:
  jdbcUrl: jdbc:postgresql://${ENV:DB_HOSTNAME}:5432/${ENV:DB_NAME}
  user: ${ENV:DB_USER}
  password: ${ENV:DB_PASSWORD}
  driver: org.postgresql.Driver
  queryHistoryHoursRetention: 24
```

Even taking this into account, we shouldn't put all of these changes into this config file. This one is meant just as a reference -- it is expected that each installation will customize with their own configs, which will be environment-dependent.

gateway-ha/pom.xml
Akanksha-kedia: Screenshot 2025-06-26 at 3 34 05 PM

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -70,16 +70,28 @@ public static void prepareMockBackend(
...
+            handle.createUpdate(HaGatewayTestUtils.getResourceFileContent("gateway-ha-persistence-postgres.sql"))
                     .execute();
         }
     }
 
+    /**
+     * @deprecated Use seedRequiredDataPostgres instead
+     */
+    @Deprecated
+    public static void seedRequiredData(String h2DbFilePath)
```
ebyhr: Remove unused methods. Trino Gateway isn't a library. We can remove such methods without deprecation.

ebyhr:
Could you rebase on main to resolve conflicts?

Issue #708, opened 2025-06-25
Flyway Migration Errors


yardenc2003:
I’m trying to run trino gateway v15 alongside a trino cluster v476, both configured to use the same postgres db for resource group management.
When starting the trino cluster, I encounter the following flyway error:

```
2025-06-24T12:32:10.093Z	ERROR	main	io.trino.server.Server	Validate failed: Migrations have failed validation
Migration checksum mismatch for migration version 1
-> Applied to database : 542221347
-> Resolved locally    : -25491552
Either revert the changes to the migration, or run repair to update the schema history.
Need more flexibility with validation rules? Learn more: https://rd.gt/3AbJUZE
org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
Migration checksum mismatch for migration version 1
-> Applied to database : 542221347
-> Resolved locally    : -25491552
Either revert the changes to the migration, or run repair to update the schema history.
Need more flexibility with validation rules? Learn more: https://rd.gt/3AbJUZE
	at org.flywaydb.core.Flyway.lambda$migrate$1(Flyway.java:201)
	at org.flywaydb.core.FlywayExecutor.execute(FlywayExecutor.java:210)
	at org.flywaydb.core.Flyway.migrate(Flyway.java:188)
	at io.trino.plugin.resourcegroups.db.FlywayMigration.migrate(FlywayMigration.java:58)
	at io.trino.plugin.resourcegroups.db.DbResourceGroupConfigurationManagerFactory.create(DbResourceGroupConfigurationManagerFactory.java:44)
	at io.trino.execution.resourcegroups.InternalResourceGroupManager.setConfigurationManager(InternalResourceGroupManager.java:170)
	at io.trino.execution.resourcegroups.InternalResourceGroupManager.loadConfigurationManager(InternalResourceGroupManager.java:154)
	at io.trino.server.Server.doStart(Server.java:163)
	at io.trino.server.Server.lambda$start$0(Server.java:90)
	at io.trino.$gen.Trino_476____20250624_123202_1.run(Unknown Source)
	at io.trino.server.Server.start(Server.java:90)
	at io.trino.server.TrinoServer.main(TrinoServer.java:37)
I came across this PR and tried to set resource-groups.db-migrations-enabled=false , but then I get errors related to the incompatible schema of the resource groups-related tables, such as:
[...]

Issue #656, opened 2025-04-01
Remove resource groups


mosabua:
As discussed in the Trino Gateway dev sync there is a mismatch in the location of the code for resource groups (in Trino), the use (mostly in Trino and NOT in Trino Gateway) and the UI for editing (in Trino Gateway).

Given this situation, we plan to remove the UI code from Trino Gateway and ideally someone can port the UI over to Trino, specifically the preview UI in Trino since that has similar tooling.

Here are some more details from @willmostly :
---
* Resource Groups are not used by the Trino Gateway, so they cannot be properly tested
* Managing a Trino component through the Gateway introduces coupling that can cause issues such as https://github.com/trinodb/trino-gateway/issues/637
* Our current test framework has no capability for testing Trino Gateway against different Trino versions, so we will be unaware of any compatibility issues
* The existing implementation can create an incorrect impression that Trino Gateway uses resource groups for load balancing and routing, which it does not

Considering these factors, the convenience of the resource group UI does not justify the maintenance overhead and potential for issues.
---

If we want to have something akin to resource groups for different cluster in Trino Gateway is a separate topic. The answer to that is probably yes in the long run .. but it would NOT be based on the resource groups in Trino itself.

PR #713, opened 2025-06-26
Align Resource Group Tables with Trino v476

yardenc2003:
## Description

This PR updates the resource groups–related tables to match the schemas used by Trino (v476).

Since version 14, automatic database migrations have been enabled in the gateway (I believe in [this PR](https://github.com/trinodb/trino-gateway/pull/575)).
This causes conflicts (different Flyway migration checksums) with the automatic migrations triggered during Trino’s startup. As a result, when sharing the database between the gateway and one or more Trino instances, migration immediately fails.

If we decide to enable only the gateway migrations, we must ensure the schemas match Trino’s. Otherwise, it fails due to missing columns.  This PR addresses that schema alignment.

### a Few Disclaimers

* I believe we need to **disable** one of the migrations (either Trino's or the gateway's) or at least document this. Currently, it does not work with the default configuration if resource groups are enabled and Trino shares the database with the gateway.
* I have not verified compatibility with older Trino versions that haven’t updated their schemas while the gateway has, only with v476 for the different db implementations.
* I had to reduce the maximum length of the environment column in `exact_match_source_selectors` in Oracle to match Trino’s schema. This change might be unsafe and prone to failures.
* I have not updated either the UI forms for resource groups/selectors creation or the underlying SQL statements (DAO classes). I wasn’t sure if this effort was worthwhile, given that this feature is about to be removed from the gateway and the existing implementation has issues updating via the UI.

Refers [this issue](https://github.com/trinodb/trino-gateway/issues/708).

## Release notes

( ) Release notes are required, with the following suggested text.

### DIFF
--- a/gateway-ha/src/main/resources/mysql/V3__align_with_trino_476.sql
+++ b/gateway-ha/src/main/resources/mysql/V3__align_with_trino_476.sql
@@ -1,0 +1,51 @@
+-- Align `exact_match_source_selectors` table
+
+-- Adjust primary key
+ALTER TABLE exact_match_source_selectors
+  DROP PRIMARY KEY;
+
+ALTER TABLE exact_match_source_selectors
+  ADD PRIMARY KEY (environment, source(128), resource_group_id);
+
+-- Modify `query_type` to be nullable
+ALTER TABLE exact_match_source_selectors
+  MODIFY COLUMN query_type VARCHAR(512) DEFAULT NULL;
+
+
+-- Align `resource_groups` table
+
+-- Drop the unique constraint on `name`
+ALTER TABLE resource_groups
+  DROP INDEX name;
+
+-- Make `soft_memory_limit` nullable
+ALTER TABLE resource_groups
+  MODIFY COLUMN soft_memory_limit VARCHAR(128) DEFAULT NULL;
+
+-- Drop and recreate the `parent` foreign key with ON DELETE CASCADE
+ALTER TABLE resource_groups
+  DROP FOREIGN KEY resource_groups_ibfk_1;
+
+ALTER TABLE resource_groups
+  ADD CONSTRAINT resource_groups_ibfk_1
+  FOREIGN KEY (parent) REFERENCES resource_groups(resource_group_id)
+  ON DELETE CASCADE;
+
+
+-- Align `selectors` table
+
+-- Add missing columns
+ALTER TABLE selectors
+  ADD COLUMN user_group_regex VARCHAR(2048) DEFAULT NULL,
+  ADD COLUMN original_user_regex VARCHAR(512) DEFAULT NULL,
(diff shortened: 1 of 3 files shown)


oneonestar:
We should allow gateway to be used with old and new version of Trino (or a mix of both).
I'd prefer disable the migration on Resource Group tables.

xkrogen:
> I'd prefer disable the migration on Resource Group tables.

+1 -- the owner of these tables is the Trino Coordinator, not the GW. The GW should not attempt to perform schema changes on these tables.

> We should allow gateway to be used with old and new version of Trino (or a mix of both). 

+1 the GW should be able to handle multiple schema versions of this DB

mosabua:
I agree that we should disable automatic updates from Trino Gateway. 

I would go one step further .. and we agreed on that already .. Trino Gateway should not have UI to manage the resource group data .. it should move to the Trino UI, specifically the new preview UI.

Separately if we need some sort of concept and tooling around resource use in Trino Gateway (and we do) we should implement something new that works properly for multiple clusters and more.

mosabua:
So what is the plan here? Are the migrations already disabled by default and we can bring this change in as it stands .. or do we need to disable the migration from Trino Gateway as well as part of this PR in a separate commit?

yardenc2003:
> So what is the plan here? Are the migrations already disabled by default and we can bring this change in as it stands .. or do we need to disable the migration from Trino Gateway as well as part of this PR in a separate commit?

Flyway table migration is still enabled by default - https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java#L23

Also, regarding the PR changes - how do we ensue the shared table schemas stay aligned with those in Trino?

mosabua:
Can you rebase this PR so we can merge .. and then ideally as a follow up we remove the resource group tables with a migration and also remove the UI for it.

yardenc2003:
@mosabua I think these changes should not be merged as you are planning on removing the resource groups feature from the gateway anyway. The right thing to do is probably abandon the resource-groups-related tables by the gateway and let trino manage them. For now I manage to overcome the incompatibility caused by the resource groups tables flyway migration, by using a separate databases for trino & gateway so they don't interfere each other, and the gateway resource groups tables are empty as they don't have affect on the trino clusters.

Would you like me to open a PR for removing the resource groups feature from gateway completely or is it blocked by something else?

mosabua:
If you could create a PR that removes the resource group stuff that would be AWESOME @yardenc2003

mosabua:
We discussed this also in the dev sync and agreed to go ahead with closing this and waiting for a PR from @yardenc2003 to remove the resource groups stuff to implement https://github.com/trinodb/trino-gateway/issues/656

PR #716, opened 2025-06-28
Remove unused method

ebyhr:
## Description

This method was called only from tests. 

## Release notes


mosabua (approved):
This is good to go .. eventually we want to get rid of the whole resource groups as discussed but we might want a replacement UI in Trino itself before that..

PR #718, opened 2025-07-01
logout when restart the server

### LINKED ISSUE #717: Trino Gateway page not getting logged out even after restart of gateway or forever until we manually click on logout
As per the Trino UI, whenever Trino gateway service is restarted it should logout the session. The session should have some duration if it is ideal the it should logout.

sug-ghosh:
## Description

fixes https://github.com/trinodb/trino-gateway/issues/717

Features Implemented

1. 15-minute Session Timeout

Backend: JWT tokens now expire after 15 minutes instead of 24 hours
Frontend: Client-side session management tracks user activity and automatically logs out after 15 minutes of inactivity
Activity Detection: Mouse movements, clicks, keyboard input, scrolling, and touch events reset the timeout

2. Server Restart Detection

Backend: JWT tokens include a server start timestamp claim
Frontend: Checks server start time on every API call to detect server restarts immediately
Automatic Logout: Users are automatically logged out on any action after server restart

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
@@ -185,4 +185,18 @@ public class LoginResource
         }
         return Response.ok(Result.ok("Ok", loginType)).build();
     }
+
+    @POST
+    @Path("serverInfo")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response serverInfo()
+    {
+        long serverStartTime = System.currentTimeMillis();
+        if (formAuthManager != null) {
+            serverStartTime = formAuthManager.getServerStartTime();
+        }
+        Map<String, Object> serverInfo = Map.of("serverStart", serverStartTime);
+        return Response.ok(Result.ok(serverInfo)).build();
+    }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
@@ -19,6 +19,7 @@ import com.auth0.jwt.exceptions.JWTCreationException;
 import com.auth0.jwt.interfaces.Claim;
 import com.auth0.jwt.interfaces.DecodedJWT;
 import io.airlift.log.Logger;
+import io.airlift.units.Duration;
 import io.trino.gateway.ha.config.FormAuthConfiguration;
 import io.trino.gateway.ha.config.LdapConfiguration;
 import io.trino.gateway.ha.config.UserConfiguration;
@@ -26,10 +27,13 @@ import io.trino.gateway.ha.domain.Result;
 import io.trino.gateway.ha.domain.request.RestLoginRequest;
 import io.trino.gateway.ha.security.util.BasicCredentials;
 
+import java.time.Instant;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
@@ -38,6 +42,8 @@ import static java.util.Locale.ENGLISH;
 public class LbFormAuthManager
 {
     private static final Logger log = Logger.get(LbFormAuthManager.class);
+    private static final long SERVER_START_TIME = System.currentTimeMillis();
+    private static final Duration DEFAULT_SESSION_TIMEOUT = new Duration(30, TimeUnit.MINUTES);
     /**
      * Cookie key to pass the token.
      */
@@ -45,6 +51,7 @@ public class LbFormAuthManager
     private final Map<String, UserConfiguration> presetUsers;
     private final Map<String, String> pagePermissions;
     private final LbLdapClient lbLdapClient;
+    private final Duration sessionTimeout;
 
     public LbFormAuthManager(FormAuthConfiguration configuration,
             Map<String, UserConfiguration> presetUsers,
@@ -58,9 +65,12 @@ public class LbFormAuthManager
         if (configuration != null) {
             this.lbKeyProvider = new LbKeyProvider(configuration
                     .getSelfSignKeyPair());
+            this.sessionTimeout = configuration.getSessionTimeout() != null ?
+                    configuration.getSessionTimeout() : DEFAULT_SESSION_TIMEOUT;
         }
         else {
             this.lbKeyProvider = null;
+            this.sessionTimeout = DEFAULT_SESSION_TIMEOUT;
         }
 
         if (configuration != null && configuration.getLdapConfigPath() != null) {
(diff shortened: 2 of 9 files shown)


sug-ghosh:
@mosabua @ebyhr please review this pr for logout users when Trino-gateway server restarts.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -38,6 +40,8 @@
 public class LbFormAuthManager
 {
     private static final Logger log = Logger.get(LbFormAuthManager.class);
+    private static final long SERVER_START_TIME = System.currentTimeMillis();
+    private static final int SESSION_TIMEOUT_MINUTES = 15;
```
Chaho12: Is there a specific reason for 15 min? If not, why don't we make this configurable? + docs would be welcome :)

sug-ghosh: @Chaho12 mostly pentester recommend 15 minutes for UI session. That's why. otherwise I can make it configurable. and which docs I should update?

Chaho12: Since Tomcat, Spring is 30 min, i thought 30 min was default. 
https://trinodb.github.io/trino-gateway/security/ this page sounds fit.

sug-ghosh: @Chaho12 okay, I will make this 30. and do I make this configurable as well?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
```diff
@@ -185,4 +185,26 @@ else if (oauthManager != null) {
...
+    @Path("serverInfo")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response serverInfo()
+    {
+        long serverStartTime = System.currentTimeMillis();
+        if (formAuthManager != null) {
+            // Get server start time from form auth manager
+            try {
+                java.lang.reflect.Field field = formAuthManager.getClass().getDeclaredField("SERVER_START_TIME");
```
andythsu: let's not use reflection for this

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -38,6 +40,8 @@
 public class LbFormAuthManager
 {
     private static final Logger log = Logger.get(LbFormAuthManager.class);
+    private static final long SERVER_START_TIME = System.currentTimeMillis();
```
andythsu, line 45: if this is intended to be shared by multiple classes, we can move it out from this class and then create a separate config class. OR we can create a getter for this field

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
andythsu: let's make this configurable

andythsu: also, please use airlift's `Duration` class

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -105,6 +109,21 @@ public Optional<Map<String, Claim>> getClaimsFromIdToken(String idToken)
             DecodedJWT jwt = JWT.decode(idToken);
 
             if (LbTokenUtil.validateToken(idToken, lbKeyProvider.getRsaPublicKey(), jwt.getIssuer(), Optional.empty())) {
+                // Check if token was issued before server restart
+                Claim serverStartClaim = jwt.getClaim("server_start");
```
andythsu: can we wrap it in `Optional` so it's clear that it's nullable?

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -105,6 +109,21 @@ public Optional<Map<String, Claim>> getClaimsFromIdToken(String idToken)
...
+                if (serverStartClaim != null && !serverStartClaim.isNull()) {
+                    long tokenServerStart = serverStartClaim.asLong();
+                    if (tokenServerStart != SERVER_START_TIME) {
+                        log.info("Token invalidated due to server restart");
+                        return Optional.empty();
+                    }
+                }
+                // Check token expiration
+                Date expiresAt = jwt.getExpiresAt();
```
andythsu: can we wrap it in Optional so it's clear that it's nullable?

webapp/src/api/base.ts
```diff
@@ -104,6 +111,26 @@ export class ClientApi {
...
+        const accessStore = useAccessStore.getState();
+        if (accessStore.token) {
+            try {
+                const isValid = await accessStore.checkTokenValidity();
+                if (!isValid) {
+                    throw new Error('Token validation failed');
+                }
+            } catch (error) {
+                // Token validation failed, user will be logged out
+                throw error;
```
andythsu, line 129: Are we doing anything in catch? Otherwise it may be unnecessary to catch the error

webapp/src/store/access.ts
```diff
@@ -78,6 +81,51 @@ export const useAccessStore = create<AccessControlStore>()(
...
+            try {
+                const serverInfo = await serverInfoApi();
+                if (sessionManager.checkServerRestart(token, serverInfo.serverStart)) {
+                    console.log('Server restart detected, logging out');
+                    get().logout();
+                    return false;
+                }
+            } catch (error) {
+                console.error('Error checking server info:', error);
+                // Don't logout on API error, just continue
```
andythsu, line 124: should we log out here? technically we should never end up in this state, but if we do, it means the server is having issues.

andythsu (changes requested):

webapp/src/store/access.ts
sug-ghosh: @andythsu No, we are not logging out here, we are logging the server error here

webapp/src/api/base.ts
sug-ghosh: @andythsu  yes, we are throwing error as token validation failed and user will be logged out in that case

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
sug-ghosh: @andythsu I have resolved other comments in this forced pushed pr.

sug-ghosh:
@andythsu can you please review once?

mosabua:
As it stands this change is not useful. Users should NOT be logged out just because one Trino Gateway server shuts down. Trino Gateway can run as a cluster and users should not be logged out but instead remain active and continue to be able to use everything.

sug-ghosh:
> As it stands this change is not useful. Users should NOT be logged out just because one Trino Gateway server shuts down. Trino Gateway can run as a cluster and users should not be logged out but instead remain active and continue to be able to use everything.

Hello @mosabua , understand, will it be helpful if I discard the changes related to logout in server detection and keep the changes of 30 minute session timeout if user remains idle. Can this be taken if Trino Gateway Run as a cluster?

mosabua:
Yes .. I think an idle logout might still be a good thing .. as long as it also works if Trino Gateway runs as cluster

sug-ghosh:
Hello @mosabua if you suggesting Gateway run as a cluster means, HA proxy above gateway cluster, then the above changes works fine. I have already validated that. If one of the server restarts it does not throw away users logged off. only if HA cluster is restarted then only it logged off the user for a new session.

Chaho12:
@sug-ghosh Can you rebase so that it get be ready for final review?

mosabua:
We still want this feature to be added. Could you rebase and address the comments @sug-ghosh ?

Chaho12:
@sug-ghosh Knock knock~

Chaho12:
@sug-ghosh If you are busy, i hope you don't mind if we take this PR and change it ourselves

PR #719, opened 2025-07-08
Replace H2 use for testing with PostgreSQL

Akanksha-kedia:
## Release notes

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
@@ -26,8 +26,6 @@ import okhttp3.Response;
 import okhttp3.ResponseBody;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
-import org.jdbi.v3.core.Handle;
-import org.jdbi.v3.core.Jdbi;
 import org.testcontainers.oracle.OracleContainer;
 import org.testcontainers.postgresql.PostgreSQLContainer;
 
@@ -68,16 +66,6 @@ public class HaGatewayTestUtils
                 .setBody(expectedResponse)
                 .addHeader(CONTENT_ENCODING, PLAIN_TEXT_UTF_8)
                 .setResponseCode(200));
-    }
-
-    public static void seedRequiredData(String h2DbFilePath)
-    {
-        String jdbcUrl = "jdbc:h2:" + h2DbFilePath + ";NON_KEYWORDS=NAME,VALUE";
-        Jdbi jdbi = Jdbi.create(jdbcUrl, "sa", "sa");
-        try (Handle handle = jdbi.open()) {
-            handle.createUpdate(HaGatewayTestUtils.getResourceFileContent("gateway-ha-persistence-mysql.sql"))
-                    .execute();
-        }
     }
 
     public static File buildGatewayConfig(PostgreSQLContainer postgreSqlContainer, int routerPort, String configFile)
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
@@ -17,49 +17,43 @@ import io.trino.gateway.ha.config.DataStoreConfiguration;
 import io.trino.gateway.ha.module.HaGatewayProviderModule;
 import io.trino.gateway.ha.persistence.JdbcConnectionManager;
 import org.jdbi.v3.core.Jdbi;
+import org.testcontainers.postgresql.PostgreSQLContainer;
 
-import java.io.File;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
+import static io.trino.gateway.ha.util.TestcontainersUtils.createPostgreSqlContainer;
 
 public final class TestingJdbcConnectionManager
 {
     private TestingJdbcConnectionManager() {}
 
+    public static PostgreSQLContainer createTestingPostgresContainer()
+    {
+        PostgreSQLContainer postgres = createPostgreSqlContainer()
+                .withDatabaseName("testdb")
+                .withInitScript("gateway-ha-persistence-postgres.sql");
+        postgres.start();
+        return postgres;
+    }
+
+    public static DataStoreConfiguration dataStoreConfig(PostgreSQLContainer postgres)
+    {
+        return new DataStoreConfiguration(
+                postgres.getJdbcUrl(),
+                postgres.getUsername(),
+                postgres.getPassword(),
+                postgres.getDriverClassName(),
+                true,
+                4,
+                false);
+    }
+
     public static DataStoreConfiguration dataStoreConfig()
     {
-        File tempH2DbDir = Path.of(System.getProperty("java.io.tmpdir"), "h2db-" + System.currentTimeMillis()).toFile();
-        tempH2DbDir.deleteOnExit();
-        String jdbcUrl = "jdbc:h2:" + tempH2DbDir.getAbsolutePath() + ";NON_KEYWORDS=NAME,VALUE";
-        HaGatewayTestUtils.seedRequiredData(tempH2DbDir.getAbsolutePath());
-        return new DataStoreConfiguration(jdbcUrl, "sa", "sa", "org.h2.Driver", true, 4, false);
+        return dataStoreConfig(createTestingPostgresContainer());
     }
 
     public static JdbcConnectionManager createTestingJdbcConnectionManager(DataStoreConfiguration config)
     {
         Jdbi jdbi = HaGatewayProviderModule.createJdbi(config);
         return new JdbcConnectionManager(jdbi, config);
-    }
-
-    public static void destroyTestingDatabase(DataStoreConfiguration config)
-    {
-        String tempH2DbDirPath = config.getJdbcUrl().replace("jdbc:h2:", "").replace(";NON_KEYWORDS=NAME,VALUE", "");
-        File dbFile = Path.of(tempH2DbDirPath).toFile();
-        File parentDir = dbFile.getParentFile();
-
-        if (parentDir != null && parentDir.exists()) {
-            File[] files = parentDir.listFiles((_, name) -> name.startsWith(dbFile.getName()));
-            if (files != null) {
-                for (File file : files) {
-                    try {
-                        Files.deleteIfExists(file.toPath());
-                    }
-                    catch (IOException e) {
-                        // Ignore deletion errors in test cleanup
-                    }
-                }
-            }
-        }
     }
 }
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManager.java
@@ -25,20 +25,6 @@ import static org.mockito.Mockito.when;
 final class TestJdbcConnectionManager
 {
     @Test
-    void testBuildJdbcUrlWithH2AndNoRoutingGroupDatabase()
-    {
-        JdbcConnectionManager connectionManager = createConnectionManager("jdbc:h2:/mydb");
-        assertThat(connectionManager.buildJdbcUrl(null)).isEqualTo("jdbc:h2:/mydb");
-    }
-
-    @Test
-    void testBuildJdbcUrlWithH2AndRoutingGroupDatabase()
-    {
-        JdbcConnectionManager connectionManager = createConnectionManager("jdbc:h2:/mydb");
-        assertThat(connectionManager.buildJdbcUrl("newdb")).isEqualTo("jdbc:h2:/newdb");
-    }
-
-    @Test
     void testBuildJdbcUrlWithMySQLAndNoRoutingGroupDatabase()
     {
         JdbcConnectionManager connectionManager = createConnectionManager("jdbc:mysql://localhost:3306/mydb");
@@ -108,7 +94,7 @@ final class TestJdbcConnectionManager
         DataStoreConfiguration dataStoreConfiguration = Mockito.mock(DataStoreConfiguration.class);
         when(dataStoreConfiguration.getJdbcUrl()).thenReturn(null);
 
-        JdbcConnectionManager connectionManager = new JdbcConnectionManager(Jdbi.create("jdbc:h2:/mydb", "sa", "sa"), dataStoreConfiguration);
+        JdbcConnectionManager connectionManager = new JdbcConnectionManager(Jdbi.create("jdbc:postgresql://localhost:5432/mydb", "postgres", "postgres"), dataStoreConfiguration);
         assertThatThrownBy(() -> connectionManager.buildJdbcUrl(null))
                 .isInstanceOf(IllegalArgumentException.class)
                 .hasMessage("JDBC URL cannot be null");
@@ -117,10 +103,10 @@ final class TestJdbcConnectionManager
     @Test
     void testBuildJdbcUrlWithNoSlashThrowsException()
     {
-        JdbcConnectionManager connectionManager = createConnectionManager("jdbc:h2:mem:test");
+        JdbcConnectionManager connectionManager = createConnectionManager("jdbc:postgresql:mydb");
         assertThatThrownBy(() -> connectionManager.buildJdbcUrl("newdb"))
                 .isInstanceOf(IllegalArgumentException.class)
-                .hasMessage("Invalid JDBC URL: no '/' found in jdbc:h2:mem:test");
+                .hasMessage("Invalid JDBC URL: no '/' found in jdbc:postgresql:mydb");
     }
 
     private static JdbcConnectionManager createConnectionManager(String jdbcUrl)
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
@@ -23,13 +23,14 @@ import io.trino.gateway.ha.persistence.JdbcConnectionManager;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.testcontainers.postgresql.PostgreSQLContainer;
 
 import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static io.trino.gateway.ha.TestingJdbcConnectionManager.createTestingJdbcConnectionManager;
+import static io.trino.gateway.ha.TestingJdbcConnectionManager.createTestingPostgresContainer;
 import static io.trino.gateway.ha.TestingJdbcConnectionManager.dataStoreConfig;
-import static io.trino.gateway.ha.TestingJdbcConnectionManager.destroyTestingDatabase;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -93,9 +94,12 @@ final class TestHaGatewayManager
         adhoc.setExternalUrl("adhoc2.trino.gateway.io");
         haGatewayManager.updateBackend(adhoc);
         assertThat(haGatewayManager.getActiveBackends("adhoc")).isEmpty();
+        // Do not assert on the order here. The query behind getAllBackends has no ORDER BY, so the
+        // order is up to the database. PostgreSQL writes an updated row as a new tuple at the end of
+        // the table, which moves the backend updated above to the end of the result.
         assertThat(haGatewayManager.getAllBackends())
                 .extracting(ProxyBackendConfiguration::getRoutingGroup)
-                .containsExactly("etl", "adhoc");
+                .containsExactlyInAnyOrder("etl", "adhoc");
 
         // Delete a backend
         haGatewayManager.deleteBackend("adhoc1");
@@ -117,7 +121,8 @@ final class TestHaGatewayManager
     @Test
     void testGatewayManagerCacheExpire()
     {
-        DataStoreConfiguration dataStoreConfig = dataStoreConfig();
+        PostgreSQLContainer postgres = createTestingPostgresContainer();
+        DataStoreConfiguration dataStoreConfig = dataStoreConfig(postgres);
         JdbcConnectionManager connectionManager = createTestingJdbcConnectionManager(dataStoreConfig);
         DatabaseCacheConfiguration cacheConfiguration = new DatabaseCacheConfiguration();
         cacheConfiguration.setEnabled(true);
@@ -138,7 +143,7 @@ final class TestHaGatewayManager
         assertThat(haGatewayManager.getBackendByName("new-etl1").map(ProxyBackendConfiguration::getProxyTo)).hasValue("https://etl1.trino.gateway.io:443");
 
         // Test read from cache when DB is not available
-        destroyTestingDatabase(dataStoreConfig);
+        postgres.stop();
         assertThat(haGatewayManager.getBackendByName("new-etl1").map(ProxyBackendConfiguration::getProxyTo)).hasValue("https://etl1.trino.gateway.io:443");
 
         // Failed to refresh from DB, but still read from cache
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -338,13 +338,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.h2database</groupId>
-            <artifactId>h2</artifactId>
-            <version>2.4.240</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>mockwebserver</artifactId>
             <version>${dep.okhttp3.version}</version>


Akanksha-kedia:
@ebyhr

webapp/pnpm-lock.yaml
```diff
@@ -1,2936 +0,0 @@
-lockfileVersion: '9.0'
```
ebyhr, line 1: Is this change related to H2 removal?

Akanksha-kedia:
I ll have to remove that

ebyhr:
> I ll have to remove that

Why do we need to remove `webapp/pnpm-lock.yaml`?

Akanksha-kedia:
I mean i ll remove that commit

Akanksha-kedia:
@ebyhr please review

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
```diff
@@ -14,14 +14,45 @@
...
+        {
+            String columnName = r.getMetaData().getColumnName(columnNumber);
+            if ("update_time".equals(columnName)) {
+                try {
+                    Timestamp timestamp = r.getTimestamp(columnNumber);
+                    return timestamp != null ? timestamp.toLocalDateTime().format(FORMATTER) : null;
+                }
+                catch (SQLException e) {
+                    // Handle case when timestamp is invalid
+                    return "2020-07-06 00:00:00";
```
ebyhr: Why do we need this logic, and how did you choose this timestamp?

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
```diff
@@ -14,14 +14,45 @@
...
+@RegisterColumnMapper(ExactMatchSourceSelectorsDao.TimestampColumnMapper.class)
 public interface ExactMatchSourceSelectorsDao
 {
+    class TimestampColumnMapper
+            implements ColumnMapper<String>
+    {
+        private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+
+        @Override
+        public String map(ResultSet r, int columnNumber, StatementContext ctx)
```
ebyhr: Rename `r` to `result`. https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -18,21 +18,28 @@
...
 
     public static JdbcConnectionManager createTestingJdbcConnectionManager()
     {
-        File tempH2DbDir = Path.of(System.getProperty("java.io.tmpdir"), "h2db-" + System.currentTimeMillis()).toFile();
-        tempH2DbDir.deleteOnExit();
-        String jdbcUrl = "jdbc:h2:" + tempH2DbDir.getAbsolutePath();
-        HaGatewayTestUtils.seedRequiredData(tempH2DbDir.getAbsolutePath());
-        DataStoreConfiguration db = new DataStoreConfiguration(jdbcUrl, "sa", "sa", "org.h2.Driver", 4, false);
-        Jdbi jdbi = Jdbi.create(jdbcUrl, "sa", "sa");
+        org.testcontainers.containers.PostgreSQLContainer<?> postgres = new org.testcontainers.containers.PostgreSQLContainer<>("postgres:14-alpine");
```
ebyhr: Static import `PostgreSQLContainer`.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -18,21 +18,28 @@
...
+
+        String jdbcUrl = postgres.getJdbcUrl();
+        String username = postgres.getUsername();
+        String password = postgres.getPassword();
+
+        DataStoreConfiguration db = new DataStoreConfiguration(jdbcUrl, username, password, "org.postgresql.Driver", 4, false);
+        Jdbi jdbi = Jdbi.create(jdbcUrl, username, password);
+
+        // Initialize the database with required schema
+        try (var handle = jdbi.open()) {
```
ebyhr: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-var

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -100,8 +157,13 @@ void testReadSpecificDbSelector()
...
 
         assertThat(newSelector).isEqualTo(selector);
-        resourceGroupManager
-                .deleteSelector(selector, specificDb);
-        resourceGroupManager.deleteResourceGroup(1, specificDb);
+
+        // Verify we can read the selector
+        List<SelectorsDetail> selectors = resourceGroupManager.readAllSelectors(specificDb);
+        assertThat(selectors).hasSize(1);
+        assertThat(selectors.get(0)).isEqualTo(selector);
```
ebyhr: Use `containsExactly` instead.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -76,19 +99,53 @@ void testReadSpecificDbResourceGroupCauseException()
...
+        // Print the group ID to debug
+        System.out.println("Specific DB group ID: " + group.getResourceGroupId());
+
+        // Skip the test if we couldn't create a resource group with a non-zero ID
+        if (group.getResourceGroupId() == 0) {
+            System.out.println("Skipping test because resource group ID is 0");
+            return;
+        }
+
+        // Create a selector for this group
```
ebyhr: Why is this test non-deterministic?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -76,19 +99,53 @@ void testReadSpecificDbResourceGroupCauseException()
...
+            }
+            resourceGroupManager.deleteResourceGroup(existingGroup.getResourceGroupId(), specificDb);
+        }
+
+        // Create a new resource group with a unique name
+        String uniqueName = "admin3-" + System.currentTimeMillis();
+        ResourceGroupsDetail group = this.createResourceGroup(uniqueName);
+
+        // Print the group ID to debug
+        System.out.println("Specific DB group ID: " + group.getResourceGroupId());
```
ebyhr: Remove or use logger instead.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -76,19 +99,53 @@ void testReadSpecificDbResourceGroupCauseException()
...
+
+        // Create a new resource group with a unique name
+        String uniqueName = "admin2-" + System.currentTimeMillis();
+        ResourceGroupsDetail group = this.createResourceGroup(uniqueName);
         List<ResourceGroupsDetail> resourceGroups = resourceGroupManager
                 .readAllResourceGroups(specificDb);
         assertThat(resourceGroups).isNotNull();
-        resourceGroupManager.deleteResourceGroup(1, specificDb);
+        assertThat(resourceGroups).hasSize(1);
+        assertThat(resourceGroups.get(0).getName()).isEqualTo(uniqueName);
```
ebyhr: Replace get(0) with getFirst().

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -40,30 +40,53 @@ final class TestSpecificDbResourceGroupsManager
...
-        Jdbi jdbi = Jdbi.create(jdbcUrl, "sa", "sa");
+        specificDb = "test_db_" + System.currentTimeMillis();
+        PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>("postgres:14-alpine");
+        postgres.start();
+
+        String jdbcUrl = postgres.getJdbcUrl();
+        String username = postgres.getUsername();
+        String password = postgres.getPassword();
+
+        // Create the specific database
```
ebyhr: Remove unhelpful comment. It's obvious from the below code.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -40,30 +40,53 @@ final class TestSpecificDbResourceGroupsManager
...
+        try (var connection = postgres.createConnection("")) {
+            try (var statement = connection.createStatement()) {
```
ebyhr: Merge into single `try`.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -40,30 +40,53 @@ final class TestSpecificDbResourceGroupsManager
...
+                statement.execute("CREATE DATABASE " + specificDb);
```
ebyhr: Why not use `postgres.withDatabaseName()` method?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -40,30 +40,53 @@ final class TestSpecificDbResourceGroupsManager
...
+
+        // Seed data in the main database
+        HaGatewayTestUtils.seedRequiredData(postgres);
+
+        // Also seed data in the specific database
+        String specificDbJdbcUrl = jdbcUrl.replaceAll("/[^/]*$", "/" + specificDb);
+        Jdbi specificJdbi = Jdbi.create(specificDbJdbcUrl, username, password);
+        try (var handle = specificJdbi.open()) {
+            handle.createUpdate(HaGatewayTestUtils.getResourceFileContent("gateway-ha-persistence-postgres.sql"))
+                    .execute();
```
ebyhr: Please try initializing the database within the container. https://hub.docker.com/_/postgres#initialization-scripts

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -190,82 +224,139 @@ void testCreateSelector()
...
     }
 
     @Test
     @Order(8)
     void testDeleteSelector()
     {
         List<SelectorsDetail> selectors = resourceGroupManager.readAllSelectors(null);
-        assertThat(selectors).hasSize(3);
-        assertThat(selectors.get(0).getResourceGroupId()).isEqualTo(1L);
+        if (selectors.isEmpty()) {
```
ebyhr: Why is this test non-deterministic?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -190,82 +224,139 @@ void testCreateSelector()
...
         selector.setSelectorResourceEstimate(null);
 
-        updated = resourceGroupManager.updateSelector(new SelectorsDetail(), selector, null);
+        updated = resourceGroupManager.createSelector(selector, null);
         selectors = resourceGroupManager.readAllSelectors(null);
 
-        assertThat(selectors).hasSize(3)
-                .element(2).isEqualTo(updated);
+        assertThat(selectors).contains(updated);
+        assertThat(selectors).hasSize(3);
```
ebyhr:
```java
assertThat(selectors).hasSize(3).contains(updated);
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -190,82 +224,139 @@ void testCreateSelector()
...
     }
 
     @Test
     @Order(7)
     void testUpdateSelector()
     {
-        SelectorsDetail selector = new SelectorsDetail();
+        // First ensure we have selectors to update
+        List<SelectorsDetail> existingSelectors = resourceGroupManager.readAllSelectors(null);
+        if (existingSelectors.isEmpty()) {
```
ebyhr: Why is this test non-deterministic?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -190,82 +224,139 @@ void testCreateSelector()
     @Order(6)
     void testReadSelector()
     {
+        // First ensure we have a selector
         List<SelectorsDetail> selectors = resourceGroupManager.readAllSelectors(null);
+        if (selectors.isEmpty()) {
```
ebyhr: Why is this test non-deterministic?

Akanksha-kedia:
@ebyhr have resolved please review.

Akanksha-kedia:
- __Database Configuration Changes__:

  - In `gateway-ha/config.yaml`: Changed database connection parameters to use environment variables with defaults, making the configuration more flexible for different environments.

- __Timestamp Handling Improvements__:

  - In `ExactMatchSourceSelectorsDao.java`: Added a `TimestampColumnMapper` class to properly handle timestamp formatting.
  - In `HaResourceGroupsManager.java`: Added automatic timestamp generation for `updateTime` when it's null or empty.

- __Test Framework Modernization__:

  - In `TestingJdbcConnectionManager.java`: Replaced H2 in-memory database with PostgreSQL using TestContainers.
  - In `TestSpecificDbResourceGroupsManager.java`: Removed file-based setup and switched to TestContainers.
  - In `TestResourceGroupsManager.java`: Updated test assertions to use newer Java methods (e.g., `getFirst()` instead of `get(0)`), and improved resource group ID lookup.

Akanksha-kedia:
@mosabua please review.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
```diff
@@ -14,14 +14,46 @@
...
+                throws SQLException
+        {
+            String columnName = result.getMetaData().getColumnName(columnNumber);
+            if ("update_time".equals(columnName)) {
+                try {
+                    Timestamp timestamp = result.getTimestamp(columnNumber);
+                    return timestamp != null ? timestamp.toLocalDateTime().format(FORMATTER) : null;
+                }
+                catch (SQLException e) {
+                    return LocalDateTime.now(ZoneId.systemDefault()).format(FORMATTER);
```
ebyhr: Please leave a code comment why we need this fallback logic.

Akanksha-kedia:
@mosabua

Akanksha-kedia:
Please review and help to merge 
@ebyhr @mosabua

Akanksha-kedia:
Screenshot 2025-08-08 at 11 18 58 AM
https://nexus-iq.visa.com/assets/index.html#/vulnerabilities/CVE-2021-42392

Akanksha-kedia:
@mosabua @electrum @ebyhr please review

Akanksha-kedia:
https://github.com/trinodb/trino-gateway/security/advisories/GHSA-hmf5-v5xc-323v#advisory-comment-134493

mosabua:
We cut a new release now .. will hopefully be able to test soon .. can you rebase in the meantime

Akanksha-kedia:
Sure thing!

Akanksha-kedia:
@mosabua  please review

mosabua:
With https://github.com/trinodb/trino-gateway/pull/780 merged now .. can you rebase and adjust and make sure CI passes?

Akanksha-kedia:
Sure

Akanksha-kedia:
please review @mosabua

mosabua:
Can you rebase and resolve conflicts here @Akanksha-kedia ? It would be good to get this in

Akanksha-kedia:
Sure let me do that

Akanksha-kedia:
@mosabua please review  i have rebased and resolved conflicts

ebyhr:
@Akanksha-kedia Could you fix CI failure?

Akanksha-kedia:
Sure!

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -22,49 +21,57 @@
...
-        File tempH2DbDir = Path.of(System.getProperty("java.io.tmpdir"), specificDb).toFile();
-        tempH2DbDir.deleteOnExit();
-        String jdbcUrl = "jdbc:h2:" + tempH2DbDir.getAbsolutePath() + ";NON_KEYWORDS=NAME,VALUE";
-        HaGatewayTestUtils.seedRequiredData(tempH2DbDir.getAbsolutePath());
-        DataStoreConfiguration db = new DataStoreConfiguration(jdbcUrl, "sa",
-                "sa", "org.h2.Driver", 4, false);
-        Jdbi jdbi = HaGatewayProviderModule.createJdbi(db);
-        JdbcConnectionManager connectionManager = new JdbcConnectionManager(jdbi, db);
-        super.resourceGroupManager = new HaResourceGroupsManager(connectionManager);
+        PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>("postgres:14-alpine")
```
mosabua: why such an old version of PostgreSQL? In either case .. it should be same as in other tests and usage within the codebase

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManager.java
```diff
@@ -117,10 +103,10 @@ void testBuildJdbcUrlWithNullJdbcUrlThrowsException()
     @Test
     void testBuildJdbcUrlWithNoSlashThrowsException()
```
mosabua, line 104: Its kinda odd that we have this test and there is no jdbc url validation method in the driver itself we can use but 🤷

mosabua (commented):
Overall I agree with the idea and PR. We definitely need to make sure all tests pass before merging.

One aspect to keep in mind also is that we will hopefully delete all the resource groups table stuff and the related code from Trino Gateway since it really is outside of the scope of the project as has been discussed in our team syncs before

mosabua:
Ping @Akanksha-kedia

Akanksha-kedia:
Yes will try asap

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -77,75 +76,87 @@ void testReadResourceGroup()
         List<ResourceGroupsDetail> resourceGroups = resourceGroupManager.readAllResourceGroups(null);
         assertThat(resourceGroups).hasSize(2);
 
-        assertThat(resourceGroups.get(0).getResourceGroupId()).isEqualTo(1L);
-        assertThat(resourceGroups.get(0).getName()).isEqualTo("admin");
-        assertThat(resourceGroups.get(0).getHardConcurrencyLimit()).isEqualTo(20);
-        assertThat(resourceGroups.get(0).getMaxQueued()).isEqualTo(200);
-        assertThat(resourceGroups.get(0).getJmxExport()).isTrue();
-        assertThat(resourceGroups.get(0).getSoftMemoryLimit()).isEqualTo("80%");
+        assertThat(resourceGroups.getFirst().getResourceGroupId()).isEqualTo(1L);
```
ebyhr: This change is unrelated to H2->PostgreSQL change. Please separate a commit.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -77,75 +76,87 @@ void testReadResourceGroup()
...
+        assertThat(resourceGroups.getFirst().getSoftMemoryLimit()).isEqualTo("80%");
     }
 
     @Test
     @Order(3)
     void testUpdateResourceGroup()
     {
+        List<ResourceGroupsDetail> resourceGroups = resourceGroupManager.readAllResourceGroups(null);
+        long adminResourceGroupId = resourceGroups.stream()
+                .filter(rg -> "admin".equals(rg.getName()))
```
ebyhr: rg -> resourceGroup
https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -191,82 +225,261 @@ void testCreateSelector()
...
+        }
+
+        // Check if selector-test-group already exists
+        List<ResourceGroupsDetail> existingGroups = resourceGroupManager.readAllResourceGroups(null);
+        ResourceGroupsDetail testGroup = existingGroups.stream()
+                .filter(rg -> "selector-test-group".equals(rg.getName()))
+                .findFirst()
+                .orElse(null);
+
+        if (testGroup == null) {
```
ebyhr: Why is this test non-deterministic?

gateway-ha/src/test/java/io/trino/gateway/ha/HaGatewayTestUtils.java
```diff
@@ -166,7 +166,16 @@ public static void setUpBackend(
...
+        // gvenzl/oracle-xe:18.4.0-slim is x86 only, and tests using this image will most likely timeout if
+        // run on other CPU architectures. This test should run in three circumstances:
+        // * in CI, defined by the presence of GitHub environment variables
+        // * if the CPU architecture is x86_64
+        // * if they are explicitly enabled by setting TG_RUN_ORACLE_TESTS=true in the test environment
+        // reference: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
+        assumeTrue("true".equals(System.getenv("GITHUB_ACTIONS"))
+                || "x86_64".equalsIgnoreCase(System.getProperty("os.arch"))
+                || "true".equals(System.getenv("TG_RUN_ORACLE_TESTS")));
+        return new OracleContainer("gvenzl/oracle-xe:18.4.0-slim");
```
ebyhr: Why do you downgrade the version? Please revert this change.

Akanksha-kedia:
> Overall I agree with the idea and PR. We definitely need to make sure all tests pass before merging.
> 
> One aspect to keep in mind also is that we will hopefully delete all the resource groups table stuff and the related code from Trino Gateway since it really is outside of the scope of the project as has been discussed in our team syncs before

by when or anyhelp needed from my end

Akanksha-kedia:
please review @mosabua

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
ebyhr: When does this issue happen? Is it covered by tests?

Akanksha-kedia: Based on my analysis, the tests do exercise the timestamp mapper (through `testReadExactMatchSourceSelectors()`), but they don't specifically test the fallback scenario where `getTimestamp()` throws a SQLException. The fallback is a defensive measure for edge cases with certain database drivers.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
Akanksha-kedia: ##

The test code in `TestResourceGroupsManager.java` at line 237-240 is __non-deterministic__ because it uses `.stream().filter().findFirst()` on a __List without guaranteed ordering__:

```java
ResourceGroupsDetail testGroup = existingGroups.stream()
    .filter(rg -> "selector-test-group".equals(rg.getName()))
    .findFirst()
    .orElse(null);
```

### Root Cause:

1. __Database ordering is not guaranteed__: The `readAllResourceGroups(null)` method returns a List from a database query, but without an explicit `ORDER BY` clause, the database can return rows in any order.

2. __Test execution order dependency__: The test uses `@TestMethodOrder(OrderAnnotation.class)` and `@Order` annotations, but previous tests create multiple resource groups. The order in which these groups are returned from the database is unpredictable.

3. __Stream.findFirst() behavior__: When filtering a stream, `findFirst()` returns the first element that matches the filter predicate. If multiple elements could match (or if the list order changes), the result becomes non-deterministic.

### The Problem:

If there are multiple resource groups in the database when this test runs, and the database returns them in different orders on different executions, the test behavior becomes unpredictable. Even though the filter looks for a specific name ("selector-test-group"), the __order__ in which the database returns all groups can vary, affecting test execution timing and state.

### Solution:

The code should either:

1. Use a deterministic query with explicit ordering in `readAllResourceGroups()`
2. Not rely on list ordering when searching for specific items
3. Clean up test data more reliably between test methods

The pattern appears in multiple places in the test file (lines 90-94, 237-240, 268-271, 305-310), making the entire test suite potentially flaky.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -81,71 +80,83 @@ void testReadResourceGroup()
         assertThat(resourceGroups.get(0).getName()).isEqualTo("admin");
         assertThat(resourceGroups.get(0).getHardConcurrencyLimit()).isEqualTo(20);
         assertThat(resourceGroups.get(0).getMaxQueued()).isEqualTo(200);
-        assertThat(resourceGroups.get(0).getJmxExport()).isTrue();
+        assertThat(resourceGroups.get(0).getJmxExport()).isEqualTo(true);
```
ebyhr: Revert.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
```diff
@@ -81,71 +80,83 @@ void testReadResourceGroup()
...
-        assertThat(resourceGroups.get(2).getMaxQueued()).isEqualTo(70);
-        assertThat(resourceGroups.get(2).getJmxExport()).isTrue();
-        assertThat(resourceGroups.get(2).getSoftMemoryLimit()).isEqualTo("20%");
-        assertThat(resourceGroups.get(2).getSoftConcurrencyLimit()).isEqualTo(Integer.valueOf(20));
+        assertThat(resourceGroups).hasSize(4); // created 2 new groups, so count should be 4
+
+        // Find the admin resource group and verify its properties
+        ResourceGroupsDetail adminGroup = resourceGroups.stream()
+                .filter(rg -> "admin".equals(rg.getName()))
+                .findFirst()
```
ebyhr: Please avoid `findFirst` as much  as possible. Consider using `collect(toOptional())` instead.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaResourceGroupsManager.java
```diff
@@ -237,11 +240,18 @@ public void deleteGlobalProperty(String name, @Nullable String routingGroupDatab
...
     /**
      * Creates exact match source selector for db.
+     * If no updateTime is provided, the current timestamp will be used automatically.
      */
     @Override
     public ExactSelectorsDetail createExactMatchSourceSelector(
             ExactSelectorsDetail exactSelectorDetail)
     {
+        // If updateTime is null or empty, set current timestamp
+        if (exactSelectorDetail.getUpdateTime() == null || exactSelectorDetail.getUpdateTime().trim().isEmpty()) {
```
ebyhr: `getUpdateTime()` contains `@Nonnull` annotation.

ebyhr: Change `DateTimeFormatter formatter` to a constant.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -22,49 +21,58 @@
...
+import static io.trino.gateway.ha.util.TestcontainersUtils.createPostgreSqlContainer;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 @TestInstance(Lifecycle.PER_CLASS)
 final class TestSpecificDbResourceGroupsManager
         extends TestResourceGroupsManager
 {
-    private String specificDb;
+    private String specificDb = "test_db_specific";
```
ebyhr: Make it static final and uppercase the constant name.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -22,49 +21,58 @@
...
+        try {
+            DataStoreConfiguration db = new DataStoreConfiguration(jdbcUrl, username,
+                    password, "org.postgresql.Driver", 4, false);
+            Jdbi jdbi = HaGatewayProviderModule.createJdbi(db);
+            JdbcConnectionManager connectionManager = new JdbcConnectionManager(jdbi, db);
+            super.resourceGroupManager = new HaResourceGroupsManager(connectionManager);
+        }
+        catch (Exception e) {
+            throw new RuntimeException("Failed to setup test database: " + specificDb, e);
+        }
```
ebyhr: Remove a redundant catch block.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestSpecificDbResourceGroupsManager.java
```diff
@@ -22,49 +21,58 @@
...
-        JdbcConnectionManager connectionManager = new JdbcConnectionManager(jdbi, db);
-        super.resourceGroupManager = new HaResourceGroupsManager(connectionManager);
+        PostgreSQLContainer postgres = createPostgreSqlContainer()
+                .withDatabaseName(specificDb)
+                .withInitScript("gateway-ha-persistence-postgres.sql");
+        postgres.start();
+
+        String jdbcUrl = postgres.getJdbcUrl();
+        String username = postgres.getUsername();
+        String password = postgres.getPassword();
```
ebyhr: Inline.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,26 +17,55 @@
...
+    public static JdbcConnectionManager createTestingJdbcConnectionManager()
+    {
+        PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>("postgres:14-alpine")
+                .withDatabaseName("testdb")
+                .withInitScript("gateway-ha-persistence-postgres.sql");
+        postgres.start();
+
+        String jdbcUrl = postgres.getJdbcUrl();
+        String username = postgres.getUsername();
+        String password = postgres.getPassword();
```
ebyhr: Inline.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,26 +17,55 @@
 import io.trino.gateway.ha.module.HaGatewayProviderModule;
 import io.trino.gateway.ha.persistence.JdbcConnectionManager;
 import org.jdbi.v3.core.Jdbi;
-
-import java.io.File;
-import java.nio.file.Path;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.containers.PostgreSQLContainer;
```
ebyhr: Please avoid a deprecated class.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,26 +17,55 @@
...
+        PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>("postgres:14-alpine")
+                .withDatabaseName("testdb")
+                .withInitScript("gateway-ha-persistence-postgres.sql");
+        postgres.start();
+
+        return new DataStoreConfiguration(
+                postgres.getJdbcUrl(),
+                postgres.getUsername(),
+                postgres.getPassword(),
+                "org.postgresql.Driver",
```
ebyhr: `PostgreSQLContainer` offers `getDriverClassName` method.

ebyhr (changes requested):
Replacing H2 with PostgreSQL is totally fine. However, some test changes don't make sense to me.

Akanksha-kedia:
@ebyhr please review all and let me know wat more chnages i need to do

ebyhr:
Please fix build failures:
```
Error:  src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java:[26,15] (imports) UnusedImports: Unused import - com.google.common.collect.MoreObjects.firstNonNull.
Error:  src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java:[32,15] (imports) UnusedImports: Unused import - java.util.stream.Collectors.collectingAndThen.
Error:  src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java:[33,15] (imports) UnusedImports: Unused import - java.util.stream.Collectors.toList.
```

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
ebyhr: Reminder.

Akanksha-kedia:
> Please fix build failures:
> 
> ```
> Error:  src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java:[26,15] (imports) UnusedImports: Unused import - com.google.common.collect.MoreObjects.firstNonNull.
> Error:  src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java:[32,15] (imports) UnusedImports: Unused import - java.util.stream.Collectors.collectingAndThen.
> Error:  src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java:[33,15] (imports) UnusedImports: Unused import - java.util.stream.Collectors.toList.
> ```

done @ebyhr do review its a long pending pr thanks in advance.

Akanksha-kedia:
@mosabua please help to review and merge or else again we have spend time in rebasing and all

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestResourceGroupsManager.java
ebyhr: How does this change relate to the H2 → PostgreSQL migration?

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,26 +17,57 @@
...
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.postgresql.PostgreSQLContainer;
 
 public final class TestingJdbcConnectionManager
 {
     private TestingJdbcConnectionManager() {}
 
+    public static JdbcConnectionManager createTestingJdbcConnectionManager()
+    {
+        PostgreSQLContainer postgres = new PostgreSQLContainer("postgres:14-alpine")
```
ebyhr: Why not use TestcontainersUtils.createPostgreSqlContainer?

mosabua:
We definitely want to go ahead with this so I am adding the stale-ignore label to keep the PR from getting closed/lost.

mosabua:
Can you @Akanksha-kedia rebase and address the last few commits from @ebyhr so we can maybe get this merged and into a release soon

Akanksha-kedia:
Sure! Please i ll do that rebase and all hust kindly help to review and
merge so that i dnt need to rebase again n again. Or may be someone can co
author to get this done ?

Akanksha-kedia:
@mosabua please check

mosabua:
Could you fix the failing tests (or rebase if that is the source of the problem) @Akanksha-kedia .. or push again in case its a temporary failure and you verified it works locally for you

Also could you have a look @ebyhr so we can get this finally in ..

Akanksha-kedia:
@mosabua please go ahead with pr merge.

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -56,7 +57,12 @@ public HaGatewayConfiguration() {}
 
     public Map<String, String> getServerConfig()
     {
-        return this.serverConfig;
+        /* START GENAI@CLINE */
```
ebyhr: What's this? Please remove. 

Also, how is this change related to H2 replacement? Please revert unrelated changes.

ebyhr:
I don't think this PR is ready for merge.

mosabua:
Can you remove the `* START GENAI@CLINE */` and similar comments @Akanksha-kedia and address any other concerns from @ebyhr  .. I would like to get this in soon

Akanksha-kedia:
Okay

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/ExactMatchSourceSelectorsDao.java
ebyhr: > The fallback is a defensive measure for edge cases with certain database drivers.

I don't understand. This project doesn't support arbitrary JDBC drivers. It only supports MySQL, PostgreSQL, and Oracle.

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
mosabua: All of these are removed now

mosabua:
@Akanksha-kedia I rebased this now .. much smaller change set since the resource group stuff was deleted. 

@ebyhr the order dependence in one test is in the javadoc now .. we might want to change the query / method for getting the backends itself in a separate commit or maybe better a follow up PR. Thoughts?

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -93,9 +94,12 @@ static void testGatewayManager(HaGatewayManager haGatewayManager)
         adhoc.setExternalUrl("adhoc2.trino.gateway.io");
         haGatewayManager.updateBackend(adhoc);
         assertThat(haGatewayManager.getActiveBackends("adhoc")).isEmpty();
+        // Do not assert on the order here. The query behind getAllBackends has no ORDER BY, so the
```
mosabua, line 97: Fyi @ebyhr .. so maybe we should make getAllBackends deterministic by adding some order by clause .. thoughts @trinodb/maintainers-gateway ?

mosabua (approved):
I am approving now but I did some of the changes so I need approval from another @trinodb/maintainers-gateway 

I plan to send a follow up PR about the ordering separately.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
Chaho12: Yes, we should add ORDER BY name to findAll(), since name is primary key -> fully deterministic.
There is also index on PK anyway so zero cost.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,49 +17,43 @@
...
+                false);
     }
 
-    public static void destroyTestingDatabase(DataStoreConfiguration config)
+    public static DataStoreConfiguration dataStoreConfig()
     {
-        String tempH2DbDirPath = config.getJdbcUrl().replace("jdbc:h2:", "").replace(";NON_KEYWORDS=NAME,VALUE", "");
-        File dbFile = Path.of(tempH2DbDirPath).toFile();
-        File parentDir = dbFile.getParentFile();
+        return dataStoreConfig(createTestingPostgresContainer());
```
Chaho12, line 51: Can you take a look the leak? it gave a red flag here.

---

🔴 #1 — The container leak

Mechanism
```
TestingJdbcConnectionManager.java:49-52
public static DataStoreConfiguration dataStoreConfig()
{
    return dataStoreConfig(createTestingPostgresContainer());
}
```
createTestingPostgresContainer() calls postgres.start() (:31) and returns the handle. The no-arg overload reads getJdbcUrl()/getUsername()/getPassword()/getDriverClassName() out of it into a DataStoreConfiguration, then drops the only reference. Nothing downstream can reach the container, so nothing can ever stop()/close() it. It stays up for the life of the surefire JVM. Testcontainers' Ryuk sidecar reaps it at JVM exit, so it's not a permanent host leak — but for the whole test run you're holding N live Postgres servers.

Census at head, per JDK leg
```
┌───────────────────────────────────┬─────────────────────────────────────────┬──────────────────────┬─────────────┬────────────┐
│               class               │                call site                │   JUnit lifecycle    │ containers  │  stopped?  │
├───────────────────────────────────┼─────────────────────────────────────────┼──────────────────────┼─────────────┼────────────┤
│ TestQueryCountBasedRouter:176-179 │ @BeforeEach                             │ default → per method │ 7 (7 @Test) │ no         │
├───────────────────────────────────┼─────────────────────────────────────────┼──────────────────────┼─────────────┼────────────┤
│ TestHaGatewayManager:44,54,161    │ 3× in test bodies                       │ PER_CLASS            │ 3           │ no         │
├───────────────────────────────────┼─────────────────────────────────────────┼──────────────────────┼─────────────┼────────────┤
│ TestHaGatewayManager:124          │ direct createTestingPostgresContainer() │ —                    │ 1           │ yes (:146) │
├───────────────────────────────────┼─────────────────────────────────────────┼──────────────────────┼─────────────┼────────────┤
│ TestRoutingManagerNotFound:32     │ constructor                             │ default → per method │ 1           │ no         │
├───────────────────────────────────┼─────────────────────────────────────────┼──────────────────────┼─────────────┼────────────┤
│ TestStochasticRoutingManager:41   │ @BeforeAll                              │ PER_CLASS            │ 1           │ no         │
└───────────────────────────────────┴─────────────────────────────────────────┴──────────────────────┴─────────────┴────────────┘
```
13 containers, 12 never stopped.

Two things make this worse than the raw number:

- The @BeforeEach one. TestQueryCountBasedRouter spins a fresh Postgres server per test method. Under H2 that was a temp file — near free. Now it's 7 database servers to run 7 assertions about routing math.
- 3 of the 5 rows are in files not in the diff. TestQueryCountBasedRouter, TestRoutingManagerNotFound, TestStochasticRoutingManager are untouched by this PR; they silently converted from H2-file to Postgres-container because they call dataStoreConfig(). The no-arg overload is precisely what hides the cost from a reviewer reading the diff.

Measured cost

- postgres:17 to "ready to accept connections" ×2 (the TestcontainersUtils:19-23 wait condition): ~1.0s on my machine, warm image cache, 3 runs (1.2 / 1.1 / 0.8s). CI's shared runner is slower, and Testcontainers adds image inspect + port mapping + Ryuk on top.
- CI maven-build on main (baseline): 10, 11, 10, 11, 10, 11 min across the last 3 runs × 3 JDKs.
- CI maven-build on this PR: 11m15s / 11m32s / 12m11s.

So roughly +1 min per leg, ×3 legs. Coarse (runner variance, minute-floored), so treat it as consistent-with rather than proof-of. Against a 30-min timeout it's not fatal — the objection is that it's pure waste that grows with every future test calling dataStoreConfig().

Fix

Delete the no-arg overload. That forces every caller to hold the container and therefore to close it. Then per class, use the pattern already sitting in the repo at BaseTestQueryHistoryManager:41-74:
```
final class TestQueryCountBasedRouter
{
    private final PostgreSQLContainer container = createTestingPostgresContainer();

    @AfterAll
    public final void close()
    {
        container.close();
    }

    @BeforeEach
    public void init()
    {
        DataStoreConfiguration dataStoreConfig = dataStoreConfig(container);
        ...
    }
}
```
7 containers → 1. Same shape for the other three classes: 13 → 4.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,49 +17,43 @@
...
-        Jdbi jdbi = HaGatewayProviderModule.createJdbi(config);
-        return new JdbcConnectionManager(jdbi, config);
+        return new DataStoreConfiguration(
+                postgres.getJdbcUrl(),
+                postgres.getUsername(),
+                postgres.getPassword(),
+                postgres.getDriverClassName(),
+                true,
+                4,
+                false);
```
Chaho12, line 46:
```suggestion
                true);
FlywayMigration.migrate(config);
```

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -17,49 +17,43 @@
...
+    public static PostgreSQLContainer createTestingPostgresContainer()
     {
-        File tempH2DbDir = Path.of(System.getProperty("java.io.tmpdir"), "h2db-" + System.currentTimeMillis()).toFile();
-        tempH2DbDir.deleteOnExit();
-        String jdbcUrl = "jdbc:h2:" + tempH2DbDir.getAbsolutePath() + ";NON_KEYWORDS=NAME,VALUE";
-        HaGatewayTestUtils.seedRequiredData(tempH2DbDir.getAbsolutePath());
-        return new DataStoreConfiguration(jdbcUrl, "sa", "sa", "org.h2.Driver", true, 4, false);
+        PostgreSQLContainer postgres = createPostgreSqlContainer()
+                .withDatabaseName("testdb")
+                .withInitScript("gateway-ha-persistence-postgres.sql");
```
Chaho12, line 32:
```suggestion
        PostgreSQLContainer postgres = createPostgreSqlContainer()
                .withDatabaseName("testdb");
```

gateway-ha-persistence-postgres.sql is legacy file.
We should have SSOT and have Flyway do the job.

```
Order of operations, before and after

As-is:
docker run postgres:17
  → Testcontainers connects, runs gateway-ha-persistence-postgres.sql (6 tables)
  → dataStoreConfig(runMigrationsEnabled = false)
  → migrate() never called anyway
  → tests run          [no flyway_schema_history]

Proposed:
docker run postgres:17                       (empty database)
  → dataStoreConfig(runMigrationsEnabled = true)
  → FlywayMigration.migrate(config)
       location "postgresql" derived from the jdbc: prefix   (FlywayMigration:29-31)
       baselineOnMigrate(true), baselineVersion("0")         (FlywayMigration:51-52)
       applies V1 → V2 → V3 → V4
  → tests run          [2 tables + flyway_schema_history]

Chaho12 (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
Chaho12: Without doing init script, we should let flyway do the job

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
mosabua: Done in a separate commit, `Order gateway backends by name`. `findAll()` now has `ORDER BY name`.

This also turned up a second test relying on the order PostgreSQL happened to return: `TestGatewayHaMultipleBackend.testBackendConfiguration` asserted the backends by array index in insertion order. Both that test and the workaround here now assert the deterministic order.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
mosabua: Fixed, thanks for the detailed census — it matched what is in the code.

The no-arg `dataStoreConfig()` overload is deleted, so every caller has to hold its container and can therefore close it. Per class, following the `BaseTestQueryHistoryManager` pattern you pointed at:

- `TestQueryCountBasedRouter`: 7 → 1, now `PER_CLASS` with the tables cleared in `@BeforeEach` instead of a fresh container per test method
- `TestHaGatewayManager`: 3 → 1 shared, plus the dedicated one for `testGatewayManagerCacheExpire`, which needs its own because it stops the database
- `TestRoutingManagerNotFound`: 1, now closed
- `TestStochasticRoutingManager`: 1, now closed

13 containers → 4, all closed. `TestQueryCountBasedRouter` runs its 7 tests in about 4s locally.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
mosabua: Done. `dataStoreConfig` now passes `runMigrationsEnabled = true` and calls `FlywayMigration.migrate(config)`.

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
mosabua: Agreed, and done — the init script is gone and Flyway is the single source of truth.

Worth recording why this is safe: the extra tables that the legacy script creates and the Flyway migrations do not are the resource group ones, which #969 removed along with the feature. No remaining test touches them. I will send a separate PR to drop those unused tables from `gateway-ha-persistence-postgres.sql` and `gateway-ha-persistence-mysql.sql`; the scripts themselves stay, since `TrinoGatewayRunner` seeds backends through them at database init.

mosabua:
All fixed now @Chaho12 .. can I get this in and move to the next one?

PR #720, opened 2025-07-09
Refactor LbAuthorizer logging: log role identification only after authorization check

vishnu-lab:
This PR refactors the logging logic in LbAuthorizer to ensure that log messages about user role identification (ADMIN, USER, API) are only written after the corresponding authorization check passes. Previously, the log messages could misleadingly indicate a user was identified as a role before the check was performed. Now, the logs clearly state whether a user is or is not identified as a given role, based on the actual result of the authorization check. The log message wording and style have also been improved for clarity and consistency.

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
@@ -37,27 +37,27 @@ public class LbAuthorizer
     {
         switch (role) {
             case "ADMIN":
-                log.info("User '%s' with memberOf(%s) was identified as ADMIN(%s)",
-                        principal.getName(), principal.getMemberOf(), configuration.getAdmin());
-                return principal.getMemberOf()
-                        .filter(m -> m.matches(configuration.getAdmin()))
-                        .isPresent();
+                return hasRole(principal, role, configuration.getAdmin());
             case "USER":
-                log.info("User '%s' with memberOf(%s) identified as USER(%s)",
-                        principal.getName(), principal.getMemberOf(), configuration.getUser());
-                return principal.getMemberOf()
-                        .filter(m -> m.matches(configuration.getUser()))
-                        .isPresent();
+                return hasRole(principal, role, configuration.getUser());
             case "API":
-                log.info("User '%s' with memberOf(%s) identified as API(%s)",
-                        principal.getName(), principal.getMemberOf(), configuration.getApi());
-                return principal.getMemberOf()
-                        .filter(m -> m.matches(configuration.getApi()))
-                        .isPresent();
+                return hasRole(principal, role, configuration.getApi());
             default:
                 log.warn("User '%s' with role %s has no regex match based on ldap search",
                         principal.getName(), role);
                 return false;
         }
+    }
+
+    private static boolean hasRole(LbPrincipal principal, String role, String regex)
+    {
+        boolean matched = principal.getMemberOf()
+                .filter(m -> m.matches(regex))
+                .isPresent();
+        if (matched) {
+            log.info("User '%s' with memberOf(%s) is identified as %s(%s)",
+                    principal.getName(), principal.getMemberOf(), role, regex);
+        }
+        return matched;
     }
 }


vishnu-lab:
@Chaho12 @ebyhr  can you guys please take a look

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
```diff
@@ -37,23 +37,44 @@ public boolean authorize(LbPrincipal principal,
...
+                boolean isAdmin = principal.getMemberOf()
                         .filter(m -> m.matches(configuration.getAdmin()))
                         .isPresent();
+                if (isAdmin) {
+                    log.info("User '%s' with memberOf(%s) is identified as ADMIN(%s)",
+                            principal.getName(), principal.getMemberOf(), configuration.getAdmin());
+                }
+                else {
+                    log.info("User '%s' with memberOf(%s) is NOT identified as ADMIN(%s)",
+                            principal.getName(), principal.getMemberOf(), configuration.getAdmin());
```
oneonestar: I don't think we need to log the `NOT` cases.
The current authorization failed message should be enough.

Chaho12: What about keeping it as debug level if you think it is needed.

Chaho12 (approved):
I don't mind the changes, but i would probably eliminate code duplication, extract helper methods and therefore simplify the main logic part. I can do this later. But its ok anyway :)

Btw i don't think we need a release note for this

ebyhr (changes requested):
The existing message is sufficient in my opinion. I don't think we want to make the logic redundant just for logging.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
vishnu-lab: resolved it, please check

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
```diff
@@ -30,30 +30,30 @@ public LbAuthorizer(AuthorizationConfiguration configuration)
         this.configuration = configuration;
     }
 
+    private boolean checkRole(LbPrincipal principal, String role, String regex)
```
ebyhr: I would rename to `isMemberOf`. 

Note that we put helper methods under the callers.

vishnu-lab: resolved

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
```diff
@@ -37,27 +37,27 @@ public boolean authorize(LbPrincipal principal,
...
-                        .isPresent();
+                return isMemberOf(principal, role, configuration.getApi());
             default:
                 log.warn("User '%s' with role %s has no regex match based on ldap search",
                         principal.getName(), role);
                 return false;
         }
     }
+
+    private boolean isMemberOf(LbPrincipal principal, String role, String regex)
```
ebyhr: This method can be static.

ebyhr (commented):
The commit title is too long. Please follow https://trino.io/development/process.html#pull-request-and-commit-guidelines-

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
```diff
@@ -37,27 +37,27 @@ public boolean authorize(LbPrincipal principal,
...
+    {
+        boolean matched = principal.getMemberOf()
```
Chaho12, line 54: hasRole or isMember?

Chaho12 (approved):
To be honest, I was thinking more about isolation and simplicity.

https://github.com/Chaho12/trino-gateway/blob/fb547356c52bf8637e9e7c9e667854f7ed16e30e/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
vishnu-lab: okay

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbAuthorizer.java
vishnu-lab: my bad , it should be hasRole only

andythsu (approved):
LGTM

PR #721, opened 2025-07-09
Enhance the MultiReadHttpServletRequest stream

vishalya:
## Description
- The return values of the function `isFinished` should be true only no bytes are available in the input stream to read
- `isReady` should always return true, since the inputstream should be ready all the time and never be false.
- Added methods to so that multiple bytes can be read for better performance.

@Chaho12 has idea on the issue with this code. @willmostly is the original author.

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/MultiReadHttpServletRequest.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/MultiReadHttpServletRequest.java
@@ -45,13 +45,15 @@ public class MultiReadHttpServletRequest
             @Override
             public boolean isFinished()
             {
-                return byteArrayInputStream.available() > 0;
+                // Determine if data is available for reading.
+                return byteArrayInputStream.available() <= 0;
             }
 
             @Override
             public boolean isReady()
             {
-                return false;
+                // It's an in memory stream, so it's ready to be read
+                return true;
             }
 
             @Override
@@ -62,6 +64,21 @@ public class MultiReadHttpServletRequest
                     throws IOException
             {
                 return byteArrayInputStream.read();
+            }
+
+            // Add multibyte read versions for efficiency
+            @Override
+            public int read(byte[] b, int off, int len)
+                    throws IOException
+            {
+                return byteArrayInputStream.read(b, off, len);
+            }
+
+            @Override
+            public int read(byte[] b)
+                    throws IOException
+            {
+                return byteArrayInputStream.read(b);
             }
         };
     }


Chaho12:
Typo in commit message. MultiReadHttpServletReques`t` misses t there. 
Btw how about "Fix ServletInputStream behavior for MultiReadHttpServletRequest and add read override" or you can remove MultiReadHttpServletRequest.

gateway-ha/src/main/java/io/trino/gateway/proxyserver/MultiReadHttpServletRequest.java
```diff
@@ -63,6 +63,13 @@ public int read()
             {
                 return byteArrayInputStream.read();
             }
+
+            @Override
+            public int read(byte[] b, int off, int len)
```
oneonestar, line 71: Should we also implement `public int read(byte[] b)`?

gateway-ha/src/main/java/io/trino/gateway/proxyserver/MultiReadHttpServletRequest.java
```diff
@@ -63,6 +63,13 @@ public int read()
...
+                    throws IOException
+            {
+                return byteArrayInputStream.read(b, off, len);
```
RoeyoOgen, line 74: do we want to make this null-safe?

vishalya: byteArrayInputStream is locally allocated, if that's what your concern is. It should not be null.

mosabua:
Is this good to to @vishalya @Chaho12 @oneonestar ?

Chaho12:
Seems ok. But don't forget to fix the commit message :)

andythsu (commented):
LGTM except for the commit msg

Chaho12 (approved):
LGTM 👍

Chaho12:
Build failed because of this issue https://github.com/trinodb/trino-gateway/pull/755

ebyhr:
> Enhance the MultiReadHttpServletRequest stream

I don't think developers can understand changes from this commit title. Please add the commit body explaining what was enhanced.

vishalya:
Updated the PR description.

ebyhr:
Thanks, please update the commit body too.

mosabua:
Lets finish this off @vishalya  .. add the info from the description to the commit message as discussed in team sync and then you can go ahead with merge.

mosabua (approved):

ebyhr:
@vishalya Not sure why you ignored the above comment.

Issue #715, opened 2025-06-27
Excessive LDAP lookups on each UI API call due to repeated memberOf resolution


vishnu-lab:
Currently, when navigating between pages in the Trino-gateway UI, each page transition triggers API calls to the gateway. For every API call, a new **_LbPrincipal_** is generated by decoding the JWT and fetching the **_memberOf_** attribute from LDAP. This causes unnecessary and excessive LDAP queries on every API request, even within the same authenticated session.

Proposed improvements to reduce redundant LDAP calls:

1. **Add an evictable cache**: Maintain a cache mapping users to their memberOf attribute values, reducing repeated lookups for the same user within a reasonable TTL (time-to-live).
2. **Include roles in JWT claims**: Encode the user’s roles (e.g., memberOf information) directly into the JWT at login time. This would avoid the need for repeated LDAP calls during subsequent API requests.

vishnu-lab:
@Chaho12  @oneonestar  can you take a look on this

Chaho12:
Even if Option 1 is applied, it wil still incur LDAP queries, though greatly reduced. Cache adss complexity (eviction, consistency) too. 
Hence Option 2 sound sbetter if we use short-lived JWTs (maybe 5 or 10 min?) to minimize the risk of stale role data (this doesn't change often so doesn't have to be immediate)

vishnu-lab:
got it , i would like to implement this

Chaho12:
@vishalya Any thoughts on this? 
Yes, we would appreciate if you send a PR on this. But please wait a moment as there might be other thoughts.

oneonestar:
Yea, I think Option 2 is better. The current implementation of OAuth is also using Option 2.

vishalya:
Thanks for the PR, yes, options is better as it bypasses the LDAP altogether.

PR #722, opened 2025-07-10
Encode the user’s roles directly into the JWT

vishnu-lab:
## Description
Currently, when navigating between pages in the Trino-gateway UI, each page transition triggers API calls to the gateway. For every API call, a new LbPrincipal is generated by decoding the JWT and fetching the memberOf attribute from LDAP. This causes unnecessary and excessive LDAP queries on every API request, even within the same authenticated session.

To reduce the LDAP calls , i have encoded the roles in the JWT claims and decode it instead of calling LDAP server.

Note: we should use short-lived JWTs  to minimize the risk of stale role data
https://github.com/trinodb/trino-gateway/issues/715
## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java
@@ -104,9 +104,10 @@ public class HaGatewayProviderModule
         Map<String, UserConfiguration> presetUsers = configuration.getPresetUsers();
 
         oauthManager = getOAuthManager(configuration);
-        formAuthManager = getFormAuthManager(configuration);
 
         authorizationManager = new AuthorizationManager(configuration.getAuthorization(), presetUsers);
+        formAuthManager = getFormAuthManager(configuration);
+
         resourceSecurityDynamicFeature = getAuthFilter(configuration);
         backendStateConnectionManager = new BackendStateManager();
 
@@ -137,7 +138,7 @@ public class HaGatewayProviderModule
         AuthenticationConfiguration authenticationConfiguration = configuration.getAuthentication();
         if (authenticationConfiguration != null && authenticationConfiguration.getForm() != null) {
             return new LbFormAuthManager(authenticationConfiguration.getForm(),
-                    configuration.getPresetUsers(), configuration.getPagePermissions());
+                    configuration.getPresetUsers(), configuration.getPagePermissions(), authorizationManager);
         }
         return null;
     }
@@ -156,7 +157,7 @@ public class HaGatewayProviderModule
 
         if (formAuthManager != null) {
             authFilters.add(new LbFilter(
-                    new FormAuthenticator(formAuthManager, authorizationManager),
+                    new FormAuthenticator(formAuthManager),
                     authorizer,
                     "Bearer",
                     new LbUnauthorizedHandler(defaultType)));
@@ -174,7 +175,7 @@ public class HaGatewayProviderModule
     {
         AuthorizationConfiguration authorizationConfig = configuration.getAuthorization();
         Authorizer authorizer = (authorizationConfig != null)
-                ? new LbAuthorizer(authorizationConfig) : new NoopAuthorizer();
+                ? new LbAuthorizer() : new NoopAuthorizer();
 
         AuthenticationConfiguration authenticationConfig = configuration.getAuthentication();
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
@@ -13,10 +13,13 @@
  */
 package io.trino.gateway.ha.security;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.trino.gateway.ha.config.AuthorizationConfiguration;
 import io.trino.gateway.ha.config.LdapConfiguration;
 import io.trino.gateway.ha.config.UserConfiguration;
 
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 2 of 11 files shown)


Chaho12 (commented):
Please follow PR guideline and write proper message.

gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java
```diff
@@ -104,9 +104,10 @@ public HaGatewayProviderModule(HaGatewayConfiguration configuration)
         Map<String, UserConfiguration> presetUsers = configuration.getPresetUsers();
 
         oauthManager = getOAuthManager(configuration);
-        formAuthManager = getFormAuthManager(configuration);
```
andythsu, line 107: is this change necessary?

gateway-ha/src/main/java/io/trino/gateway/ha/security/FormAuthenticator.java
```diff
@@ -43,11 +44,27 @@ public Optional<LbPrincipal> authenticate(String idToken)
...
+        String privilegesField = formAuthManager.getPrivilegesField();
+
+        Map<String, Claim> claims = null;
+        try {
+            claims = formAuthManager.getClaimsFromIdToken(idToken).orElseThrow();
+        }
+        catch (Exception e) {
+            return Optional.empty();
+        }
+        String userId = claims.get(userIdField).asString().replace("\"", "");
```
andythsu, line 56: why do we need to replace `\"` ?

gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
```diff
@@ -35,9 +39,42 @@ public AuthorizationManager(AuthorizationConfiguration configuration,
...
+
+    public String getPrivileges(String username)
+    {
+        if (authorizationConfiguration == null) {
+            return "ADMIN_USER_API";
+        }
+        Optional<String> memberOf = getMemberOf(username);
+        List<String> privileges = new ArrayList<String>();
+
+        if (authorizationConfiguration.getAdmin() != null) {
```
andythsu, line 61: would it be cleaner to replace the ifs with 

```
Map<String, String> rolePatterns = Map.of(
    "ADMIN", authorizationConfiguration.getAdmin(),
    "USER", authorizationConfiguration.getUser(),
    "API", authorizationConfiguration.getApi()
);

rolePatterns.forEach((role, pattern) -> {
    if (pattern != null) {
        memberOf.filter(m -> m.matches(pattern)).ifPresent(m -> privileges.add(role));
    }
});
```

gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java
vishnu-lab: inside **getFormAuthManager(configuration)** i am using **authorizationManager** variable. that is why i want to initialize **authorizationManager** first and then call that method

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbPrincipal.java
```diff
@@ -15,18 +15,17 @@
...
     private final String name;
-    private final Optional<String> memberOf;
+    private final String privileges;
 
-    public LbPrincipal(String name, Optional<String> memberOf)
+    public LbPrincipal(String name, String privileges)
     {
         this.name = name;
-        this.memberOf = memberOf;
+        this.privileges = privileges;
```
vishalya, line 28: It's not a good idea to set the privileges directly in the Principal. The privileges should be calculated dynamically. If any privilege for a user changes, we need to invalidate the session which is going to be difficult. This is especially true with the LDAP based privileges.

gateway-ha/src/main/java/io/trino/gateway/ha/security/FormAuthenticator.java
```diff
@@ -43,11 +44,27 @@ public Optional<LbPrincipal> authenticate(String idToken)
...
+
+        Claim claim = claims.get(privilegesField);
+        if (claim == null || claim.asString() == null) {
+            log.warn("No privileges found for user %s in idToken", userId);
+            throw new AuthenticationException("No privileges found for user " + userId + " in idToken");
+        }
+
+        String privileges = claim.asString();
```
vishalya, line 64: Rather than setting the privileges, let's set the membefOf and everything should fall in place.

vishalya (changes requested):
We want to use cache for the performance reason and we also need to prevent stale data, so we should cache the results of the LbLdapClient.getMemberOf method using a Guava cache that invalidates at regular intervals.
Similarly, when using token based claims for authorization, we should treat these claims as memberOf attributes. This would distinguish the roles supported by the gateway from the external system providing the privileges.

gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
```diff
@@ -35,9 +39,42 @@ public AuthorizationManager(AuthorizationConfiguration configuration,
...
+    {
+        this.presetUsers = presetUsers;
+        this.lbLdapClient = lbLdapClient;
+        this.authorizationConfiguration = authorizationConfiguration;
+    }
+
+    public String getPrivileges(String username)
+    {
+        if (authorizationConfiguration == null) {
+            return "ADMIN_USER_API";
```
oneonestar, line 56: I’d prefer not to give any privileges when `authorizationConfiguration` is not set.

mosabua:
Do we want to reopen and continue @vishnu-lab and @trinodb/maintainers-gateway

PR #726, opened 2025-07-16
Fix query distribution chart keeping the legends more than an hour

sug-ghosh:
## Description

previously , it is showing more than 1 hour data like the screenshot : - 

image (2) (1)

this pr fixes this issue which shows only last 1 hour's query distribution.

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/webapp/src/components/dashboard.tsx
+++ b/webapp/src/components/dashboard.tsx
@@ -18,7 +18,7 @@ export function Dashboard() {
   const [distributionDetail, setDistributionDetail] = useState<DistributionDetail>();
 
   useEffect(() => {
-    distributionApi({})
+    distributionApi({ latestHour: 1 })
       .then(data => {
         setDistributionDetail(data);
       }).catch(() => { });


sug-ghosh:
> This seems wierd because it is fine on ours. (using 11 though) What version of trino gateway are you using? image

I'm also using 11 only

Chaho12:
Does anyone else see this issue? Because the root cause seems to be sth else

sug-ghosh:
> Does anyone else see this issue? Because the root cause seems to be sth else

yes, validated by hitting query for 8 hours, it shows result of 24 hours instead.

andythsu:
We are on latest trino-gateway 15 and it shows correctly

image

RoeyoOgen:
@sug-ghosh we also see only 1 hour...

prakhar10:
I have seen this issue sometimes, but it seems more because of browser cache/cookies, If I clear those it shows last 1 hour data.

If the `latestHour` is not set then by default it will be 1 - https://github.com/trinodb/trino-gateway/blob/d6fce939b60a78c5327a542ebfa3dc2cdc8be8a8/gateway-ha/src/main/java/io/trino/gateway/ha/domain/request/QueryDistributionRequest.java#L30

andythsu:
looks like the default is 1 already. Do we still need this PR? @sug-ghosh

Chaho12:
Please open PR if you see the issue again.

PR #729, opened 2025-07-21
Add forward proto header configuration for cluster monitoring

raj-manvar:
## Description

As documented at https://trino.io/docs/current/security/tls.html#use-a-load-balancer-to-terminate-tls-https, when Trino is behind a loadbalancer or proxy like Trino Gateway, it's common that the TLS is terminated at Trino Gateway. 

Correspondingly when Trino Gateway forwards the request to Trino clusters, Gateway adds X-Forwarded-* headers in code [here](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java#L317). Relevant documentation is [here](https://trinodb.github.io/trino-gateway/installation/#trino-configuration) where users can optionally disable this by setting routing.addXForwardedHeaders to false.

This MR is to add the same Header while making [health check calls](https://trinodb.github.io/trino-gateway/installation/#health-checks-on-trino-clusters) to get cluster stats like queued queries or running queries. Since it's possible that TLS is terminated at Gateway, a similar header would be required when making the http calls to fetch the cluster stats, for example using the /metrics or /v1/jmx/mbean endpoints 

If such a header isn't added, the http call to fetch metrics would fail with an error like:
```
UnexpectedResponseException{message=Request is missing required keys: 
presto_execution_name_QueryManager_RunningQueries
presto_execution_name_QueryManager_QueuedQueries
trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount
in response: 'Error 403 Forbidden: Authentication over HTTP is not enabled', request=GET http://xxxxxxx/metrics?xxxx
```

## Release notes

```markdown
* Add option to add `X-Forwarded-Proto` when fetching cluster stats.
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -362,6 +362,23 @@ that are marked as active.
 See [TrinoStatus](routing-rules.md#trinostatus) for more details on 
 what each Trino status means.
 
+Username and password for the health check can be configured by adding 
+`backendState` to your configuration. The username and password must be valid
+across all backends.
+
+SSL and xForwardProtoHeader can be configured based on whether the 
+connection between the Trino Gateway and the backend is secure. 
+By default, both are set to false.
+Find more information in [the related Trino documentation](https://trino.io/docs/current/security/tls.html#use-a-load-balancer-to-terminate-tls-https).
+
+```yaml
+backendState:
+  username: "user"
+  password: "password"
+  ssl: <false/true>
+  xForwardedProtoHeader: <false/true>  
+```
+
 The type of health check is configured by setting
 
 ```yaml
@@ -442,15 +459,7 @@ monitor:
 This uses a JDBC connection to query `system.runtime` tables for cluster 
 information. It is required for the query count based routing strategy. This is
 recommended over `UI_API` since it does not restrict the Web UI authentication
-method of backend clusters. Configure a username and password by adding
-`backendState` to your configuration. The username and password must be valid 
-across all backends.
-
-```yaml
-backendState:
-  username: "user"
-  password: "password"
-```
+method of backend clusters. 
 
 Trino Gateway uses `explicitPrepare=false` by default. This property was introduced
 in Trino 431, and uses a single query for prepared statements, instead of a 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static com.google.common.net.HttpHeaders.X_FORWARDED_PROTO;
 import static io.airlift.http.client.HttpStatus.fromStatusCode;
 import static io.trino.gateway.ha.handler.HttpUtils.UI_API_QUEUED_LIST_PATH;
 import static io.trino.gateway.ha.handler.HttpUtils.UI_API_STATS_PATH;
@@ -48,11 +49,13 @@ public class ClusterStatsHttpMonitor
 
     private final String username;
     private final String password;
+    private final boolean xForwardedProtoHeader;
 
     public ClusterStatsHttpMonitor(BackendStateConfiguration backendStateConfiguration)
     {
         username = backendStateConfiguration.getUsername();
         password = backendStateConfiguration.getPassword();
+        xForwardedProtoHeader = backendStateConfiguration.getXForwardedProtoHeader();
     }
 
     @Override
@@ -137,10 +140,13 @@ public class ClusterStatsHttpMonitor
         }
 
         String targetUrl = backend.getProxyTo() + path;
-        Request request = new Request.Builder()
+        Request.Builder requestBuilder = new Request.Builder()
                 .url(HttpUrl.parse(targetUrl))
-                .get()
-                .build();
+                .get();
+        if (xForwardedProtoHeader) {
+            requestBuilder.addHeader(X_FORWARDED_PROTO, "https");
+        }
+        Request request = requestBuilder.build();
 
         Call call = client.newCall(request);
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/BackendStateConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/BackendStateConfiguration.java
@@ -18,6 +18,7 @@ public class BackendStateConfiguration
     private String username;
     private String password = "";
     private Boolean ssl = false;
+    private boolean xForwardedProtoHeader;
 
     public BackendStateConfiguration() {}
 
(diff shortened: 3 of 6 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/config/BackendStateConfiguration.java
```diff
@@ -18,6 +18,7 @@ public class BackendStateConfiguration
     private String username;
     private String password = "";
     private Boolean ssl = false;
+    private Boolean addXForwardedProtoHeader = false;
```
Chaho12: We should use boolean(primitive) instead of Boolean(object) unless we need null, which I think we don't.

docs/installation.md
```diff
@@ -362,6 +362,20 @@ that are marked as active.
...
+connection between the Trino Gateway and the backend is secure.
+
+```yaml
+backendState:
+  username: "user"
+  password: "password"
+  ssl: <false/true>
+  addXForwardedProtoHeader: <false/true>  
+```
+
```
Chaho12, line 381: Can you add when to enable it? Also mention default value too.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
```diff
@@ -137,10 +140,13 @@ private String queryCluster(ProxyBackendConfiguration backend, String path)
...
 
         String targetUrl = backend.getProxyTo() + path;
-        Request request = new Request.Builder()
+        Request.Builder requestBuilder = new Request.Builder()
                 .url(HttpUrl.parse(targetUrl))
-                .get()
-                .build();
+                .get();
+        if (addXForwardedProtoHeader) {
+            requestBuilder.addHeader(X_FORWARDED_PROTO, "HTTPS");
```
Chaho12: What about https instead of HTTPS? [RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1) saids that it should be in lowercase although most servers are case-sensitive

   When a URI uses components of the generic syntax, the component
   syntax equivalence rules always apply; namely, that the scheme and
   host are case-insensitive and therefore should be normalized to
   lowercase.  For example, the URI <HTTP://www.EXAMPLE.com/> is
   equivalent to <http://www.example.com/>.

gateway-ha/src/main/java/io/trino/gateway/ha/config/BackendStateConfiguration.java
raj-manvar: Updated to primitive type! I can also update the `ssl` to primitive type as a separate commit before the final merge. ( just incase current MR has more commits that need to be squashed )

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
raj-manvar: Thanks for checking! updated to lowercase.

docs/installation.md
raj-manvar: I added a link to Trino docs that explains this much more clearly. Let me know if that works :)

vishalya (approved):
LGTM

docs/installation.md
andythsu: can we add default is `false` for both options?

andythsu:
On a second thought, is this PR necessary? We are currently using `UI_API` for healthchecks which calls Trino via HTTPS. We had to turn on `ssl` for this. How is making calls to `/metrics` different from `UI_API`?

raj-manvar:
@andythsu  https://trino.io/docs/current/security/tls.html#use-a-load-balancer-to-terminate-tls-https has more details, but when a Trino cluster is behind a LoadBalancer/Gateway it usually just doesn't accept https connections ( and only accepts non-secure http connections) . I think in your case, both Gateway and Trino clusters maybe accepting https connections, but that maynot always be the case.
Copy pasting from the above page. 
```
The load balancer or proxy server accepts TLS connections and forwards them to the Trino coordinator, which typically runs with default HTTP configuration on the default port, 8080.
```
This is because the Trino clusters don't have a certificate which is globally trusted, only the LoadBalancer/Gateway is mounted with a trusted certificate. 
For Trino to accept any non-secure http connection with credentials ( via /metrics endpoints or /v1/query) a X-Forwarded-Proto header is required

docs/installation.md
raj-manvar: Updated docs have a line `By default, both are set to false.` stating this

raj-manvar:
We could also consider adding this header by default for all cluster health http calls. For the /v1/statement endpoints this header is added by default [here](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java#L317) ( unless users specifically set routing.addXForwardedHeaders to false ).  
I think having the header maybe harmless incase of secure https connections anyways so we could add it by default

raj-manvar:
Hello all, is this good to merge now ? Thanks!

gateway-ha/src/main/java/io/trino/gateway/ha/config/BackendStateConfiguration.java
```diff
@@ -18,6 +18,7 @@ public class BackendStateConfiguration
     private String username;
     private String password = "";
     private Boolean ssl = false;
+    private boolean addXForwardedProtoHeader;
```
ebyhr: Could you change to String `xForwardedProtoHeader` instead?

raj-manvar: hi, sorry to clarify, you mean the property would take a string value like `http` or `https` ? or the property would take string value of `true` or `false` ?

ebyhr: The former.

raj-manvar: Thanks for the clarification! Could there be any scenario where the value of the header is anything except `https` ? :thinking:  I don't know of any such cases.

ebyhr: I'm okay with boolean, but we usually don't use "add" prefix. I think we can simply rename it to  "xForwardedProtoHeader"

raj-manvar: Renamed the config to xForwardedProtoHeader!

docs/installation.md
```diff
@@ -362,6 +362,22 @@ that are marked as active.
 See [TrinoStatus](routing-rules.md#trinostatus) for more details on 
 what each Trino status means.
 
+Username and password for the health check can be configured by adding 
+`backendState` to your configuration. The username and password must be valid
+across all backends.
+
+SSL and xForwardProtoHeader can be configured based on whether the 
+connection between the Trino Gateway and the backend is secure. By default, both are set to false.
```
ebyhr: Wrap at 80 characters.

raj-manvar: Thanks! resolved

PR #730, opened 2025-07-22
Update to Trino 476 dependency

mosabua:
## Description

See title.

## Release notes


mosabua:
This is gonna need more work .. might not get around to it so I wont be in the way for others to look.

PR #732, opened 2025-07-22
Make user filter in history page more configurable

raj-manvar:
## Description

Currently only admins are allowed to filter the username for the query history field. However there are cases where queries aren't always tied to the user who is looking for the query information. 

For example a user is looking for a query submitted by a scheduled non-human workload which maybe using separate authentication mechanism like [jwt auth](https://trino.io/docs/current/security/jwt.html) or [kerberos](https://trino.io/docs/current/security/kerberos.html) which is more common for non-human workloads and not tied to the human looking for routing history. 

As part of this issue, provide alternatives for users to be able to configure user filter in history query page.

~~[Trino web UI ](https://trino.io/docs/current/admin/web-interface.html) also doesn't restrict the query filtering and allows all users to filter all queries submitted to the Trino cluster.~~ This is incorrect. Thanks to @Chaho12 for correction! 

## Release notes

```markdown
* All all users to filter query history page based on user field. 
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayWebAppResource.java
@@ -126,18 +126,7 @@ public class GatewayWebAppResource
     public Response findQueryHistory(QueryHistoryRequest query, @Context SecurityContext securityContext)
     {
         TableData<?> queryHistory;
-        if (!securityContext.isUserInRole("ADMIN")) {
-            queryHistory = queryHistoryManager.findQueryHistory(new QueryHistoryRequest(
-                    query.page(),
-                    query.size(),
-                    securityContext.getUserPrincipal().getName(),
-                    query.backendUrl(),
-                    query.queryId(),
-                    query.source()));
-        }
-        else {
-            queryHistory = queryHistoryManager.findQueryHistory(query);
-        }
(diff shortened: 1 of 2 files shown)


Chaho12 (changes requested):
> [Trino web UI ](https://trino.io/docs/current/admin/web-interface.html)also doesn't restrict the query filtering and allows all users to filter all queries submitted to the Trino cluster.

This is not true. Trino Web UI does support restriction, and we enabled it so that normal users can only see query that they ran. 
https://trino.io/docs/current/security/file-system-access-control.html#query-rules

I disagree on allowing normal users to filter/search for other peoples query or run information etc. I understand that QueryText does not show full query (which needs to be fixed) but there could be other data privacy issues when a person see's other person's query without consent.

raj-manvar:
Oh, Thanks @Chaho12  for the correction! so sorry for the misinformation. I didn't know this was the case. Also hadn't thought for privacy issues. 

I'll update the title and message of the PR.  

This PR definitely won't work in this case. Few alternatives I can think of:
* Add a configuration so that user role can view the query.
* Add an extra page permission like history_filter which would give permission to all filters. 

Not sure if any of these collide with some long term initiatives tho, let me know what do you think of these. Sorry again for the misinfo

andythsu:
I think it's better to introduce another role that grant access to filter user queries

raj-manvar:
I'll close this PR for now, current branch attached doesn't fit the description so it's misleading when opening the open PRs page.

PR #741, opened 2025-07-31
Refactor RoutingManager

maswin:
## Description
This commit addresses a number of items
1. QueryCountBasedRouter was broken. The change introduced in the following commit - https://github.com/trinodb/trino-gateway/commit/3430f352800e852715613fd3619bf700938dfb07 introduced a bug. provideBackendConfiguration was used to get the cluster instead of provideClusterForRoutingGroup. But this method was not overridden in QueryCountBasedRouter class. So this led to a bug and QueryCountBasedRouter was failing. Fixed it.
2. While trying to create a new RoutingManager, it became complex to understand what all method needs to be overridden since the class was populated with lot of internal methods. So extracted them all out to an interface and modified existing class as BaseRoutingManager. selectBackend method can be overridden to modify the cluster selection part.
3. Modified QueryCountBasedRouter to use ConcurrentHashMap instead of GaurdedBy("this") which locks the entire object. This reduces lock contention. 
4. Fixed related tests

## Release notes

```markdown
* Fix some things.
```


gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -13,290 +13,29 @@
...
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
 
-/**
- * This class performs health check, stats counts for each backend and provides a backend given
- * request object. Default implementation comes here.
- */
-public abstract class RoutingManager
+public interface RoutingManager
```
jalpan-randeri, line 22: Can we have java doc for this and method as this becomes our primary interface for all RoutingManager implementations. ?

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,306 @@
...
+                                    @Override
+                                    public String load(String queryId)
+                                    {
+                                        return findRoutingGroupForUnknownQueryId(queryId);
+                                    }
+                                });
+        queryIdExternalUrlCache =
+                CacheBuilder.newBuilder()
+                        .maximumSize(10000)
+                        .expireAfterAccess(30, TimeUnit.MINUTES)
```
jalpan-randeri: nit: you can extract this into separate builder and reuse 3 time while building cache. 

for example
```
    private final CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder()
            .maximumSize(10000)
            .expireAfterAccess(30, TimeUnit.MINUTES);


queryIdBackendCache = builder.build(
                                new CacheLoader<>()
                                {
                                    @Override
                                    public String load(String queryId)
                                    {
                                        return findBackendForUnknownQueryId(queryId);
                                    }
                                });
        queryIdRoutingGroupCache = builder.build(
                                new CacheLoader<>()
                                {
                                    @Override
                                    public String load(String queryId)
                                    {
                                        return findRoutingGroupForUnknownQueryId(queryId);
                                    }
                                });
        queryIdExternalUrlCache = builder.build(
                                new CacheLoader<>()
                                {
                                    @Override
                                    public String load(String queryId)
                                    {
                                        return findExternalUrlForUnknownQueryId(queryId);
                                    }
                                });

```

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,306 @@
...
+    private boolean isBackendNotHealthy(String backendId)
+    {
+        if (backendToStatus.isEmpty()) {
+            log.error("backends can not be empty");
+            return true;
+        }
+        TrinoStatus status = backendToStatus.get(backendId);
+        if (status == null) {
+            return true;
+        }
```
jalpan-randeri:
```suggestion
        TrinoStatus status = backendToStatus.getOrDefault(backendId, TrinoStatus.UNKNOWN);
        return status != TrinoStatus.HEALTHY;
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,306 @@
...
+     * backend is found.
+     */
+    @Override
+    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
+    {
+        List<ProxyBackendConfiguration> backends = gatewayBackendManager.getActiveBackends(routingGroup);
+        backends.removeIf(backend -> isBackendNotHealthy(backend.getName()));
+        return selectBackend(backends, user).orElseGet(() -> provideDefaultBackendConfiguration(user));
+    }
+
```
jalpan-randeri, line 113: Since we are removing the unhealthy backends from candidate list. You can use lambda to filter out unwanted entries. It will simplify the predicate function that identifies healthy backends. 

For example:
```suggestion
      public ProxyBackendConfiguration provideDefaultBackendConfiguration(String user) {
        var backends = gatewayBackendManager.getActiveDefaultBackends()
                .stream()
                .filter(backend -> isBackendHealthy(backend.getName()))
                .toList();

        return selectBackend(backends, user).orElseThrow(() -> new IllegalStateException("Number of active backends found zero"));
    }

  
    @Override
    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user) {
        var backends = gatewayBackendManager.getActiveBackends(routingGroup)
                .stream()
                .filter(backend -> isBackendHealthy(backend.getName()))
                .toList();
        
        return selectBackend(backends, user).orElseGet(() -> provideDefaultBackendConfiguration(user));
    }
    
    private boolean isBackendHealthy(String backendId) {
        TrinoStatus status = backendToStatus.getOrDefault(backendId, TrinoStatus.UNKNOWN);
        return status == TrinoStatus.HEALTHY;
    }
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,306 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * This class performs health check, stats counts for each backend and provides a backend given
+ * request object. Default implementation comes here.
+ */
+public abstract class BaseRoutingManager
```
jalpan-randeri, line 49: Suggestion IMO, for ease of reading you may want to group methods by visibility and order them. 

1/ constructor
2/ abstract methods
3/ public method
4/ protected - package 
5/ private

I see the abstract methods as an interface for subclasses so it is easier to discover them by future maintainers.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,306 @@
...
+        try {
+            routingGroup = queryIdRoutingGroupCache.get(queryId);
+        }
+        catch (ExecutionException e) {
+            log.warn("Exception while loading queryId from routing group cache %s", e.getLocalizedMessage());
+        }
+        return routingGroup;
+    }
+
+    protected void updateBackEndHealth(List<ClusterStats> stats)
```
jalpan-randeri: can we delete this method ? 

It get confusing as we have 3 methods that update the state in cache.  
You move this method logic into the interface method `public void updateClusterStats` . This would simplify and we avoid method overloading of interface method.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,306 @@
...
+            }
+            for (Map.Entry<String, Future<Integer>> entry : responseCodes.entrySet()) {
+                if (entry.getValue().isDone()) {
+                    int responseCode = entry.getValue().get();
+                    if (responseCode == 200) {
+                        log.info("Found query [%s] on backend [%s]", queryId, entry.getKey());
+                        setBackendForQueryId(queryId, entry.getKey());
+                        return entry.getKey();
+                    }
+                }
```
jalpan-randeri, line 230: Q: Curious, Why did you add condition check for `future.isDone()` ?

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
vishalya: It's a good idea to have an interface to enforce the rules.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: Most of these code are existing - https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
I changed the file name from RoutingManager to BaseRoutingManager and created interface file with name RoutingManager. So git instead of marking it as file name change, assumed everything in the file as new changes.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: Done

maswin:
> Can we separate out this PR into 2 different ones. (1) Just fix the bug and add the minimal interface needed, and the fixed tests. (2) Rest of the refactoring

Tried splitting it into 2 commits but feels bit complicated as the refactoring kind of took care of the bug. 
The changes are relatively less, but git for some reason assumes BaseRoutingManager as a new file rather than the a name change from previous RoutingManager class (probably since there is a new interface with that name)

vishalya:
I'm concerned that having a default implementation in the `BaseRoutingManager` base class makes our routing logic brittle. We saw this when the interface change broke the query-count-based router.

A better approach would be to move the default logic into its own class, say `DefaultRoutingManager`. Then, concrete classes could use composition (a has-a relationship) to include this default behavior instead of inheriting it directly. This will decouple our concrete routers from the base class implementation, preventing similar breaks in the future.

On a related note, could you detail what new runtime tests are being added to catch these kinds of integration failures?

maswin:
> I'm concerned that having a default implementation in the `BaseRoutingManager` base class makes our routing logic brittle. We saw this when the interface change broke the query-count-based router.
> 
> A better approach would be to move the default logic into its own class, say `DefaultRoutingManager`. Then, concrete classes could use composition (a has-a relationship) to include this default behavior instead of inheriting it directly. This will decouple our concrete routers from the base class implementation, preventing similar breaks in the future.
> 
> On a related note, could you detail what new runtime tests are being added to catch these kinds of integration failures?

composition might makes things very complicated. Interface with an abstract base implementation is a common pattern which should be ok. For instance in Trino there is a [TrinoCatalog](https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/TrinoCatalog.java) interface with [AbtsractTrinoCatalog](https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/AbstractTrinoCatalog.java) implementation that has common methods implemented. 

The primary problem I see is the interface is bloated and can further be made lean. There should only be 4 methods - 
```
void updateBackEndHealth(String backendId, TrinoStatus value); // When user marks a backend unhealthy
void updateClusterStats(List<ClusterStats> stats); // Update based on JMX metrics
ProxyBackendConfiguration getBackendConfiguration(String routingGroup, String user); // Get for the first time and if not in local cache
ProxyBackendConfiguration setBackendConfiguration(String routingGroup, String user); // Set if not in local cache
```

Maintaining 3 separate cache and separately setting every cache makes no sense as they all point to same data. One cache with all data together should be enough and exposing just one method to get and set the backEnd configuration. 

This should make things less confusing to be overridden and implemented. 

If this new lean interface sounds good I can make the changes.

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
maswin: Added javadoc

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,279 @@
...
+     * Attempts to look up the external url associated with the query id from query history table
+     */
+    private String findExternalUrlForUnknownQueryId(String queryId)
+    {
+        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
+        setExternalUrlForQueryId(queryId, externalUrl);
+        return externalUrl;
+    }
+
+    private static LoadingCache<String, String> buildCache(Function<String, String> loader)
```
andythsu: does this have to be a `static` method?

andythsu:
this is definitely a good improvement to our routing logic. Thanks for the java doc.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,279 @@
...
+
+        Map<String, Future<Integer>> responseCodes = new HashMap<>();
+        try {
+            for (ProxyBackendConfiguration backend : backends) {
+                String target = backend.getProxyTo() + "/v1/query/" + queryId;
+
+                Future<Integer> call =
+                        executorService.submit(
+                                () -> {
+                                    URL url = new URL(target);
```
Chaho12: This is deprecated in java 20.
https://stackoverflow.com/questions/75966165/how-to-replace-the-deprecated-url-constructors-in-java-20
```suggestion
                                    URL url = URI.create(target).toURL();
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,279 @@
...
+                        return entry.getKey();
+                    }
+                }
+            }
+        }
+        catch (Exception e) {
+            log.warn("Query id [%s] not found", queryId);
+        }
+        // Fallback on first active backend if queryId mapping not found.
+        return gatewayBackendManager.getActiveBackends(defaultRoutingGroup).get(0).getProxyTo();
```
Chaho12: Handle Index Out of Bounds Risk?
```suggestion
                List<ProxyBackendConfiguration> fallbackBackends = gatewayBackendManager.getActiveBackends(defaultRoutingGroup);
        if (fallbackBackends.isEmpty()) {
            throw new IllegalStateException("No active backends available for default routing group: " + defaultRoutingGroup);
        }
        return fallbackBackends.get(0).getProxyTo();
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -243,24 +200,39 @@ private void updateLocalStats(LocalStats stats, String user)
...
+    // updated stats for all the clusters.
+    // if a user has queries queued then we assume that the routed query will be also queued or
+    // else we assume it would be scheduled immediately and we increment the stats for the running
+    // queries
     @Override
-    public synchronized void updateBackEndStats(List<ClusterStats> stats)
+    protected synchronized Optional<ProxyBackendConfiguration> selectBackend(List<ProxyBackendConfiguration> backends, String user)
     {
-        clusterStats = stats.stream().map(a -> new LocalStats(a)).collect(Collectors.toList());
+        Optional<ProxyBackendConfiguration> cluster = backends.stream().min((a, b) -> compareStats(clusterStats.get(a.getName()), clusterStats.get(b.getName()), user));
```
Chaho12: NPE could happen because clusterStats.get(a.getName()) or clusterStats.get(b.getName()) can return null, and then this null value gets passed to the compareStats() method.

We should filter and find availableBackends and use that to stream instead.

andythsu:
@maswin Do you still have bandwidth to work on this?

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
maswin: fixed

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
maswin: done

maswin:
> @maswin Do you still have bandwidth to work on this?

Sorry about the delay. Just addressed the review comments

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -13,280 +13,75 @@
...
-    {
-        List<ProxyBackendConfiguration> backends = gatewayBackendManager.getActiveDefaultBackends();
-        backends.removeIf(backend -> isBackendNotHealthy(backend.getName()));
-        if (backends.isEmpty()) {
-            throw new IllegalStateException("Number of active backends found zero");
-        }
-        int backendId = Math.abs(RANDOM.nextInt()) % backends.size();
-        return backends.get(backendId);
-    }
+    void updateBackEndHealth(String backendId, TrinoStatus value);
```
andythsu, line 30: can we rename this to `updateClusterHealth` since we want to refer Trino clusters as "cluster" instead of "backend"

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -13,280 +13,75 @@
...
-        String externalUrl = null;
-        try {
-            externalUrl = queryIdExternalUrlCache.get(queryId);
-        }
-        catch (ExecutionException e) {
-            log.warn("Exception while loading queryId from cache %s", e.getLocalizedMessage());
-        }
-        return externalUrl;
-    }
+    void setBackendForQueryId(String queryId, String backend);
```
andythsu, line 45: `setClusterForQueryId`

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
```diff
@@ -13,280 +13,75 @@
...
-                }
-            }
-        }
-        catch (Exception e) {
-            log.warn("Query id [%s] not found", queryId);
-        }
-        // Fallback on first active backend if queryId mapping not found.
-        return gatewayBackendManager.getActiveBackends(defaultRoutingGroup).get(0).getProxyTo();
-    }
+    String findBackendForQueryId(String queryId);
```
andythsu, line 61: `findClusterForQueryId`

gateway-ha/src/main/java/io/trino/gateway/ha/router/RoutingManager.java
maswin: We have a number of places in which we use backend, including the UI code - 
i.e,
ProxyBackendConfiguration
BackendStateManager
GatewayBackendManager

Isn't it better to change them all at once in a separate commit?

oneonestar:
@sourcery-ai review

vishalya:
sourcery-ai has 2 good suggestions. Otherwise LGTM, don't forget to rebase with the main branch.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,283 @@
...
+                        {
+                            @Override
+                            public String load(String queryId)
+                            {
+                                return loader.apply(queryId);
+                            }
+                        });
+    }
+
+    private boolean isBackendHealthy(String backendId)
```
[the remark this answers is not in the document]
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * This class performs health check, stats counts for each backend and provides a backend given
+ * request object. Default implementation comes here.
+ */
+public abstract class BaseRoutingManager
+        implements RoutingManager
+{
+    private static final Logger log = Logger.get(BaseRoutingManager.class);
+    private final ExecutorService executorService = Executors.newFixedThreadPool(5);
```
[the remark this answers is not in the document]
maswin: Done

maswin:
> sourcery-ai has 2 good suggestions. Otherwise LGTM, don't forget to rebase with the main branch.

Done

andythsu (approved):
🚀

andythsu:
I think before we merge it we just need to fix the commit message and then rebase

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,294 @@
...
+    {
+        queryIdExternalUrlCache.put(queryId, externalUrl);
+    }
+
+    @VisibleForTesting
+    String findBackendForUnknownQueryId(String queryId)
+    {
+        String backend;
+        backend = queryHistoryManager.getBackendForQueryId(queryId);
+        if (Strings.isNullOrEmpty(backend)) {
```
ebyhr: Static import isNullOrEmpty method.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,294 @@
...
+     * Attempts to look up the external url associated with the query id from query history table
+     */
+    private String findExternalUrlForUnknownQueryId(String queryId)
+    {
+        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
+        setExternalUrlForQueryId(queryId, externalUrl);
+        return externalUrl;
+    }
+
+    private LoadingCache<String, String> buildCache(Function<String, String> loader)
```
ebyhr: This method can be static.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -0,0 +1,294 @@
...
+    {
+        return CacheBuilder.newBuilder()
+                .maximumSize(10000)
+                .expireAfterAccess(30, TimeUnit.MINUTES)
```
ebyhr: Static import MINUTES.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -15,35 +15,28 @@
...
 
 public class QueryCountBasedRouter
-        extends StochasticRoutingManager
+        extends BaseRoutingManager
 {
-    private static final Logger log = Logger.get(QueryCountBasedRouter.class);
-    @GuardedBy("this")
-    private List<LocalStats> clusterStats;
-    private final String defaultRoutingGroup;
+    private ConcurrentHashMap<String, LocalStats> clusterStats;
```
ebyhr: This field can be final.

PR #748, opened 2025-08-11
Add request filter for early user and query parsing

vishalya:
## Description
The parsing of the user and query was happening at different places and at a later stages, this PR structures it to happen at a early stage and a single point in time. 

- Early Request Parsing:  New JAX-RS filters have been added to intercept incoming requests. This filter is responsible for parsing essential user and query information at the earliest stage of processing under the PRE_AUTHENTICATION and PRE_AUTHORIZATION.
- Use of ContainerRequestContext: The implementation now uses ContainerRequestContext instead of HttpServletRequest to access request details, as this is the object available within the JAX-RS filter context.
- Injecting Request Attributes: The extracted user and query information is now passed along via request attributes, making it easily accessible in the later stages of request routing

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static io.trino.gateway.ha.handler.HttpUtils.OAUTH_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.TRINO_UI_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.UI_API_STATS_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_INFO_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_NODE_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_QUERY_PATH;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A filter component that determines whether a given path should be whitelisted
+ * for routing to Trino clusters.
+ */
+@Singleton
+public class PathFilter
+{
+    private final Set<String> statementPaths;
+    private final List<Pattern> extraWhitelistPatterns;
+
+    @Inject
+    public PathFilter(
+            List<String> statementPaths,
+            List<String> extraWhitelistPaths)
+    {
+        this.statementPaths = Set.copyOf(requireNonNull(statementPaths, "Required configuration 'statementPaths' can't be null"));
+        this.extraWhitelistPatterns = requireNonNull(extraWhitelistPaths, "extraWhitelistPaths cannot be null").stream()
+                .map(pattern -> {
+                    try {
+                        return Pattern.compile(pattern);
+                    }
+                    catch (PatternSyntaxException e) {
+                        throw new IllegalArgumentException("Invalid regex pattern: " + pattern, e);
+                    }
+                })
+                .collect(toImmutableList());
+    }
+
+    /**
+     * Determines if the given path is whitelisted for routing to backend.
+     *
+     * @param path the request path to check
+     * @return true if the path should be routed to backend, false otherwise
+     */
+    public boolean isPathWhiteListed(String path)
+    {
+        return statementPaths.stream().anyMatch(path::startsWith)
+                || path.startsWith(V1_QUERY_PATH)
+                || path.startsWith(TRINO_UI_PATH)
+                || path.startsWith(V1_INFO_PATH)
+                || path.startsWith(V1_NODE_PATH)
+                || path.startsWith(UI_API_STATS_PATH)
+                || path.startsWith(OAUTH_PATH)
+                || extraWhitelistPatterns.stream().anyMatch(pattern -> pattern.matcher(path).matches());
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
@@ -65,13 +65,17 @@ import io.trino.sql.tree.StringLiteral;
 import io.trino.sql.tree.Table;
 import io.trino.sql.tree.TableFunctionInvocation;
 import io.trino.sql.tree.WithQuery;
-import jakarta.servlet.http.HttpServletRequest;
 import jakarta.ws.rs.HttpMethod;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.core.MediaType;
 
 import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.URLDecoder;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
@@ -142,35 +146,44 @@ public class TrinoQueryProperties
         maxBodySize = -1;
     }
 
-    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
+    public TrinoQueryProperties()
     {
-        requireNonNull(request, "request is null");
+        this("", "", "", ImmutableList.of(), Optional.empty(), Optional.empty(),
+                ImmutableSet.of(), ImmutableSet.of(), ImmutableSet.of(), false, Optional.empty());
+    }
+
+    public TrinoQueryProperties(ContainerRequestContext requestContext, boolean isClientsUseV2Format, int maxBodySize)
+    {
+        requireNonNull(requestContext, "requestContext is null");
         this.isClientsUseV2Format = isClientsUseV2Format;
         this.maxBodySize = maxBodySize;
 
-        defaultCatalog = Optional.ofNullable(request.getHeader(TRINO_CATALOG_HEADER_NAME));
-        defaultSchema = Optional.ofNullable(request.getHeader(TRINO_SCHEMA_HEADER_NAME));
-        if (request.getMethod().equals(HttpMethod.POST)) {
+        defaultCatalog = Optional.ofNullable(requestContext.getHeaderString(TRINO_CATALOG_HEADER_NAME));
+        defaultSchema = Optional.ofNullable(requestContext.getHeaderString(TRINO_SCHEMA_HEADER_NAME));
+        if (requestContext.getMethod().equals(HttpMethod.POST)) {
             isNewQuerySubmission = true;
-            processRequestBody(request);
+            processRequestBody(requestContext);
         }
     }
 
-    private void processRequestBody(HttpServletRequest request)
+    private void processRequestBody(BufferedReader reader, Map<String, String> preparedStatements)
     {
-        try (BufferedReader reader = request.getReader()) {
+        try (reader) {
             if (reader == null) {
                 log.warn("HTTP request returned null reader");
                 body = "";
                 return;
             }
 
-            Map<String, String> preparedStatements = getPreparedStatements(request);
             SqlParser parser = new SqlParser();
             reader.mark(maxBodySize);
             char[] buffer = new char[maxBodySize];
             int nChars = reader.read(buffer, 0, maxBodySize);
             reader.reset();
+            if (nChars <= 0) {
+                log.warn("query text is empty");
+                return;
+            }
             if (nChars == maxBodySize) {
                 log.warn("Query length greater or equal to requestAnalyzerConfig.maxBodySize detected");
                 return;
@@ -238,11 +251,51 @@ public class TrinoQueryProperties
         }
     }
 
-    private Map<String, String> getPreparedStatements(HttpServletRequest request)
+    private void processRequestBody(ContainerRequestContext requestContext)
+    {
+        if (!requestContext.hasEntity()) {
+            return;
+        }
+
+        MediaType mediaType = requestContext.getMediaType();
+        if (mediaType == null) {
+            return;
+        }
+
+        String charset = mediaType.getParameters().get("charset");
+        if (charset == null) {
+            log.debug("charset is not set in the request");
+            return;
+        }
+
+        if (!UTF_8.name().equalsIgnoreCase(charset)) {
+            log.debug("Request charset is not UTF-8 (%s), skipping query parsing", charset);
+            return;
+        }
+
+        InputStream entityStream = requestContext.getEntityStream();
+        try (InputStreamReader entityReader = new InputStreamReader(entityStream, UTF_8);
+                BufferedReader reader = new BufferedReader(entityReader)) {
+            processRequestBody(reader, getPreparedStatements(requestContext));
+        }
+        catch (IOException e) {
+            log.warn("Error extracting request body for rules processing: %s", e.getMessage());
+            errorMessage = Optional.of(e.getMessage());
+        }
+        catch (ParsingException e) {
+            log.info("Could not parse request body as SQL: %s; Message: %s", body, e.getMessage());
+            errorMessage = Optional.of(e.getMessage());
+        }
+        catch (RequestParsingException e) {
+            log.warn(e, "Error parsing request for rules");
+            errorMessage = Optional.of(e.getMessage());
+        }
+    }
+
+    private Map<String, String> getPreparedStatements(Enumeration<String> headers)
             throws RequestParsingException
     {
         ImmutableMap.Builder<String, String> preparedStatementsMapBuilder = ImmutableMap.builder();
-        Enumeration<String> headers = request.getHeaders(TRINO_PREPARED_STATEMENT_HEADER_NAME);
         if (headers == null) {
             return preparedStatementsMapBuilder.build();
         }
@@ -257,6 +310,19 @@ public class TrinoQueryProperties
             }
         }
         return preparedStatementsMapBuilder.build();
+    }
+
+    private Map<String, String> getPreparedStatements(ContainerRequestContext requestContext)
+            throws RequestParsingException
+    {
+        if (requestContext.getHeaders() == null) {
+            return ImmutableMap.of();
+        }
+        List<String> headers = requestContext.getHeaders().get(TRINO_PREPARED_STATEMENT_HEADER_NAME);
+        if (headers == null || headers.isEmpty()) {
+            return ImmutableMap.of();
+        }
+        return getPreparedStatements(Collections.enumeration(headers));
     }
 
     private String decodePreparedStatementFromHeader(String headerValue)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
@@ -33,14 +33,14 @@ import com.nimbusds.openid.connect.sdk.UserInfoResponse;
 import com.nimbusds.openid.connect.sdk.claims.UserInfo;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.RequestAnalyzerConfig;
-import jakarta.servlet.http.Cookie;
-import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.core.HttpHeaders;
 
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 import java.util.Base64;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -62,7 +62,7 @@ public class TrinoRequestUser
 
     private final Optional<LoadingCache<String, UserInfo>> userInfoCache;
 
-    private TrinoRequestUser(HttpServletRequest request, String userField, Optional<LoadingCache<String, UserInfo>> userInfoCache)
+    private TrinoRequestUser(ContainerRequestContext request, String userField, Optional<LoadingCache<String, UserInfo>> userInfoCache)
     {
         this.userInfoCache = requireNonNull(userInfoCache);
         user = extractUser(request, userField);
@@ -106,15 +106,17 @@ public class TrinoRequestUser
         return user.filter(testUser::equals).isPresent();
     }
 
-    private Optional<String> extractUserFromCookies(HttpServletRequest request, String userField)
+    private Optional<String> extractUserFromCookies(ContainerRequestContext requestContext, String userField)
     {
-        if (request.getCookies() == null) {
+        Map<String, jakarta.ws.rs.core.Cookie> cookies = requestContext.getCookies();
+        if (cookies == null || cookies.isEmpty()) {
+            log.debug("cookies are empty");
             return Optional.empty();
         }
-        log.debug("Trying to get user from cookie");
-        Optional<Cookie> uiToken = Arrays.stream(request.getCookies())
-                .filter(cookie -> cookie.getName().equals(TRINO_UI_TOKEN_NAME) || cookie.getName().equals(TRINO_SECURE_UI_TOKEN_NAME))
-                .findAny();
+
+        log.debug("Trying to get user from cookie from ContainerRequestContext");
+        Optional<jakarta.ws.rs.core.Cookie> uiToken = Optional.ofNullable(cookies.get(TRINO_UI_TOKEN_NAME))
+                .or(() -> Optional.ofNullable(cookies.get(TRINO_SECURE_UI_TOKEN_NAME)));
 
         return uiToken.map(t -> {
             try {
@@ -129,20 +131,20 @@ public class TrinoRequestUser
         });
     }
 
-    private Optional<String> extractUser(HttpServletRequest request, String userField)
+    private Optional<String> extractUser(ContainerRequestContext requestContext, String userField)
     {
         String header;
-        header = request.getHeader(TRINO_USER_HEADER_NAME);
+        header = requestContext.getHeaderString(TRINO_USER_HEADER_NAME);
         if (header != null) {
             return Optional.of(header);
         }
 
-        Optional<String> user = extractUserFromAuthorizationHeader(request.getHeader("Authorization"), userField);
+        Optional<String> user = extractUserFromAuthorizationHeader(requestContext.getHeaderString(HttpHeaders.AUTHORIZATION), userField);
         if (user.isPresent()) {
             return user;
         }
 
-        return extractUserFromCookies(request, userField);
+        return extractUserFromCookies(requestContext, userField);
     }
 
     private Optional<String> extractUserFromAuthorizationHeader(String header, String userField)
@@ -225,7 +227,7 @@ public class TrinoRequestUser
             }
         }
 
-        public TrinoRequestUser getInstance(HttpServletRequest request)
+        public TrinoRequestUser getInstance(ContainerRequestContext request)
         {
             return new TrinoRequestUser(request, userField, userInfoCache);
         }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.security;
+
+import com.google.inject.Inject;
+import io.airlift.log.Logger;
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.config.RequestAnalyzerConfig;
+import io.trino.gateway.ha.router.PathFilter;
+import io.trino.gateway.ha.router.TrinoQueryProperties;
+import io.trino.gateway.ha.security.util.GatewayFilterPriorities;
+import jakarta.annotation.Priority;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.container.PreMatching;
+import org.glassfish.jersey.server.ContainerRequest;
+
+import java.io.IOException;
+
+import static io.trino.gateway.ha.handler.HttpUtils.TRINO_QUERY_PROPERTIES;
+
+/**
+ *
+ * This filter parses the query statement and stores the TrinoQueryProperties object
+ * as a property to be accessed in the later processing
+ */
+@PreMatching
+@Priority(GatewayFilterPriorities.PRE_AUTHORIZATION)
+public class QueryMetadataParser
+        implements ContainerRequestFilter
+{
+    private static final Logger log = Logger.get(QueryMetadataParser.class);
+    private static final int MAX_QUERY_TEXT_LOG_LENGTH = 100;
+    private final boolean isAnalyzeRequest;
+    private final boolean isClientsUseV2Format;
+    private final int maxBodySize;
+    private final PathFilter pathFilter;
+
+    @Inject
+    public QueryMetadataParser(HaGatewayConfiguration config, PathFilter pathFilter)
+    {
+        RequestAnalyzerConfig analyzerConfig = config.getRequestAnalyzerConfig();
+        this.isAnalyzeRequest = analyzerConfig.isAnalyzeRequest();
+        this.isClientsUseV2Format = analyzerConfig.isClientsUseV2Format();
+        this.maxBodySize = analyzerConfig.getMaxBodySize();
+        this.pathFilter = pathFilter;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        if (path == null || !isAnalyzeRequest || !pathFilter.isPathWhiteListed(path)) {
+            return;
+        }
+
+        log.debug("Processing query metadata for path: %s", path);
+        // Buffer the entity (aka body of the request) for future reads during request processing
+        ContainerRequest jerseyRequest = (ContainerRequest) requestContext;
+        jerseyRequest.bufferEntity();
+
+        TrinoQueryProperties queryProps;
+        try {
+            queryProps = new TrinoQueryProperties(requestContext, isClientsUseV2Format, maxBodySize);
+        }
+        catch (Exception ex) {
+            log.warn(ex, "Failed to parse query properties for query text: [%s]. Error: %s. Using empty properties.",
+                    getQueryTextForLogging(requestContext), ex.getMessage());
+            queryProps = new TrinoQueryProperties();
+        }
+
+        requestContext.setProperty(TRINO_QUERY_PROPERTIES, queryProps);
+    }
+
+    private String getQueryTextForLogging(ContainerRequestContext requestContext)
+    {
+        try {
+            ContainerRequest jerseyRequest = (ContainerRequest) requestContext;
+
+            String body = jerseyRequest.readEntity(String.class);
+            if (body == null || body.isEmpty()) {
+                return "<empty>";
+            }
+            else if (body.length() > MAX_QUERY_TEXT_LOG_LENGTH) {
+                return body.substring(0, MAX_QUERY_TEXT_LOG_LENGTH) + "...";
+            }
+
+            return body;
+        }
+
+        catch (Exception e) {
+            log.error(e, "unable to read query text");
+            return "<error reading query>";
+        }
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.trino.gateway.ha.security;
+
+import com.google.inject.Inject;
+import io.airlift.log.Logger;
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.config.RequestAnalyzerConfig;
+import io.trino.gateway.ha.router.PathFilter;
+import io.trino.gateway.ha.router.TrinoRequestUser;
+import io.trino.gateway.ha.security.util.GatewayFilterPriorities;
+import jakarta.annotation.Priority;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.container.PreMatching;
+
+import java.io.IOException;
+
+import static io.trino.gateway.ha.handler.HttpUtils.TRINO_REQUEST_USER;
+
+/**
+ * A filter which parses and extracts Trino user identity from incoming request headers
+ * and stores it in the request context property TRINO_REQUEST_USER
+ * for downstream filters and handlers to use.
+ */
+
+@PreMatching
+@Priority(GatewayFilterPriorities.PRE_AUTHENTICATION)
+public class QueryUserInfoParser
+        implements ContainerRequestFilter
+{
+    private static final Logger log = Logger.get(QueryUserInfoParser.class);
+    private final String tokenUserField;
+    private final String oauthTokenInfoUrl;
+    private final PathFilter pathFilter;
+
+    @Inject
+    public QueryUserInfoParser(HaGatewayConfiguration config, PathFilter pathFilter)
+    {
+        RequestAnalyzerConfig requestAnalyzerConfig = config.getRequestAnalyzerConfig();
+        this.tokenUserField = requestAnalyzerConfig.getTokenUserField();
+        this.oauthTokenInfoUrl = requestAnalyzerConfig.getOauthTokenInfoUrl();
+        this.pathFilter = pathFilter;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        if (!pathFilter.isPathWhiteListed(path)) {
+            return;
+        }
+
+        RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
+        requestAnalyzerConfig.setTokenUserField(tokenUserField);
+        requestAnalyzerConfig.setOauthTokenInfoUrl(oauthTokenInfoUrl);
+
+        TrinoRequestUser user = new TrinoRequestUser.TrinoRequestUserProvider(requestAnalyzerConfig).getInstance(requestContext);
+        requestContext.setProperty(TRINO_REQUEST_USER, user);
+        log.debug("Parsed user %s", user.getUser().orElse("None"));
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouterPreMatchContainerRequestFilter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouterPreMatchContainerRequestFilter.java
@@ -14,15 +14,13 @@
 package io.trino.gateway.proxyserver;
 
 import com.google.inject.Inject;
-import io.trino.gateway.ha.handler.RoutingTargetHandler;
+import io.trino.gateway.ha.router.PathFilter;
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.container.ContainerRequestFilter;
 import jakarta.ws.rs.container.PreMatching;
 
 import java.io.IOException;
 import java.net.URI;
-
-import static java.util.Objects.requireNonNull;
 
 /**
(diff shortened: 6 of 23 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -289,6 +298,19 @@ private void validateStatementPath(String statementPath, List<String> statementP
...
+    {
+        List<Pattern> extraWhitelistPaths = requireNonNull(this.getExtraWhitelistPaths()).stream().map(Pattern::compile).collect(toImmutableList());
+        return statementPaths.stream().anyMatch(path::startsWith)
+                || path.startsWith(V1_QUERY_PATH)
+                || path.startsWith(TRINO_UI_PATH)
+                || path.startsWith(V1_INFO_PATH)
+                || path.startsWith(V1_NODE_PATH)
+                || path.startsWith(UI_API_STATS_PATH)
+                || path.startsWith(OAUTH_PATH)
+                || extraWhitelistPaths.stream().anyMatch(pattern -> pattern.matcher(path).matches());
```
Chaho12: This code is duplicate at [RoutingTargetHandler](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L121-L131).

If this is needed at early stage, why don't we remove it at RoutingTargetHandler and use this instead?
Also, we could update [RouterPreMatchContainerRequestFilter.java](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouterPreMatchContainerRequestFilter.java)

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -142,30 +146,29 @@ public TrinoQueryProperties(
         maxBodySize = -1;
     }
 
-    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
+    public TrinoQueryProperties(ContainerRequestContext reqContext, boolean isClientsUseV2Format, int maxBodySize)
```
Chaho12: I think it is allright to simply use `requestContext` for more descriptive purpose.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -238,11 +241,37 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
...
+        if (entityStream == null ||
+                !requestContext.hasEntity() || requestContext.getMediaType() == null ||
+                !requestContext.getMediaType().getParameters().containsKey("charset") ||
+                !StandardCharsets.UTF_8.name().equalsIgnoreCase(requestContext.getMediaType().getParameters().get("charset"))) {
+            return;
+        }
+        InputStreamReader entityReader = new InputStreamReader(entityStream, StandardCharsets.UTF_8);
+        try (BufferedReader reader = new BufferedReader(entityReader)) {
+            processRequestBody(reader, getPreparedStatements(requestContext));
+        }
```
Chaho12: We should close InputStreamReader explicitly or use try-with-resources for both `InputStreamReader` and `BufferedReader`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -238,11 +241,37 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
...
     }
 
-    private Map<String, String> getPreparedStatements(HttpServletRequest request)
+    private void processRequestBody(ContainerRequestContext requestContext)
+    {
+        InputStream entityStream = requestContext.getEntityStream();
+        if (entityStream == null ||
+                !requestContext.hasEntity() || requestContext.getMediaType() == null ||
+                !requestContext.getMediaType().getParameters().containsKey("charset") ||
+                !StandardCharsets.UTF_8.name().equalsIgnoreCase(requestContext.getMediaType().getParameters().get("charset"))) {
```
Chaho12: How about separating to a different function and simplify it to

```suggestion
    if (!shouldProcessBody(requestContext)) {
        return;
    }

```
```java
private boolean shouldProcessBody(ContainerRequestContext requestContext) {
    if (!requestContext.hasEntity()) return false;
    MediaType mediaType = requestContext.getMediaType();
    if (mediaType == null) return false;
    String charset = mediaType.getParameters().get("charset");
    return StandardCharsets.UTF_8.name().equalsIgnoreCase(charset);
}
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -238,11 +241,37 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
         }
     }
 
-    private Map<String, String> getPreparedStatements(HttpServletRequest request)
+    private void processRequestBody(ContainerRequestContext requestContext)
```
Chaho12, line 254: Maybe clarify name such as "processRequestBodyFromContext"?

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
vishalya: Good point, I will bring all the filters in sync with it.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
vishalya: 👍

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
vishalya: The parameter kinda tells it, so it would be redundant.

vishalya:
Addressed the review comments.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
vishalya: On a second thought, I simplified and made the checks more readable.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,70 @@
...
+        haGatewayConfiguration = config;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        RequestAnalyzerConfig reqAnalyzerconfig = haGatewayConfiguration.getRequestAnalyzerConfig();
+        if (!reqAnalyzerconfig.isAnalyzeRequest() ||
```
RoeyoOgen: why do you in some places (like here) explicitly call `reqAnalyzerconfig.isAnalyzeRequest()`, and in others (FileBasedRoutingGroupSelector innit) do you do `analyzeRequest = requestAnalyzerConfig.isAnalyzeRequest();`

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,70 @@
...
+                !haGatewayConfiguration.isPathWhiteListed(path)) {
+            return;
+        }
+
+        log.info("Processing query metadata");
```
RoeyoOgen: 1. maybe add queryID?
2. maybe change to debug and not info? could flood logs as its done for every req.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.IOException;
+
+import static io.trino.gateway.ha.handler.HttpUtils.TRINO_REQUEST_USER;
+
+/**
+ *
+ * A filter to parse the query request headers related to user and store the user information as a property
+ * in the request for later use
+ */
```
RoeyoOgen, line 37:
```suggestion
/**
 * A filter which parses and extracts Trino user identity from incoming request headers
 * and stores it in the request context property TRINO_REQUEST_USER
 * for downstream filters and handlers to use.
 */
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -0,0 +1,63 @@
...
+    private final HaGatewayConfiguration haGatewayConfiguration;
+
+    @Inject
+    public QueryUserInfoParser(HaGatewayConfiguration config)
+    {
+        haGatewayConfiguration = config;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
```
RoeyoOgen, line 59: NIT: `filter` both parses user info and stores it in request context. Perhaps split user extraction logic into a util func?

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,70 @@
...
+    {
+        haGatewayConfiguration = config;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        RequestAnalyzerConfig reqAnalyzerconfig = haGatewayConfiguration.getRequestAnalyzerConfig();
```
RoeyoOgen:
```suggestion
        RequestAnalyzerConfig reqAnalyzerConfig = haGatewayConfiguration.getRequestAnalyzerConfig();
```
missing caps

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,70 @@
...
+    private final HaGatewayConfiguration haGatewayConfiguration;
+
+    @Inject
+    public QueryMetadataParser(HaGatewayConfiguration config)
+    {
+        haGatewayConfiguration = config;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
```
RoeyoOgen, line 61: NIT: `Filter` both buffers the entity and constructs TrinoQueryProperties maybe split into two?

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/Priorities.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.security.util;
+
+public final class Priorities
```
RoeyoOgen: 1. why not have this as an enum?
2. maybe add a Javadoc explaining ordering?

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
```diff
@@ -194,63 +187,36 @@ private static Optional<String> extractQueryId(HttpServletRequest request)
...
-        // a minimal mock for this test.
-        HttpServletRequest request = request(query);
-        when(request.getHeader("X-Trino-Catalog")).thenReturn(defaultCatalog);
-        when(request.getHeader("X-trino-catalog")).thenReturn(defaultCatalog);
-        return request;
+        RequestAnalyzerConfig config = new RequestAnalyzerConfig();
+        config.setAnalyzeRequest(true);
+        return new QueryRequestMock().query(query)
+                .httpHeader("X-Trino-Catalog", defaultCatalog)
+                .httpHeader("X-trino-catalog", defaultCatalog)
```
RoeyoOgen: why is there both capital T and lower t? in `-Trino-`?
official docs only has uppercase.
(applies to below as well)

image

gateway-ha/src/test/java/io/trino/gateway/ha/util/QueryRequestMock.java
```diff
@@ -0,0 +1,170 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class QueryRequestMock
+{
+    private RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
+    private ContainerRequestContext requestContext = mock(ContainerRequest.class);
+    private HttpServletRequest mockRequest = mock(HttpServletRequest.class);
```
RoeyoOgen, line 57: mockRequest and requestContext are reused, do you want to retain state across multiple test runs ?

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
+        when(requestContext.getUriInfo()).thenReturn(uriInfo);
+
+        MediaType mediaType = new MediaType("application", "json", java.util.Map.of("charset", "UTF-8"));
+        when(requestContext.getMediaType()).thenReturn(mediaType);
+
+        String json = "Select xyz from cat1.schema1.table1";
```
RoeyoOgen:
```suggestion
        String queryText = "Select xyz from cat1.schema1.table1";
```

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,87 @@
...
+        InputStream entityStream = new ByteArrayInputStream(json.getBytes(StandardCharsets.UTF_8));
+        when(requestContext.getEntityStream()).thenReturn(entityStream);
+        when(requestContext.hasEntity()).thenReturn(true);
+        filter.filter(requestContext);
+
+        ArgumentCaptor<TrinoQueryProperties> captor = ArgumentCaptor.forClass(TrinoQueryProperties.class);
+        verify(requestContext).setProperty(eq(TRINO_QUERY_PROPERTIES), captor.capture());
+
+        // Optionally, check that the entity stream was read (if your filter does so)
+        verify(requestContext).getEntityStream();
```
RoeyoOgen, line 177: you check here the mock was called, perhaps check also that buffering worked or query was parsed?

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,87 @@
...
+{
+    private QueryMetadataParser filter;
+    private RequestAnalyzerConfig requestAnalyzerConfig;
+
+    @BeforeEach
+    void setUp()
+    {
+        HaGatewayConfiguration config = new HaGatewayConfiguration();
+        requestAnalyzerConfig = new RequestAnalyzerConfig();
+        requestAnalyzerConfig.setAnalyzeRequest(true);
```
RoeyoOgen, line 52: Only tests the happy path when = true. What about false?

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryUserInfoParser.java
```diff
@@ -0,0 +1,81 @@
...
+import java.net.URISyntaxException;
+import java.util.Base64;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class TestQueryUserInfoParser
```
ebyhr: https://trino.io/docs/current/develop/tests.html
> Test classes should be defined as package-private and final.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryUserInfoParser.java
```diff
@@ -0,0 +1,81 @@
...
+{
+    private QueryUserInfoParser filter;
+
+    @BeforeEach
```
ebyhr: https://trino.io/docs/current/develop/tests.html#keep-test-setup-and-teardown-simple

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryUserInfoParser.java
```diff
@@ -0,0 +1,81 @@
...
+        ContainerRequestContext requestContext = mock(ContainerRequest.class);
+        when(requestContext.getMethod()).thenReturn("POST");
+
+        UriInfo uriInfo = mock(ExtendedUriInfo.class);
+        try {
+            when(uriInfo.getRequestUri()).thenReturn(new URI("http://localhost" + HttpUtils.V1_STATEMENT_PATH));
+        }
+        catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
```
ebyhr: Remove redundant catch block.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryUserInfoParser.java
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    void testFilterSetsQueryUserInfo()
+            throws Exception
+    {
+        ContainerRequestContext requestContext = mock(ContainerRequest.class);
+        when(requestContext.getMethod()).thenReturn("POST");
+
+        UriInfo uriInfo = mock(ExtendedUriInfo.class);
+        try {
+            when(uriInfo.getRequestUri()).thenReturn(new URI("http://localhost" + HttpUtils.V1_STATEMENT_PATH));
```
ebyhr: Static import V1_STATEMENT_PATH.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestExternalRoutingGroupSelector.java
```diff
@@ -89,7 +89,7 @@ static RulesExternalConfiguration provideRoutingRuleExternalConfig()
 
     @Test
     void testByRoutingRulesExternalEngine()
-            throws URISyntaxException
+            throws URISyntaxException, IOException
```
ebyhr: Throwing `Exception` is fine in test methods.

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/Priorities.java
ebyhr: The name is too general.

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/Priorities.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public static final int PRE_AUTHENTICATION = 500;
+    public static final int AUTHENTICATION = 1000;
+    public static final int PRE_AUTHORIZATION = 1500;
+    public static final int AUTHORIZATION = 2000;
+    public static final int HEADER_VERIFIER = 2500;
+    public static final int HEADER_DECORATOR = 3000;
+    public static final int ENTITY_CODER = 4000;
+    public static final int USER = 5000;
+    public static final int ROUTE_TO_BACKEND = 6000;
```
ebyhr: Remove unsued codes.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -0,0 +1,63 @@
...
+ * A filter to parse the query request headers related to user and store the user information as a property
+ * in the request for later use
+ */
+@PreMatching
+@Priority(Priorities.PRE_AUTHENTICATION)
+public class QueryUserInfoParser
+        implements ContainerRequestFilter
+{
+    private static final Logger log = Logger.get(QueryUserInfoParser.class);
+    private final HaGatewayConfiguration haGatewayConfiguration;
```
ebyhr: Don't put mutable objects in fields. Same for other classes.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
ebyhr: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
vishalya: Here, I need a reference to haGatewayConfiguration for the `isPathWhiteListed` in addition to RequestAnalyzerConfig, so the pattern of usage is different.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
vishalya: We may not have query id on certain requests but converting it into debug.

gateway-ha/src/test/java/io/trino/gateway/ha/handler/TestQueryIdCachingProxyHandler.java
vishalya: Making it consistent.

gateway-ha/src/test/java/io/trino/gateway/ha/util/QueryRequestMock.java
vishalya: The idea is that a new object of QueryRequestMock will be created for each test.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
vishalya: Added the buffering check.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
vishalya: Added some false paths too.

oneonestar:
How about using a single filter to parse all the data instead of having multiple filters?
We can store all the necessary data in a single object.
I think this would simplify the code and also help avoid dependency or execution order issues between multiple filters in the future.

vishalya:
I would like to keep it separate as they are different and will be parsed at different priorities (user info might need to be parsed at pre-authentication and query might be parsed at pre-authorization), The query information may not be present for all the requests.

vishalya:
Added a new `class PathFilter` to reuse the logic in the filters.

vishalya:
@ebyhr - I have addressed the review comments.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -238,11 +246,45 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
...
+        }
+
+        MediaType mediaType = requestContext.getMediaType();
+        if (mediaType == null) {
+            return;
+        }
+
+        String charset = mediaType.getParameters().get("charset");
+        if (!StandardCharsets.UTF_8.name().equalsIgnoreCase(charset)) {
+            return;
```
Chaho12, line 273: should we log a warning that it is missing charset? because request without charset is quite common isn't it?
sth like this?
```
log.debug("Request charset is not UTF-8 (%s), skipping query parsing", charset);
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -65,13 +65,18 @@
...
 import jakarta.ws.rs.HttpMethod;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.core.MediaType;
 
 import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
```
Chaho12: in like 93, we already imported static StandardCharsets.UTF_8. We should use that instead

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -0,0 +1,75 @@
...
+            return;
+        }
+
+        RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
+        requestAnalyzerConfig.setTokenUserField(tokenUserField);
+        requestAnalyzerConfig.setOauthTokenInfoUrl(oauthTokenInfoUrl);
+
+        TrinoRequestUser user = new TrinoRequestUser.TrinoRequestUserProvider(requestAnalyzerConfig).getInstance(requestContext);
+        requestContext.setProperty(TRINO_REQUEST_USER, user);
+        log.debug("Parsed user %s", user.getUser().orElse("None"));
```
Chaho12, line 73: Nit.
```suggestion
        log.debug("Parsed user: %s", user.getUser().orElse("None"));
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        if (!pathFilter.isPathWhiteListed(path)) {
+            return;
+        }
+
+        RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
```
Chaho12, line 67: Do we need to create a new RequestAnalyzerConfig object instead of reusing the configuration that was already injected into the class? Then there is no need to extract and recreate fields.

gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
```diff
@@ -0,0 +1,71 @@
...
+{
+    private final Set<String> statementPaths;
+    private final List<Pattern> extraWhitelistPatterns;
+
+    @Inject
+    public PathFilter(
+            @Named("statementPaths") List<String> statementPaths,
+            @Named("extraWhitelistPaths") List<String> extraWhitelistPaths)
+    {
+        this.statementPaths = Set.copyOf(statementPaths);
```
Chaho12: nit. log that it cannot be null?
```suggestion
        this.statementPaths = Set.copyOf(requireNonNull(statementPaths, "statementPaths cannot be null"));
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
```diff
@@ -0,0 +1,71 @@
...
+        this.extraWhitelistPatterns = requireNonNull(extraWhitelistPaths).stream()
+                .map(Pattern::compile)
+                .collect(toImmutableList());
```
Chaho12: same. tell pattern cannot be null and throw the pattern exception log
```suggestion
        this.extraWhitelistPatterns = requireNonNull(extraWhitelistPaths, "extraWhitelistPaths cannot be null").stream()
                .map(pattern -> {
                    try {
                        return Pattern.compile(pattern);
                    } catch (PatternSyntaxException e) {
                        throw new IllegalArgumentException("Invalid regex pattern: " + pattern, e);
                    }
                })
                .collect(toImmutableList());
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,79 @@
...
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        if (!isAnalyzeRequest || !pathFilter.isPathWhiteListed(path)) {
+            return;
+        }
+
+        log.debug("Processing query metadata");
```
Chaho12: nit. make it for useful and tell path
```suggestion
        log.debug("Processing query metadata for path: %s", path);
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,79 @@
...
+        //Buffer the entity (aka body of the request) for future reads during request processing
+        ContainerRequest jerseyRequest = (ContainerRequest) requestContext;
+        jerseyRequest.bufferEntity();
+
+        TrinoQueryProperties queryProps = new TrinoQueryProperties(requestContext,
+                isClientsUseV2Format,
+                maxBodySize);
+
+        requestContext.setProperty(TRINO_QUERY_PROPERTIES, queryProps);
```
Chaho12: How about try catch exception as TrinoQueryProperties could throw exception? 
Cause gateway should continue processing even if query parsing fails

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -238,11 +246,45 @@ else if (statement instanceof ExecuteImmediate executeImmediate) {
...
+            return;
+        }
+
+        MediaType mediaType = requestContext.getMediaType();
+        if (mediaType == null) {
+            return;
+        }
+
+        String charset = mediaType.getParameters().get("charset");
+        if (!StandardCharsets.UTF_8.name().equalsIgnoreCase(charset)) {
```
Chaho12: We should first check if charset is null first.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
```diff
@@ -0,0 +1,177 @@
...
+        // Should allow configured statement paths
+        assertThat(statementOnlyFilter.isPathWhiteListed(V1_STATEMENT_PATH)).isTrue();
+        assertThat(statementOnlyFilter.isPathWhiteListed("/v2/statement")).isTrue();
+        assertThat(statementOnlyFilter.isPathWhiteListed("/custom/execute")).isTrue();
+        assertThat(statementOnlyFilter.isPathWhiteListed("/custom/execute/batch")).isTrue();
+
+        // Should not allow other paths
+        assertThat(statementOnlyFilter.isPathWhiteListed("/api/custom")).isFalse();
+        assertThat(statementOnlyFilter.isPathWhiteListed("/health")).isFalse();
+    }
```
Chaho12, line 178: can you add test for configuration validatinons? Test null statement paths, null whitelist paths, invalid regex pattern etc.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
vishalya: The query parsing failure should be a hard one, since a lot of rules/routing depends on it.

Chaho12: Yes I agree that Gateway depends on it. However, it shouldn't prevent users from executing queries. Users should expect the same result whether they run a query through the Gateway or directly against Trino. There are a few potential concerns:

1. Our parser implementation may have bugs or could be outdated compared to Trino's latest version. Since we cannot stay perfectly in sync with Trino at all times, discrepancies might occur.
2. Maintaining parsing logic in both the Gateway and Trino introduces redundancy and potential inconsistency. Ideally, query validation or parsing should be handled by a single source of truth. In this context, I consider the Gateway's parser more of an auxiliary parser for information rather than the primary validator.

vishalya: You have some valid points there, let me see handle the failure and continue as you have suggested.

xkrogen:
@vishalya are you still working on this? Would love to see it land -- this is a great cleanup

vishalya:
I’ll spend some time on the code comments this week

gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
Chaho12: @vishalya WDYT?

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -0,0 +1,75 @@
...
+    private final String tokenUserField;
+    private final String oauthTokenInfoUrl;
+    private final PathFilter pathFilter;
+
+    @Inject
+    public QueryUserInfoParser(HaGatewayConfiguration config, PathFilter pathFilter)
+    {
+        RequestAnalyzerConfig requestAnalyzerConfig = config.getRequestAnalyzerConfig();
+        this.tokenUserField = requestAnalyzerConfig.getTokenUserField();
+        this.oauthTokenInfoUrl = requestAnalyzerConfig.getOauthTokenInfoUrl();
```
Chaho12, line 54:
```suggestion
private final RequestAnalyzerConfig requestAnalyzerConfig;
...
        this.requestAnalyzerConfig = config.getRequestAnalyzerConfig();
```
We could get rid for tokenUserField and oauthTokenInfoUrl and have a requestAnalyzerConfig object instead

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
Chaho12: Then we could get rid of line 67~69, because requestAnalyzerConfig is private variable

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,108 @@
...
+        this.maxBodySize = analyzerConfig.getMaxBodySize();
+        this.pathFilter = pathFilter;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+            throws IOException
+    {
+        String path = requestContext.getUriInfo().getRequestUri().getPath();
+        if (!isAnalyzeRequest || !pathFilter.isPathWhiteListed(path)) {
```
Chaho12:
```suggestion
        if (path == null || !isAnalyzeRequest || !pathFilter.isPathWhiteListed(path)) {
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,108 @@
...
+            return;
+        }
+
+        log.debug("Processing query metadata for path: %s", path);
+        //Buffer the entity (aka body of the request) for future reads during request processing
```
Chaho12: nit.
```suggestion
        // Buffer the entity (aka body of the request) for future reads during request processing
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,108 @@
...
+import org.glassfish.jersey.server.ContainerRequest;
+
+import java.io.IOException;
+
+import static io.trino.gateway.ha.handler.HttpUtils.TRINO_QUERY_PROPERTIES;
+
+/**
+ *
+ * This filter parses the query statement and stores the TrinoQueryProperties object as a property to be accessed
+ * in the later processing
```
Chaho12: This comment seems to be long. How about new line after object?

Chaho12 (approved):
LGTM. Can you resolve comments if they are addressed :)

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
vishalya: I'm copying the field to address the issue raised [here](https://github.com/trinodb/trino-gateway/pull/748#discussion_r2300082810).

Chaho12: i see!

Chaho12:
Let's get this merged. This refactoring would affect others quite a bit.

gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
```diff
@@ -0,0 +1,78 @@
...
+@Singleton
+public class PathFilter
+{
+    private final Set<String> statementPaths;
+    private final List<Pattern> extraWhitelistPatterns;
+
+    @Inject
+    public PathFilter(
+            List<String> statementPaths,
+            List<String> extraWhitelistPaths)
```
ebyhr, line 46: Please inject `HaGatewayConfiguration` instead.

gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java
```diff
@@ -0,0 +1,78 @@
...
+    {
+        this.statementPaths = Set.copyOf(requireNonNull(statementPaths, "Required configuration 'statementPaths' can't be null"));
```
ebyhr, line 48: Use `ImmutableSet.copyOf` and remove `requireNonNull`.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoQueryProperties.java
```diff
@@ -142,35 +146,44 @@ public TrinoQueryProperties(
         maxBodySize = -1;
     }
 
-    public TrinoQueryProperties(HttpServletRequest request, boolean isClientsUseV2Format, int maxBodySize)
+    public TrinoQueryProperties()
     {
-        requireNonNull(request, "request is null");
+        this("", "", "", ImmutableList.of(), Optional.empty(), Optional.empty(),
+                ImmutableSet.of(), ImmutableSet.of(), ImmutableSet.of(), false, Optional.empty());
```
ebyhr, line 152: The convention is single line or separate lines. 
```java
        this(
                "",
                "",
                "",
                ImmutableList.of(),
                Optional.empty(),
                Optional.empty(),
                ImmutableSet.of(),
                ImmutableSet.of(),
                ImmutableSet.of(),
                false,
                Optional.empty());
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -106,15 +106,17 @@ public boolean userExistsAndEquals(String testUser)
         return user.filter(testUser::equals).isPresent();
     }
 
-    private Optional<String> extractUserFromCookies(HttpServletRequest request, String userField)
+    private Optional<String> extractUserFromCookies(ContainerRequestContext requestContext, String userField)
     {
-        if (request.getCookies() == null) {
+        Map<String, jakarta.ws.rs.core.Cookie> cookies = requestContext.getCookies();
```
ebyhr, line 111: Import jakarta.ws.rs.core.Cookie.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,108 @@
...
+ * as a property to be accessed in the later processing
+ */
+@PreMatching
+@Priority(GatewayFilterPriorities.PRE_AUTHORIZATION)
+public class QueryMetadataParser
+        implements ContainerRequestFilter
+{
+    private static final Logger log = Logger.get(QueryMetadataParser.class);
+    private static final int MAX_QUERY_TEXT_LOG_LENGTH = 100;
+    private final boolean isAnalyzeRequest;
```
ebyhr, line 45: Add an empty line between constants and fields.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,108 @@
...
+        catch (Exception ex) {
+            log.warn(ex, "Failed to parse query properties for query text: [%s]. Error: %s. Using empty properties.",
+                    getQueryTextForLogging(requestContext), ex.getMessage());
+            queryProps = new TrinoQueryProperties();
+        }
+
+        requestContext.setProperty(TRINO_QUERY_PROPERTIES, queryProps);
+    }
+
+    private String getQueryTextForLogging(ContainerRequestContext requestContext)
```
ebyhr, line 87: This method can be static.

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryMetadataParser.java
```diff
@@ -0,0 +1,108 @@
...
+    {
+        try {
+            ContainerRequest jerseyRequest = (ContainerRequest) requestContext;
+
+            String body = jerseyRequest.readEntity(String.class);
+            if (body == null || body.isEmpty()) {
+                return "<empty>";
+            }
+            else if (body.length() > MAX_QUERY_TEXT_LOG_LENGTH) {
```
ebyhr, line 96: Remove redundant else: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#additional-ide-configuration

gateway-ha/src/main/java/io/trino/gateway/proxyserver/RouterPreMatchContainerRequestFilter.java
```diff
@@ -35,19 +33,19 @@ public class RouterPreMatchContainerRequestFilter
...
 
-    private final RoutingTargetHandler routingTargetHandler;
+    private final PathFilter pathFilter;
 
     @Inject
-    public RouterPreMatchContainerRequestFilter(RoutingTargetHandler routingTargetHandler)
+    public RouterPreMatchContainerRequestFilter(PathFilter pathFilter)
     {
-        this.routingTargetHandler = requireNonNull(routingTargetHandler);
+        this.pathFilter = pathFilter;
```
ebyhr, line 41: Add requireNonNull.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
```diff
@@ -0,0 +1,219 @@
...
+import static io.trino.gateway.ha.handler.HttpUtils.V1_INFO_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_NODE_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_QUERY_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_STATEMENT_PATH;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class TestPathFilter
+{
+    private PathFilter pathFilter;
```
ebyhr, line 34: Make this field final.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
```diff
@@ -0,0 +1,219 @@
...
+        assertThat(pathFilter.isPathWhiteListed("/v1/quer")).isFalse(); // Partial match of hardcoded path
+        assertThat(pathFilter.isPathWhiteListed("/api/v1/custo")).isFalse(); // Partial match of regex
+    }
+
+    @Test
+    void testRegexPattern()
+    {
+        List<String> complexStatementPaths = ImmutableList.of(V1_STATEMENT_PATH);
+        List<String> complexRegexPaths = ImmutableList.of(
+                "/api/v[1-9]/.*");
```
ebyhr, line 129: Unwrap.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
```diff
@@ -0,0 +1,219 @@
...
+
+    @Test
+    void testDefaultHaConfigurationNullPaths()
+    {
+        HaGatewayConfiguration configuration = new HaGatewayConfiguration();
+
+        assertThatThrownBy(() -> {
+            new PathFilter(null,
+                    configuration.getExtraWhitelistPaths());
+        }).isInstanceOf(NullPointerException.class);
```
ebyhr, line 217:
```suggestion
        assertThatThrownBy(() -> new PathFilter(null, configuration.getExtraWhitelistPaths()))
                .isInstanceOf(NullPointerException.class);
```

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,208 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+final class TestQueryMetadataParser
+{
+    private QueryMetadataParser filter;
+    private RequestAnalyzerConfig requestAnalyzerConfig;
+    private PathFilter pathFilter;
```
ebyhr, line 46: Make these fields final.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,208 @@
...
+        ContainerRequestContext requestContext = mock(ContainerRequest.class);
+        when(requestContext.getMethod()).thenReturn("POST");
+
+        UriInfo uriInfo = mock(ExtendedUriInfo.class);
+        try {
+            when(uriInfo.getRequestUri()).thenReturn(new URI("http://localhost" + HttpUtils.V1_STATEMENT_PATH));
+        }
+        catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
```
ebyhr, line 71: Remove redundant try-catch.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,208 @@
...
+        ContainerRequestContext requestContext = mock(ContainerRequest.class);
+        when(requestContext.getMethod()).thenReturn("POST");
+
+        UriInfo uriInfo = mock(ExtendedUriInfo.class);
+        try {
+            when(uriInfo.getRequestUri()).thenReturn(new URI("http://localhost" + HttpUtils.OAUTH_PATH));
+        }
+        catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
```
ebyhr, line 102: Remove redundant try-catch.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,208 @@
...
+        ContainerRequestContext requestContext = mock(ContainerRequest.class);
+        when(requestContext.getMethod()).thenReturn("POST");
+
+        UriInfo uriInfo = mock(ExtendedUriInfo.class);
+        try {
+            when(uriInfo.getRequestUri()).thenReturn(new URI("http://localhost" + HttpUtils.V1_STATEMENT_PATH));
+        }
+        catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
```
ebyhr, line 133: Remove redundant try-catch.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,208 @@
...
```
ebyhr, line 162: Remove redundant try-catch.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryMetadataParser.java
```diff
@@ -0,0 +1,208 @@
...
```
ebyhr, line 196: Remove redundant try-catch.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestQueryUserInfoParser.java
```diff
@@ -0,0 +1,78 @@
...
+import static io.trino.gateway.ha.handler.HttpUtils.V1_STATEMENT_PATH;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+final class TestQueryUserInfoParser
+{
+    private QueryUserInfoParser filter;
```
ebyhr, line 42: Make this field constant.

PR #752, opened 2025-08-15
Add sorting for clusters and query history page columns

maswin:
## Description

Added sort options for clusters based on name, queued query count and running query count.
Added filter option for routingGroup and sort options for routignGroup and submission time in query history page.

Clusters page header [3 new sort icons on Name, Queued, Running]:
Screenshot 2025-08-19 at 11 12 21 AM

History Page [Sort and Filter option on RoutingGroup] - 
Screenshot 2025-08-19 at 11 12 40 AM

## Release notes

( X) Release notes are required, with the following suggested text:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,3 +19,6 @@ ij_java_doc_align_param_comments = false
 
 [*.yml]
 indent_size = 2
+
+[*.tsx]
+indent_size = 2
--- a/webapp/src/components/history.tsx
+++ b/webapp/src/components/history.tsx
@@ -133,13 +133,33 @@ export function History() {
           onPageChange: list,
         }}>
           <Column title="QueryId" dataIndex="queryId" key="queryId" render={linkQueryRender} />
-          <Column title="RoutingGroup" dataIndex="routingGroup" key="routingGroup" render={routingGroupRender} />
+          <Column title="RoutingGroup" dataIndex="routingGroup" key="routingGroup"
+            sorter={(a, b) => {
+              if (!a || !b) return 0;
+              return a.routingGroup.localeCompare(b.routingGroup);
+            }}
+            filters={
+                [...new Set(backendData?.map(b => b.routingGroup))]
+                        .map(routingGroup => {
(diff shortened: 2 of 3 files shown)


maswin:
CI build seemed to have failed with an unrelated error - 
`dependency failed to start: container docker-postgres-1 is unhealthy`

ebyhr:
Could you add screenshots to PR description when you modify UI code?

maswin:
> Could you add screenshots to PR description when you modify UI code?

Sure, done

webapp/src/components/history.tsx
```diff
@@ -133,13 +133,33 @@ export function History() {
...
         }}>
           <Column title="QueryId" dataIndex="queryId" key="queryId" render={linkQueryRender} />
-          <Column title="RoutingGroup" dataIndex="routingGroup" key="routingGroup" render={routingGroupRender} />
+          <Column title="RoutingGroup" dataIndex="routingGroup" key="routingGroup"
+            sorter={(a, b) => {
+              if (!a || !b) return 0;
+              return a.routingGroup.localeCompare(b.routingGroup);
+            }}
+            filters={
+                [...new Set(backendData?.map(b => b.routingGroup))]
```
RoeyoOgen, line 142: if `backendData` is undefined this may become `new Set([undefined])`, perhaps default to empty array

Chaho12: if backendData  is undefined, it will be new Set(undefined) which will give an empty set. So filters array will be simply empty array so i think it is ok like this.

Chaho12 (approved):
LGTM 👍

.editorconfig
```diff
@@ -19,3 +19,6 @@ ij_java_doc_align_param_comments = false
 
 [*.yml]
 indent_size = 2
+
+[*.tsx]
+indent_size = 2
```
ebyhr, line 24: Please separate a commit. This change is unrelated to "Add sorting for clusters and query history page columns".

PR #754, opened 2025-08-18
Fix querycount based router

vishalya:
## Description
The query count based router was broken after the feature ["Use external url for query routing history"](https://github.com/trinodb/trino-gateway/commit/3430f352800e852715613fd3619bf700938dfb07) was added. The main reason being that the new interface methods were not overridden correctly.
This PR would fix that issue by adding the correct implementation.

Also, we are trying to address the root cause by refactoring the log in [this PR ](https://github.com/trinodb/trino-gateway/pull/741). 

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
@@ -21,6 +21,7 @@ import com.google.inject.Inject;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
 import io.trino.gateway.ha.clustermonitor.TrinoStatus;
+import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.config.RoutingConfiguration;
 
 import java.util.ArrayList;
@@ -51,6 +52,7 @@ public class QueryCountBasedRouter
         private int queuedQueryCount;
         private TrinoStatus trinoStatus;
         private String proxyTo;
+        private String externalUrl;
         private String routingGroup;
         private String clusterId;
         private Map<String, Integer> userQueuedCount;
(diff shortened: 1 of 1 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -233,6 +256,20 @@ public String provideDefaultCluster(String user)
...
+        Optional<LocalStats> cluster = getClusterToRoute(user, routingGroup);
+        cluster.ifPresent(c -> updateLocalStats(c, user));
+        return cluster.map(c -> c.backendConfiguration());
+    }
+
+    @Override
+    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
+    {
+        return getBackendConfigurationForRoutingGroup(routingGroup, user)
+                .orElseGet(() -> getBackendConfigurationForRoutingGroup(defaultRoutingGroup, user).orElseThrow(() -> new RouterException("did not find any cluster for the default routing group: " + defaultRoutingGroup)));
```
Chaho12: This seem to be too long. We should wrap it so it is much shorter.

gateway-ha/src/main/java/io/trino/gateway/ha/router/QueryCountBasedRouter.java
```diff
@@ -233,6 +256,20 @@ public String provideDefaultCluster(String user)
         return getBackendForRoutingGroup(defaultRoutingGroup, user).orElseThrow(() -> new RouterException("did not find any cluster for the default routing group: " + defaultRoutingGroup));
     }
 
+    public Optional<ProxyBackendConfiguration> getBackendConfigurationForRoutingGroup(String routingGroup, String user)
+    {
+        Optional<LocalStats> cluster = getClusterToRoute(user, routingGroup);
+        cluster.ifPresent(c -> updateLocalStats(c, user));
+        return cluster.map(c -> c.backendConfiguration());
```
ebyhr: Avoid abbreviations and replace lambda with method reference.

mosabua:
Wont need release notes since it got broken after 15 was released

PR #763, opened 2025-08-31
Add username info on Non OK HTTP Status code

ignitz:
## Description

Currently, when Trino Gateway logs non-OK HTTP status codes, such as 401 Unauthorized, it does not include information about the user who initiated the request. This lack of context makes troubleshooting difficult, as it requires manually cross-referencing timestamps with external logs (e.g., Ingress) to identify the source of the problematic requests.

Just to include the username in the log output for these error scenarios, significantly speeding up diagnostics and reducing needed to find those users in another place.

## Release notes


ebyhr:
The commit body is too long. Please follow https://trino.io/development/process.html#pull-request-and-commit-guidelines- from next time.

Chaho12:
Got it. Will make sure to check if commit body is not too long

PR #709, opened 2025-06-25
Exclude mina-core from Apache Directory LDAP API library

Akanksha-kedia:
## Description

- Mina-core (org.apache.mina:mina-core):2.2.3

  - Not directly referenced in the code
  -  transitive dependency through org.apache.directory.api:api-all
  - No direct evidence of usage

Screenshot 2025-06-25 at 9 17 37 PM
even on excluding from org.apache.directory.api:api-all and adding explicit 

        <dependency>
            <groupId>org.apache.mina</groupId>
            <artifactId>mina-core</artifactId>
            <version>2.2.4</version>
        </dependency> 
        [ERROR] Unused declared dependencies found:
[ERROR] org.apache.mina:mina-core:jar:2.2.4:compile
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 31:31 min
[INFO] Finished at: 2025-06-25T16:58:59+05:30
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.8.1:analyze-only (default) on project gateway-ha: Dependency problems found -> [Help 1]
[ERROR] and also checked all the classes no import or this dependency is being used.
        

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required, with the following suggested text:

```markdown
* Fix some things.
```

PR #769, opened 2025-09-29
Fix the vulnerability in postgresql

Akanksha-kedia:
## Description

Screenshot 2025-09-29 at 2 41 46 PM

Screenshot 2025-09-29 at 3 10 46 PM

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -323,7 +323,7 @@
         <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
-            <version>42.7.7</version>
+            <version>42.7.8</version>
             <scope>runtime</scope>
         </dependency>
 


Akanksha-kedia:
@mosabua please review.

gateway-ha/pom.xml
```diff
@@ -128,6 +128,16 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.jetty.http2</groupId>
+                    <artifactId>jetty-http2-common</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.eclipse.jetty.http2</groupId>
+                    <artifactId>jetty-http2-server</artifactId>
+                </exclusion>
+            </exclusions>
```
ebyhr, line 140: Revert and update airlift/jetty instead.

ebyhr: Please explain why this exclusion is safe. We recently broke this project by dependency exclusion https://github.com/trinodb/trino-gateway/pull/709.

ebyhr (changes requested):

gateway-ha/pom.xml
mosabua: Agreed .. we do NOT want randomly add exclusions like that just because some security issue reporting shows up without any possible exploit.

gateway-ha/pom.xml
```diff
@@ -323,7 +333,7 @@
         <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
-            <version>42.7.7</version>
+            <version>42.7.8</version>
```
mosabua, line 326: This might be ok in a separate PR.

gateway-ha/pom.xml
mosabua: In this case .. jetty is required..

Akanksha-kedia:
have avoided jetty and only kept postgresql
@mosabua @ebyhr please review

ebyhr:
@Akanksha-kedia Please update the commit title.

mosabua:
In the future please do NOT use a misleading commit message like this. The change just does an update of the dependency. There was no exploit documented as possible before this change. If you really want to include more info you can do it as a separate text after the commit message and then link to the actual CVE report.

Maybe like so:

```
Update PostgreSQL JDBC driver to 42.7.8

- Removes reporting for CVE xyz.
- No known exploit existed before this change.
```

PR #770, opened 2025-09-30
Improve code quality by addressing TODO comments

Akanksha-kedia:
### __src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java__

- __Change:__ Fix silent failure on invalid backend URL configuration
- __Impact:__ Fail-fast behavior with descriptive error messages
- __Lines:__ Replace `return clusterStats.build()` with `throw new IllegalArgumentException()`

### __src/test/java/io/trino/gateway/ha/security/TestOIDC.java__

- __Change:__ Replace `System.out.println()` with proper logging
- __Impact:__ Professional logging practices in test code
- __Lines:__ Add Logger import and replace print statement with `log.info()`

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
@@ -78,8 +78,7 @@ public class ClusterStatsJdbcMonitor
             properties.setProperty("SSL", String.valueOf(parsedUrl.getProtocol().equals("https")));
         }
         catch (MalformedURLException e) {
-            log.error("could not parse backend url %s ", url);
-            return clusterStats.build(); // TODO Invalid configuration should fail
+            throw new IllegalArgumentException("Invalid backend URL: " + url, e);
         }
 
         try (Connection conn = DriverManager.getConnection(jdbcUrl, properties);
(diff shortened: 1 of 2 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsJdbcMonitor.java
```diff
@@ -78,8 +78,8 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
             properties.setProperty("SSL", String.valueOf(parsedUrl.getProtocol().equals("https")));
         }
         catch (MalformedURLException e) {
-            log.error("could not parse backend url %s ", url);
-            return clusterStats.build(); // TODO Invalid configuration should fail
+            log.error("Invalid backend URL configuration: %s", url);
+            throw new IllegalArgumentException("Invalid backend URL: " + url, e);
```
ebyhr, line 81: We avoid both logging and throwing exceptions - either one should be sufficient. Remove the above logging.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOIDC.java
```diff
@@ -158,7 +161,7 @@ void setup()
         File testConfigFile =
                 HaGatewayTestUtils.buildGatewayConfig("auth/oauth-test-config.yml", additionalVars);
         String[] args = {testConfigFile.getAbsolutePath()};
-        System.out.println(ROUTER_PORT);
+        log.info("Starting gateway on port: %d", ROUTER_PORT);
```
ebyhr: I don't think this log is helpful. Remove.

ebyhr (changes requested):

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestOIDC.java
mosabua: Its just a test also .. the system out can be removed as well

PR #776, opened 2025-10-07
Fix partialCancel statement path redirection

maswin:
## Description
Partial cancel url path is of form `/v1/statement/executing/partialCancel/<query_id>/...`
Ref - https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/server/protocol/Query.java#L726

But in the code we are considering it to be of form - `/v1/statement/partialCancel/<query_id>/...` without the `executing` prefix which causes failures while extracting query id. Fixed it with some minor refactoring. 

## Release notes

```markdown
* bug fix: Fix partialCancel statement path redirection
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
@@ -13,6 +13,7 @@
  */
 package io.trino.gateway.ha.handler;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.common.io.CharStreams;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.router.TrinoQueryProperties;
@@ -24,6 +25,7 @@ import java.io.InputStreamReader;
 import java.net.URI;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -52,6 +54,8 @@ public final class ProxyUtils
      * capitalization.
      */
     private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
+    private static final Set<String> QUERY_STATE_PATH = ImmutableSet.of("queued", "scheduled", "executing");
+    private static final String PARTIAL_CANCEL_PATH = "partialCancel";
 
     private ProxyUtils() {}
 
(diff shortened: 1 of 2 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -52,6 +53,8 @@ public final class ProxyUtils
      * capitalization.
      */
     private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
+    public static final ImmutableSet<String> QUERY_STATE_PATH = ImmutableSet.of("queued", "scheduled", "executing");
+    public static final String PARTIAL_CANCEL_PATH = "partialCancel";
```
ebyhr: Change to private.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -52,6 +53,8 @@ public final class ProxyUtils
      * capitalization.
      */
     private static final Pattern QUERY_ID_PARAM_PATTERN = Pattern.compile(".*(?:%2F|(?i)query_?id(?-i)=|^)(\\d+_\\d+_\\d+_\\w+).*");
+    public static final ImmutableSet<String> QUERY_STATE_PATH = ImmutableSet.of("queued", "scheduled", "executing");
```
ebyhr: Use an interface for declaration:
```suggestion
    public static final Set<String> QUERY_STATE_PATH = ImmutableSet.of("queued", "scheduled", "executing");
```

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -100,10 +103,10 @@ public static Optional<String> extractQueryIdIfPresent(String path, String query
...
             String[] tokens = path.split("/");
             if (tokens.length >= 2) {
-                if (tokens[1].equals("queued")
-                        || tokens[1].equals("scheduled")
-                        || tokens[1].equals("executing")
-                        || tokens[1].equals("partialCancel")) {
+                if (tokens.length >= 3 && QUERY_STATE_PATH.contains(tokens[1])) {
+                    if (tokens.length >= 4 && tokens[2].equals(PARTIAL_CANCEL_PATH)) {
+                        return Optional.of(tokens[3]);
+                    }
```
ebyhr, line 110: Can we add a test?

maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
maswin: Done

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -100,10 +104,10 @@ public static Optional<String> extractQueryIdIfPresent(String path, String query
             path = path.replace(matchingStatementPath.orElse(V1_QUERY_PATH), "");
             String[] tokens = path.split("/");
             if (tokens.length >= 2) {
-                if (tokens[1].equals("queued")
-                        || tokens[1].equals("scheduled")
-                        || tokens[1].equals("executing")
-                        || tokens[1].equals("partialCancel")) {
+                if (tokens.length >= 3 && QUERY_STATE_PATH.contains(tokens[1])) {
```
vishalya, line 107: I am fine with the current fix to go in as is, but we should definitely refactor this block of the code as some point.

maswin:
@ebyhr - Is this good to be merged?

PR #781, opened 2025-10-09
Add time-based cluster scheduling feature

hashroute3:
## Description

This PR introduces time-based cluster scheduling to Trino Gateway, allowing administrators to automatically activate and deactivate clusters based on cron schedules. This feature is particularly useful for:

This feature is implemented as a separate, optional module that is not included by default — it can be added on-demand for deployments that require time-based scheduling capabilities.

- Cost optimization by dynamically activating or deactivating clusters during off-hours. While Trino Gateway itself does not shut down clusters, it can stop routing queries to them, allowing external autoscaling mechanisms to scale the clusters down naturally.
- Workload management by activating specific clusters during peak hours
- Maintenance windows by scheduling periods when clusters are excluded from routing.
- Different timezone support for global deployments

Key features:
- Cron-based scheduling with UNIX cron expression support
- Configurable timezone support (defaults to America/Los_Angeles)
- Flexible activation logic with `activeDuringCron` flag
- Configurable check intervals for schedule evaluation
- Graceful error handling for invalid configurations
- Comprehensive logging for visibility into scheduling decisions and state transitions
- Modular design — provided as a standalone module that is not enabled by default and can be included only when needed

Example configuration:
```yaml
scheduleConfiguration:
  enabled: true
  checkInterval: 5m
  timezone: America/New_York  # Override default PST timezone if needed
  schedules:
    - clusterName: production-cluster
      cronExpression: "* 9-17 * * 1-5"  # Active during business hours (M-F)
      activeDuringCron: true
    - clusterName: dev-cluster
      cronExpression: "* 8-20 * * *"    # Active 8 AM to 8 PM daily
      activeDuringCron: true
```

The scheduler works by:
1. Parsing cron expressions at startup using the `cronutils` library
2. Periodically checking each cluster's schedule against the current time
3. Activating/deactivating clusters based on:
   - Whether the current time matches the cron schedule
   - The `activeDuringCron` flag (determines if cluster should be active during or outside the schedule)
   - The cluster's current state

Implementation details:
- Implemented as a separate module that is not loaded by default — deployments can include it only
- Uses `ScheduledExecutorService` for reliable scheduling
- Thread-safe using `ConcurrentHashMap` for execution times
- Graceful handling of configuration changes and errors
- Comprehensive unit tests covering various scenarios
- Default timezone set to America/Los_Angeles (PST/PDT) for backward compatibility

## Release notes

```markdown
*Add time-based cluster scheduling as an optional module, allowing clusters to be automatically activated or deactivated based on cron schedules.
This feature is not enabled by default and can be included on-demand for deployments that need time-based scheduling.
It enables cost optimization by aligning cluster routing availability with predictable usage patterns — while Trino Gateway does not directly shut down clusters, disabling routing allows external autoscaling systems to scale resources down naturally.
  Supports:
  - UNIX cron expressions for flexible scheduling
  - Configurable timezones (defaults to America/Los_Angeles)
  - Per-cluster schedules with customizable activation logic
  - Configurable check intervals
```

## Summary by Sourcery

Implement an optional cron-based cluster scheduler that parses schedules from configuration, evaluates them at fixed intervals, and uses the backend manager to toggle routing for clusters according to defined time windows

New Features:
- Introduce an optional time-based cluster scheduling module to automatically activate or deactivate clusters based on cron expressions
- Support configurable timezones for schedule evaluation (defaulting to America/Los_Angeles)
- Allow per-cluster schedules with flexible activation logic (activeDuringCron flag) and customizable check intervals

Enhancements:
- Integrate the scheduler into the application via a Guice module and lifecycle hooks
- Extend HaGatewayConfiguration to include ScheduleConfiguration for optional scheduling support

Build:
- Add cron-utils, jakarta.inject-api, and slf4j-api dependencies to the POM

Documentation:
- Provide an example schedule-config.yaml demonstrating time-based activation settings

Tests:
- Add comprehensive unit tests for ClusterScheduler covering activation, deactivation, timezone handling, invalid configurations, and multiple clusters

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -279,11 +279,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.weakref</groupId>
             <artifactId>jmxutils</artifactId>
         </dependency>
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/ClusterSchedulerConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/ClusterSchedulerConfiguration.java
@@ -14,16 +14,15 @@
 package io.trino.gateway.ha.config;
 
 import com.google.inject.Inject;
+import io.airlift.log.Logger;
 import io.trino.gateway.ha.scheduler.ClusterScheduler;
 import jakarta.annotation.Nullable;
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class ClusterSchedulerConfiguration
 {
-    private static final Logger log = LoggerFactory.getLogger(ClusterSchedulerConfiguration.class);
+    private static final Logger log = Logger.get(ClusterSchedulerConfiguration.class);
 
     private final ClusterScheduler scheduler;
 
@@ -49,7 +48,7 @@ public class ClusterSchedulerConfiguration
                 scheduler.close();
             }
             catch (Exception e) {
-                log.error("Exception occurred while shutting down ClusterScheduler", e);
+                log.error(e, "Exception occurred while shutting down ClusterScheduler");
             }
         }
     }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/scheduler/ClusterScheduler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/scheduler/ClusterScheduler.java
@@ -19,14 +19,13 @@ import com.cronutils.model.definition.CronDefinition;
 import com.cronutils.model.definition.CronDefinitionBuilder;
 import com.cronutils.model.time.ExecutionTime;
 import com.cronutils.parser.CronParser;
+import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.config.ScheduleConfiguration;
 import io.trino.gateway.ha.router.GatewayBackendManager;
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
 import jakarta.inject.Inject;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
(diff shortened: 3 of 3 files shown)


Chaho12:
ps. Check on theses docs for 
- graceful shutdown: https://trinodb.github.io/trino-gateway/operation/#start-trino-gateway-server
- health check: https://trinodb.github.io/trino-gateway/installation/?h=health#additional-options

andythsu:
dang, the sequence diagram from sourcery.ai is 💯

andythsu:
To answer some of your motivations behind this change:

```
Cost optimization by dynamically activating or deactivating clusters during off-hours. While Trino Gateway itself does not shut down clusters, it can stop routing queries to them, allowing external autoscaling mechanisms to scale the clusters down naturally.
```
We used to let Trino Gateway handle activation/deactivation of the clusters but then we found some issues
1. If the clusters are **deactivated**, they are considered gone from Trino Gateway's perspective, and admins will have to intervene to activate the clusters, which is easily forgotten and therefore causing more problems (no queries can be routed)
2. If the clusters are activated/deactivated automatically by Trino Gateway (whether intentional or unintentional), often times we don't get any notification, or even if we do, those notifications will be overlooked. This will cause problems for query routing as well.

Therefore, we've decided to **disable automatic activation/deactivation from Trino Gateway completely** since it's a very important switch that only admins should decide when to make the decision.

Even if in the ideal world we fix these issues, it would still be easier and more straightforward to just shut down the clusters instead of "deactivating" from Trino Gateway because at the end of the day Trino Gateway is just routing requests to the clusters. If you want to optimize cost, you should just turn off the clusters. Blinding Trino Gateway from the cluster isn't going to optimize your cost

```
Workload management by activating specific clusters during peak hours
```
We talked about dynamically bringing up Trino clusters during peak hours and killing them during non-peak hours, but this feature was not implemented. It may not even be a good feature in Trino Gateway. IMO, the best approach is to have some sort of monitoring system that monitors your traffic, brings up more Trino clusters if needed, and registers the Trino clusters to Trino Gateway.

For some context/background knowledge, current Trino Gateway has a health check that checks Trino clusters' health. It can be `HEALTHY`, `UNHEALTHY`, `PENDING`. Definition on each type can be found https://trinodb.github.io/trino-gateway/routing-rules/?h=healthy#trinostatus

tl;dr, if you shut down the Trino cluster, Trino Gateway will **not** deactivate the cluster, but will mark the cluster as `UNHEALTHY`. Trino Gateway will **not** route the requests to `UNHEALTHY` clusters, but will continue probing them. However, if you deactivate the clusters on Trino Gateway, it loses the ability to "revive" the cluster without admin's intervention.

hashroute3:
Thanks for the detailed context,  helps clarify the rationale behind disabling automatic activation/deactivation in Gateway.

To clarify, this feature does not start or stop Trino clusters. It simply provides time-based routing control, letting Gateway temporarily include or exclude clusters from routing based on configurable schedules. This aligns Gateway behavior with autoscaling systems that already bring clusters up/down during predictable windows, reducing unnecessary routing attempts or failures.
few points for clarification:
The module is optional and disabled by default, preserving existing behavior.
It doesn’t modify cluster activation state — only routing eligibility.

From testing cluster lifecycle management relying solely on HEALTHY/UNHEALTHY states, I observed a few challenges:

* Graceful termination only applies to workers, shutting down a cluster kills in-flight queries since the coordinator also goes down.

* When a cluster restarts, Gateway often marks it HEALTHY before all workers are ready, briefly increasing latency compared to other clusters.

We are already using this mechanism and helped mitigate both by coordinating routing availability with known cluster schedules.

PR #783, opened 2025-10-15
Add cache in HaGatewayManager

oneonestar:
## Description
Add cache in HaGatewayManager.
Continue the work in #501 

Set `databaseCacheTTL` to a non-zero [Airlift duration](https://airlift.github.io/airlift/units/) value to enable in-memory caching of backend metadata retrieved from the gateway database. Trino Gateway caches the list of backend clusters for the specified time and refreshes it asynchronously. Use this setting to reduce database load and improve routing performance.

A value of `0s` (the default) disables the cache and queries the database on
every request.

## Release notes

```markdown
* Allow setting `databaseCacheTTL` to enable in-memory caching of backend metadata retrieved from the gateway database.
```

## Summary by Sourcery

Introduce an optional in-memory cache for gateway backends in HaGatewayManager with configurable TTL and automatic invalidation on data changes.

New Features:
- Add optional Guava LoadingCache for all gateway backends with a databaseCacheTTL setting in RoutingConfiguration
- Expose metrics for cache lookup successes and failures

Enhancements:
- Invalidate and asynchronously refresh the backend cache on startup and whenever backends are added, updated, activated, or deleted
- Replace multiple DAO methods with a single findAll and perform in-memory filtering for active, group-specific, and named backends

Tests:
- Update TestHaGatewayManager to validate behavior with cache enabled and disabled

## Summary by Sourcery

Introduce configurable in-memory caching in HaGatewayManager to reduce database load and improve routing performance by caching backend metadata with TTL, automatic invalidation, and streamlined retrieval, while updating tests and documentation.

New Features:
- Add optional Guava-based in-memory cache for gateway backends with configurable TTL and asynchronous refresh
- Expose metrics for backend cache lookup successes and failures

Enhancements:
- Invalidate and refresh the backend cache automatically on backend additions, updates, activations, or deletions
- Simplify backend retrieval by fetching all backends once and performing in-memory filtering instead of multiple DAO queries

Documentation:
- Update routing rules documentation to include the databaseCacheTTL configuration and caching behavior

Tests:
- Extend TestHaGatewayManager to validate behavior with cache enabled and disabled

### DIFF
--- a/docs/operation.md
+++ b/docs/operation.md
@@ -81,3 +81,33 @@ taking a long time for garbage collection.
 completed initialization and is ready to serve requests. This means the initial
 connection to the database and the first round of health check on Trino clusters
 are completed. Otherwise, status code 503 is returned.
+
+## Database cache configuration
+
+Trino Gateway can cache database queries to improve performance and reduce load
+on the backend database. This also allow gateway to continue routing queries
+when the database is temporarily unavailable. Currently only the list of backend
+Trino clusters used for query routing are being cached.
+The cache can be configured using the `databaseCache` section in the config file.
+
+```yaml
+databaseCache:
+  enabled: true
+  expireAfterWrite: 60m
+  refreshAfterWrite: 5s
+```
+
+Configuration options:
+
+* `enabled` - Enable or disable the database cache. Default is `false`.
+* `expireAfterWrite` - The maximum time a cached entry is kept since it was last
+  loaded or refreshed. This ensures stale data is eventually removed.
+  If cache is not refreshed before expiration, requests will fail once the entry
+  expires (i.e. cache miss will attempt to reload data, but if the database is unavailable,
+  the request fails because there is no stale value to fall back to after
+  expiration).
+* `refreshAfterWrite` - Duration after which cache entries are eligible for
+  asynchronous refresh. When a refresh is triggered, the existing cached value
+  continues to be served while the refresh happens in the background.
+  This helps keep data fresh while serving slightly stale data to avoid blocking requests.
+
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/DatabaseCacheConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/DatabaseCacheConfiguration.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.config;
+
+import io.airlift.units.Duration;
+
+import java.util.concurrent.TimeUnit;
+
+public class DatabaseCacheConfiguration
+{
+    private boolean enabled;
+    private Duration expireAfterWrite = Duration.succinctDuration(60, TimeUnit.MINUTES);
+    private Duration refreshAfterWrite = Duration.succinctDuration(5, TimeUnit.SECONDS);
+
+    public boolean isEnabled()
+    {
+        return enabled;
+    }
+
+    public void setEnabled(boolean enabled)
+    {
+        this.enabled = enabled;
+    }
+
+    public Duration getExpireAfterWrite()
+    {
+        return expireAfterWrite;
+    }
+
+    public void setExpireAfterWrite(Duration expireAfterWrite)
+    {
+        this.expireAfterWrite = expireAfterWrite;
+    }
+
+    public Duration getRefreshAfterWrite()
+    {
+        return refreshAfterWrite;
+    }
+
+    public void setRefreshAfterWrite(Duration refreshAfterWrite)
+    {
+        this.refreshAfterWrite = refreshAfterWrite;
+    }
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/GatewayBackendDao.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/GatewayBackendDao.java
@@ -25,24 +25,6 @@ public interface GatewayBackendDao
 
     @SqlQuery("""
             SELECT * FROM gateway_backend
-            WHERE active = true
-            """)
-    List<GatewayBackend> findActiveBackend();
-
-    @SqlQuery("""
-            SELECT * FROM gateway_backend
-            WHERE active = true AND routing_group = :routingGroup
-            """)
-    List<GatewayBackend> findActiveBackendByRoutingGroup(String routingGroup);
-
-    @SqlQuery("""
-            SELECT * FROM gateway_backend
-            WHERE name = :name
-            """)
-    List<GatewayBackend> findByName(String name);
-
-    @SqlQuery("""
-            SELECT * FROM gateway_backend
             WHERE name = :name
             LIMIT 1
             """)
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
@@ -13,9 +13,15 @@
  */
 package io.trino.gateway.ha.router;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.github.benmanes.caffeine.cache.Ticker;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
+import io.airlift.stats.CounterStat;
+import io.trino.gateway.ha.config.DatabaseCacheConfiguration;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.config.RoutingConfiguration;
 import io.trino.gateway.ha.persistence.dao.GatewayBackend;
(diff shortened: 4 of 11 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -24,35 +29,92 @@
...
         this.defaultRoutingGroup = routingConfiguration.getDefaultRoutingGroup();
+        if (!routingConfiguration.getDatabaseCacheTTL().isZero()) {
+            cacheEnabled = true;
+            backendCache = CacheBuilder
+                    .newBuilder()
+                    .initialCapacity(1)
+                    .refreshAfterWrite(routingConfiguration.getDatabaseCacheTTL().toJavaTime())
+                    .build(CacheLoader.asyncReloading(
+                            CacheLoader.from(this::fetchAllBackends),
+                            MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor())));
```
Chaho12: How about using a ThreadFactory to prevent thread leaks? 
By default, newSingleThreadExecutor seems to create non-daemon threads.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -24,35 +29,92 @@
...
+    {
+        try {
+            List<GatewayBackend> backends = dao.findAll();
+            backendLookupSuccesses.update(1);
+            return backends;
+        }
+        catch (Exception e) {
+            backendLookupFailures.update(1);
+            log.warn(e, "Failed to fetch backends");
+            throw e;
```
Chaho12, line 101: Should we throw exception? or just return empty list to maintain service availability
I mean the cache will retry on the next refresh cycle anyway

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
Chaho12:
```
   ThreadFactory daemonThreadFactory = runnable -> {
       Thread thread = new Thread(runnable, "backend-cache-refresh");
       thread.setDaemon(true);
       return thread;
   };
   ...

   MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor(daemonThreadFactory))));
```

docs/routing-rules.md
```diff
@@ -10,8 +10,19 @@ defaults to the `adhoc` group.
 ```yaml
 routing:
     defaultRoutingGroup: "test-group"
+    # Optional: cache backend metadata to reduce database look-ups
+    databaseCacheTTL: "5m"
```
Peiyingy: Let's create a `CacheConfiguration` instead of putting all the configs under routing configs. We can define `cacheEnabled`, `expireAfterWrite`, and `maximumSize` explicitly instead of using `if (!routingConfiguration.getDatabaseCacheTTL().isZero())` to decide if cache is enabled, and set default values for them. Also, we can make cache behavior configurable there.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -13,8 +13,13 @@
  */
 package io.trino.gateway.ha.router;
 
+import com.google.common.cache.CacheBuilder;
```
Peiyingy: Let's also use this opportunity to upgrade the caching dependency from guava to caffeine.

Peiyingy (commented):
Thanks for putting up this PR!
At LinkedIn, we’ve already implemented caching in `GatewayBackendDao` and have it running in production. Since we run multiple Gateway instances, we found that a cache-aside approach can lead to noticeable inconsistencies across instances. To address this, we switched to a DB-first model, where the cache only serves as a fallback during DB outages.

We’d recommend making the cache behavior configurable so that users can choose between cache-aside and DB-first depending on their deployment setup.

Separately, we also added a write buffer mechanism in `QueryHistoryManager`. With the DB-first cache and write buffer in place, the Gateway can continue processing user queries even during DB outages, eliminating this single point of failure. We plan to open up that write buffer PR to OSS soon as well.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/GatewayBackendDao.java
```diff
@@ -23,24 +23,6 @@ public interface GatewayBackendDao
...
-            """)
-    List<GatewayBackend> findActiveBackendByRoutingGroup(String routingGroup);
-
-    @SqlQuery("""
-            SELECT * FROM gateway_backend
-            WHERE name = :name
-            """)
-    List<GatewayBackend> findByName(String name);
-
     @SqlQuery("""
```
Peiyingy: Let's also add cache on `findByName` to support resiliency during DB outage.

Peiyingy (commented):
Let's also add cache for `findFirstByName` to guard resiliency during DB outage

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
oneonestar: What's the benefit of switching from guava to caffeine?

Peiyingy: Caffeine has largely become the modern replacement for Guava’s cache. It continues to get active development and performance improvements, while Guava’s cache has been pretty much in maintenance mode for a while. A lot of Java projects have already switched for that reason. Since this PR is already updating the caching layer, it seems like a good moment to move to something faster and better maintained going forward.

oneonestar: I created another PR for the migration: #808

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
oneonestar: Removed w/ migration to Caffeine.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
oneonestar: Throwing an exception causes the cache to use the last known good value stored in the cache.

docs/routing-rules.md
oneonestar: Created `DatabaseCacheConfiguration.java`.

oneonestar:
> Let's also add cache for findFirstByName to guard resiliency during DB outage

`findFirstByName` only used when doing an update. We don't want to do an update when there is DB outage.

oneonestar:
@Chaho12  @Peiyingy 
Rebased to Caffeine and addressed all the comments. PTAL.

docs/operation.md
```diff
@@ -81,3 +81,27 @@ taking a long time for garbage collection.
 completed initialization and is ready to serve requests. This means the initial
 connection to the database and the first round of health check on Trino clusters
 are completed. Otherwise, status code 503 is returned.
+
+## Database cache configuration
+
+Trino Gateway can cache database queries to improve performance and reduce load
```
Chaho12, line 87: I think it is better to clarify what is cached as you wrote down in the PR.
e.g) 
Trino Gateway caches metadata retrieved from the gateway database, such as the list of backend clusters used for query routing.

docs/operation.md
```diff
@@ -81,3 +81,27 @@ taking a long time for garbage collection.
...
+```
+
+Configuration options:
+
+* `enabled` - Enable or disable the database cache. Default is `false`.
+* `expireAfterWrite` - Duration after which cache entries expire following their
+  last update. This ensures stale data is eventually removed.
+* `refreshAfterWrite` - Duration after which cache entries are eligible for
+  asynchronous refresh. This helps keep data fresh while serving slightly stale
+  data to avoid blocking requests.
```
Chaho12: The README is generally clear, but the cache behavior is still a bit ambiguous.
It would help to explicitly describe when async refresh happens, how requests
are handled during refresh, and what data is actually cached.

Chaho12: Maybe something like this, if i understood it correctly.

expireAfterWrite: The maximum time a cached entry is kept since it was last loaded or refreshed.
After this duration, the entry is removed and the next request will reload it
from the database.


refreshAfterWrite: The duration after which a cached entry is refreshed asynchronously.
When a refresh is triggered, the existing cached value continues to be served
while the refresh happens in the background.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DatabaseCacheConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+package io.trino.gateway.ha.config;
+
+import io.airlift.units.Duration;
+
+import java.util.concurrent.TimeUnit;
+
+public class DatabaseCacheConfiguration
+{
+    private boolean enabled;
+    private Duration expireAfterWrite = Duration.succinctDuration(10, TimeUnit.MINUTES);
```
Chaho12: How about matching this value with README.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -28,34 +34,111 @@
...
+    private void invalidateBackendCache()
+    {
+        if (cacheEnabled) {
+            // Avoid using bulk invalidation like invalidateAll(), in order to invalidate in-flight loads properly.
+            // See https://github.com/trinodb/trino/issues/10512#issuecomment-1016398117
+            backendCache.invalidate(ALL_BACKEND_CACHE_KEY);
+        }
+    }
+
+    private List<GatewayBackend> getOrFetchAllBackends()
```
Chaho12: i think we can keep this as `getAllBackends()`

docs/operation.md
oneonestar: Updated.

docs/operation.md
oneonestar: Updated.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
oneonestar: `getAllBackends()` already exist. I renamed it to `getAllBackendsInternal()`.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DatabaseCacheConfiguration.java
oneonestar: Updated.

docs/operation.md
```diff
@@ -81,3 +81,31 @@ taking a long time for garbage collection.
...
+  expireAfterWrite: 60m
+  refreshAfterWrite: 5s
+```
+
+Configuration options:
+
+* `enabled` - Enable or disable the database cache. Default is `false`.
+* `expireAfterWrite` - The maximum time a cached entry is kept since it was last
+  loaded or refreshed. This ensures stale data is eventually removed.
+  If cache is not refreshed before expiration, requests will fail once the entry
```
Chaho12, line 105: I know this is intended and technically correct, but adding a bit more context would make it clearer and more complete.
e.g)
A cache miss will attempt to reload data, but if the database is unavailable,
the request fails because there is no stale value to fall back to after
expiration.

Chaho12 (approved):
LGTM 👍

docs/operation.md
oneonestar: Updated

oneonestar:
Merging to avoid conflicts with other cache improvement PRs.
Any additional comments will be addressed in another PR.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DatabaseCacheConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.airlift.units.Duration;
+
+import java.util.concurrent.TimeUnit;
+
+public class DatabaseCacheConfiguration
+{
+    private boolean enabled;
+    private Duration expireAfterWrite = Duration.succinctDuration(60, TimeUnit.MINUTES);
+    private Duration refreshAfterWrite = Duration.succinctDuration(5, TimeUnit.SECONDS);
```
ebyhr, line 24:
```suggestion
    private Duration expireAfterWrite = Duration.succinctDuration(1, HOURS);
    private Duration refreshAfterWrite = Duration.succinctDuration(5, SECONDS);
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -28,34 +34,111 @@
...
     {
         dao = requireNonNull(jdbi, "jdbi is null").onDemand(GatewayBackendDao.class);
-        this.defaultRoutingGroup = routingConfiguration.getDefaultRoutingGroup();
+        defaultRoutingGroup = routingConfiguration.getDefaultRoutingGroup();
+        cacheEnabled = databaseCacheConfiguration.isEnabled();
+        if (cacheEnabled) {
+            Caffeine<Object, Object> caffeineBuilder = Caffeine.newBuilder()
+                    .initialCapacity(1)
+                    .ticker(ticker);
+            if (databaseCacheConfiguration.getExpireAfterWrite() != null) {
```
ebyhr, line 70: When does this getter return null when cacheEnabled is true?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -28,34 +34,111 @@
...
+    {
+        if (cacheEnabled) {
+            try {
+                return backendCache.get(ALL_BACKEND_CACHE_KEY);
+            }
+            catch (Exception e) {
+                throw new RuntimeException("Failed to load backends from database to cache", e);
+            }
+        }
+        else {
```
ebyhr, line 124: Remove redundant else. 

https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#additional-ide-configuration
> Java | Control flow issues | Redundant 'else' (including Report when there are no more statements after the 'if' statement option),

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -13,33 +13,47 @@
...
 
     @Test
-    void testGatewayManager()
+    void testGatewayManagerWithoutCache()
+    {
+        JdbcConnectionManager connectionManager = createTestingJdbcConnectionManager(dataStoreConfig());
+        testGatewayManager(new HaGatewayManager(connectionManager.getJdbi(), new RoutingConfiguration(), new DatabaseCacheConfiguration()));
+    }
+
+    void testGatewayManager(HaGatewayManager haGatewayManager)
```
ebyhr, line 56: This method can be static.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -13,33 +13,47 @@
...
-    void setUp()
+    @Test
+    void testGatewayManagerWithCache()
     {
         JdbcConnectionManager connectionManager = createTestingJdbcConnectionManager(dataStoreConfig());
-        RoutingConfiguration routingConfiguration = new RoutingConfiguration();
-        haGatewayManager = new HaGatewayManager(connectionManager.getJdbi(), routingConfiguration);
+        DatabaseCacheConfiguration cacheConfiguration = new DatabaseCacheConfiguration();
+        cacheConfiguration.setEnabled(true);
+        cacheConfiguration.setRefreshAfterWrite(new Duration(5, TimeUnit.SECONDS));
```
ebyhr, line 45: Static import `SECONDS`.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -99,9 +113,48 @@ void testGatewayManager()
...
+        ProxyBackendConfiguration etl = new ProxyBackendConfiguration();
+        etl.setActive(false);
+        etl.setRoutingGroup("etl");
+        etl.setName("new-etl1");
+        etl.setProxyTo("https://etl1.trino.gateway.io:443/");
+        etl.setExternalUrl("https://etl1.trino.gateway.io:443/");
+        haGatewayManager.addBackend(etl);
+
+        // Initial fetch
+        assertThat(haGatewayManager.getBackendByName("new-etl1").map(ProxyBackendConfiguration::getProxyTo).orElseThrow()).isEqualTo("https://etl1.trino.gateway.io:443");
```
ebyhr, line 137: Use `AbstractOptionalAssert#contains` or `AbstractOptionalAssert#hasValue` instead of calling `orElseThrow`. Same for others.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -124,4 +177,22 @@ void testRemoveTrailingSlashInUrl()
         assertThat(haGatewayManager.getBackendByName("new-etl1").map(ProxyBackendConfiguration::getProxyTo).orElseThrow()).isEqualTo("https://etl2.trino.gateway.io:443");
         assertThat(haGatewayManager.getBackendByName("new-etl1").map(ProxyBackendConfiguration::getExternalUrl).orElseThrow()).isEqualTo("https://etl2.trino.gateway.io:443");
     }
+
+    public static class TestingTicker
```
ebyhr, line 181: This test class and `increment` method can be private.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestHaGatewayManager.java
```diff
@@ -124,4 +177,22 @@ void testRemoveTrailingSlashInUrl()
...
+        @Override
+        public synchronized long read()
+        {
+            return this.time;
+        }
+
+        public synchronized void increment(long delta, TimeUnit unit)
+        {
+            checkArgument(delta >= 0L, "delta is negative");
+            this.time += unit.toNanos(delta);
```
ebyhr, line 195: Redundant `this.`

gateway-ha/src/test/java/io/trino/gateway/ha/TestingJdbcConnectionManager.java
```diff
@@ -39,4 +41,25 @@ public static JdbcConnectionManager createTestingJdbcConnectionManager(DataStore
...
         return new JdbcConnectionManager(jdbi, config);
     }
+
+    public static void destroyTestingDatabase(DataStoreConfiguration config)
+    {
+        String tempH2DbDirPath = config.getJdbcUrl().replace("jdbc:h2:", "").replace(";NON_KEYWORDS=NAME,VALUE", "");
+        File dbFile = Path.of(tempH2DbDirPath).toFile();
+        File parentDir = dbFile.getParentFile();
+
+        if (parentDir != null && parentDir.exists()) {
```
ebyhr, line 51: `destroyTestingDatabase` is called from single place. Why do we need this condition? Could you leave a code comment?

PR #788, opened 2025-10-30
Fix UI Query Distribution chart showing zero queries inaccurately

Peiyingy:
## Description

The Trino Gateway UI currently has a bug in the Query Distribution chart: query counts display correctly for even minutes, but always show zero for odd minutes due to the incorrect query count calculation.

Screenshot 2025-10-23 at 22 35 23
Screenshot 2025-10-23 at 22 35 46

The issue is caused by using `Float.parseFloat` when converting the minute value. Since float has limited precision, the last bit of the value was lost, results wrong minute buckets. This PR replaces `Float.parseFloat` with `BigDecimal longvalue` to ensure full precision and correct query counts for all minutes.

## Testing Done
- [x] `mvn clean install`
- [x] Tested the new query distribution chart

Screenshot 2025-10-23 at 22 39 49
Screenshot 2025-10-23 at 22 40 00

## Release notes

```markdown
* Fix UI Query Distribution chart showing zero queries inaccurately.
```

## Summary by Sourcery

Bug Fixes:
- Replace Float.parseFloat with BigDecimal.longValue to avoid precision loss when computing minute buckets in the chart

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
@@ -24,6 +24,7 @@ import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.testcontainers.containers.JdbcDatabaseContainer;
 
+import java.math.BigDecimal;
 import java.util.List;
 import java.util.Optional;
 
@@ -117,16 +118,17 @@ abstract class BaseTestQueryHistoryManager
     @Test
     void testTimestampParsing()
     {
-        long result = 30338640;
+        // This ensures odd-minute values remain precision when converted from different formats.
+        long expectedMinuteBucket = 30338641;
 
-        // postgres: minute -> {Double@9333} 3.033864E7
-        String postgresTimestamp = "3.033864E7";
-        long parsedLongTimestamp = (long) Float.parseFloat(postgresTimestamp);
-        assertThat(parsedLongTimestamp).isEqualTo(result);
+        // postgres: minute -> {Double@9333} 3.0338641E7
+        String postgresTimestamp = "3.0338641E7";
+        long parsedPostgresMinute = new BigDecimal(postgresTimestamp).longValue();
+        assertThat(parsedPostgresMinute).isEqualTo(expectedMinuteBucket);
 
-        // mysql: minute -> {BigDecimal@9775} "30338640"
-        String mysqlTimestamp = "30338640";
-        long parsedLongTimestamp2 = (long) Float.parseFloat(mysqlTimestamp);
-        assertThat(parsedLongTimestamp2).isEqualTo(result);
+        // mysql: minute -> {BigDecimal@9775} "30338641"
+        String mysqlTimestamp = "30338641";
+        long parsedMysqlMinute = new BigDecimal(mysqlTimestamp).longValue();
+        assertThat(parsedMysqlMinute).isEqualTo(expectedMinuteBucket);
     }
 }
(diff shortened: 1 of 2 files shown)


maswin:
Verified that the fix works. Can we add a test case?

Chaho12:
@maswin Thx for verifying!
@Peiyingy I changed the comments in Release notes section if you don't mind.

Chaho12:
You don't need unnecessary description there. You can remove `.` at the end @Peiyingy 
image

ebyhr:
@Chaho12 Please remove the PR number from commit title next time.
Screenshot 2025-12-18 at 8 54 09

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
```diff
@@ -117,16 +118,17 @@ void testFindDistribution()
...
-        assertThat(parsedLongTimestamp2).isEqualTo(result);
+        // This ensures odd-minute values remain precision when converted from different formats.
+        long expectedMinuteBucket = 30338641;
+
+        // postgres: minute -> {Double@9333} 3.0338641E7
+        String postgresTimestamp = "3.0338641E7";
+        long parsedPostgresMinute = new BigDecimal(postgresTimestamp).longValue();
+        assertThat(parsedPostgresMinute).isEqualTo(expectedMinuteBucket);
+
+        // mysql: minute -> {BigDecimal@9775} "30338641"
```
ebyhr, line 129: What is the meaning of `9333` and `9775` after `@`?

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
```diff
@@ -117,16 +118,17 @@ void testFindDistribution()
...
+        String mysqlTimestamp = "30338641";
+        long parsedMysqlMinute = new BigDecimal(mysqlTimestamp).longValue();
+        assertThat(parsedMysqlMinute).isEqualTo(expectedMinuteBucket);
```
ebyhr, line 132: The modified test passes without this PR's change. We should rewrite the code to test product flow.

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
Peiyingy: Those parts are not changed by this PR. This PR only changed the minute number to an odd number.

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
Peiyingy: The original tests did not invoke the timestamp parsing logic in `HaQueryHistoryManager` either. For the purpose of this bug fix, I updated the tests only to align with the existing implementation. We can certainly follow up with a separate PR to refactor and improve those tests.

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
ebyhr: I knew. I recommend removing the part if the number doesn't give us any context.

gateway-ha/src/test/java/io/trino/gateway/ha/router/BaseTestQueryHistoryManager.java
ebyhr: I knew. However, the PR fixing a bug should contain the test covering the change.

Issue #377, opened 2024-06-03
Add connection pooling to Trino Gateway backend datastore


rdsarvar:
Out-of-the-box JDBI does not seem to leverage connection pools and will create a new connection for every query against the backend datastore as seen in the public Slack channel [here](https://trinodb.slack.com/archives/CG9K9MX1V/p1714195241951439). During high traffic this can lead to reaching maximum connection counts against the backend datastore.

It has been noted that there are multiple ways of getting around the limitation of maximum number of connections:

1. Optimize the underlying datastore configurations so that connections drop off quicker
2. Use pgbouncer as a service in between Trino Gateway and PostgreSQL (assuming postgres is used as the backend store instead of MySQL or similar)
3. Implement connection pooling as part of the Trino Gateway code itself. Example PostgresSQL connection pooling provided by JDBI docs: https://jdbi.org/#_high_availability

As an enhancement we are looking to minimize overall connections required by Trino Gateway against its backend datastore.

mosabua:
Trino already has connection pooling in a number of places. Not sure if they involved JDBI though ... we definitely have some for JDBC connections though. We can probably take some inspiration from there. Or maybe there even is something in airlift (although I dont remember seeing something along that line).

arturmkr:
@mosabua 
It seems that for each query the gateway opens a new JDBC connection to store query metadata.  
There is no connection pooling layer (e.g. HikariCP or shared DataSource), so connections are created per request.

During load testing, I observed around 600 concurrent connections from `trino-gateway` to PostgreSQL.  
Typical statements executed are:

```
INSERT INTO query_history (...)
SELECT * FROM gateway_backend WHERE active = ? AND routing_group = ?
SELECT external_url FROM query_history WHERE query_id = ?
```

This looks like a missing connection pooling issue in `JdbcConnectionManager`,  since `Jdbi.create()` is called per routing group / request without a shared pool.

mosabua:
Thanks for investigating this more @arturmkr .. we definitely should add connection pooling. Will either @rdsarvar or @arturmkr work on a PR that implements that?

arturmkr:
@mosabua , I prepared the [PR](https://github.com/trinodb/trino-gateway/pull/791) for this.

rdsarvar:
Thanks @arturmkr, I was going to do this a while back but after having a baby I've been tight on spare time 😄

PR #791, opened 2025-11-04
Add db connection pool support

arturmkr:
### Summary
This PR introduces optional JDBC connection pooling using HikariCP in `JdbcConnectionManager`, solves issue #377 

### Details
- Added an optional field `maxPoolSize` to `DataStoreConfiguration`.
- When `maxPoolSize` is configured (non-null and > 0):
  - `JdbcConnectionManager` uses a Hikari connection pool for routing-group database connections.
- When `maxPoolSize` is not set (null):
  - Connections are created directly (no pooling).

### Tests
Added `TestJdbcConnectionManagerPool`:
1. **blocksWhenExceedingMaxPoolSize**
   - Verifies that when a pool is enabled, new connections block after the pool limit.
2. **doesNotBlockWhenMaxPoolSizeIsNull**
   - Verifies that when the pool is disabled, connections are created directly and do not block.

### Why
This allows controlling connection usage and improving DB efficiency under load, while keeping the old behavior as the default (no pooling).

## Summary by Sourcery

Introduce optional JDBC connection pooling using HikariCP controlled by a new maxPoolSize setting, while preserving default direct connections, and add tests to verify pooling behavior

New Features:
- Enable JDBC connection pooling via HikariCP when maxPoolSize > 0

Enhancements:
- Add maxPoolSize field with constructors, getter, and setter to DataStoreConfiguration
- Update JdbcConnectionManager#getJdbi to use a HikariCP DataSource when pooling is enabled

Build:
- Add HikariCP dependency to the project

Tests:
- Introduce TestJdbcConnectionManagerPool with tests for pool blocking and no-pool scenarios

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -55,6 +55,14 @@ You can also disable query history recording to the database by setting
 `queryHistoryEnabled` to `false`. This can be useful in scenarios where you
 want to reduce database load or don't need query history tracking.
 
+If `maxPoolSize` is configured and greater than 0, Trino Gateway uses a
+connection pool for data store connections, including gateway metadata and
+query history.
+If `maxPoolSize` is not configured, Trino Gateway creates a new JDBC connection
+per request.
+A value of `10` is a reasonable starting point for many deployments, but the
+optimal value depends on the database capacity and expected concurrency.
+
 For example:
 
 ```yaml
@@ -66,6 +74,7 @@ dataStore:
   queryHistoryHoursRetention: 24
   runMigrationsEnabled: false
   queryHistoryEnabled: true  # Set to false to disable query history recording
+  maxPoolSize: 10            # Optional: enables JDBC connection pooling for the data store
 ```
 
 `Flyway` uses a transactional lock in databases that support it such as 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -55,6 +55,20 @@ You can also disable query history recording to the database by setting
 `queryHistoryEnabled` to `false`. This can be useful in scenarios where you
 want to reduce database load or don't need query history tracking.
 
+If `maxPoolSize` is configured and greater than 0, Trino Gateway uses a
+connection pool for data store connections, including gateway metadata and
+query history.
+If `maxPoolSize` is not configured, Trino Gateway creates a new JDBC connection
+per request.
+A value of `10` is a reasonable starting point for many deployments, but the
+optimal value depends on the database capacity and expected concurrency.
+The optional `keepaliveTime` and `maxLifetime` settings control how frequently
+idle pooled connections are checked and how long a pooled connection can be
+reused. Both accept Airlift-style durations such as `2m` or `30m`. If they are
+not configured, the connection pool defaults are used. Non-zero values must be
+at least `30s`, and `keepaliveTime` must be less than `maxLifetime` when both
+are enabled.
+
 For example:
 
 ```yaml
@@ -66,6 +80,9 @@ dataStore:
   queryHistoryHoursRetention: 24
   runMigrationsEnabled: false
   queryHistoryEnabled: true  # Set to false to disable query history recording
+  maxPoolSize: 10            # Optional: enables JDBC connection pooling for the data store
+  keepaliveTime: 2m          # Optional: interval for checking idle pooled connections
+  maxLifetime: 30m           # Optional: maximum lifetime of a pooled connection
 ```
 
 `Flyway` uses a transactional lock in databases that support it such as 
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -109,6 +109,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>7.0.2</version>
+        </dependency>
+
+        <dependency>
             <groupId>io.airlift</groupId>
             <artifactId>aircompressor-v3</artifactId>
             <version>3.7</version>
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
@@ -22,8 +22,9 @@ public class DataStoreConfiguration
     private boolean queryHistoryEnabled = true;
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
+    private Integer maxPoolSize;
 
-    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, boolean queryHistoryEnabled, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled)
+    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, boolean queryHistoryEnabled, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled, Integer maxPoolSize)
     {
         this.jdbcUrl = jdbcUrl;
         this.user = user;
@@ -32,6 +33,7 @@ public class DataStoreConfiguration
         this.queryHistoryEnabled = queryHistoryEnabled;
         this.queryHistoryHoursRetention = queryHistoryHoursRetention;
         this.runMigrationsEnabled = runMigrationsEnabled;
+        this.maxPoolSize = maxPoolSize;
     }
 
     public DataStoreConfiguration() {}
@@ -104,5 +106,15 @@ public class DataStoreConfiguration
     public void setRunMigrationsEnabled(boolean runMigrationsEnabled)
     {
         this.runMigrationsEnabled = runMigrationsEnabled;
+    }
+
+    public Integer getMaxPoolSize()
+    {
+        return maxPoolSize;
+    }
+
+    public void setMaxPoolSize(Integer maxPoolSize)
+    {
+        this.maxPoolSize = maxPoolSize;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
@@ -13,6 +13,8 @@
  */
 package io.trino.gateway.ha.config;
 
+import io.airlift.units.Duration;
+
 public class DataStoreConfiguration
 {
     private String jdbcUrl;
@@ -22,8 +24,11 @@ public class DataStoreConfiguration
     private boolean queryHistoryEnabled = true;
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
+    private Integer maxPoolSize;
+    private Duration keepaliveTime;
+    private Duration maxLifetime;
 
-    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, boolean queryHistoryEnabled, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled)
+    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, boolean queryHistoryEnabled, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled, Integer maxPoolSize)
     {
         this.jdbcUrl = jdbcUrl;
         this.user = user;
@@ -32,6 +37,7 @@ public class DataStoreConfiguration
         this.queryHistoryEnabled = queryHistoryEnabled;
         this.queryHistoryHoursRetention = queryHistoryHoursRetention;
         this.runMigrationsEnabled = runMigrationsEnabled;
+        this.maxPoolSize = maxPoolSize;
     }
 
     public DataStoreConfiguration() {}
@@ -104,5 +110,35 @@ public class DataStoreConfiguration
     public void setRunMigrationsEnabled(boolean runMigrationsEnabled)
     {
         this.runMigrationsEnabled = runMigrationsEnabled;
+    }
+
+    public Integer getMaxPoolSize()
+    {
+        return maxPoolSize;
+    }
+
+    public void setMaxPoolSize(Integer maxPoolSize)
+    {
+        this.maxPoolSize = maxPoolSize;
+    }
+
+    public Duration getKeepaliveTime()
+    {
+        return keepaliveTime;
+    }
+
+    public void setKeepaliveTime(Duration keepaliveTime)
+    {
+        this.keepaliveTime = keepaliveTime;
+    }
+
+    public Duration getMaxLifetime()
+    {
+        return maxLifetime;
+    }
+
+    public void setMaxLifetime(Duration maxLifetime)
+    {
+        this.maxLifetime = maxLifetime;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
@@ -15,20 +15,27 @@ package io.trino.gateway.ha.persistence;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Inject;
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.DataStoreConfiguration;
 import io.trino.gateway.ha.persistence.dao.QueryHistoryDao;
 import jakarta.annotation.Nullable;
+import jakarta.annotation.PreDestroy;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
 public class JdbcConnectionManager
@@ -39,27 +46,39 @@ public class JdbcConnectionManager
     private final DataStoreConfiguration configuration;
     private final ScheduledExecutorService executorService =
             Executors.newSingleThreadScheduledExecutor();
+    private final ScheduledFuture<?> cleanupTask;
+
+    private final Map<String, HikariDataSource> pools = new ConcurrentHashMap<>();
 
     @Inject
     public JdbcConnectionManager(Jdbi jdbi, DataStoreConfiguration configuration)
     {
         this.jdbi = requireNonNull(jdbi, "jdbi is null");
-        this.configuration = configuration;
-        startCleanUps();
+        this.configuration = requireNonNull(configuration, "configuration is null");
+        cleanupTask = startCleanUps();
     }
 
     public Jdbi getJdbi()
     {
-        return jdbi;
+        return getJdbi(null);
     }
 
     public Jdbi getJdbi(@Nullable String routingGroupDatabase)
     {
-        if (routingGroupDatabase == null) {
+        Jdbi result;
+        Integer maxPoolSize = configuration.getMaxPoolSize();
+        if (maxPoolSize != null) {
+            HikariDataSource dataSource = getOrCreateDataSource(routingGroupDatabase, maxPoolSize);
+            result = Jdbi.create(dataSource);
+        }
+        else if (routingGroupDatabase == null) {
             return jdbi;
         }
+        else {
+            result = Jdbi.create(buildJdbcUrl(routingGroupDatabase), configuration.getUser(), configuration.getPassword());
+        }
 
-        return Jdbi.create(buildJdbcUrl(routingGroupDatabase), configuration.getUser(), configuration.getPassword())
+        return result
                 .installPlugin(new SqlObjectPlugin())
                 .registerRowMapper(new RecordAndAnnotatedConstructorMapper());
     }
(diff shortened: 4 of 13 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -107,11 +124,51 @@ private void startCleanUps()
         executorService.scheduleWithFixedDelay(
                 () -> {
                     log.info("Performing query history cleanup task");
-                    long created = System.currentTimeMillis() - TimeUnit.HOURS.toMillis(this.configuration.getQueryHistoryHoursRetention());
+                    long created = System.currentTimeMillis()
+                            - TimeUnit.HOURS.toMillis(this.configuration.getQueryHistoryHoursRetention());
```
ebyhr: Revert unrelated change.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -21,15 +21,32 @@ public class DataStoreConfiguration
     private String driver;
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
+    private Integer maxPoolSize;
```
rdsarvar, line 27: nit: IMO if we're adding HikariCP we should be giving the users complete flexibility to all of it's settings and not just max pool size:

Example of what I mean from previously unpushed code I had laying around:
```
+    // HikariCP connection pool settings
+    private int maximumPoolSize = 10;
+    private int minimumIdle = 5;
+    private long connectionTimeout = 30000; // 30 seconds
+    private long idleTimeout = 600000; // 10 minutes
+    private long maxLifetime = 1800000; // 30 minutes
+    private long leakDetectionThreshold = 60000; // 1 minute
+    private boolean autoCommit = true;
+    private String connectionTestQuery;
```

For the above to work with your logic max pool size can default to 0

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -107,11 +124,51 @@ private void startCleanUps()
...
+                cfg.setDriverClassName(configuration.getDriver());
+            }
+            cfg.setMaximumPoolSize(maxPoolSize);
+            cfg.setPoolName("gateway-ha-" + key);
+
+            return new HikariDataSource(cfg);
+        });
+    }
+
+    public void close()
```
rdsarvar, line 167: question: is this only used by tests? I don't see where it's being called otherwise

rdsarvar (commented):
small nitpick on configurability but open to @mosabua et al opinions

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
arturmkr: fixed

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
arturmkr: Thanks for the suggestion! I think exposing all HikariCP parameters would add unnecessary complexity for this use case as Hikari’s default settings are already well-optimized and sufficient.
For now I kept it minimal with maxPoolSize, since it’s the only setting that typically needs adjustment.

arturmkr:
@mosabua , do you have any comments about this PR?

gateway-ha/pom.xml
```diff
@@ -104,6 +104,12 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
```
mosabua, line 113: Is this widely used and a mature project? What is used for connection pooling in Trino? Ideally we should stick to the same approach as Trino itself

kbhatianr: HikariCP is a very widely used lib for connection pooling. The only reference to connection pooling I see in the Trino project is in the Oracle connector - which [seems to use an Oracle-specific lib for pooling](https://github.com/trinodb/trino/blob/master/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OraclePoolConnectionFactory.java#L23). 


Airlift as well does not yet seem to have[ Database Connection Pooling](https://github.com/airlift/airlift?tab=readme-ov-file#reference) implemented - it's marked as TBD. 

I would hypothesize that once Airlift adopts it, it would be a wrapper on top of Hikari lib.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -59,7 +66,18 @@ public Jdbi getJdbi(@Nullable String routingGroupDatabase)
             return jdbi;
         }
 
-        return Jdbi.create(buildJdbcUrl(routingGroupDatabase), configuration.getUser(), configuration.getPassword())
+        Integer maxPoolSize = configuration.getMaxPoolSize();
```
kbhatianr, line 69: Instead of having properties specific to `maxPoolSize`, and potentially others in the future, my suggestion would be to consider storing all the Hikari properties in a config, and then parsing that config. Whether you utilize all the properties or not [as per your other comment] is a separate discussion - I'd lean towards using all the properties - however, regardless it would clean up the code and make it more extensible rather than having to write code for every new property we want to consider in the future.

ebyhr: Disagreed, we should expose only important properties.

gateway-ha/pom.xml
ebyhr: Using HikariCP is totally fine.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
arturmkr: fixed

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
kbhatianr: My comment wasn't what we should expose / not expose (well, I did put my opinion that imo we should expose the entire Hikari config), but rather that we shouldn't have individual properties that we pass around - instead, creating a container object for this allows easier extensibility in the future when we may want additional properties.

Chaho12 (commented):
Can you update docs here? regarding pool size? I  haven't used HikariCP so am not too famliar with this library.
https://github.com/trinodb/trino-gateway/blob/3a1553a401a430fa7f4ce3d00791c6e3f0af4f81/docs/installation.md?plain=1#L54-L63

A reference to Hikari might be good too, suggesting 10 is suitable number for decent performance for db connection pool support.
https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/HikariConfig.java#L56

mosabua:
Will you be able to rebase this PR and address comments @arturmkr ?

arturmkr:
@mosabua, will do in a couple of days

arturmkr:
@mosabua 
I have merged the latest main into this branch to resolve conflicts and make sure it works with the current codebase.

Since this PR has been open for about 4 months, I would really appreciate it if someone could take a look at it.

Thanks!

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix:
         java-version: [25]
-    timeout-minutes: 20
+    timeout-minutes: 30
```
Chaho12, line 46: Why is this changed needed?

arturmkr: @Chaho12 , I increased the timeout because some CI runs legitimately take 21–22 minutes and fail due to the 20-minute limit. This is not specific to my PR, it happens in other runs as well:

https://github.com/trinodb/trino-gateway/actions/workflows/ci.yml?query=is%3Acancelled

Setting it to 30 minutes simply avoids unnecessary failures.

Chaho12: Oh ok. Let's increase it for now. I'll try to find a way to reduce CI time. 
It seems that even those with success is near the 20 min limit.
image
https://github.com/trinodb/trino-gateway/actions/workflows/ci.yml?query=is%3Asuccess

ebyhr: @arturmkr Could you send a new PR to increase the timeout so we can fix it soon?

arturmkr: @ebyhr , https://github.com/trinodb/trino-gateway/pull/953

docs/installation.md
```diff
@@ -54,6 +54,14 @@ You can also disable query history recording to the database by setting
 `queryHistoryEnabled` to `false`. This can be useful in scenarios where you
 want to reduce database load or don't need query history tracking.
 
+If `maxPoolSize` is configured and greater than 0, Trino Gateway uses a
+HikariCP connection pool for routing-group database connections created via
+`JdbcConnectionManager#getJdbi(routingGroupDatabase)`.
```
ebyhr: Let's avoid mentioning the method name. It's an implementation detail, and the name can be changed in the future.

docs/installation.md
```diff
@@ -54,6 +54,14 @@ You can also disable query history recording to the database by setting
 `queryHistoryEnabled` to `false`. This can be useful in scenarios where you
 want to reduce database load or don't need query history tracking.
 
+If `maxPoolSize` is configured and greater than 0, Trino Gateway uses a
+HikariCP connection pool for routing-group database connections created via
```
ebyhr: Let's avoid mentioning the library name here. It's an implementation detail, and the library can be changed in the future.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -32,6 +33,12 @@ public DataStoreConfiguration(String jdbcUrl, String user, String password, Stri
         this.queryHistoryEnabled = queryHistoryEnabled;
         this.queryHistoryHoursRetention = queryHistoryHoursRetention;
         this.runMigrationsEnabled = runMigrationsEnabled;
+        this.maxPoolSize = maxPoolSize;
+    }
+
+    public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, boolean queryHistoryEnabled, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled)
+    {
+        this(jdbcUrl, user, password, driver, queryHistoryEnabled, queryHistoryHoursRetention, runMigrationsEnabled, null);
```
ebyhr: This constructor is called only from tests. Could you remove this, and update the test code instead?

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -59,6 +66,14 @@ public Jdbi getJdbi(@Nullable String routingGroupDatabase)
             return jdbi;
         }
 
+        Integer maxPoolSize = configuration.getMaxPoolSize();
+        if (maxPoolSize != null && maxPoolSize > 0) {
+            HikariDataSource ds = getOrCreateDataSource(routingGroupDatabase, maxPoolSize);
```
ebyhr: > ds

Please avoid abbreviations in this project. Same for other places. 

https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestJdbcConnectionManagerPool
+{
+    @Test
+    void blocksWhenExceedingMaxPoolSize()
```
ebyhr: https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations 
* Test method must start with test, for example testExplain()

arturmkr: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
arturmkr: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
arturmkr: Fixed

docs/installation.md
arturmkr: Fixed

docs/installation.md
arturmkr: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -114,4 +129,44 @@ private void startCleanUps()
...
     }
+
+    private HikariDataSource getOrCreateDataSource(String routingGroupDatabase, int maxPoolSize)
+    {
+        return pools.compute(routingGroupDatabase, (key, existing) -> {
+            if (existing != null && !existing.isClosed()) {
+                return existing;
+            }
+
+            HikariConfig cfg = new HikariConfig();
```
ebyhr: Could you avoid abbreviations? https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -105,4 +107,14 @@ public void setRunMigrationsEnabled(boolean runMigrationsEnabled)
     {
         this.runMigrationsEnabled = runMigrationsEnabled;
     }
+
+    public Integer getMaxPoolSize()
+    {
+        return this.maxPoolSize;
```
ebyhr:
```suggestion
        return maxPoolSize;
```

arturmkr: Other getters in this class use `this` as well (e.g. getJdbcUrl(), getUser(), getDriver(), etc), so I kept the same style here for consistency.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
arturmkr: Fixed

arturmkr:
@mosabua , @ebyhr , are there any blockers to merging this PR?

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
ebyhr: There is no need to follow the redundant pattern. We'll fix other places in a follow-up.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -59,6 +66,14 @@ public Jdbi getJdbi(@Nullable String routingGroupDatabase)
             return jdbi;
         }
 
+        Integer maxPoolSize = configuration.getMaxPoolSize();
+        if (maxPoolSize != null && maxPoolSize > 0) {
```
ebyhr: We should throw an exception if maxPoolSize <= 0, instead of skipping silently.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -114,4 +129,44 @@ private void startCleanUps()
...
+            return new HikariDataSource(hikariConfig);
+        });
+    }
+
+    @PreDestroy
+    public void close()
+    {
+        for (Map.Entry<String, HikariDataSource> e : pools.entrySet()) {
+            HikariDataSource dataSource = e.getValue();
+            if (dataSource != null && !dataSource.isClosed()) {
```
ebyhr: When does dataSource become null?

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
```diff
@@ -0,0 +1,176 @@
...
+        try (ExecutorService es = Executors.newFixedThreadPool(3)) {
+            List<Future<Connection>> acquired = new ArrayList<>();
+
+            CountDownLatch hold = new CountDownLatch(1);
+            CountDownLatch acquiredLatch = new CountDownLatch(2);
+
+            // Open exactly maxPoolSize connections and keep them open
+            for (int i = 0; i < 2; i++) {
+                acquired.add(es.submit(() -> {
+                    try (var h = jdbi.open()) {
```
ebyhr: * https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-var
* https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
arturmkr: fixed

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
arturmkr: fixed

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
arturmkr: fixed

arturmkr:
@mosabua , @ebyhr are there any blockers to merging this PR?

mosabua (approved):
Looks good to me. Any concerns with merging @ebyhr and @trinodb/maintainers-gateway ?

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -59,6 +66,17 @@ public Jdbi getJdbi(@Nullable String routingGroupDatabase)
             return jdbi;
         }
 
+        Integer maxPoolSize = configuration.getMaxPoolSize();
+        if (maxPoolSize != null) {
+            if (maxPoolSize <= 0) {
+                throw new IllegalArgumentException("maxPoolSize must be greater than 0");
+            }
```
ebyhr: You can use `checkArgument` method. Also, this check can be moved to `getOrCreateDataSource`. The method might be called from a different place in the future.

docs/installation.md
```diff
@@ -54,6 +54,13 @@ You can also disable query history recording to the database by setting
 `queryHistoryEnabled` to `false`. This can be useful in scenarios where you
 want to reduce database load or don't need query history tracking.
 
+If `maxPoolSize` is configured and greater than 0, Trino Gateway uses a
```
ebyhr, line 58: > add db connection pool support

The commit title can be changed to "Add support for connection pool with HikariCP"

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -59,6 +66,17 @@ public Jdbi getJdbi(@Nullable String routingGroupDatabase)
...
+            HikariDataSource dataSource = getOrCreateDataSource(routingGroupDatabase, maxPoolSize);
+            return Jdbi.create(dataSource)
+                    .installPlugin(new SqlObjectPlugin())
+                    .registerRowMapper(new RecordAndAnnotatedConstructorMapper());
+        }
+
```
ebyhr, line 80: Let's avoid redundant code of installPlugin x registerRowMapper something like:
```java
        Jdbi jdbi;
        if (maxPoolSize != null) {
            jdbi = Jdbi.create(getOrCreateDataSource(routingGroupDatabase, maxPoolSize));
        }
        else {
            jdbi = Jdbi.create(buildJdbcUrl(routingGroupDatabase), configuration.getUser(), configuration.getPassword());
        }

        return jdbi
                .installPlugin(new SqlObjectPlugin())
                .registerRowMapper(new RecordAndAnnotatedConstructorMapper());
```

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
```diff
@@ -0,0 +1,177 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestJdbcConnectionManagerPool
+{
+    @Test
+    void testBlocksWhenExceedingMaxPoolSize()
+            throws Exception
+    {
+        String dbPath = Path.of(System.getProperty("java.io.tmpdir"), "h2db-pool-" + System.currentTimeMillis()).toString();
```
ebyhr: You can use `@TempDir` of JUnit.

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
```diff
@@ -0,0 +1,177 @@
...
+        String jdbcUrl = "jdbc:h2:" + dbPath;
+
+        DataStoreConfiguration configuration = new DataStoreConfiguration(
+                jdbcUrl, "sa", "sa", "org.h2.Driver",
+                true, 4, true,
+                2);
```
ebyhr: The convention of parameter is one line or wrap per argument. 
```java
DataStoreConfiguration configuration = new DataStoreConfiguration(jdbcUrl, "sa", "sa", "org.h2.Driver", true, 4, true, 2);
```
or
```java
        DataStoreConfiguration configuration = new DataStoreConfiguration(
                jdbcUrl,
                "sa",
                "sa",
                "org.h2.Driver",
                true,
                4,
                true,
                2);
```

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
```diff
@@ -0,0 +1,177 @@
...
+        try (ExecutorService executorService = Executors.newFixedThreadPool(3)) {
+            List<Future<Connection>> acquired = new ArrayList<>();
+
+            CountDownLatch hold = new CountDownLatch(1);
+            CountDownLatch acquiredLatch = new CountDownLatch(2);
+
+            // Open exactly maxPoolSize connections and keep them open
+            for (int attempt = 0; attempt < 2; attempt++) {
+                acquired.add(executorService.submit(() -> {
+                    try (Handle handle = jdbi.open()) {
```
ebyhr:
```suggestion
                    try (Handle _ = jdbi.open()) {
```

gateway-ha/src/test/java/io/trino/gateway/ha/persistence/TestJdbcConnectionManagerPool.java
```diff
@@ -0,0 +1,177 @@
...
+                        acquiredLatch.countDown();
+                        boolean released = hold.await(10, TimeUnit.SECONDS);
```
ebyhr: Static import `SECONDS`.

mosabua:
@arturmkr can you address the feedback from @ebyhr so we can maybe get this into the 19 release

arturmkr:
@mosabua , will address all issues today

Chaho12:
@arturmkr Let us know when you are ready :)

arturmkr:
Hi @Chaho12 @ebyhr 

Sorry for the long delay in following up on this PR and thank you for your patience and for the detailed review feedback.

I have now updated the branch and addressed all the actionable review comments. While revisiting the implementation, I also realized that the original change only applied connection pooling to routing-group database connections, while the gateway metadata, query history, and scheduled query-history cleanup were still using non-pooled connections.

The connection pool is now used consistently for all gateway datastore access. I’ve also added coverage for pool reuse, pool limits, invalid configuration, shutdown, and the gateway metadata/query-history paths.

I merged the latest main, squashed the changes into a single commit, and ran the complete gateway-ha test suite: 376 tests passed with no failures or errors.

Could you please take another look when you have a chance? 

Thanks again!

oh0873:
Hi @arturmkr, would you be open to also making `keepaliveTime` and `maxLifetime` configurable?

arturmkr:
> Hi @arturmkr, would you be open to also making `keepaliveTime` and `maxLifetime` configurable?

Done! Both `keepaliveTime` and `maxLifetime` are now configurable as optional data store settings. HikariCP defaults are preserved when they are not specified. I also added configuration tests and updated the documentation.

arturmkr:
@mosabua , can we proceed with merging the PR? are there any blockers?

Aaron-ML:
Would love to see this get into a release. Currently seeing memory exhaustion issues due to open memory per connection right now.

Issue #571, opened 2024-12-13
Error 431 Request Header Fields Too Large after adding trino gateway


lozbrown:
After implementing Trino Gateway  and putting it between trino and clients users are seeing rejects with 431, 

> Caused by: java.sql.SQLException: Error executing query: Error starting query at https://quasar-trino-prd.macbank/v1/statement returned an invalid response: JsonResponse{statusCode=431, headers={cache-control=[must-revalidate,no-cache,no-store], content-length=[2057], content-type=[text/html;charset=iso-8859-1], date=[Fri, 13 Dec 2024 14:12:30 GMT]}, hasValue=false} [Error: <html>\n<head>\n<meta http-equiv=\"Content-Type\" content=\"text/html;charset=ISO-8859-1\"/>\n<title>Error 431 Request Header Fields Too Large</title>\n</head>\n<body>\n<h2>HTTP ERROR 431 Request Header Fields Too Large</h2>\n<table>\n<tr><th>URI:</th><td>/v1/statement</td></tr>\n<tr><th>STATUS:</th><td>431</td></tr>\n<tr><th>MESSAGE:</th><td>Request Header Fields Too Large</td></tr>\n</table>\n<h3>

trino version 453
JDBC driver 439

Any ideas what could be causing this?

prakhar10:
Adding `proxy.http-client.request-buffer-size` in the `serverConfig` seems to have helped in my setup.

lozbrown:
> Adding `proxy.http-client.request-buffer-size` in the `serverConfig` seems to have helped in my setup.
> 
> I've noticed this for large queries in PREPARE statements only.

Thanks, what value did you say it to

prakhar10:
> > Adding `proxy.http-client.request-buffer-size` in the `serverConfig` seems to have helped in my setup.
> > 
> > I've noticed this for large queries in PREPARE statements only.
> 
> Thanks @prakhar10 , what value did you set it to, I'm not sure what it defaults to.

I've set it to 20mb.

lozbrown:
Unfortunately this does not seem to have resolved the issue.

danuwasin-id:
try with `http-server.max-request-header-size` instead, it solved for me.

nileshsinha1993:
this also doesn't work

http-server.max-request-header-size

sftwrdvlpr:
[...]

PR #792, opened 2025-11-07
Add Trino Gateway 17 release notes

mosabua:
## Description

Assemble the release notes for Trino Gateway 17 release and adjust rest of docs to version 17.

Release date is set to the planned date following our monthly release cadence.

## Additional context and related issues

* Helm chart release PR to merge after the release - https://github.com/trinodb/charts/pull/386 

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

## Verification for each pull request

Format: PR/issue number, ✅ / ❌ rn ✅ / ❌ docs
✅ rn - release note added and verified, or assessed to be not necessary, set to ❌ rn before completion
✅ docs - need for docs assessed and merged, or assessed to be not necessary, set to ❌ docs before completion

Any dates missing in the list just had no merged PRs.

## 19 Sep 2025

* #767 ✅ rn ✅ docs

## 30 Sep 2025

* #769 ✅ rn ✅ docs

## 1 Oct 2025

* #774 ✅ rn ✅ docs
* #770 ✅ rn ✅ docs

## 9 Oct 2025

* #779 ✅ rn ✅ docs
* #782 ✅ rn ✅ docs
* #780 ✅ rn ✅ docs

## 24 Oct 2025

* #776 ✅ rn ✅ docs

## 29 Oct 2025

* #775 ✅ rn ✅ docs

## 30 Oct 2025

* #741 ✅ rn ✅ docs

## 3 Nov 2025

* #748 ✅ rn ✅ docs

## 28 Nov 2025

* #793 ✅ rn ✅ docs

## 29 Nov 2025

* #802 ✅ rn ✅ docs

## 10 Dec 2025

* #766 ✅ rn ✅ docs

## 11 Dec 2025

* #809 ✅ rn ✅ docs

## 14 Dec 2025

* #788 ✅ rn ✅ docs

## 18 Dec 2025

* #806 ✅ rn ✅ docs

## 30 Dec 2025

* #813 ✅ rn ✅ docs

## 31 Dec 2025

* #814 ✅ rn ✅ docs

## 2 Jan 2026

* #817 ✅ rn ✅ docs

## 3 Jan 2026

* #818 ✅ rn ✅ docs
* #819 ✅ rn ✅ docs
* #820 ✅ rn ✅ docs
* #821 ✅ rn ✅ docs

## 4 Jan 2026

* #823 ✅ rn ✅ docs
* #824 ✅ rn ✅ docs
* #825 ✅ rn ✅ docs
[...]

PR #793, opened 2025-11-08
Update airbase to 325 and airlift bom to 372

sftwrdvlpr:
## Description

- Updated `airbase` to 325
- Updated `airlift` BOM to 372
- Updated `okhttp3` to 5.3.0
- Added `okhttp3` BOM
- Added `okhttp-jvm` dependency
- Updated `jersey-server` to 3.1.11
- Updated `jetty` to 12.1.3

Fixes issue #571 by setting configured `max-request-header-size` and `max-response-header-size` parameters on the HttpClient.

## Release notes

```markdown
* Set `max-request-header-size` and `max-response-header-size` parameters on the HttpClient to support large HTTP headers.
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>io.airlift</groupId>
         <artifactId>airbase</artifactId>
-        <version>277</version>
+        <version>325</version>
     </parent>
 
     <groupId>io.trino.gateway</groupId>
@@ -46,7 +46,7 @@
         <air.modernizer.java-version>8</air.modernizer.java-version>
         <air.release.preparation-goals>clean verify -DskipTests</air.release.preparation-goals>
 
-        <dep.jetty.version>12.0.25</dep.jetty.version>
+        <dep.jetty.version>12.1.3</dep.jetty.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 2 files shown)


mosabua:
Thank you so much for submitting this PR. I think we will try to get this in before the next release (see #792 ) and also include the upgrade to Java 25. 

I am a bit surprised that not more changes are needed .. last time I looked at that upgrade (only for a few minutes and a while ago but anyway) I kinda remember to running into other things, but maybe the bigger jump fixed things.

sftwrdvlpr:
Yes, usually such a big gap between versions doesn't go that smoothly, though I was prepared for bigger changes.

sftwrdvlpr:
@mosabua just out of curiosity, what is the planned release date?

pom.xml
```diff
@@ -46,15 +46,15 @@
         <air.modernizer.java-version>8</air.modernizer.java-version>
         <air.release.preparation-goals>clean verify -DskipTests</air.release.preparation-goals>
 
-        <dep.jetty.version>12.0.25</dep.jetty.version>
+        <dep.jetty.version>12.1.3</dep.jetty.version>
```
wendigo, line 49: 12.1.4

mosabua: Will do that in follow up PR since I cant push to the branch and want to do more work anyway.

PR #794, opened 2025-11-11
Introduce query history write buffer for DB resiliency

Peiyingy:
## Description

Following the work in https://github.com/trinodb/trino-gateway/pull/783 to introduce caching in `HaGatewayManager`, this PR adds a write buffer mechanism to `QueryHistoryManager`. With this change, if the database becomes unavailable, Trino-Gateway can continue to route queries using the cached Trino cluster data while temporarily storing query history records in the write buffer. When the database is available again, the buffered history entries will be flushed. This allows Trino-Gateway to avoid using the database as a single point of failure and improves overall resiliency.  

This approach has already been implemented and validated in production at LinkedIn.

## Testing
- [x] `mvn clean install`


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
```diff
@@ -29,21 +35,45 @@
...
+    private final WriteBuffer<QueryDetail> writeBuffer;
+    private final ScheduledExecutorService scheduledExecutor;
 
-    public HaQueryHistoryManager(Jdbi jdbi, boolean isOracleBackend)
+    @Inject
+    public HaQueryHistoryManager(Jdbi jdbi, boolean isOracleBackend, WriteBufferConfiguration writeBufferConfig)
     {
         dao = requireNonNull(jdbi, "jdbi is null").onDemand(QueryHistoryDao.class);
         this.isOracleBackend = isOracleBackend;
+        if (writeBufferConfig != null && writeBufferConfig.isEnabled()) {
```
oneonestar:
```suggestion
        if (writeBufferConfig.isEnabled()) {
```
`writeBufferConfig` should never be `null`. If it's `null` for whatever reason, we should fail fast.

gateway-ha/src/main/java/io/trino/gateway/ha/router/WriteBuffer.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /** Buffer an item for later flush. Drops the oldest if full. */
+    public void buffer(T item)
+    {
+        if (!deque.offerLast(item)) {
+            deque.pollFirst();
+            deque.offerLast(item);
+        }
+    }
```
oneonestar: This needs `synchronized` to be thread safe.

gateway-ha/src/main/java/io/trino/gateway/ha/router/WriteBuffer.java
```diff
@@ -0,0 +1,66 @@
...
+        for (T next; (next = deque.pollFirst()) != null; ) {
+            try {
+                flusher.accept(next);
+                flushed++;
+            }
+            catch (RuntimeException e) {
+                deque.offerFirst(next);
+                break; // stop after first failure
+            }
+        }
```
oneonestar, line 64: This may fail to insert back to the queue when it's full.
Maybe `peek() -> accept() -> remove()` ?

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+package io.trino.gateway.ha.config;
+
+import io.airlift.units.Duration;
+
+import java.util.concurrent.TimeUnit;
+
+public class WriteBufferConfiguration
+{
+    private boolean enabled;
+    private int maxCapacity = 10000;
```
Chaho12: What is the reason for this number 10,000?

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+    private Duration flushInterval = new Duration(2, TimeUnit.SECONDS);
```
Chaho12, line 24: Same for 2 seconds. What is the reason for this number?

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public int getMaxCapacity()
+    {
+        return maxCapacity;
+    }
+
+    public void setMaxCapacity(int maxCapacity)
+    {
+        this.maxCapacity = maxCapacity;
```
Chaho12, line 46: Maybe check for negative/zero number?

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+    private Duration flushInterval = new Duration(2, TimeUnit.SECONDS);
+
+    public WriteBufferConfiguration() {}
+
+    public WriteBufferConfiguration(boolean enabled, int maxCapacity, Duration flushInterval)
+    {
+        this.enabled = enabled;
+        this.maxCapacity = maxCapacity;
+        this.flushInterval = flushInterval;
+    }
```
Chaho12: Why would you need two constructors?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
```diff
@@ -54,15 +84,27 @@ public void submitQueryDetail(QueryDetail queryDetail)
...
+                    queryDetail.getBackendUrl(),
+                    queryDetail.getUser(),
+                    queryDetail.getSource(),
+                    queryDetail.getCaptureTime(),
+                    queryDetail.getRoutingGroup(),
+                    queryDetail.getExternalUrl());
+        }
+        catch (RuntimeException e) {
+            if (isConnectionIssue(e) && writeBuffer != null) {
+                writeBuffer.buffer(queryDetail);
```
Chaho12: What happens if writer buffer is full?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
```diff
@@ -29,21 +35,45 @@
...
+                Thread t = new Thread(r, "query-history-write-buffer");
+                t.setDaemon(true);
+                return t;
+            });
+            scheduledExecutor.scheduleWithFixedDelay(
+                    this::flushBufferedWrites,
+                    writeBufferConfig.getFlushInterval().toMillis(),
+                    writeBufferConfig.getFlushInterval().toMillis(),
+                    TimeUnit.MILLISECONDS);
+        }
```
Chaho12, line 75: Couldn't there be a race condition

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
Peiyingy: This is just a default value, same for the `flushInterval`. You can always overwrite them in the config file. I'll also update the default value settings in the doc. Do you recommend to set different default values for them?

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
Peiyingy: See https://github.com/trinodb/trino-gateway/pull/794#discussion_r2608648160

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
Peiyingy: Then we won't buffer this record and the query submission would fail - same behavior as the current code without this change.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
Peiyingy: Could you please elaborate on the race condition here? The flusher `scheduledExecutor` is a single thread here, and the buffering thread safe issue is handled in the `WriteBuffer` class

gateway-ha/src/main/java/io/trino/gateway/ha/config/WriteBufferConfiguration.java
Chaho12: I was just wondering what is the justification for this  number 10,000. Maybe other projects use this number for buffer?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
Chaho12: Since this is an improvement to possible issue of loosing record, how about we at least log error that information is lost and increase buffer and/or change flush interval?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
Chaho12: The single-threaded scheduler doesn't eliminate the race condition—it just means only one thread calls flushAll(), but that thread still races with multiple threads calling buffer().

Even though the scheduledExecutor is single-threaded, the race condition occurs between:
- Multiple application threads calling buffer() (synchronized)
- Single scheduler thread calling flushAll() (NOT synchronized)
- Any thread calling size() (NOT synchronized)

gateway-ha/src/main/java/io/trino/gateway/ha/router/WriteBuffer.java
```diff
@@ -0,0 +1,69 @@
...
+            flushed++;
+        }
+        return flushed;
+    }
+
+    public int size()
+    {
+        return deque.size();
+    }
+}
```
Chaho12:
```suggestion
    public synchronized int flushAll(Consumer<T> flusher) // Add synchronized
    {
        int flushed = 0;
        for (T next; (next = deque.peekFirst()) != null; ) {
            try {
                flusher.accept(next);
            }
            catch (RuntimeException e) {
                break; // stop after first failure
            }
            // Only remove after a successful flush
            deque.pollFirst();
            flushed++;
        }
        return flushed;
    }

    public synchronized int size() // Add synchronized
    {
        return deque.size();
    }
}
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaQueryHistoryManager.java
Chaho12: https://github.com/trinodb/trino-gateway/pull/794/files#r2608879369 Check this part

Chaho12:
@Peiyingy Hey, any updates?

Peiyingy:
> @Peiyingy Hey, any updates?

Sorry for the delay. Just updated the PR

Chaho12:
Can you rebase and fix conflicts before we review the PR? Sorry for delayed response.

Issue #789, opened 2025-11-01
Support Backend Cluster Based Pinning


Peiyingy:
### Background
Currently, routing rules in Trino Gateway only support directing traffic to a routing group. In environments where a routing group contains multiple backend clusters, there’s no straightforward way to route queries to a specific backend when needed (for example, for testing, debugging, or gradual rollout purposes). Currently, the only workaround would be registering a cluster multiple times under different routing groups, which is really hacky and complicates configuration management.

Supporting direct routing to a backend cluster would provide finer control and flexibility in query distribution without requiring redundant backend registrations or workarounds.

### Proposal
Extend the routing rule framework to allow rules to specify either:
- `routingGroup`  or
- `backendCluster`

We can still rely on the priority mechanism to determine which setting takes precedence when both are defined.  This feature would enable targeted query routing to a specific backend cluster while maintaining compatibility with the existing routing group based design.

---

### Example YAML rule
```yaml
---
name: "user1_pinning_rule"
description: "Pin user1 directly to adhoc01 backend"
priority: 1
condition: "request.getHeader(\"X-Trino-User\") == \"user1\""
actions:
  - "result.put(\"backendCluster\", \"adhoc01\")"

mosabua:
This looks like a good idea to have as a feature, not necessarily for production use at scale but definitely useful.

My one suggestion is that we opted in the past to remove the term "backend cluster" and instead just use "Trino cluster" .. the renaming is not complete but we should not pile on more.

Also I think theoretically it would be sufficient to just allow naming a routing group and then users would just have to set up a specific group for the one cluster but that seems clumsy. So I vote for implementing both group and cluster.

PR #796, opened 2025-11-18
Support Cluster Based Pinning for Routing Rules

Peiyingy:
## Description

This PR adds Trino-cluster-based pinning as proposed in https://github.com/trinodb/trino-gateway/issues/789.  
File-based routing rules can now specify actions like `"result.put(\"routingCluster\", \"trino-1\")"` to pin directly to a specific backend cluster. The rules engine evaluates all rules and selects the highest-priority match, which can either be a routing group or a routing cluster. 

- If the result is a routing group, the existing behavior applies.
- If the result is a routing cluster, we verify that the cluster exists, is active, and is healthy. If it passes these checks, the query is routed to that cluster. If not, we fall back to the default routing group, matching the current behavior for routing group.

We also support external routing selector to return `routingCluster` as a `routingDestination`.

### Logic Changes

- File-based routing rules now support `"result.put(\"routingCluster\", \"trino-1\")"`.
- In `FileBasedRoutingSelector.findRoutingDestination`, we now ensure the result map contains exactly one entry so that only the highest-priority decision is kept.

### Key Change Explanation 

All routing rules are sorted by priority. In `FileBasedRoutingSelector`, we iterate through them from low to high priority, checking whether each rule matches. Each match overwrites the previous decision. By the end, we have the routing decision associated with the highest-priority rule.

Previously, we only had one possible result key: `routingGroup`, so following the behavior as described above, the result `HashMap` would return us only one `routingGroup` key, with the value of the highest priority routing group.

Now, we support *both* `routingGroup` and `routingCluster`, and using a regular `HashMap` can return **two** results: 

- `routingGroup` key with the value of highest priority group
- `routingCluster` key with the value of the highest priority cluster. 

At this point, we won't know which one actually has the higher priority. 

To solve this issue, we switched the result map to a `LinkedHashMap`, with the eviction mechanism of `removeEldestEntry` to keep the map size at one. This ensures that as we traverse rules, a higher-priority routing cluster can override a lower-priority routing group, and vice versa. Thus, after all the rules are traversed, we end up with exactly one result - either a routing group or a routing cluster - with the globally highest priority.

### Refactoring

- Renamed `RoutingGroupSelector` to the more general `RoutingSelector`, with corresponding updates to all the classes implement it.
- Renamed `RoutingGroupResponse` to `RoutingResponse` and added a `routingCluster` field.
- Added a `routingCluster` field in `RoutingSelectorResponse`
- Replaced `routingGroup` in `RoutingDestination` with a more general variable name `routingDecision`, which can be either a routing group or a routing cluster. This change also applies to:
  - the `query_history` table column  
  - cache key  
  - UI display

## Testing

- [x] `mvn clean install`
- [x] Added new test cases in `TestRoutingSelector`
- [x] Local test to verify the routing cluster and routing group pinning both work
Screenshot 2025-11-17 at 23 47 48


RoeyoOgen:
in PR description you say:
> If the result is a routing cluster, we verify that the cluster exists, is active, and is healthy. If it passes these checks, the query is routed to that cluster. If not, we fall back to the adhoc routing group, matching the current behavior for routing group.

did you make sure to [accommodate](https://trinodb.github.io/trino-gateway/routing-rules/):
`
routing:
    defaultRoutingGroup:
` config ?

gateway-ha/config.yaml
```diff
@@ -4,7 +4,7 @@ serverConfig:
 
 routingRules:
   rulesEngineEnabled: False
-  # rulesConfigPath: "src/main/resources/rules/routing_rules.yml"
+  # rulesConfigPath: "gateway-ha/src/main/resources/rules/routing_rules.yml"
```
RoeyoOgen, line 10: why have this change?
also why `-ha?`

RoeyoOgen, line 10: why have this change?
also why `-ha?`

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingSelector.java
```diff
@@ -59,7 +61,13 @@ public FileBasedRoutingGroupSelector(String rulesPath, Duration rulesRefreshPeri
     @Override
     public RoutingSelectorResponse findRoutingDestination(HttpServletRequest request)
     {
-        Map<String, String> result = new HashMap<>();
+        // Keep only the highest-priority rule result by limiting the map to a single entry.
+        LinkedHashMap<String, String> result = new LinkedHashMap<>(1) { @Override
```
RoeyoOgen, line 65: why this this change needed? is this a current issue or changed by added behaviour?

RoeyoOgen (commented):
also, i might have missed but i don't see any new tests just fixing existing ones

Peiyingy:
> in PR description you say:
> 
> > If the result is a routing cluster, we verify that the cluster exists, is active, and is healthy. If it passes these checks, the query is routed to that cluster. If not, we fall back to the adhoc routing group, matching the current behavior for routing group.
> 
> did you make sure to [accommodate](https://trinodb.github.io/trino-gateway/routing-rules/): `routing: defaultRoutingGroup:` config ?

Yes, If there's no cluster matching the rule exists, we would fall back to the `defaultRoutingGroup`. The logic is written here.
https://github.com/trinodb/trino-gateway/blob/e28d4c2a1019ebb666f500529f34f60595942a18/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java#L120

I've also modified the PR description to make it clear.

Peiyingy:
> also, i might have missed but i don't see any new tests just fixing existing ones

New tests are added in `TestRoutingSelector`

https://github.com/trinodb/trino-gateway/blob/e28d4c2a1019ebb666f500529f34f60595942a18/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingSelector.java#L562-L615

gateway-ha/config.yaml
Peiyingy: When I'm testing with `TrinoGatewayRunner`, it sets the working directory to the repository root. For instance, to read the `config.yaml` file, we need to put `gateway-ha` directory in the path
https://github.com/trinodb/trino-gateway/blob/5a36940c3e7c5a69af0b29376141c20ae78f020a/gateway-ha/src/test/java/io/trino/gateway/TrinoGatewayRunner.java#L64

Wondering what's the reason we'd like to set it to `rulesConfigPath: "src/main/resources/rules/routing_rules.yml"`?

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingSelector.java
Peiyingy: Good question, I've added the `Key Change Explanation`  section in the PR description to explain this change

Issue #797, opened 2025-11-21
Add Stronger Isolation Semantics for Routing Rules


Peiyingy:
## Proposal: Add Stronger Isolation Semantics for Routing Rules

Today, Trino-Gateway supports routing rules that can pin traffic to either a routing group. However, the current behavior still allows fallback routing when the pinned backend becomes unavailable. 

At LinkedIn, we have some use cases where certain workloads must be strictly isolated from the general adhoc clusters. To support this, we propose introducing an optional `enforceIsolation` flag in routing-rule definitions. 
- When `enforceIsolation = false (default)`, the routing behavior remains unchanged.
- When `enforceIsolation = true`, the gateway would *not* fall back to adhoc clusters if the pinned backend becomes unavailable. Instead, the query should fail immediately at the gateway layer, ensuring true isolation.

#### Sample Routing Rule
```
---
name: "user_pinning_rule"
priority: 1
enforceIsolation: True
condition: "request.getHeader(\"X-Trino-User\") == \"user\""
actions:
  - "result.put(\"routingGroup\", \"Trino-1\")"
```

PR #804, opened 2025-12-04
Implement Strict Routing Semantics for Routing Rules

Peiyingy:
## Description

This PR adds `strictRouting ` flag for routing as proposed in https://github.com/trinodb/trino-gateway/issues/797.  
Routing rules can now support `strictRouting: True` to force pinning the query to the routing group

When `strictRouting = false (default)`, the routing behavior remains unchanged.
When `strictRouting = true`, the gateway would **not** fall back to default clusters if the pinned backend becomes unavailable. Instead, the query should fail immediately with a 404 error.

## Testing

- [x] `mvn clean install`
- [x] Added new test cases in `TestStochasticRoutingManager`
- [x] Local test to verify the `strictRouting ` flag works

```
trino> select 1;
Error starting query at http://localhost:8080/v1/statement returned an invalid response: JsonResponse{statusCode=404, headers={content-length=[96], content-type=[text/plain], date=[Thu, 04 Dec 2025 03:45:04 GMT], vary=[Accept-Encoding]}, hasValue=false} [Error: No healthy backends available for routing group 'adhoc1' under strict routing for user 'pye']
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
@@ -33,7 +33,7 @@ public record RoutingRule(
         String name,
         String description,
         Integer priority,
-        Boolean enforceIsolation,
+        Boolean strictRouting,
         List<String> actions,
         String condition)
 {
@@ -41,7 +41,7 @@ public record RoutingRule(
         requireNonNull(name, "name is null");
         description = requireNonNullElse(description, "");
         priority = requireNonNullElse(priority, 0);
-        enforceIsolation = requireNonNullElse(enforceIsolation, false);
+        strictRouting = requireNonNullElse(strictRouting, false);
         actions = ImmutableList.copyOf(actions);
         requireNonNull(condition, "condition is null");
     }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
@@ -33,6 +33,7 @@ public record RoutingRule(
         String name,
         String description,
         Integer priority,
+        Boolean enforceIsolation,
         List<String> actions,
         String condition)
 {
@@ -40,6 +41,7 @@ public record RoutingRule(
         requireNonNull(name, "name is null");
         description = requireNonNullElse(description, "");
         priority = requireNonNullElse(priority, 0);
+        enforceIsolation = requireNonNullElse(enforceIsolation, false);
         actions = ImmutableList.copyOf(actions);
         requireNonNull(condition, "condition is null");
     }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
@@ -26,6 +26,7 @@ import static java.util.Objects.requireNonNullElse;
  * @param name name of the routing rule
  * @param description description of the routing rule. Defaults to an empty string if not provided, indicating the user intends it to be blank.
  * @param priority priority of the routing rule. Higher number represents higher priority. If two rules have same priority then order of execution is not guaranteed.
+ * @param strictRouting if true, return 503 instead of falling back to the default backend when no healthy backend is available for the routing group
  * @param actions actions of the routing rule
  * @param condition condition of the routing rule
  */
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -95,7 +95,7 @@ public class RoutingTargetHandler
         String routingGroup = !isNullOrEmpty(routingDestination.routingGroup())
                 ? routingDestination.routingGroup()
                 : defaultRoutingGroup;
-        ProxyBackendConfiguration backendConfiguration = routingManager.provideBackendConfiguration(routingGroup, user, routingDestination.enforceIsolation());
+        ProxyBackendConfiguration backendConfiguration = routingManager.provideBackendConfiguration(routingGroup, user, routingDestination.strictRouting());
         String clusterHost = backendConfiguration.getProxyTo();
         String externalUrl = backendConfiguration.getExternalUrl();
         // Apply headers from RoutingDestination if there are any
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -91,11 +91,14 @@ public class RoutingTargetHandler
         RoutingSelectorResponse routingDestination = routingGroupSelector.findRoutingDestination(request);
         String user = request.getHeader(USER_HEADER);
 
-        // This falls back on default routing group backend if there is no cluster found for the routing group.
+        // When no cluster is found:
+        //   - If strictRouting is false, fall back to the default routing group backend.
+        //   - If strictRouting is true, return a 404 response.
         String routingGroup = !isNullOrEmpty(routingDestination.routingGroup())
                 ? routingDestination.routingGroup()
                 : defaultRoutingGroup;
-        ProxyBackendConfiguration backendConfiguration = routingManager.provideBackendConfiguration(routingGroup, user, routingDestination.strictRouting());
+        boolean strictRouting = Optional.ofNullable(routingDestination.strictRouting()).orElse(false);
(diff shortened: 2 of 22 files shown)


RoeyoOgen:
Just a few remarks:
1. why "enforceIsolation" and not _strictRouting_? or _noFallback_?
2. does this differ if the routing group is a single cluster rather than a group? - does this clash with your other PR #[796](https://github.com/trinodb/trino-gateway/pull/796)?

xkrogen:
> why "enforceIsolation" and not strictRouting? or noFallback?

+1 `strictRouting` is a good name

Peiyingy:
> Just a few remarks:
> 
> 1. why "enforceIsolation" and not _strictRouting_? or _noFallback_?
> 2. does this differ if the routing group is a single cluster rather than a group? - does this clash with your other PR #[796](https://github.com/trinodb/trino-gateway/pull/796)?

1. Thanks for the suggesting. I'll change the name to `strictRouting`
2. This would also support routing cluster rules. After either of the PR is merged, we just need to rebase the other PR to make sure both `strictRouting` and cluster routing are supported.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -99,15 +103,21 @@ public ProxyBackendConfiguration provideDefaultBackendConfiguration(String user)
...
+     * has no suitable backend unless {@code strictRouting} is true, in which case a 404 is returned.
      */
     @Override
-    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
+    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user, Boolean strictRouting)
     {
         List<ProxyBackendConfiguration> backends = gatewayBackendManager.getActiveBackends(routingGroup).stream()
                 .filter(backEnd -> isBackendHealthy(backEnd.getName()))
                 .toList();
+        if (backends.isEmpty() && strictRouting) {
```
kbhatianr: This may be a nitpick, but evaluating `strictRouting` first can short circuit the `backends.isEmpty()` one, in case it's false. A teeny performance optimization would be to re-order these so it's `if (strictRouting && backends.isEmpty())`

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -99,15 +103,21 @@ public ProxyBackendConfiguration provideDefaultBackendConfiguration(String user)
...
      */
     @Override
-    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
+    public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user, boolean strictRouting)
     {
         List<ProxyBackendConfiguration> backends = gatewayBackendManager.getActiveBackends(routingGroup).stream()
                 .filter(backEnd -> isBackendHealthy(backEnd.getName()))
                 .toList();
+        if (strictRouting && backends.isEmpty()) {
+            throw new WebApplicationException(
```
RoeyoOgen, line 116: why WebApplicationException and not IllegalStateException like in [IllegalStateException("Number of active backends found zero"))](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java)

or maybe a custom Explicit Exception inherited from [RouterException](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/RouterException.java)?

felicity3786, line 116: We can add a WARN or ERROR with context when the 404 is triggered? Also +1 on maybe making it a `NoBackendAvailableException`?

felicity3786 (commented):
Let's also update the official doc?

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -99,15 +103,21 @@ public ProxyBackendConfiguration provideDefaultBackendConfiguration(String user)
...
+                             Response.status(NOT_FOUND)
```
oneonestar: Client did nothing wrong and shouldn't receive a 4xx error. We should return a 5XX indicate a server side error, maybe `503 Service Unavailable`.

oneonestar (commented):
It's ambiguous what you mean `strictRouting`.

There are two fallback logics in routing.
1. If `RoutingGroupSelector` failed to select a routing group for a query, fallback to `defaultRoutingGroup`. https://github.com/trinodb/trino-gateway/blob/eb736ee547d7cca82b78026a4cd275b718c49d07/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L97-L99

2. If there are no available active healthy cluster for the selected routing group, fallback to `ActiveDefaultBackends` (ie. active healthy cluster in `defaultRoutingGroup`). https://github.com/trinodb/trino-gateway/blob/a4159fcd135b988edad6ec92a97be5a088925b2b/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java#L121

It's not clear which fallback logic (or both) are disabled. If you decided to disable both (currently you only handled 2), please make sure the error messages are clearly distinguishable for these cases. Docs are also required to explain the exact affect of this config.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -91,11 +91,14 @@ private RoutingTargetResponse getRoutingTargetResponse(HttpServletRequest reques
...
 
-        // This falls back on default routing group backend if there is no cluster found for the routing group.
+        // When no cluster is found:
+        //   - If strictRouting is false, fall back to the default routing group backend.
+        //   - If strictRouting is true, return a 404 response.
         String routingGroup = !isNullOrEmpty(routingDestination.routingGroup())
                 ? routingDestination.routingGroup()
                 : defaultRoutingGroup;
-        ProxyBackendConfiguration backendConfiguration = routingManager.provideBackendConfiguration(routingGroup, user);
+        boolean strictRouting = Optional.ofNullable(routingDestination.strictRouting()).orElse(false);
```
oneonestar:
```suggestion
        boolean strictRouting = requireNonNullElse(routingDestination.strictRouting(), false);
```

gateway-ha/src/main/java/io/trino/gateway/ha/router/schema/ExternalRouterResponse.java
oneonestar, line 26: Update comment.

gateway-ha/src/main/java/io/trino/gateway/ha/domain/RoutingRule.java
oneonestar, line 32: Update comment.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -91,11 +91,14 @@ private RoutingTargetResponse getRoutingTargetResponse(HttpServletRequest reques
         RoutingSelectorResponse routingDestination = routingGroupSelector.findRoutingDestination(request);
         String user = request.getHeader(USER_HEADER);
 
-        // This falls back on default routing group backend if there is no cluster found for the routing group.
+        // When no cluster is found:
+        //   - If strictRouting is false, fall back to the default routing group backend.
+        //   - If strictRouting is true, return a 404 response.
```
oneonestar: Move the comment to `provideBackendConfiguration()`.
Also, we return 503 now, not 404.

gateway-ha/src/main/java/io/trino/gateway/ha/router/FileBasedRoutingGroupSelector.java
```diff
@@ -72,13 +73,15 @@ public RoutingSelectorResponse findRoutingDestination(HttpServletRequest request
...
+        for (RoutingRule rule : requireNonNull(rules.get())) {
             if (rule.evaluateCondition(data, state)) {
                 log.debug("%s evaluated to true on request: %s", rule, request);
                 rule.evaluateAction(result, data, state);
+                strictRouting = rule.isStrictRouting();
             }
-        });
-        return new RoutingSelectorResponse(result.get(RESULTS_ROUTING_GROUP_KEY));
+        }
+        return new RoutingSelectorResponse(result.get(RESULTS_ROUTING_GROUP_KEY), ImmutableMap.of(), strictRouting);
```
oneonestar, line 88: This logic is flawed.
The value of `strictRouting` is always equal to the last evaluated rule, but the last evaluated rule may not be the one that sets the final `routingGroup` (e.g., it may only update the `state` without setting `routingGroup`).

oneonestar (commented):
I don’t like setting the default value to `false` when it’s `null` in multiple places, but it might require heavy refactoring to remove that pattern. I’m okay with it for now, and we can improve it later in another PR.

PR #805, opened 2025-12-08
Add Dropdown to Change Timezones

kjsbot:
## Description
Currently the Trino Gateway UI does not have an option to change the timezone. This PR adds a dropdown to allow the user to select any timezone they want.

## Testing
- mvnw clean install
- Tested the UI
Screenshot 2025-12-18 at 5 58 59 AM
Screenshot 2025-12-18 at 5 59 22 AM

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/domain/response/DistributionResponse.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/domain/response/DistributionResponse.java
@@ -220,20 +220,20 @@ public class DistributionResponse
 
     public static class LineChart
     {
-        private Long timestamp;
+        private Long epochMillis;
         private String backendUrl;
         private Long queryCount;
         private String name;
 
         @JsonProperty
-        public Long getTimestamp()
+        public Long getEpochMillis()
         {
-            return timestamp;
+            return epochMillis;
         }
 
-        public void setTimestamp(Long timestamp)
+        public void setEpochMillis(Long epochMillis)
         {
-            this.timestamp = timestamp;
+            this.epochMillis = epochMillis;
         }
 
         @JsonProperty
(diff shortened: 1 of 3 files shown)


mosabua:
What is the motivation for this change? I would understand if it defaults to the local timezone of the browser of the user automatically and you can set it to just display UTC as stored in the database .. but I dont understand the logic to allow selection of any timezone

oneonestar:
I think this feature make sense for teams working across multiple timezones. Grafana and Datadog also have this feature.

felicity3786:
> I think this feature make sense for teams working across multiple timezones. Grafana and Datadog also have this feature.

Agree seems like a valid dashboard feature.

Peiyingy:
+1 on this feature make sense. In LinkedIn, we have developers from different timezones, and we've gained benefit from this changing timezone feature. This is also a common feature for dashboard applications.

gateway-ha/src/main/java/io/trino/gateway/ha/domain/response/DistributionResponse.java
```diff
@@ -220,20 +220,20 @@ public void setName(String name)
...
     {
-        private String minute;
+        private String timestamp;
         private String backendUrl;
         private Long queryCount;
         private String name;
 
         @JsonProperty
-        public String getMinute()
+        public String getTimestamp()
```
kbhatianr: Should we change this from a `String` over to a `long` or something similar?

kbhatianr:
+1 - love adding this feature in. IMO, this feature should be utilized everywhere that we show a timestamp value:
1. Query History => Submission Time
2. Dashboard => Started At

mosabua:
Sounds like we have consensus to proceed. We talked about it also in the dev sync this week. The main thing we should make sure is that the setting is not isolated to one page but is applicable to the whole app and probably visible as switch everywhere .. and that any writing of data does not get messed up

gateway-ha/src/main/java/io/trino/gateway/ha/domain/response/DistributionResponse.java
```diff
@@ -220,20 +220,20 @@ public void setName(String name)
 
     public static class LineChart
     {
-        private String minute;
+        private Long timestamp;
```
ebyhr: The field name is confusing. Could you rename to `epochMillis` or something?

Chaho12:
Can you squash commits so that we can rebase&merge? 
https://trino.io/development/process.html#pull-request-and-commit-guidelines

PR #806, opened 2025-12-10
Show query text in QueryHistory UI

kbhatianr:
## Description
In PR-740, the UI was updated to show the query text in the QueryHistory component in a modal component, to account for larger query texts. This makes it more difficult for smaller query texts that would fit otherwise in the page. 
In this PR, we revert back to showing the query text, but also make it a link that's clickable for larger texts.  

Previous to the change, screenshot:
image

After the change screenshot:
image

## Release notes

```markdown
* Show truncated query text in Query History UI
```

### DIFF
--- a/webapp/src/components/history.tsx
+++ b/webapp/src/components/history.tsx
@@ -87,9 +87,10 @@ export function History() {
       <Typography.Text
          link={{ onClick: () => handleShowModal(text) }}
          underline
-         style={{ cursor: "pointer" }}
+         style={{ cursor: "pointer", width: "300px"}}
+         ellipsis={{ showTooltip: true }}
       >
-         View Query
+         {text}
       </Typography.Text>
     );
 


webapp/src/components/history.tsx
```diff
@@ -88,8 +88,9 @@ export function History() {
          link={{ onClick: () => handleShowModal(text) }}
          underline
          style={{ cursor: "pointer" }}
+         ellipsis={{ showTooltip: true }}
       >
-         View Query
+         {text}
```
mosabua, line 93: How does this trim the text? Statically based on number of characters or dynamically based on room in the UI?

Chaho12: IIRC, previously API returns as trimmed text + and {text} shows even lesser than that (hover til shows trimmed text)
e.g) `SELECT field1, field2 FROM table WHERE dt = '2025-12-11' AND field like '...`

kbhatianr: I've been unable to test this fully out, due to not being able to build the project; so to be fair I don't actually know if this works as intended. 

My understanding of how the [semi-ui ellipsis](https://semi.design/en-US/basic/typography#Ellipsis) works is that the parent needs to have some width, and the text in the `Typography.text` needs to be wider than that. 
[Over here](https://github.com/trinodb/trino-gateway/blob/main/webapp/src/components/history.tsx#L158), the column is defined as having a fixed width of 300.

kbhatianr: I tested it out, and it did not work. I submitted a subsequent change to match the size of the column to the header.

Chaho12: What is the issue with not being able to build the project?

kbhatianr: Completely unrelated; we can take that on slack and/or create a separate issue.

I did end up being able to launch the webapp locally, and put in a screenshot of the UI in the description

webapp/README.md
```diff
@@ -10,7 +10,21 @@
 
 # Development
 
+## Pre-Requisites
```
mosabua:
```suggestion
## Requirements

* pnpm installed and available on the PATH
```

webapp/README.md
```diff
@@ -10,7 +10,21 @@
...
+1. pnpm
+
+MacOS:
```
mosabua: Remove that whole thing and maybe just link to the pnpm website in the bullet item list

webapp/README.md
```diff
@@ -10,7 +10,21 @@
...
+```
+brew install pnpm
 ```
+
+## Build & Run
```
mosabua:
```suggestion
## Build and run
```

kbhatianr:
@cla-bot check

webapp/src/components/history.tsx
```diff
@@ -87,7 +87,7 @@ export function History() {
       <Typography.Text
          link={{ onClick: () => handleShowModal(text) }}
          underline
-         style={{ cursor: "pointer" }}
+         style={{ cursor: "pointer", width: "300px"}}
```
ebyhr, line 90: > Add in width to the individual text box, so it'll render correctly. This width matches the column width

Please shorten the commit title.

webapp/README.md
ebyhr: Please extract a commit for README change. This is unrelated to width change in history.tsx.

Chaho12:
Change commit mesesage, split commit and we are good to go

ebyhr:
@kbhatianr Thanks for attaching the screenshot! Can you provide 2 screenshots (before and after this change)? It will reduce reviewer's burden.

PR #807, opened 2025-12-12
Enable health check to use mtls to connect to backends

felicity3786:
## Description
Today, Trino Gateway’s health monitors authenticate to backend clusters using either:
```
Authorization: Basic header (username/password), or X-Trino-User header (when no password is configured)
```
This creates operational burdens for deployments that must rotate or manage shared credentials, and it exposes unnecessary authentication material in gateway configs.
This PR introduces optional mTLS-based authentication for health checks, allowing deployments to rely solely on client certificates when connecting to backend JMX or metrics endpoints.

`backendState.monitorMtlsEnabled: true/false`
When `monitorMtlsEnabled`=true, the health monitors switch to certificate-based auth and no longer send any identity headers. Startup validation for TLS required configuration is also added.
If `monitorMtlsEnabled` is false (default), existing behavior is unchanged, provides the backward compatibility.

### Test
Build passed.
Tested with local gateway the health check is working as expected.

NA

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsMetricsMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsMetricsMonitor.java
@@ -53,6 +53,7 @@ public class ClusterStatsMetricsMonitor
     private final int retries;
     private final MetricsResponseHandler metricsResponseHandler;
     private final Header identityHeader;
+    private final boolean monitorMtlsEnabled;
     private final String metricsEndpoint;
     private final String runningQueriesMetricName;
     private final String queuedQueriesMetricName;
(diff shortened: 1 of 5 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java
```diff
@@ -127,6 +137,28 @@ public HaGatewayProviderModule(HaGatewayConfiguration configuration)
         queryHistoryManager = new HaQueryHistoryManager(jdbi, configuration.getDataStore().getJdbcUrl().startsWith("jdbc:oracle"));
     }
 
+    private static void validateMonitorMtlsConfig(Map<String, String> serverConfig)
+    {
+        String[] requiredKeys = new String[] {
+                MONITOR_HTTP_CLIENT_KEY_STORE_PATH,
+                MONITOR_HTTP_CLIENT_KEY_STORE_PASSWORD,
+                MONITOR_HTTP_CLIENT_TRUST_STORE_PATH,
+                MONITOR_HTTP_CLIENT_TRUST_STORE_PASSWORD
```
oneonestar, line 146: A user could skip setting truststore. Airlift would fallback to some system default truststore.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsMetricsMonitor.java
```diff
@@ -65,12 +66,18 @@ public ClusterStatsMetricsMonitor(HttpClient httpClient, BackendStateConfigurati
...
+                identityHeader = new Header("Authorization",
+                        new BasicCredentials(backendStateConfiguration.getUsername(), backendStateConfiguration.getPassword()).getBasicAuthHeader());
+            }
+            else {
+                identityHeader = new Header("X-Trino-User", backendStateConfiguration.getUsername());
+            }
         }
         else {
-            identityHeader = new Header("X-Trino-User", backendStateConfiguration.getUsername());
+            identityHeader = null;
```
oneonestar, line 80: Please use `Optional` instead of `null`.

gateway-ha/src/main/java/io/trino/gateway/ha/module/HaGatewayProviderModule.java
```diff
@@ -81,6 +82,11 @@
 public class HaGatewayProviderModule
         extends AbstractModule
 {
+    private static final String MONITOR_HTTP_CLIENT_KEY_STORE_PATH = "monitor.http-client.key-store-path";
+    private static final String MONITOR_HTTP_CLIENT_KEY_STORE_PASSWORD = "monitor.http-client.key-store-password";
+    private static final String MONITOR_HTTP_CLIENT_TRUST_STORE_PATH = "monitor.http-client.trust-store-path";
+    private static final String MONITOR_HTTP_CLIENT_TRUST_STORE_PASSWORD = "monitor.http-client.trust-store-password";
+
```
Peiyingy, line 89: Actually I was thinking of something like this, WDYT?

```suggestion
    private static final List<String> REQUIRED_MONITOR_MTLS_KEYS = List.of(
            "monitor.http-client.key-store-path",
            "monitor.http-client.key-store-password",
            "monitor.http-client.trust-store-path",
            "monitor.http-client.trust-store-password");
```

mosabua:
I think this PR still provides a useful feature. What do you think @felicity3786 @oneonestar  .. if yes we should reopen and continue work on it ideally.

ebyhr:
Closing as there is no response over three weeks.

PR #808, opened 2025-12-12
Migrate from Guava Cache to Caffeine

oneonestar:
## Description
Migrate from Guava Cache to Caffeine.

Motivation:
https://github.com/trinodb/trino-gateway/pull/783#discussion_r2608489696

Removed the call to `setRoutingGroupForQueryId()` and `setExternalUrlForQueryId()` because `cache.get()` already insert the value into the cache if absent.

## Release notes

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -75,6 +75,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>3.2.3</version>
+        </dependency>
+
+        <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
@@ -13,12 +13,11 @@
  */
 package io.trino.gateway.ha.router;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Strings;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
 import io.trino.gateway.ha.clustermonitor.TrinoStatus;
@@ -36,7 +35,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -123,7 +121,7 @@ public abstract class BaseRoutingManager
         try {
             backendAddress = queryIdBackendCache.get(queryId);
         }
-        catch (ExecutionException e) {
+        catch (RuntimeException e) {
             log.warn("Exception while loading queryId from cache %s", e.getLocalizedMessage());
         }
         return backendAddress;
@@ -137,7 +135,7 @@ public abstract class BaseRoutingManager
         try {
             externalUrl = queryIdExternalUrlCache.get(queryId);
         }
-        catch (ExecutionException e) {
+        catch (RuntimeException e) {
             log.warn("Exception while loading queryId from cache %s", e.getLocalizedMessage());
         }
         return externalUrl;
@@ -155,7 +153,7 @@ public abstract class BaseRoutingManager
         try {
             routingGroup = queryIdRoutingGroupCache.get(queryId);
         }
-        catch (ExecutionException e) {
+        catch (RuntimeException e) {
             log.warn("Exception while loading queryId from routing group cache %s", e.getLocalizedMessage());
         }
         return routingGroup;
@@ -245,9 +243,7 @@ public abstract class BaseRoutingManager
      */
     private String findRoutingGroupForUnknownQueryId(String queryId)
     {
-        String routingGroup = queryHistoryManager.getRoutingGroupForQueryId(queryId);
-        setRoutingGroupForQueryId(queryId, routingGroup);
-        return routingGroup;
+        return queryHistoryManager.getRoutingGroupForQueryId(queryId);
     }
 
     /**
@@ -255,25 +251,15 @@ public abstract class BaseRoutingManager
      */
     private String findExternalUrlForUnknownQueryId(String queryId)
     {
-        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
-        setExternalUrlForQueryId(queryId, externalUrl);
-        return externalUrl;
+        return queryHistoryManager.getExternalUrlForQueryId(queryId);
     }
 
     private LoadingCache<String, String> buildCache(Function<String, String> loader)
     {
-        return CacheBuilder.newBuilder()
+        return Caffeine.newBuilder()
                 .maximumSize(10000)
                 .expireAfterAccess(30, TimeUnit.MINUTES)
-                .build(
-                        new CacheLoader<>()
-                        {
-                            @Override
-                            public String load(String queryId)
-                            {
-                                return loader.apply(queryId);
-                            }
-                        });
+                .build(loader::apply);
     }
 
     private boolean isBackendHealthy(String backendId)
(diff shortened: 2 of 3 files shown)


ebyhr (dismissed):
In Trino, we intentionally avoid Caffeine (see https://github.com/trinodb/trino/issues/10512#issuecomment-1016221168). Please hold off on this PR until we reach consensus.

xkrogen:
@ebyhr I don't think the discussion from that issue is relevant here. GW cache doesn't have any invalidations to be concerned about. Even if we did have such concerns, GW is using the plain Guava cache, not Trino's [EvictableCache](https://github.com/trinodb/trino/blob/master/lib/trino-cache/src/main/java/io/trino/cache/EvictableCache.java), so (a) we are already not standardized, and (b) the behavior will at least be improved when moving from Guava to Caffeine.

I see two options here:
* Move to Caffeine -- This lets GW use the current best known in-memory Java caching solution, at the cost of some divergence from Trino (though no worse than the current situation).
* Move to `EvictableCache` -- `trino-cache` is already exposed as a separate module which we could pull in here. This at least unifies the implementation, so that any future improvements in Trino (e.g. moving the underlying implementation to Caffeine) will benefit GW as well.

Leaving things as-is only provides the illusion of standardization across Trino / GW ..

mosabua:
Can you start a thread about Caffeine vs Guava cache use on the core-dev channel on Trino slack please so we can move this forward.

Moving to caffeine might be easier .. but we also already have a dependency on some trino aspects so trino-cache use might also be fine .. we would have to see what is heavier as well maybe.

Also the avoidance of caffeine might be invalid by now .. that mention was over 3 years ago.

oneonestar:
From https://github.com/trinodb/trino/issues/10512#issuecomment-1012153981, Trino avoided using Caffeine for the following reasons:

> Caffeine does not solve the problem fully for us. We still need to handle invalidateAll() anyway. So migration is out of the scope here..

We don't have the invalidation problem in the gateway. If such a problem arises in the future, we could try `AsyncCache` or the generation ID approach as mentioned in the cache discussion thread.

> One change at a time. Changing caching implementation will surely affect users somehow. So we prefer to avoid changing cache backend now to make sure that we are able do diagnose issues in future. Once we get sure that current fix is working as expected we might consider changing the backend.

Irrelevant.

> I am not sure how to benchmark cost and benefit tradeoffs for Guava cache vs Caffeine.
> We are concerned a bit about project future. Guava is supported by giant company so future sounds a bit safer here. On the other hand I am disappointed that they haven't fixed/addressed the long standing serious issue.

It appears that time has already proven Caffeine to have wider adoption and better maintenance than Guava cache.

Instead of moving to `EvictableCache`, I'd prefer moving to Caffeine in the gateway. Trying out Caffeine in the gateway could potentially help Trino transition towards Caffeine as well.

Peiyingy:
I'd also prefer moving to `Caffeine` instead of `EvictableCache`. In https://github.com/trinodb/trino-gateway/pull/783, given we want to support multiple caching strategies (e.g. cache-aside and DB-first), it would be more extendable to build on top of the `Caffeine` cache

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -245,35 +243,23 @@ private String searchAllBackendForQuery(String queryId)
...
+        return queryHistoryManager.getRoutingGroupForQueryId(queryId);
     }
 
     /**
      * Attempts to look up the external url associated with the query id from query history table
      */
     private String findExternalUrlForUnknownQueryId(String queryId)
     {
-        String externalUrl = queryHistoryManager.getExternalUrlForQueryId(queryId);
-        setExternalUrlForQueryId(queryId, externalUrl);
```
kbhatianr, line 259: Removing the `set...` means you are no longer interacting with the cache. Where do you insert the `externalUrl` into the cache now?

oneonestar: https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/LoadingCache.html#get(K)
> If the specified key is not already associated with a value, attempts to compute its value and enters it into this cache unless null.

This means that when calling get() on the cache, the value will be inserted into the cache.
```
queryIdExternalUrlCache.get() -> CacheLoader.load() -> findExternalUrlForUnknownQueryId()
```

vishalya: Then how does it impact the testing, do we still need the function `setExternalUrlForQueryId` or those tests should be modified.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -198,7 +196,7 @@ private Optional<String> extractUserFromBearerAuth(String header, String userFie
                 userInfo = Optional.of(userInfoCache.orElseThrow().get(token));
                 return Optional.of(userInfo.orElseThrow().getSubject().toString());
             }
-            catch (ExecutionException e) {
+            catch (RuntimeException e) {
```
vishalya, line 199: Any reason why `ExecutionException` is not good here?

gateway-ha/pom.xml
```diff
@@ -74,6 +74,12 @@
...
 
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>3.2.3</version>
+        </dependency>
+
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
```
vishalya, line 85: can you remove guava dependency?

gateway-ha/pom.xml
```diff
@@ -74,6 +74,12 @@
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>3.2.3</version>
```
vishalya: how about adding the version [here ](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/pom.xml#L24)and using it?

vishalya (dismissed):

gateway-ha/pom.xml
oneonestar: We still need guava for `ImmutableList` and others.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
oneonestar: Guava throws `ExecutionException` but Caffeine throws `RuntimeException`.
 
* [guava#get()](https://github.com/google/guava/blob/8df92320363fa0934d97167dc9312afbf61aa7f6/guava/src/com/google/common/cache/LoadingCache.java#L71)
* [caffeine#get()](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/LoadingCache.html#get(K))

gateway-ha/pom.xml
oneonestar: Done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
oneonestar: > Then how does it impact the testing, do we still need the function setExternalUrlForQueryId or those tests should be modified.

`@VisibleForTesting setExternalUrlForQueryId()` still is a quick and dirty way to do test, by setting cache explicitly instead of load the value through `CacheLoader`.

gateway-ha/pom.xml
```diff
@@ -22,11 +22,12 @@
...
 
         <!-- dependency versions -->
-        <dep.flyway.version>11.20.0</dep.flyway.version>
+        <dep.flyway.version>11.8.2</dep.flyway.version>
         <dep.jeasy.version>4.1.0</dep.jeasy.version>
-        <dep.mockito.version>5.21.0</dep.mockito.version>
+        <dep.mockito.version>5.18.0</dep.mockito.version>
         <dep.okhttp3.version>5.3.2</dep.okhttp3.version>
         <dep.trino.version>474</dep.trino.version>
+        <dep.caffeine.version>3.2.3</dep.caffeine.version>
```
ebyhr: We usually avoid defining a property if it's only used in a single place. Could you inline this version?

oneonestar: Done.

gateway-ha/pom.xml
ebyhr: @vishalya We usually avoid defining a property if it's only used in a single place.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
ebyhr: The `setExternalUrlForQueryId()` method becomes unused in the product code after this change.

Perhaps we could add another constructor to the `BaseRoutingManager` class that exposes the `LoadingCache`, and modify the cache object in tests. That is a common pattern we test in Trino projects.

ebyhr (approved):

mosabua:
In the dev sync on 21 Jan we agreed to merge this after the Trino Gateway 17 release to reduce immediate risk.

mosabua:
@oneonestar this is merged now so you can potentially proceed on #783 .. one thing I am wondering though .. does this PR need a release notes entry after all .. it should improve caching reliability right?

oneonestar:
> does this PR need a release notes entry after all .. it should improve caching reliability right?

It’s unlikely to have any noticeable effect on users.

PR #809, opened 2025-12-12
Check Health status for all clusters regradless of activation

felicity3786:
## Description
Health checks should run for all clusters (active and inactive), so UI TrinoStatus is always up-to-date irrespective of activation state.

Because the health check is lightweight and inexpensive, it can safely be applied across all backends. This improves the accuracy and observability of backend health while clearly separating health status from activation state.

This change does not affect routing behavior—queries continue to be routed only to clusters that are both healthy and active.

## Release notes

```markdown
* Check health status for all clusters regardless of activation status.
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -375,8 +375,7 @@ it is marked as UNHEALTHY, and the Trino Gateway stops routing requests to it.
 It is important to distinguish TrinoStatus from the active/inactive
 state of a backend. The active/inactive state indicates whether a backend is
 manually turned on or off, whereas TrinoStatus is programmatically determined
-by the health check process. Health checks are only performed on backends
-that are marked as active.
+by the health check process.
 
 See [TrinoStatus](routing-rules.md#trinostatus) for more details on 
 what each Trino status means.


oneonestar:
LGTM. Could you squash the commits?

mosabua:
Fyi @oneonestar .. you can use the GitHub UI to squash and merge .. just make sure you update the commit message and get rid of the PR number

vaultah:
In environments that use blue-green zero downtime deployment strategy for Trino clusters, there are usually two trino-gateway backends ("blue" and "green") for a single routing group, and only one backend is active at the time while the other is deactivated and fully offline.

One effect of this change is that the cluster monitoring process now also checks the intentionally offline backends, and generates error logs -- with full tracebacks too -- when it can't reach them. Backend deactivation was previously enough to remove the noise, and it's surprising that this behaviour is enabled by default, with no configuration option to disable it.

vaultah:
Another interesting consequence related to zero-downtime deployment (or environments with many inactive backends in general), is that offline backends can delay monitoring of active backends.

`ActiveClusterMonitor` uses `DEFAULT_THREAD_POOL_SIZE = 20` threads, and if the total number of backends is greater than that, health checks begin to queue. For metrics-based checks, connection-establishment failures time out after 5 seconds by default, while request timeouts default to 5 minutes. As a result, many unreachable backends can lengthen the monitor cycle, and because results are only published after the entire cycle finishes, active backend status updates are delayed as well.

PR #810, opened 2025-12-13
[Darft WIP]Add Best-Effort Routing Fallback When All Backends in Default Group Are Unhealthy

felicity3786:
## Description
Today, routing only selects from backends that are marked HEALTHY.
If all backends in a routing group are marked UNHEALTHY, the router throws an exception:
`Number of active backends found zero`
This behavior is correct in most cases. However, in real-world environments, health checks can be temporarily flaky due to factors such as transient network issues, endpoint slowness, or bugs in health check discovery. In these scenarios, false-positive outages can cause the gateway to reject queries even though a backend may still be able to serve traffic.
If every backend in a routing group is unhealthy, the query will ultimately fail whether it is routed or not. Failing earlier at the gateway provides no additional safety in this case and can unnecessarily reduce availability.

This PR introduces a new configuration flag:
`RoutingConfiguration.bestEffortRouting`
When best-effort routing is enabled and all active backends in the routing group are marked UNHEALTHY, the router will still select one active backend as a last-resort fallback, instead of immediately throwing an exception.

Best-effort routing only applies to the default routing group. Customized routing groups continue to enforce strict HEALTHY-only routing. This ensures the behavior acts strictly as a safety net for the final fallback group, rather than weakening routing guarantees across all groups. The default value is false, preserving full backward compatibility.

New unit tests and corresponding docs.

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
@@ -113,15 +113,10 @@ public abstract class BaseRoutingManager
     @Override
     public ProxyBackendConfiguration provideBackendConfiguration(String routingGroup, String user)
     {
-        List<ProxyBackendConfiguration> activeBackends = gatewayBackendManager.getActiveBackends(routingGroup);
-        List<ProxyBackendConfiguration> healthyBackends = activeBackends.stream()
+        List<ProxyBackendConfiguration> backends = gatewayBackendManager.getActiveBackends(routingGroup).stream()
                 .filter(backEnd -> isBackendHealthy(backEnd.getName()))
                 .toList();
-        // If no healthy backends in group, optionally route among all active backends when enabled
-        List<ProxyBackendConfiguration> candidates = !healthyBackends.isEmpty()
-                ? healthyBackends
-                : (bestEffortRouting ? activeBackends : healthyBackends);
-        return selectBackend(candidates, user).orElseGet(() -> provideDefaultBackendConfiguration(user));
+        return selectBackend(backends, user).orElseGet(() -> provideDefaultBackendConfiguration(user));
     }
 
     /**
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestBestEffortRouting.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestBestEffortRouting.java
@@ -21,12 +21,11 @@ import org.junit.jupiter.api.Test;
 
 import static io.trino.gateway.ha.TestingJdbcConnectionManager.createTestingJdbcConnectionManager;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 final class TestBestEffortRouting
 {
     @Test
-    void testBestEffortRoutingEnabled_SelectsActiveWhenAllUnhealthyInGroup()
+    void testBestEffortRoutingEnabledAllUnhealthy()
     {
         JdbcConnectionManager connectionManager = createTestingJdbcConnectionManager();
         RoutingConfiguration routingConfiguration = new RoutingConfiguration();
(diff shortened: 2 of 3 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -92,10 +94,16 @@ public void setRoutingGroupForQueryId(String queryId, String routingGroup)
...
+        List<ProxyBackendConfiguration> healthyDefaults = activeDefaults.stream()
                 .filter(backEnd -> isBackendHealthy(backEnd.getName()))
                 .toList();
-        return selectBackend(backends, user).orElseThrow(() -> new IllegalStateException("Number of active backends found zero"));
+        // If no healthy defaults, optionally route among all active defaults when enabled
+        List<ProxyBackendConfiguration> candidates = !healthyDefaults.isEmpty()
+                ? healthyDefaults
+                : (bestEffortRouting ? activeDefaults : healthyDefaults);
+        return selectBackend(candidates, user)
+                .orElseThrow(() -> new IllegalStateException("Number of active backends found zero"));
```
Peiyingy, line 106: I'm a bit concerned with the readability of the ternary branch. Shall we change it to something like this ⬇️  We can also benefit from different exception messages

```java
List<ProxyBackendConfiguration> activeDefaults = gatewayBackendManager.getActiveDefaultBackends();
List<ProxyBackendConfiguration> healthyDefaults = activeDefaults.stream()
        .filter(backEnd -> isBackendHealthy(backEnd.getName()))
        .toList();

if (bestEffortRouting && healthyDefaults.isEmpty()) {
    return selectBackend(activeDefaults, user)
            .orElseThrow(() -> new IllegalStateException(
                    "No active default backend found under best-effort routing"));
}

return selectBackend(healthyDefaults, user)
        .orElseThrow(() -> new IllegalStateException(
                "No healthy default backend found"));
```

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestBestEffortRouting.java
Peiyingy, line 1: I'm not sure if this new standalone test class is necessary. Shall we integrate these tests in `TestStochasticRoutingManager`? At least we can create some common test util methods for both of them.

gateway-ha/src/main/java/io/trino/gateway/ha/config/RoutingConfiguration.java
```diff
@@ -54,4 +58,14 @@ public void setDefaultRoutingGroup(String defaultRoutingGroup)
...
+
+    public boolean isBestEffortRouting()
+    {
+        return bestEffortRouting;
+    }
+
+    public void setBestEffortRouting(boolean bestEffortRouting)
+    {
+        this.bestEffortRouting = bestEffortRouting;
+    }
```
xkrogen, line 70: Similar to how we are introducing `strictRouting` in https://github.com/trinodb/trino-gateway/pull/804, I am wondering if this config should be a per-rule basis instead of global. We can think of each routing rule as having a varying level of "strictness", from least to most:
* Strict: Route _only_ to the target group
* Default: Fall back to default group if target group is all unhealthy
* Best Effort: Fall back to unhealthy clusters if all of target group + default group are unhealthy

xkrogen: cc @Peiyingy thoughts?

Peiyingy: Different levels of strictness makes sense. However, if we move that to routing rule definition, is won't cover the default routing case at all. Maybe we can keep this global config, and allow per-rule config to override it? WDYT?

xkrogen: I wonder if we should move it so that the default rule is specified in the rule configs along the other rules, instead of being treated separately? Kind of like resource group definitions in Trino -- where the last selector is typically a "catch all" that you fall back to if no other rules match ([example](https://trino.io/docs/current/admin/resource-groups.html#selector-rules:~:text=%7B%0A%20%20%20%20%20%20%22group%22%3A%20%22global.adhoc.other.%24%7BUSER%7D%22%0A%20%20%20%20%7D)). Not sure if that totally makes sense for routing rules, though -- I forget all the places that the default is used.

Global config with override makes sense to me as well, if we make it a `strictnessLevel` config (same as the per-rule config)

felicity3786: > Maybe we can keep this global config, and allow per-rule config to override it?

Not sure I understand correctly, the best effort will not make query choose a unhealthy target within the customized group, it only apply to adhoc/default group. What do we need the override for?

If we set best effort at global level, and the rule targeted group has no healthy cluster:

- without `strictRouting`,  query goes to adhoc, may fail if adhocs are also unhealthy.
- with `strictRouting`, query fail at target group unhealthy clusters.

if we override the best effort per this rule, I think the result is gonna be the same?

But I agree that thinking in terms of routing “strictness levels” (STRICT / DEFAULT / BEST_EFFORT) is the right long-term model, and that it maps well to how we want routing behavior to evolve.
As for moving the rule to same fallback as resource group, I don't think MVEL has any blocker on this maybe it will add one more default rule to traverse every time we inspect a rule. Maybe we can align more on this and make a separate PR if needed since that remodels the rule behavior a bit,  then come back and generalize the strictness.

Peiyingy: > if we override the best effort per this rule, I think the result is gonna be the same?

Thanks for clarifying that. Yes that would be the same here. I guess in rules, we can still decide that when there's fail back to default, do we want to make it `bestEffort` or not. Say globally that's disabled, but for this rule, if it falls back to defaults, and defaults are unhealthy, we still want it to fall back to unhealthy clusters. I'm not sure if this granularity is necessary, though.

Peiyingy: > I wonder if we should move it so that the default rule is specified in the rule configs along the other rules, instead of being treated separately?

Having a implicit default routing rule makes sense to me. I don't know if it's necessary to specify that.

I think under the current design, we've already handled all possible fallbacks:

1. If there are no available active healthy cluster for the selected routing group, fallback to ActiveDefaultBackends (ie. active healthy cluster in defaultRoutingGroup).
-> Handled by `StrictRouting`

2. If there's no  If there are no available active healthy cluster for the default routing group, fallback to unhealthy ones
-> Handled by `BestEffortRouting`

I agree we should unify the naming, but the current code design makes sense to me.

mosabua:
I think this is still reasonable idea and worth pursuing. If you agree and are willing to continue please reopen and update the PR @felicity3786

PR #815, opened 2026-01-02
Add Jaeger Tracing to query runner

ebyhr:
## Description

Screenshot 2026-01-02 at 10 13 05

## Release notes

### DIFF
--- a/gateway-ha/config.yaml
+++ b/gateway-ha/config.yaml
@@ -1,6 +1,9 @@
 serverConfig:
   node.environment: test
   http-server.http.port: 8080
+  tracing.enabled: true
+  otel.exporter.endpoint: http://localhost:4318/v1/traces
+  otel.exporter.protocol: http/protobuf
 
 routingRules:
   rulesEngineEnabled: False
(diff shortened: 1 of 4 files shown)


ebyhr:
@Chaho12 Yes, tracing helps verifying HTTP requests.

wendigo (approved):

mosabua:
We should reword commit message since this is about OpenTelemetry ..Jaeger is just one app that supports the format.

Also add to the docs..

ebyhr:
> We should reword commit message since this is about OpenTelemetry ..Jaeger is just one app that supports the format.

There is no reason to hide the app in the commit message. 

> Also add to the docs..

I don't think we need to add the doc. It's just a sidecar providing tracing.

Chaho12:
I also thought it would be nice for newbies to have doc or some sort of annotations clarifying that this option is intended for testing.

mosabua:
So just to clarify .. this does NOT affect normal Trino Gateway use since this is only wired into the test class right? Also .. how does this relate to the OpenTelemetry support we get from Airlift that you seem to be using with the tracing enabled property .. I don't think this is really documented yet so we should follow up on this PR. And commit message is still kinda misleading imho.

mosabua:
I think it would be good to add some docs about using this in https://trinodb.github.io/trino-gateway/development/

PR #817, opened 2026-01-02
Use `@Inject` instead of manual initialization

ebyhr:
## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
@@ -13,6 +13,19 @@
  */
 package io.trino.gateway.ha.security.util;
 
+import com.google.common.collect.ImmutableList;
+import com.google.inject.Inject;
+import io.trino.gateway.ha.config.HaGatewayConfiguration;
+import io.trino.gateway.ha.security.ApiAuthenticator;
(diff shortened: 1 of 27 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
```diff
@@ -30,9 +43,40 @@ public class ChainedAuthFilter
...
+            HaGatewayConfiguration config,
+            Authorizer authorizer)
     {
-        this.filters = requireNonNull(filters);
+        requireNonNull(authorizationManager, "authorizationManager is null");
+        requireNonNull(authorizer, "authorizer is null");
+
+        ImmutableList.Builder<ContainerRequestFilter> authFilters = ImmutableList.builder();
+        String defaultType = config.getAuthentication().getDefaultType();
+        if (oauthManager != null) {
```
vishalya, line 59: `config.getAuthentication()` could be null.

wendigo (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
ebyhr: See https://github.com/trinodb/trino-gateway/pull/817/changes#diff-f3357e75489fbd2e79087d91f13f8adb879bbbf289dd40788a779f53e2a1b90cR95-R97

PR #901, opened 2026-02-09
Improve database cache and tests

oneonestar:
## Description
Follow up comments in #783 

## Release notes

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
@@ -45,7 +45,6 @@ public class HaGatewayManager
 
     private final GatewayBackendDao dao;
     private final String defaultRoutingGroup;
-    private final boolean cacheEnabled;
     private final LoadingCache<Object, List<GatewayBackend>> backendCache;
 
     private final CounterStat backendLookupSuccesses = new CounterStat();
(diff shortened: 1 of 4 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/HaGatewayManager.java
```diff
@@ -121,9 +121,7 @@ private List<GatewayBackend> getAllBackendsInternal()
                 throw new RuntimeException("Failed to load backends from database to cache", e);
             }
         }
-        else {
-            return fetchAllBackends(ALL_BACKEND_CACHE_KEY);
-        }
+        return fetchAllBackends(ALL_BACKEND_CACHE_KEY);
```
ebyhr: I really dislike having a `boolean cacheEnabled` flag and a nullable `backendCache` field.

Would it be possible to use a no-op Caffeine cache instead, such as `Caffeine#maximumSize(0)`, when caching is disabled?

oneonestar: Updated.

oneonestar:
> For example, Trino1 is deactivated on gateway1, but not on gateway2 for a short period of time, so that all the queries routed to gateway2 would all fail, leading to an overall unstable system.

@Peiyingy Is this your use case?

1. `Trino1` is deactivated on `gateway1`
2. `Trino1` is shut down
3. `gateway2` continues routing traffic to `Trino1` until the cache is refreshed

I think you can set `refreshAfterWrite` to a short interval, such as `1s`, and then delay shutting down `Trino1` for a few seconds to allow the cache to refresh on all gateways. You can also set `expireAfterWrite` to a few hours to handle potential DB outages.

This will not handle 100% of cases, since cache refreshes only occur when a new request comes in and the refresh is asynchronous. If a gateway receives requests at a low frequency, the refresh could be delayed. However, I think combining a delayed shutdown with a short `refreshAfterWrite` should be sufficient to cover most cases.

Chaho12:
@oneonestar @Peiyingy 
 How about a way to sync/flush cache among gateways when there is activate/deactivate cluster event?

Peiyingy:
> This will not handle 100% of cases, since cache refreshes only occur when a new request comes in and the refresh is asynchronous. If a gateway receives requests at a low frequency, the refresh could be delayed. However, I think combining a delayed shutdown with a short refreshAfterWrite should be sufficient to cover most cases.

Yeah, that would be the issue. Say we are in a system with multiple gateway instances because the QPS is high, and we have a high reliability requirement; then this caching issue can easily breach the 99% availability SLO.

> How about a way to sync/flush cache among gateways when there is activate/deactivate cluster event?

That requires introducing communication between gateway instances. While DB was originally the SOT we used to sync between instances, introducing another communication to solve this stale cache issue seems like overkill.

Also, activate/deactivate is just an example I made here. In reality, any DB change (delete cluster, routing group change, etc.) would cause this inconsistent routing behavior.

oneonestar:
> Yeah, that would be the issue. Say we are in a system with multiple gateway instances because the QPS is high, and we have a high reliability requirement; then this caching issue can easily breach the 99% availability SLO.

When QPS is high and queries are evenly routed across different gateway instances, cache refreshes are triggered reliably.
The issue arises only when QPS is low, as delayed cache refreshes can then cause problems.

Peiyingy:
> > Yeah, that would be the issue. Say we are in a system with multiple gateway instances because the QPS is high, and we have a high reliability requirement; then this caching issue can easily breach the 99% availability SLO.
> 
> When QPS is high and queries are evenly routed across different gateway instances, cache refreshes are triggered reliably. The issue arises only when QPS is low, as delayed cache refreshes can then cause problems.

I think the issue here is, when the QPS is really high, even though the cache refresh would be trigger reliably, dozens of queries that are routed during those several seconds to the wrong cluster would encounter error.

If the OSS community doesn't have enough resources for this feature, I can contribute a PR on it sometime later.

Issue #920, opened 2026-02-16
infoUri broken likely due to eager evaluation failure on /ui/query.html


zeckem19:
RoutingTargetHandler.resolveRouting() uses Optional.orElse(getRoutingTargetResponse(request)) https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L83 which eagerly evaluates the fallback even when the query ID is found in the cache. If getRoutingTargetResponse() throws (e.g. no backends match the routing rules), the exception propagates and the request fails with a 500, despite the query ID being successfully resolved.

Steps to reproduce:

1. Configure routing rules that require a specific header (e.g. X-Trino-DC) to determine a routing group, with a catch-all rule that routes to a non-existent group (e.g. drop) when the header is missing
2. Submit a query through the gateway with the required headers — query succeeds and query ID is cached
3. Open the infoUri in a browser: GET /ui/query.html?{queryId} — this request has no X-Trino-DC header

zeckem19:
continued in #922

PR #921, opened 2026-02-16
Use orElseGet for lazy evaluation in RoutingTargetHandler

zeckem19:
Fixes https://github.com/trinodb/trino-gateway/issues/920

## Description
`RoutingTargetHandler.resolveRouting()` uses `Optional.orElse(getRoutingTargetResponse(request))`  which eagerly evaluates the fallback even when the query ID is found in the cache. If getRoutingTargetResponse() throws (e.g. no backends match the routing rules), the exception propagates and the request fails with a 500, despite the query ID being successfully resolved.

## Release notes

Fix eager evaluation in RoutingTargetHandler

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -80,7 +80,7 @@ public class RoutingTargetHandler
             return new RoutingTargetResponse(
                     new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request), externalUrl),
                     request);
-        }).orElse(getRoutingTargetResponse(request));
+        }).orElseGet(() -> getRoutingTargetResponse(request));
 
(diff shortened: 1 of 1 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -80,7 +80,7 @@ public RoutingTargetResponse resolveRouting(HttpServletRequest request)
             return new RoutingTargetResponse(
                     new RoutingDestination(routingGroup, cluster, buildUriWithNewCluster(cluster, request), externalUrl),
                     request);
-        }).orElse(getRoutingTargetResponse(request));
+        }).orElseGet(() -> getRoutingTargetResponse(request));
```
ebyhr, line 83: Is it possible to add a test? Let me know if the test is too complicated.

ebyhr (commented):
Could you please submit CLA?

zeckem19:
Superseded by #922 which includes the same fix plus a test case.

PR #877, opened 2026-01-27
Add distributed cache for horizontal scaling

hpopuri2:
---
  ##Add Valkey Distributed Cache for Horizontal Scaling

 ##Summary

  This PR implements distributed caching using Valkey to enable horizontal scaling of Trino Gateway. Multiple gateway instances can now share query metadata through a distributed cache layer, ensuring consistent query routing across all instances.

 ##Motivation

  Currently, Trino Gateway uses local Guava caches that are not shared between instances. In multi-instance deployments, this can lead to:
  - Inconsistent query routing when requests hit different gateway instances
  - Cache misses requiring expensive database lookups
  - Inability to leverage cache across horizontally scaled deployments

  This implementation addresses these limitations while maintaining backward compatibility and graceful degradation.

  ##Architecture

  3-Tier Caching Strategy

  Request Flow:
  1. L1 Cache (Local Guava) → ~1ms
    - Hit: Return immediately
    - Miss: Check L2
  2. L2 Cache (Valkey Distributed) → ~5ms
    - Hit: Populate L1, return
    - Miss: Check L3
  3. L3 Cache (PostgreSQL Database) → ~50ms
    - Found: Populate L2 + L1, return
    - Not Found: Search backends via HTTP (~200ms)

  Cache Keys

  Three values are cached for each query:
  - trino:query:backend:{queryId} - Backend URL for query routing
  - trino:query:routing_group:{queryId} - Routing group assignment
  - trino:query:external_url:{queryId} - External URL for query access

  All keys use configurable TTL (default 30 minutes / 1800 seconds).

  ##Implementation Details

  Core Components

  ValkeyConfiguration (gateway-ha/src/main/java/io/trino/gateway/ha/config/ValkeyConfiguration.java)
  - 9 configurable parameters with sensible defaults
  - Input validation (port range, positive values)
  - Convention over Configuration - only enabled, host, and port required
  - Fixed: cacheTtlSeconds parameter now properly used (was previously hardcoded)

  Cache Interface (gateway-ha/src/main/java/io/trino/gateway/ha/cache/Cache.java)
  - Generic caching abstraction: get(), set(), invalidate(), isEnabled()
  - Implementation-agnostic design (name describes contract, not implementation)
  - Enables future alternative implementations (e.g., Redis Cluster, Memcached)
  - Located in dedicated io.trino.gateway.ha.cache package for better organization

[...]

PR #923, opened 2026-02-17
Refactor routing logics and remove unused lookups

oneonestar:
## Description
Refactor the code to separate Query ID/Cookie-based routing from new query submission routing.

Remove the unnecessary `getExternalUrlForQueryId` and `getRoutingGroupForQueryId` methods, along with their related DAOs and caches. The `externalUrl` and `routingGroup` fields are stored for newly submitted queries as before.

`getExternalUrlForQueryId` and `getRoutingGroupForQueryId` are only used when a Query ID exists. However, their results are never persisted back to the database. We only store query history for newly submitted queries:
[https://github.com/trinodb/trino-gateway/blob/5bf67f35776b51a1113653b3333ce9bb520e81e2/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java#L190-L196](https://github.com/trinodb/trino-gateway/blob/5bf67f35776b51a1113653b3333ce9bb520e81e2/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java#L190-L196)

For newly submitted queries, `externalUrl` and `routingGroup` are stored here, so this refactoring won't affect the existing behavoir:
[https://github.com/trinodb/trino-gateway/blob/5bf67f35776b51a1113653b3333ce9bb520e81e2/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L94-L114](https://github.com/trinodb/trino-gateway/blob/5bf67f35776b51a1113653b3333ce9bb520e81e2/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L94-L114)

## Release notes


hpopuri2:
My Understanding of This PR (different cases i thought)

  1. POST Request (First-time Query Execution)

  - Caches backend URL only in local Caffeine cache
  - Stores full query history (backend, routingGroup, externalUrl) to database as before
  - No change to query history persistence

  2. Follow-up Requests

  - Only need backend URL to route correctly
  - routingGroup and externalUrl were looked up but never actually used
  - Follow-up requests don't record to query history (not POST requests)
  - Safe to remove these unnecessary lookups

  3. Multi-Gateway Scenario

  - Gateway B retrieves backend URL only from database
  - Routes to correct backend cluster
  - Doesn't need routingGroup/externalUrl for routing
  - Safe because routing depends solely on backend URL

  4. Same proxyTo, Different Routing Groups

  - Multiple routing groups can point to same backend URL
  - Follow-up routing uses backend URL only, not routing group
  - Safe because routing group isn't used in routing decisions


LGTM

vishalya:
Looks like the caches - `queryIdRoutingGroupCache` and `queryIdExternalUrlCache` are safe to be removed.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java
```diff
@@ -62,9 +62,7 @@ private ProxyUtils() {}
 
     public static Optional<String> extractQueryIdIfPresent(
             HttpServletRequest request,
-            List<String> statementPaths,
-            boolean requestAnalyserClientsUseV2Format,
```
andythsu, line 66: It looks like we used to use `requestAnalyserClientsUseV2Format` and `requestAnalyserMaxBodySize`  to create `TrinoQueryProprties` class but since now we are storing it in request object we just never cleaned it up. https://github.com/trinodb/trino-gateway/blob/abe47c278b8cc8d9c0691fd27abb082292e5c9cf/gateway-ha/src/main/java/io/trino/gateway/ha/handler/ProxyUtils.java#L77

gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java
andythsu: in here, the `routingDestination.routingGroup()` would point to the `defaultRoutingGroup`, which would be wrong

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -62,28 +60,35 @@ public RoutingTargetHandler(
...
-        logRewrite(routingTargetResponse.routingDestination().clusterHost(), request);
-        return routingTargetResponse;
+        Optional<String> queryId = extractQueryIdIfPresent(request, statementPaths);
+        Optional<String> previousCluster;
+        Optional<RoutingTargetResponse> routingTargetResponse;
+        if (queryId.isPresent()) {
+            // Query ID based routing
+            previousCluster = queryId.map(routingManager::findBackendForQueryId);
+            routingTargetResponse = previousCluster.map(cluster -> new RoutingTargetResponse(
+                    new RoutingDestination(defaultRoutingGroup, cluster, buildUriWithNewCluster(cluster, request), cluster),
```
andythsu: The whole point of these two seemingly unnecessary cache is to not use `defaultRoutingGroup` because it might not be the right routing group for subsequent queries. 

For example:
first query: `POST/ /v1/statement` goes to cluster `foo_cluster` in routing group `foo_routing_group` with cluster url `https://foo-cluster`
second query: `GET/ ...`, instead of using `foo_cluster` it would use whatever value is set to `defaultRoutingGroup` as routing group. 

It may not break the existing workflow as it doesn't use routing group to route. It's routed with cluster uri, but the information being carried around with subsequent GETs is wrong.

oneonestar: > second query: GET/ ...

I assume you're referring to the `nextUri` from the first request. The subsequent requests must route to the exact Trino cluster in order to process the query. Routing the subsequent requests to other Trino clusters in the same routing group doesn't make sense.

andythsu: yes I understand. This PR still routes the subsequent requests to the same Trino cluster as it's using `clusterUri`. However, `routingDestination.routingGroup` now points to `defaultRoutingGroup` instead of the actual routing group for the first POST request

For example, it would end up with
1st request: `POST/ /v1/statement` `RoutingDestination` class holds `routingGroup = foo, clusterUri = foo.com`
2nd request: `GET/ /...` `RoutingDestination` class holds `routingGroup = <defaultRoutingGroup>, clusterUri = foo.com`
it doesn't break the flow as it only uses `clusterUri`, but now `routingGroup` is wrong.

Please correct me if I'm wrong

oneonestar: Instead of passing an unused `defaultRoutingGroup` around, I updated the code to use empty string to indicate it's not used.

oneonestar:
CI error related to #966

andythsu (approved):
🚀 not the best solution, but the benefits outweighs having to maintain two caches just for this info

oneonestar:
Rebase to latest main branch.
@vishalya Would you like to take a look?

hpopuri2:
@oneonestar  can we push a ahead with this ?
so that it unblocks this PR https://github.com/trinodb/trino-gateway/pull/877

Chaho12:
@oneonestar can you fix conflicts?

PR #934, opened 2026-02-20
Standardize backend endpoints and clarify routing-rules auth scope in docs

koszti:
## Description

Standardize backend endpoints and clarify routing-rules auth scope in docs

Updates `docs/gateway-api.md` to consistently document backend management via `/gateway/backend/...` endpoints and to explicitly mark `/webapp/updateRoutingRules` as an `ADMIN` endpoint.

## Release notes

### DIFF
--- a/docs/gateway-api.md
+++ b/docs/gateway-api.md
@@ -10,10 +10,11 @@ The example commands are for a Trino Gateway server running at
 If there are duplicate `proxyTo` URLs in the configuration, the `Name` in the
 **Query History** page of the UI might not show correctly.
 
-## Add or update a Trino cluster
+## Create a Trino cluster
 
 ```shell
-curl -X POST http://localhost:8080/entity?entityType=GATEWAY_BACKEND \
+curl -X POST http://localhost:8080/gateway/backend/modify/add \
+ -H "Content-Type: application/json" \
  -d '{  "name": "trino-3",
         "proxyTo": "http://localhost:8083",
         "active": true,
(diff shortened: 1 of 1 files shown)


docs/gateway-api.md
```diff
@@ -10,10 +10,11 @@ The example commands are for a Trino Gateway server running at
 If there are duplicate `proxyTo` URLs in the configuration, the `Name` in the
 **Query History** page of the UI might not show correctly.
 
-## Add or update a Trino cluster
+## Create a Trino cluster
 
 ```shell
-curl -X POST http://localhost:8080/entity?entityType=GATEWAY_BACKEND \
+curl -X POST http://localhost:8080/gateway/backend/modify/add \
```
mosabua, line 16: Thats a very different endpoint .. was the docs just plain wrong?

mosabua (commented):
In principal this is good. Just wondering if we need to clarify the reason / origin for the `gateway` context and name, and same fot the `backend` and `modify` .. and then of course add and update as verbs/actions.

docs/gateway-api.md
koszti: both do the same logical thing and are working fine. The main difference is that the currently documented `/entity?entityType=GATEWAY_BACKEND` endpoint requires the `ADMIN` role, so it’s not really meant for pure API clients and isn’t in sync with the other cluster CRUD operations. It seems more geared toward the UI (?).

mosabua: Fair enough .. keeping it consistent is good.

mosabua (approved):

PR #938, opened 2026-02-23
Add API specific endpoints for routing rules

koszti:
## Description

Add API specific routing rules endpoints under the API-role `/gateway` surface:
* `GET /gateway/routing-rules/all`
* `POST /gateway/routing-rules/modify/update`

**Reason**: API clients use API-role credentials and routing-rules management should be available on the API-role endpoint surface (not only under `/webapp` which requires `ADMIN` privileges).

Keep existing /webapp routing-rules endpoints for compatibility (no webapp breakage):
* `GET /webapp/getRoutingRules`
* `POST /webapp/updateRoutingRules`

Share routing-rules logic via RoutingRulesResourceHandler so `/gateway` and `/webapp` use the same implementation.

Update docs/gateway-api.md with the new API-role endpoints and compatibility note.

* Named list endpoint as `/gateway/routing-rules/all` for consistency with `/gateway/backend/all`.
* Updated TestRoutingAPI to cover `/gateway endpoints` and `/webapp/getRoutingRules compatibility`.

Side note: in the future we could converge on one API surface and treat the web UI as another API client.

## Release notes

* Add API-role routing rules endpoints at `/gateway/routing-rules/all` and `/gateway/routing-rules/modify/update`, while retaining existing `/webapp` routing rules endpoints for compatibility.

### DIFF
--- a/docs/gateway-api.md
+++ b/docs/gateway-api.md
@@ -124,20 +124,25 @@ curl -X POST http://localhost:8080/gateway/backend/activate/trino-2
 
 ## Update routing rules
 
-This endpoint is part of the `/webapp` endpoint family and requires the
-`ADMIN` role.
-
 The API can be used to programmatically update the routing rules. Rule are
 updated based on the rule name. Storage of the rules must use a writeable file
-and the configuration 'rulesType: FILE'.
+and the configuration `rulesType: FILE`.
 
 For this feature to work with multiple replicas of the Trino Gateway, you must
 provide a shared storage that supports file locking for the routing rules file.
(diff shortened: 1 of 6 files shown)


Chaho12:
I like this idea — what do you guys think?
Btw, do we normally support backward compatibility, or do we just go with breaking changes and update the API spec?
I think we usually do the latter for Trino, right?

mosabua:
I am reopening for @koszti as we still want this fix to go in. Ping us for review please

mosabua:
We should get this in .. for now I reopened so it doesnt get lost

PR #945, opened 2026-03-02
Add spooling protocol support and limitations to docs

koszti:
## Description

Add spooling protocol support and limitations to docs

Spooling protocol is not fully supported. Document limitations and current behavior.

## Release notes


mosabua:
Just a first feedback .. we should make a general page about the client protocol instead and then have a spooling protocol section on that page .. just like in the trino docs. The client protocol content can be fairly simple for now .. just so we get the right framework in place.

mosabua:
We still want to document this, but as I mentioned in a general client protocol page like in Trino. Also note that this might change as soon as we merge https://github.com/trinodb/trino-gateway/pull/1142

PR #960, opened 2026-03-13
Fix corruption of binary responses caused by UTF-8 decoding in proxy response

SelvaRM04:
## Description

The gateway currently reads backend responses using UTF-8 string decoding:

`new String(response.getInputStream().readNBytes((int) responseSize.toBytes()), StandardCharsets.UTF_8)`

This will work for all text responses. But for binary assets like image/png, decoding binary content as UTF-8 modifies the byte stream, because invalid UTF-8 sequences are replaced or re-encoded. As a result, gateway returns corrupted/broken image to client.

One observable example is `'/ui/assets/logo.png`'. This returns a broken image.

## Solution

Avoiding the conversion of response body to a UTF-8 string fixes this issue. Forward the raw bytes return by the backend ensures images are transmitted without modification.

`response.getInputStream().readNBytes((int) responseSize.toBytes())`

## Result
Before this change:

Gateway 3

Gateway 5

After this change:
`/ui/assets/logo.png` loads correctly through gateway

Gateway 4

Gateway 6

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyResponseHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyResponseHandler.java
@@ -23,7 +23,6 @@ import io.trino.gateway.ha.config.ProxyResponseConfiguration;
 import io.trino.gateway.proxyserver.ProxyResponseHandler.ProxyResponse;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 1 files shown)


SelvaRM04:
Submitted the signed copy of CLA. Please review and add my name to contributors.

oneonestar:
@cla-bot check

Chaho12:
@coderabbitai full review

Chaho12 (commented):
I agree with the fix, where raw is the correct, but can you add a test in TestProxyRequestHandler to verify that a binary response body passes through without corruption?

Chaho12:
Let me know if you are ready for review

vishalya (commented):
The fix looks good, but the downstream logging code in `ProxyRequestHandler::recordBackendForQueryId` needs to be updated.

Could you add a utility function that checks the `CONTENT_TYPE` header and logs the request body as text when appropriate? For all other cases, including binary data, it should log only the body length.

mosabua:
We should still pursue this.

mosabua:
Can you address comments @SelvaRM04 ?

Chaho12:
@SelvaRM04 knock knock~

Issue #787, opened 2025-10-30
Unhandled NullPointerException if request is for an unknown query ID


vaultah:
trino-gateway may fail with an unhandled NPE if it's unable to find the routing group for a particular query ID.

```
jakarta.servlet.ServletException: com.google.common.util.concurrent.UncheckedExecutionException: java.lang.NullPointerException
    ...
    at java.base/java.lang.Thread.run(Thread.java:1447)
Caused by: com.google.common.util.concurrent.UncheckedExecutionException: java.lang.NullPointerException
    at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2074)
    at com.google.common.cache.LocalCache.get(LocalCache.java:3986)
    at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:4007)
    at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:4946)
    at io.trino.gateway.ha.router.RoutingManager.findRoutingGroupForQueryId(RoutingManager.java:194)
    at java.base/java.util.Optional.map(Optional.java:260)
    at io.trino.gateway.ha.handler.RoutingTargetHandler.lambda$resolveRouting$0(RoutingTargetHandler.java:86)
    at java.base/java.util.Optional.map(Optional.java:260)
    at io.trino.gateway.ha.handler.RoutingTargetHandler.resolveRouting(RoutingTargetHandler.java:85)
    at io.trino.gateway.proxyserver.RouteToBackendResource.postHandler(RouteToBackendResource.java:67)
    ...
    ... 37 more
Caused by: java.lang.NullPointerException
    at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:902)
    at com.google.common.cache.LocalCache.put(LocalCache.java:4212)
    at com.google.common.cache.LocalCache$LocalManualCache.put(LocalCache.java:4872)
    at io.trino.gateway.ha.router.RoutingManager.setRoutingGroupForQueryId(RoutingManager.java:122)
    at io.trino.gateway.ha.router.RoutingManager.findRoutingGroupForUnknownQueryId(RoutingManager.java:264)
    at io.trino.gateway.ha.router.RoutingManager$2.load(RoutingManager.java:85)
    at io.trino.gateway.ha.router.RoutingManager$2.load(RoutingManager.java:81)
    at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3551)
[...]

PR #963, opened 2026-03-15
Avoid null cache writes for query routing mappings

panjwanirahul:
Guard query-id cache writes in routing manager against null/empty values to prevent runtime failures when query history rows have been removed by retention.

  ### What changed

  - Updated `BaseRoutingManager`:
    - `setBackendForQueryId(...)`: skip empty `queryId`; invalidate entry if backend is null/empty.
    - `setRoutingGroupForQueryId(...)`: skip empty `queryId`; invalidate entry if routing group is null/empty.
    - `setExternalUrlForQueryId(...)`: skip empty `queryId`; invalidate entry if `externalUrl` is null.
  - Added regression tests in `TestRoutingManagerExternalUrlCache`:
    - `testNullExternalUrlDoesNotThrowAndFallsBackToQueryHistory`
    - `testNullQueryIdDoesNotThrowForExternalUrlCacheSet`

  ### Why

  When query history retention deletes older records, lookup for `externalUrl` can return null.
  The previous code attempted to cache that null value, which caused an NPE in the cache path.
  The new behavior treats null/empty mappings as “not cacheable” and preserves normal fallback behavior.

  - Related issue: https://github.com/trinodb/trino-gateway/issues/787
  - Behavior impact:
    - Fixes NPE in query routing cache update path.
    - Does not change successful routing behavior for valid mappings.
    - For missing mappings, cache entry is invalidated/skipped and existing fallback logic continues.
  - Local validation:
    - `TestRoutingManagerExternalUrlCache` passed.
    - Routing-focused suite passed (`TestRoutingManagerExternalUrlCache`, `TestRoutingGroupSelector`, `TestRoutingRulesManager`, `TestPathFilter`, `TestTrinoRequestUser`, `TestTrinoQueryProperties`).

  ## Release notes

  ```markdown
  * Fix a NullPointerException in query routing cache updates when retained query history entries are removed and external URL lookup returns null.

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
@@ -17,7 +17,6 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
-import com.google.common.base.Strings;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
 import io.trino.gateway.ha.clustermonitor.TrinoStatus;
@@ -39,6 +38,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
 
 /**
  * This class performs health check, stats counts for each backend and provides a backend given
@@ -76,12 +77,30 @@ public abstract class BaseRoutingManager
     @Override
     public void setBackendForQueryId(String queryId, String backend)
     {
+        if (isNullOrEmpty(queryId)) {
+            log.debug("Skipping backend cache update for empty queryId");
+            return;
+        }
+        if (isNullOrEmpty(backend)) {
+            queryIdBackendCache.invalidate(queryId);
+            log.debug("Invalidated backend cache for queryId [%s] due to empty backend", queryId);
+            return;
+        }
         queryIdBackendCache.put(queryId, backend);
     }
 
     @Override
     public void setRoutingGroupForQueryId(String queryId, String routingGroup)
     {
+        if (isNullOrEmpty(queryId)) {
+            log.debug("Skipping routing group cache update for empty queryId");
+            return;
+        }
+        if (isNullOrEmpty(routingGroup)) {
+            queryIdRoutingGroupCache.invalidate(queryId);
+            log.debug("Invalidated routing group cache for queryId [%s] due to empty routing group", queryId);
+            return;
+        }
         queryIdRoutingGroupCache.put(queryId, routingGroup);
     }
 
@@ -177,6 +196,15 @@ public abstract class BaseRoutingManager
     @Override
     public void setExternalUrlForQueryId(String queryId, String externalUrl)
     {
+        if (isNullOrEmpty(queryId)) {
+            log.debug("Skipping externalUrl cache update for empty queryId");
+            return;
+        }
+        if (isNullOrEmpty(externalUrl)) {
+            queryIdExternalUrlCache.invalidate(queryId);
+            log.debug("Invalidated externalUrl cache for queryId [%s] due to empty externalUrl", queryId);
+            return;
+        }
         queryIdExternalUrlCache.put(queryId, externalUrl);
     }
 
@@ -185,7 +213,7 @@ public abstract class BaseRoutingManager
     {
         String backend;
         backend = queryHistoryManager.getBackendForQueryId(queryId);
-        if (Strings.isNullOrEmpty(backend)) {
+        if (isNullOrEmpty(backend)) {
             log.debug("Unable to find backend mapping for [%s]. Searching for suitable backend", queryId);
             backend = searchAllBackendForQuery(queryId);
         }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
@@ -17,7 +17,6 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
-import com.google.common.base.Strings;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.clustermonitor.ClusterStats;
 import io.trino.gateway.ha.clustermonitor.TrinoStatus;
@@ -39,6 +38,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
 
 /**
  * This class performs health check, stats counts for each backend and provides a backend given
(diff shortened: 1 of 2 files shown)


rahulpanjwani17:
@ebyhr Hey, can you please check this pull request and also re run the cla action? i just got the mail that my cla is signed
Thanks!!

ebyhr:
@cla-bot check

rahulpanjwani17:
@ebyhr thank you so much for re running the cla bot, can you also review this pull request?

kbhatianr (approved):

panjwanirahul:
@ebyhr Hey!! Please get a review on this? It would be great if you could please check this.

panjwanirahul:
@mosabua hello, can i please get a review on this? Pending since a long time, thank you so much!

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -76,12 +76,30 @@ public BaseRoutingManager(GatewayBackendManager gatewayBackendManager, QueryHist
     @Override
     public void setBackendForQueryId(String queryId, String backend)
     {
+        if (Strings.isNullOrEmpty(queryId)) {
```
ebyhr: Let's import `isNullOrEmpty` directly.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -76,12 +76,30 @@ public BaseRoutingManager(GatewayBackendManager gatewayBackendManager, QueryHist
...
+            log.warn("Skipping backend cache update for empty queryId");
```
ebyhr: Why is the log level warning? What is the expected action when Trino Gateway admin sees this message? 

> Guard query-id cache writes in routing manager against null/empty values to prevent runtime failures when query history rows have been removed by retention.

The PR description indicates this is a normal scenario. The log level should be debug if so.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
panjwanirahul: done

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
panjwanirahul: makes sense, thanks for the review, made the changes @ebyhr

panjwanirahul:
@ebyhr can you please check this again?

panjwanirahul:
@mosabua can i please get a review here? Thanks much

panjwanirahul:
@mosabua @ebyhr can you folks please check this or assign someone? Thanks much!

vishalya (commented):
**My understanding of the flow is this:** The trino-gateway first checks its in-memory cache for the query, then falls back to the query history. But if both caches have expired or don't contain the query id, the router would have to reach out directly to the actual clusters to discover where that query is currently executing. So I'm puzzled—how do we actually end up in a situation where both caches miss?
Could you explain why deleting the history would stop the search for the actual backend?

panjwanirahul:
> **My understanding of the flow is this:** The trino-gateway first checks its in-memory cache for the query, then falls back to the query history. But if both caches have expired or don't contain the query id, the router would have to reach out directly to the actual clusters to discover where that query is currently executing. So I'm puzzled—how do we actually end up in a situation where both caches miss? Could you explain why deleting the history would stop the search for the actual backend?

@vishalya 
The discovery logic is actually a bit separate from this fix. This is really just a safety fix for an NPE—caches crash if you try to .put() a null value, which is what happens once history is purged by retention. This PR just adds a guard so the gateway doesn't crash when it hits a null mapping, allowing the normal fallbacks to continue.

vishalya:
I’m trying to trace the crash path and figure out whether we’re masking the real issue somewhere along the way.

ebyhr (commented):

panjwanirahul:
@ebyhr done, please check now

panjwanirahul:
@ebyhr @mosabua please check this

Chaho12:
@panjwanirahul Can you tell us how we can reproduce the situation? Steps that makes this NPE happen?

panjwanirahul:
@Chaho12 Thanks for asking — here are reproducible steps. I also rebased this PR onto current main; the issue is still present.

### Why this happens
Caffeine (and previously Guava) does not allow null cache values. When query-history retention has deleted the row, lookups can return null. Writing that null into the query-id caches throws NPE.

To @vishalya’s earlier question: backend lookup has an extra fallback (`searchAllBackendForQuery` probes clusters). Routing-group / external-url lookup do **not** — they only consult query history. So a history miss returns null; attempting to cache that null is what crashes. This PR does not change discovery behavior; it only prevents caching null/empty mappings so the request path can continue with the existing fallbacks (`defaultRoutingGroup` / cluster host).

### Minimal unit-test reproduction (current main)
```java
routingManager.setRoutingGroupForQueryId("20260315_query", null);
// throws NullPointerException from Caffeine put

routingManager.setExternalUrlForQueryId("20260315_query", null);
// throws NullPointerException from Caffeine put
```

On this branch those calls no longer throw; the cache entry is invalidated/skipped and later `find*` falls back to query history as usual.

### End-to-end reproduction
1. Run gateway with query history enabled and a short retention window (or manually delete the history row).
2. Submit a statement so a `queryId` is recorded in history + in-memory cache.
3. Delete the history row for that `queryId` (or wait for retention), and clear/expire the in-memory cache (restart gateway, or wait for the 30-minute access expiry).
4. Send a follow-up request that includes that `queryId` (e.g. statement status / next URI), or any path that calls `set*ForQueryId(..., null)` after a history miss.
5. On unfixed main: NPE from cache `put`.
6. With this PR: no NPE; routing continues via existing fallbacks.

### Local validation on latest main
- Unfixed main: null `setRoutingGroupForQueryId` / `setExternalUrlForQueryId` throw NPE.
- This branch: `TestRoutingManagerExternalUrlCache` passes, including null/empty external URL guards.

Happy to address any further review comments.

Chaho12:
@panjwanirahul Are you sure this happens in latest version? or at least 18+? 

I need to E2E test, but when we migrated guava cache to caffeine in this [PR](https://github.com/trinodb/trino-gateway/pull/808), the issue should have been resolved as `setRoutingGroupForQueryId(queryId, routingGroup);` and ` setExternalUrlForQueryId(queryId, externalUrl);` is gone now. 

In 17 or older version [code](https://github.com/trinodb/trino-gateway/blob/17/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java#L249)

```java
protected String findRoutingGroupForUnknownQueryId(String queryId)
{
    String routingGroup = queryHistoryManager.getRoutingGroupForQueryId(queryId);
    setRoutingGroupForQueryId(queryId, routingGroup);   // put(queryId, null)
    return routingGroup;
}
```
The Guava CacheLoader wrote the loaded value back into the cache. When query history had no row for the query id, routingGroup was null, so LocalCache.put hit Preconditions.checkNotNull and threw, surfacing as UncheckedExecutionException. That matches the trace exactly. 

Also, the read path already falls back to the default routing group
- https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java#L76-L79
```java
                    .orElse(defaultRoutingGroup);
            String externalUrl = queryId.map(routingManager::findExternalUrlForQueryId)
                    .orElse(cluster);
```

Two independent guards:
1. No query id → getRoutingTargetResponse(), which forces defaultRoutingGroup when the selector returns null/empty 
2. Query id present but its history row was purged → the mapper returns null, Optional.map collapses that to an empty Optional, and orElse supplies defaultRoutingGroup / cluster.

panjwanirahul:
@Chaho12 You’re right — thanks for digging into this.

I re-checked against current main. The #787 crash came from the Guava-era loader path that called `setRoutingGroupForQueryId(queryId, routingGroup)` with a null history result. After #808 that `set*` call is gone from `findRoutingGroupForUnknownQueryId` / `findExternalUrlForUnknownQueryId`, and the read path already falls back via `Optional.map(...).orElse(defaultRoutingGroup/cluster)`.

So the original NPE from purged query history should already be resolved on 18+ / latest main. The unit-level `set*(id, null)` NPE I showed is only a residual risk if a caller passes null into the setters; it isn’t the #787 end-to-end path anymore.

We can close this PR, and we can close #787 as fixed by #808

Chaho12:
@panjwanirahul Thx for cross-checking.

PR #942, opened 2026-02-26
Fix dashboard line chart field name mismatch with API response

kimsehwan96:
## Description

Because of mismatched field name in dashboard api response the query distribution graph doens't rendered. (related with https://github.com/trinodb/trino-gateway/pull/805)

스크린샷 2026-02-26 오후 1 16 15

스크린샷 2026-02-26 오후 1 16 20

The getDistribution api call reponse has `data.lineChart[].epochMillis` filed but the frontend application expect `timestamp` to use it as x axis data. 

After change frontend application's timestamp filed into `epochMillis` and the graph working properly.

스크린샷 2026-02-26 오후 1 17 28

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required, with the following suggested text:

```markdown
* Fix some things.
```

PR #970, opened 2026-03-19
Fix bug with Linechart

kjsbot:
## Description
Fix line chart not rendering bug due to field name mismatch between frontend and backend.

The backend used epochMillis while the frontend used timestamp field name, updated the frontend to be consistent with the backend.
Screenshot 2026-03-18 at 7 20 38 PM

## Release notes

```markdown
* Fix some things.
```


ebyhr:
Closing as duplicate of #942

PR #1016, opened 2026-04-15
Expand CI runs to Java 26 and 27

mosabua:
## Description

Adopting approach from Trino.

Probably have to update docs. Will check after CI passes.

## Release notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java-version: [25]
+        include: 
+          - { java-version: 25 }
+          - { java-version: 26 }
+          - { java-version: 27-ea }
     timeout-minutes: 30
     steps:
       - name: Checkout source
@@ -33,6 +36,7 @@ jobs:
         uses: actions/setup-java@v5
         with:
           java-version: ${{ matrix.java-version }}
+          cache: 'maven'
           distribution: 'temurin'
       - name: Build with Maven
         run: $MAVEN clean verify
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,8 +21,9 @@ Consider the following requirements for your Trino Gateway installation.
 
 ### Java
 
-Trino Gateway requires a Java 25 runtime. Older versions of Java can not be
-used. Newer versions might work but are not tested.
+Trino Gateway requires a Java 25 runtime. Lower versions of Java can not be
+used. Higher versions are verified in our CI pipelines, but you may run into
+unexpected issues. 
 
 Verify the Java version on your system with `java -version`.
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -21,8 +21,8 @@ or execute the following command:
 
 #### Locally
 
-This project requires Java 25. Note that higher version of Java have not been
-verified and may run into unexpected issues.
+This project requires Java 25. Higher version of Java are verified in our CI
+pipeline, but may run into unexpected issues.
 
 Run `./mvnw clean install` to build `trino-gateway`. VM options required for
 compilation and testing are specified in `.mvn/jvm.config`.


.github/workflows/ci.yml
```diff
@@ -24,7 +24,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java-version: [25]
+        include: 
+          - { java-version: 25, cache: 'true', cleanup-node: true }
+          - { java-version: 26, cache: 'restore', cleanup-node: true }
```
vishalya: Are java versions 26, 27 also targeted in trino pipeline? What if a PR fails for 26, 27 but works for 25.

mosabua: This is the same as in Trino .. it should never fail on a newer version and if it does thats a good thing because it will show us what we have to fix early rather than have a whole bunch of issues when we eventually go to 27..

docs/installation.md
```diff
@@ -21,8 +21,9 @@ Consider the following requirements for your Trino Gateway installation.
 
 ### Java
 
-Trino Gateway requires a Java 25 runtime. Older versions of Java can not be
-used. Newer versions might work but are not tested.
+Trino Gateway requires a Java 25 runtime. Lower versions of Java can not be
+used. Higher versions are verified in our CI pipelines, but you may may run into
```
ebyhr:
```suggestion
used. Higher versions are verified in our CI pipelines, but you may run into
```

mosabua: Fixed .. thx

.github/workflows/ci.yml
```diff
@@ -24,7 +24,10 @@ jobs:
...
+          - { java-version: 27-ea, cache: 'restore', cleanup-node: true }
```
ebyhr: Which job uses `cache` and `cleanup-node`?

mosabua: From what I understand the matrix will start with 25 and then 26 and 27e will use the restored cache. This is the setting in Trino .. do you have more context @wendigo ?

ebyhr: I mean `matrix.java-version` is used at L38, but others aren't used.

mosabua: Looking closer this might be wrong and we should just use `cache: maven` and no cleanup-node stuff .. or there is some other magic in Trino build that changes this and I would have to add that if that makes sense for Trino Gateway

mosabua: I added cache use now - see https://github.com/actions/setup-java?tab=readme-ov-file#caching-maven-dependencies

and removed the cleanup .. this might be the simplest approach and still help with performance.

.github/workflows/ci.yml
```diff
@@ -24,7 +24,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java-version: [25]
+        include: 
+          - { java-version: 25, cache: 'maven' }
+          - { java-version: 26, cache: 'maven' }
+          - { java-version: 27-ea, cache: 'maven' }
```
ebyhr: Why do we need to add `cache` to the matrix if all the values are the same?

mosabua: I guess we dont .. I hardcoded it now

.github/workflows/ci.yml
```diff
@@ -24,7 +24,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java-version: [25]
```
mosabua, line 27: we could maybe even use this array syntax for multiple values but I think the current setup is clearer to read

.github/workflows/ci.yml
```diff
@@ -33,6 +36,7 @@ jobs:
         uses: actions/setup-java@v5
         with:
           java-version: ${{ matrix.java-version }}
+          cache: 'maven'
           distribution: 'temurin'
       - name: Build with Maven
         run: $MAVEN clean verify
```
ebyhr, line 42: In Trino, newer Java versions skip running tests. It only verifies successful build.

ebyhr (changes requested):

.github/workflows/ci.yml
mosabua: I think in Trino Gateway we can run tests since the builds dont take that long and its a good check for runtime with the actual Java version as compared to just building the artifacts.

Chaho12 (dismissed):
LGTM 👍

mosabua:
Any concerns with merging @ebyhr ?

.github/workflows/ci.yml
ebyhr: Please don't close unresolved comments. 

For instance, when there is a broken test, 3 jobs will be red. Contributors need to check all jobs for single test failure. This is really bad developer experience.

mosabua: I don't think so. You just fix the test locally for 25 and push.. and let CI do the job. It's better than not running the tests

mosabua: Also it the only way to see test failures specific to newer version and that value outweighs the drawbacks

mosabua:
Can I get another set of reviews here @trinodb/maintainers-gateway

wendigo (approved):

.github/workflows/ci.yml
Chaho12: If it succeeds in 25, but fails on 26/27, do you expect it to be fixed? Or at least research on why it fails (e.g change in deprecated function name or a bug in 27+)

ebyhr: @mosabua I'm not sure why you merged this PR without answering the question above.

mosabua: I already answered and we would do the same as in Trino .. if it really fails in 26 or 27, and passed in 25, thats exactly why we want to have these separate tests. We can then check what the reason could be (although it would be VERY rare to occur since it would indicate an incompatibility in the JDK or one of the dependencies).

PR #1020, opened 2026-04-17
Separate CI job into Maven and Docker

ebyhr:
## Release notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,10 @@ defaults:
 env:
   # Use Maven wrapper from repo with Maven version and other configs
   MAVEN: ./mvnw -B
+  MAVEN_FAST_INSTALL: "-B -V -T 1C -DskipTests -Dmaven.source.skip=true -Dair.check.skip-all"
 
 jobs:
-  build:
+  maven-build:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -36,6 +37,20 @@ jobs:
           distribution: 'temurin'
       - name: Build with Maven
         run: $MAVEN clean verify
+
+  docker-build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v6
+      - name: Install required Java distribution
+        uses: actions/setup-java@v5
+        with:
+          java-version: 25
+          distribution: 'temurin'
+      - name: Maven Build
+        run: $MAVEN install ${MAVEN_FAST_INSTALL}
       - uses: docker/setup-qemu-action@v4
         with:
           platforms: amd64,arm64,ppc64le


ebyhr:
This change improves the entire CI speed from 20m to 10m.

.github/workflows/ci.yml
```diff
@@ -17,9 +17,10 @@ defaults:
 env:
   # Use Maven wrapper from repo with Maven version and other configs
   MAVEN: ./mvnw -B
+  MAVEN_FAST_INSTALL: "-B -V -T 1C -DskipTests -Dmaven.source.skip=true -Dair.check.skip-all"
```
mosabua, line 20: This means we are always skipping tests and checks. Why is that a good thing? I mean .. yes its faster of course but the whole point of CI is to run the tests.

mosabua (dismissed):
This provides a lot less feedback for any problems. We are currently  not in a time crunch on the CI runs for this project so we should not make it worse just to make it faster. I could be convinced to do this in a default build and then follow a second build that verifies more to get fast failure maybe. Or do this behavior only for Java 26 and 27 .. but still also have a way to run the tests. But as this stands now we should not merge this.

.github/workflows/ci.yml
ebyhr: @mosabua ? maven-build still runs tests. This option is for docker-build.

ebyhr:
> This provides a lot less feedback for any problems.

I believe you are misunderstanding this change. 

> We are currently not in a time crunch on the CI runs for this project so we should not make it worse just to make it faster.

Strongly disagreed. Faster CI is always better. 20m is too slow for this small project in my opinion. 

> Or do this behavior only for Java 26 and 27 .. but still also have a way to run the tests. 

This change is unrelated to Java 26 / 27 since we can have 3 maven-build & 3 docker-build if needed.

.github/workflows/ci.yml
```diff
@@ -36,6 +37,20 @@ jobs:
...
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v6
+      - name: Install required Java distribution
+        uses: actions/setup-java@v5
+        with:
+          java-version: 25
+          distribution: 'temurin'
+      - name: Maven Build
+        run: $MAVEN install ${MAVEN_FAST_INSTALL}
```
Chaho12, line 53: @mosabua I think you missed out this point. 
He split build -> maven-build & docker-build, where docker-build is light. 
Since we do maven build and then run docker build, we do not really have to do run test while building docker image.

Chaho12 (commented):
Is there a reason why we don't re-use jar created in maven-build job?
Won't there be side-effects since artifact is different?

ebyhr:
> Is there a reason why we don't re-use jar created in maven-build job?

How do we reuse jar? They run in parallel. 

> Won't there be side-effects since artifact is different?

It should be no side-effect.

Chaho12:
> How do we reuse jar? They run in parallel.

Well I was thinking artifact sharing + needs option to run safely, but as you mention this runs in sequential for maven and docker build. 

But if we expand maven build for Java 26 and 27, this could run in parallel.

Chaho12 (approved):
Since you don't expect any side-effects, we should split CI to save time.

.github/workflows/ci.yml
mosabua: Yeah.. I had a closer look now. Sorry

.github/workflows/ci.yml
mosabua: Yeah .. I didnt see that.. all good.

Issue #269, opened 2024-03-01
Support certificate authentication on Trino


oneonestar:
(Feature request)
Trino Gateway currently doesn't work if the backend Trino is using certificate authentication.

There are a few ways that I could think of to solve this issue. Discussion is needed before moving to implementation.

1. TLS Termination at Trino Proxy.
   Perform authentication on Trino Proxy. Forward the request to coordinator base on the routing policies.
   Perform authentication between Trino Proxy and Coordinator.
   Allow Trino Proxy to "impersonate" the user.
2. Return HTTP 307 Redirect
   First decrypt the request, select coordinator base on routing policies.
   Return HTTP 307 Redirect to the selected coordinator.
   Client directly connect to the Trino coordinator.
3. TLS Passthrough
   (I don't think this will work since hostname wont't match)

>  Discussion about automatic retry of queries in the future, will need storage of submitted queries and state, and more, will work on that at a later stage

Looking at the dev sync's meeting notes, method 1 seems a reasonable choice.

electrum:
In `trino-proxy`, we handled this by [creating a JWT bearer token using the certificate principal ](https://github.com/trinodb/trino/blob/e9adf9414a4c6153fc40ecda601da29085351ae3/service/trino-proxy/src/main/java/io/trino/proxy/ProxyResource.java#L249-L263). `trino-proxy` is [configured](https://github.com/trinodb/trino/blob/master/service/trino-proxy/src/main/java/io/trino/proxy/JwtHandlerConfig.java) with a key to generate the JWT that is trusted by the Trino coordinator. In other words, we convert the TLS certificate into a bearer token.

Another potential option is to support the `Client-Certificate` header from [RFC 9440](https://www.rfc-editor.org/rfc/rfc9440.html). This would require support in Trino. It's similar to the standard `Forwarded` in that it requires opt-in configuration to be trusted, as we do with the `http-server.process-forwarded` config.

wendigo:
@electrum for the RFC 9440 do we need to touch jetty at all? We have an access to the headers in the `public Identity authenticate(ContainerRequestContext request)` so we can get the cert chain there

oneonestar:
> for the RFC 9440 do we need to touch jetty at all?
No. ContainerRequestContext is enough.

I think the flow would be:
1. During TLS handshake, Gateway ask for client certificate
2. Client present and proof they have the client certificate.
3. Gateway verify the certificate is valid
[...]

PR #1022, opened 2026-04-18
Add client certificate authentication via JWT bridge

InfinityHub123:
## Description

Teams running Trino at scale often reach a point where a gateway becomes necessary for load balancing, routing, and cluster management. The natural solution is Trino Gateway, but for teams that authenticate to Trino using client certificate, introducing the gateway breaks their setup and forces them to downgrade their infrastructure security level.

The issue is that mTLS is a transport-layer mechanism between the client and the server it connects to directly. Once Trino Gateway sits in the middle, it terminates the TLS connection. The backend Trino clusters never see the client certificate, so any mTLS-based authentication configured on the clusters stops working. Operators are forced to either weaken cluster authentication or give up on using the gateway - neither is acceptable.

## Solution

This PR adds an optional mTLS to JWT bridge inside Trino Gateway. The idea is to shift authentication from mTLS on the clusters to JWT, while keeping mTLS as the client-facing authentication mechanism at the gateway boundary.

When the bridge is enabled:

- Clients continue to connect to the gateway using their client certificate (no changes on the client side)
- The gateway extracts the client identity from the certificate, maps it to a trino user using configurable rules, and mints a short-lived signed JWT
- The JWT is injected as `Authorization: Bearer` into the request forwarded to the backend cluster
- The backend Trino clusters are configured to accept JWT authentication from the gateway's signing key

This keeps the overall security level intact: clients still need a valid certificate to get through the gateway, and clusters still validate a cryptographically signed token before accepting a request. It also leaves the door open for direct cluster access when needed. for example, if an operator wants to connect directly to a specific cluster for debugging, they can do so since the cluster still supports client certificate authentication.

### Configuration

- Copy the user mapping rules from your Trino cluster configuration to the gateway. Whatever pattern or file you used in Trino's Certificate authentication setup to map certificate subjects to usernames should be used also here.

```yaml
clientCertificateJwtAuthentication:
  jwtSigningKeyPair:
    privateKey: /path/to/private-key.pem   # RSA or EC
    publicKey: /path/to/public-key.pem
  jwtIssuer: trino-gateway                 # optional
  jwtAudiences:                            # optional
    - trino
  jwtPrincipalClaim: preferred_username    # default: sub
  jwtKeyId: my-key-id                      # optional, for JWKS key rotation
  clientCertificateRequired: false         # optional, default: false
  overrideTrinoUser: false                 # optional, default: false

requestAnalyzerConfig:
  clientCertificateIdentityField: SUBJECT_DN       #  same field used on the clusters
  clientCertificateUserMappingPattern: CN=([^,]+),.*   # copy from Trino's user mapping config
  # or: clientCertificateUserMappingFile: /path/to/mapping.json
```

- Add JWT as an accepted authentication type on each backend Trino cluster and point it at the gateway's public key or JWKS endpoint. The clusters no longer need to handle mTLS from clients - they trust the JWT issued by the gateway instead.

```properties
# config.properties on each Trino cluster
http-server.authentication.type=JWT
http-server.authentication.jwt.key-file=/path/to/gateway-public-key.pem
# or use a JWKS endpoint:
# http-server.authentication.jwt.key-file=https://jwks-service/.well-known/jwks.json
```

### EC Key Support

As part of this change, `LbKeyProvider` was extended to support Elliptic Curve keys in addition to RSA. Previously the gateway only supported RSA key pairs for JWT signing (used by the form-based auth flow). The key loading logic was rewritten using BouncyCastle's `PEMParser`, which handles both RSA and EC PEM files transparently.

As a result, the `selfSignKeyPair` configuration fields were renamed from `privateKeyRsa`/`publicKeyRsa` to `privateKey`/`publicKey` to reflect that they are no longer RSA-specific.

## Changes in this PR

- Added the mTLS to JWT bridge: certificate identity extraction, user mapping (Trino Compatible), JWT minting, and header injection into forwarded requests
- Wired the bridge into the existing filter chain so the mapped certificate identity is available for routing decisions
- Extended `LbKeyProvider` to support EC keys alongside RSA using BouncyCastle, and renamed the `selfSignKeyPair` fields accordingly

I took inspiration from oneonestar's issue https://github.com/trinodb/trino-gateway/issues/269
## Release Notes

* [:warning: Breaking change:](#breaking) Rename `selfSignKeyPair` configuration fields `privateKeyRsa` and `publicKeyRsa` to `privateKey` and `publicKey`.
* Add client certificate authentication support via JWT bridge.
* Add support for EC keys.


InfinityHub123:
I've submited CLA a few weeks ago

InfinityHub123:
@ebyhr can you run the CLA bot again? I’ve received the verification email

mosabua:
@cla-bot check

InfinityHub123:
@vishalya @oneonestar can I please get a review here?

InfinityHub123:
@mosabua I want to advance this PR and would appreciate reviewers' feedback. Is someone available?

Chaho12:
@InfinityHub123 Can you first squash commits to 1 ~ 2

InfinityHub123:
@Chaho12 Done

InfinityHub123:
@mosabua @Chaho12 Is someone available?

InfinityHub123:
@ebyhr can I get a review?

InfinityHub123:
@mosabua my PR is ready for review. Would appreciate if someone could take a look.
I'll be available to address any feedback quickly.
Thanks

InfinityHub123:
@vishalya @oneonestar 
I would really like to get reviews for my PR, I opened it 2 months ago.
I'm receptive to feedback and willing to make the necessary changes quickly🙏

ebyhr:
@coderabbitai full review

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -129,14 +133,20 @@ private Optional<String> extractUserFromCookies(ContainerRequestContext requestC
...
         String header;
         header = requestContext.getHeaderString(TRINO_USER_HEADER_NAME);
         if (header != null) {
             return Optional.of(header);
         }
 
+        Optional<String> clientCertificateIdentity = extractUserFromClientCertificate(servletRequest, clientCertificateUserResolver);
+        if (clientCertificateIdentity.isPresent()) {
+            return clientCertificateIdentity;
+        }
```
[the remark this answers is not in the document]
InfinityHub123: Looked into this.  `extractUser()`'s header-first priority is pre-existing behavior, not something introduced here, and it's not actually a security gap for this feature.

 The JWT subject minted by the cert-bridge comes straight from `ClientCertificateJwtRequestAuthenticator.getOrCreateIdentity()`, which calls
  `ClientCertificateUserResolver.resolveMappedUser()` directly - it never goes through `TrinoRequestUser`/`X-Trino-User` at all. So even if a client sends
  `X-Trino-User: bob` while their cert maps to `alice`, the JWT Trino validates is always signed for `alice`. `bob` is just forwarded as a routing/display hint,
  same as it's always been for OAuth/basic/cookie auth. This is covered by `TestRouteToBackendResource.testPostHandlerDecoratesRoutedQueryRequestWithJwt`, which
  asserts exactly that divergence is expected.

  `TrinoRequestUser.getUser()` itself is only consumed for routing-group selection and query-history logging, not for authorization decisions - so reordering it
  wouldn't close a real trust gap, and rejecting on mismatch would break that test and change long-standing behavior shared by all auth methods. Leaving as-is.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
```diff
@@ -0,0 +1,168 @@
...
+    public HttpServletRequest authenticate(HttpServletRequest request)
+    {
+        if (!isAuthenticatedRequest(request)) {
+            return request;
+        }
+
+        Optional<String> identity = getOrCreateIdentity(request);
+        if (identity.isEmpty()) {
+            return request;
+        }
```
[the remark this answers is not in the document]
InfinityHub123: This one I'd also push back on. The bridge is explicitly documented and tested as an opt-in overlay: it mints a JWT only when a client cert is actually presented, and otherwise leaves the request untouched so other Trino-side auth methods (password, OAuth, LDAP, etc.) keep working. See testDecoratorSkipsMissingCertificate, which asserts the request passes through unmodified when no cert is found.

Whether a cert is mandatory is a TLS-listener concern (clientAuth NEED vs WANT), not something this authenticator should enforce. Forcing a 401 here would make it impossible to run cert-bridge auth alongside other auth methods, which is a requirement for us.

Also, the "bypass" framing doesn't hold up: a client can't forge a valid Authorization: Bearer <jwt> to impersonate the bridge, since Trino verifies the JWT signature against the gateway's actual signing key. A self-supplied header would just fail Trino's verification, same as if the bridge didn't exist. Leaving it as is

gateway-ha/src/main/java/io/trino/gateway/ha/config/HaGatewayConfiguration.java
```diff
@@ -201,6 +204,7 @@ public RequestAnalyzerConfig getRequestAnalyzerConfig()
 
     public void setRequestAnalyzerConfig(RequestAnalyzerConfig requestAnalyzerConfig)
     {
+        validateRequestAnalyzerConfig(requestAnalyzerConfig);
         this.requestAnalyzerConfig = requestAnalyzerConfig;
```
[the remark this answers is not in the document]
InfinityHub123: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/handler/HeaderOverrideRequestWrapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class HeaderOverrideRequestWrapper
+        extends HttpServletRequestWrapper
+{
+    private final Map<String, String> customHeaders;
+
+    public HeaderOverrideRequestWrapper(HttpServletRequest request, Map<String, String> customHeaders)
+    {
+        super(request);
+        this.customHeaders = customHeaders;
```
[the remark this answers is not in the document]
InfinityHub123: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private boolean isAuthenticatedRequest(HttpServletRequest request)
+    {
+        String requestUri = request.getRequestURI();
+        for (String authenticatedPath : authenticatedPaths) {
+            if (requestUri.startsWith(authenticatedPath)) {
+                return true;
+            }
+        }
+        return false;
```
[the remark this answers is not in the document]
InfinityHub123: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/security/UserMapping.java
```diff
@@ -0,0 +1,143 @@
...
+                return Optional.empty();
+            }
+            if (!allow) {
+                throw new UserMappingException("Principal is not allowed");
+            }
+            String result = matcher.replaceAll(user).trim();
+            if (result.isEmpty()) {
+                throw new UserMappingException("Principal matched, but mapped user is empty");
+            }
+            return Optional.of(userCase.transform(result));
```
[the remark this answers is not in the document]
InfinityHub123: Fixed

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbTokenUtil.java
```diff
@@ -44,40 +43,56 @@ void setup()
...
-        assertThat(LbTokenUtil.validateToken(idToken, lbKeyProvider.getRsaPublicKey(), jwt.getIssuer(), Optional.of(List.of("test.com", "test1.com")))).isTrue();
+        assertThat(LbTokenUtil.validateToken(idToken, lbKeyProvider.publicKey(), jwt.getIssuer(), Optional.of(ImmutableList.of("test.com", "test1.com")))).isTrue();
+    }
+
+    @Test
+    void testEcTokenValidationWithAudiences()
+    {
+        LbKeyProvider ecKeyProvider = new LbKeyProvider(new SelfSignKeyPairConfiguration(
+                "src/test/resources/auth/test_ec_private_key.pem",
+                "src/test/resources/auth/test_ec_public_key.pem"));
```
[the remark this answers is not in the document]
InfinityHub123: Fixed

gateway-ha/src/test/java/io/trino/gateway/ha/testing/TestingHttpServletRequest.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.Map;
+
+public class TestingHttpServletRequest
+        extends HttpServletRequestWrapper
+{
+    private final Map<String, Object> attributes = new HashMap<>();
+    private final Map<String, String> headers = new HashMap<>();
+    private final String requestUri;
```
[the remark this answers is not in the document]
InfinityHub123: Fixed

InfinityHub123:
Applied fixes for all valid coderabbitAI findings. Left explanatory comments on 2 findings I didn't apply (both were based on a misreading of how the cert-bridge auth actually works (see inline replies for details)).

InfinityHub123:
Hey @vishalya @oneonestar 
Could you please continue reviewing my PR?

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -60,10 +63,11 @@ public class TrinoRequestUser
 
     private final Optional<LoadingCache<String, UserInfo>> userInfoCache;
 
-    private TrinoRequestUser(ContainerRequestContext request, String userField, Optional<LoadingCache<String, UserInfo>> userInfoCache)
+    private TrinoRequestUser(ContainerRequestContext request, HttpServletRequest servletRequest, ClientCertificateUserResolver clientCertificateUserResolver, String userField, Optional<LoadingCache<String, UserInfo>> userInfoCache)
+            throws UserMappingException
     {
         this.userInfoCache = requireNonNull(userInfoCache);
-        user = extractUser(request, userField);
+        user = extractUser(request, servletRequest, clientCertificateUserResolver, userField);
```
vishalya, line 70: We should not create the cert resolver unconditionally.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -129,14 +133,20 @@ private Optional<String> extractUserFromCookies(ContainerRequestContext requestC
...
+    private Optional<String> extractUser(ContainerRequestContext requestContext, HttpServletRequest servletRequest, ClientCertificateUserResolver clientCertificateUserResolver, String userField)
+            throws UserMappingException
     {
         String header;
         header = requestContext.getHeaderString(TRINO_USER_HEADER_NAME);
         if (header != null) {
             return Optional.of(header);
         }
 
+        Optional<String> clientCertificateIdentity = extractUserFromClientCertificate(servletRequest, clientCertificateUserResolver);
```
vishalya, line 139: The user from the cert would take a precedence over regular user information coming from basic/authorization header. Do we really want that?

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
```diff
@@ -0,0 +1,168 @@
...
+            return cachedJwt.token();
+        }
+
+        CachedJwt jwt = createJwt(request, identity, now);
+        jwtCache.put(identity, jwt);
+        return jwt.token();
+    }
+
+    @SuppressWarnings("unused") // request reserved for future claims expansion
+    private CachedJwt createJwt(HttpServletRequest request, String identity, Instant now)
```
vishalya: Drop the request parameter, if we are NOT using it and remove the warning suppression above.

vishalya (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
InfinityHub123: Fixed. The single arg TrinoRequestUserProvider constructor was unconditionally creating a ClientCertificateUserResolver. Now it passes null instead, matching the opt-in gating QueryUserInfoParser already does.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
InfinityHub123: Fixed

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
InfinityHub123: Fixed

InfinityHub123:
@ebyhr I've fixed the issues coderabbitai found. Can I get a second review/approve?

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        String subjectDn = certificate.getSubjectX500Principal().getName();
+        if (identityField.equals(SUBJECT_DN_FIELD)) {
+            return Optional.of(subjectDn);
+        }
+
+        try {
+            LdapName ldapName = new LdapName(subjectDn);
+            return ldapName.getRdns().stream()
+                    .filter(rdn -> rdn.getType().equals(identityField))
```
Chaho12: It seems that RDN attribute types are case-insensitive per RFC 4514 / RFC 2253. 
For example, CN=John Doe is functionally equivalent to cn=John Doe or Cn=John Doe.

```suggestion
                    .filter(rdn -> rdn.getType().equalsIgnoreCase(identityField))
```

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java
```diff
@@ -55,7 +55,7 @@ public LbFormAuthManager(
                 .filter(entry -> entry.getValue() != null)
                 .collect(toImmutableMap(entry -> entry.getKey().toUpperCase(ENGLISH), Map.Entry::getValue));
 
-        if (configuration != null) {
+        if (configuration != null && configuration.getSelfSignKeyPair() != null) {
```
Chaho12: Why is this changed?

Before this PR, a form config missing selfSignKeyPair failed fast at construction (reads the null path and throws). 
But in this PR, construction succeeds, gateway boots clean, then the first user login NPEs with an opaque stack trace and no config hint. Mis-config moved from startup → runtime. Worse operator experience.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java
```diff
@@ -0,0 +1,63 @@
...
+                    .map(rdn -> String.valueOf(rdn.getValue()))
+                    .filter(value -> !value.isBlank())
+                    .findFirst();
+        }
+        catch (InvalidNameException e) {
+            throw new IllegalArgumentException("Failed to parse certificate subject DN: " + subjectDn, e);
```
Chaho12, line 74: I wasn't sure about this part so I ran through Claude Code. PTAL.

---

Failure path: LdapName can't parse subjectDn → IllegalArgumentException. Every caller catches only UserMappingException:
- getOrCreateIdentity (ClientCertificateJwtRequestAuthenticator.java:766-774) — catches UserMappingException only.
- QueryUserInfoParser.filter (:1339-1346) — same.
- TrinoRequestUser.TrinoRequestUserProvider.getInstance — catches UserMappingException → rethrows IllegalArgumentException; the extractor's IAE is a different, uncaught one.

So the IAE bubbles out of the JAX-RS resource → mapped to HTTP 500, and the response/log leaks the full subject DN.

Likelihood: low. X500Principal.getName() emits RFC 2253, which LdapName parses in almost all real certs. Trigger is exotic DN encodings only. It's a robustness/consistency gap, not a live bug — a cert that clears TLS should never yield a 500. All other identity failures already return 401; this one shouldn't be the exception.

Fix — single point, converts to the 401 path everyone already handles. In ClientCertificateUserResolver.resolveMappedUser (:900-902):

```java
String rawIdentity;
try {
    rawIdentity = ClientCertificateIdentityExtractor.extractIdentity(certificate.orElseThrow(), identityField)
            .orElseThrow(() -> new UserMappingException("Client certificate does not contain required identity field '%s'".formatted(identityField)));
}
catch (IllegalArgumentException e) {
    throw new UserMappingException(e.getMessage());
}
String mappedUser = userMapping.mapUser(rawIdentity);

**Review by vishalya — APPROVED** [review]

File: gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java (hunk 12 above)
Comment (InfinityHub123) [reply gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java]: Fixed

File: gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java (hunk 13 above)
Comment (InfinityHub123) [reply gateway-ha/src/main/java/io/trino/gateway/ha/security/LbFormAuthManager.java]: Fixed

File: gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java (hunk 14 above)
Comment (InfinityHub123) [reply gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java:74]: Fixed

**Comment by InfinityHub123** [comment]
@Chaho12 can you approve? I've fixed the issues you mentioned

File: gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java [hunk 15]
```diff
@@ -0,0 +1,63 @@
...
+    public static Optional<String> extractIdentity(X509Certificate certificate, String identityField)
+    {
+        String subjectDn = certificate.getSubjectX500Principal().getName();
+        if (identityField.equals(SUBJECT_DN_FIELD)) {
+            return Optional.of(subjectDn);
+        }
+
+        try {
+            LdapName ldapName = new LdapName(subjectDn);
+            return ldapName.getRdns().stream()
```
Chaho12: LdapName.getRdns could pick wrong CN on multi-RDN DN. 
Please add test `certificate("CN=alice,OU=data,CN=legacy,O=example")` to see if it returns Alice and not legacy.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -129,7 +133,8 @@ private Optional<String> extractUserFromCookies(ContainerRequestContext requestC
         });
     }
 
-    private Optional<String> extractUser(ContainerRequestContext requestContext, String userField)
+    private Optional<String> extractUser(ContainerRequestContext requestContext, HttpServletRequest servletRequest, ClientCertificateUserResolver clientCertificateUserResolver, String userField)
```
Chaho12, line 136: I think we should have a separate note in docs/security.md that forwarded X-Trino-User still permits impersonation unless the backend has impersonation rules; stock Trino with no access-control.properties allows it.

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
```diff
@@ -142,9 +147,24 @@ private Optional<String> extractUser(ContainerRequestContext requestContext, Str
             return user;
         }
 
+        Optional<String> clientCertificateIdentity = extractUserFromClientCertificate(servletRequest, clientCertificateUserResolver);
+        if (clientCertificateIdentity.isPresent()) {
+            return clientCertificateIdentity;
+        }
+
```
Chaho12: Cert identity must outrank X-Trino-User. Otherwise routing group, external router input, and query-history user are all client-controlled while the JWT says otherwise.

Move the cert block above the header block:
```java
        Optional<String> clientCertificateIdentity = extractUserFromClientCertificate(servletRequest, clientCertificateUserResolver);
        if (clientCertificateIdentity.isPresent()) {
            return clientCertificateIdentity;
        }

        String header = requestContext.getHeaderString(TRINO_USER_HEADER_NAME);
        if (header != null) {
            return Optional.of(header);
        }
```

---

Two consequences. They're different in kind, and only the second is this PR's problem.

Consequence 1 — backend impersonation. Not the PR's fault.

Client cert CN=alice, client also sends X-Trino-User: bob

Gateway → Backend
  Authorization: Bearer <sub=alice>
  X-Trino-User: bob

Backend: principal user = alice (from JWT + user mapping)
         session user    = bob  (from header)
         alice != bob → checkCanImpersonateUser(alice, bob)
           no access-control.properties → allowed → query runs as bob

Direct mTLS to Trino does exactly this too. Fix lives in backend config (file access control with impersonation rules), not in this PR. Just document it, because the PR body claims the bridge "keeps the overall security level intact" — true only if the operator has impersonation rules.

Consequence 2 — gateway-side. This one IS the PR's problem.

QueryUserInfoParser (PreMatching filter, runs before routing)
  → TrinoRequestUser.extractUser
      step 1: X-Trino-User present? → return "bob"   ← certificate never consulted
  → stored as TRINO_REQUEST_USER

TRINO_REQUEST_USER = "bob" then drives three things, all before the backend is contacted:

- FileBasedRoutingGroupSelector.java:67-68 — routing rules evaluate with trinoRequestUser in scope. A rule like condition: "trinoRequestUser.userExistsAndEquals(\"etl_admin\")" → routing group high-priority. alice picks whichever cluster she wants by setting a header.
- ExternalRoutingGroupSelector.java:146-154 — user is POSTed to your external routing service; its decision is made on a forged value.
- ProxyRequestHandler.java:191 — query history records bob. Audit trail is wrong, and wrong in the specific direction that frames an innocent user.

None of those three touch Trino. No backend access-control configuration can protect them. That's why the fix has to be the precedence reorder in TrinoRequestUser.java:140-153, not a docs note.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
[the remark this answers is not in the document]
Chaho12: Could this be configurable instead of hardcoded? As-is, an operator running cert-only (no password/OAuth on the gateway) has no way to set clientAuth to NEED — setBinding() overrides airlift's default and there's no config property for it.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateIdentityExtractor.java
InfinityHub123: Added

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
InfinityHub123: Added a note

gateway-ha/src/main/java/io/trino/gateway/ha/router/TrinoRequestUser.java
InfinityHub123: Fixed. Cert identity comes before x-trino-user, added the note about the impersonation

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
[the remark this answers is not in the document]
InfinityHub123: `NEED` isn't scoped to the query paths,  it's a connector-wide TLS setting, enforced at handshake time before Jetty even sees the request path, so it'd also lock out the UI/OAuth/admin API of **Trino Gateway itself**  on the same port, so we must keep using  `REQUESTED`. Added `clientCertificateRequired` on `clientCertificateJwtAuthentication` instead: when true, `authenticate()` 401s on `/v1/statement`/`/v1/query` if no cert identity resolved. Default `false`. Documented in `docs/security.md`.

InfinityHub123:
@Chaho12 addressed all issues, replied inline on each. Good catches, especially the routing-group/audit-trail angle on the precedence bug. Ready for another pass/approval.

mosabua:
@Chaho12 @vishalya - can we merge this ?

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
```diff
@@ -0,0 +1,172 @@
...
+        jwtAudiences = Optional.ofNullable(authConfig.getJwtAudiences()).filter(audiences -> !audiences.isEmpty());
+        jwtKeyId = Optional.ofNullable(authConfig.getJwtKeyId()).filter(keyId -> !keyId.isBlank());
+        keyProvider = new LbKeyProvider(requireNonNull(authConfig.getJwtSigningKeyPair(), "jwtSigningKeyPair is null"));
+        jwtCache = Caffeine.newBuilder()
+                .maximumSize(10000)
+                .expireAfterWrite(JWT_EXPIRATION_SECONDS - JWT_REUSE_SAFETY_WINDOW_SECONDS, TimeUnit.SECONDS)
+                .build();
+        authenticatedPaths = Stream.concat(configuration.getStatementPaths().stream(), Stream.of(V1_QUERY_PATH))
+                .distinct()
+                .collect(toImmutableList());
```
Chaho12: Is it ok to only set authenticated path to /v1/query? I mean the [PathFilter.isPathWhiteListed](https://github.com/trinodb/trino-gateway/blob/main/gateway-ha/src/main/java/io/trino/gateway/ha/router/PathFilter.java#L72-L82) has more paths.
So, if Cluster is set as JTW, this paths wouldn't have the AUTH so wouldn't it return 401?
- especially /v1/spoold/ is needed when using spooling protocol to fetch dtata.

If so, why don't we change so that it uses same source.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
```diff
@@ -0,0 +1,172 @@
...
+        jwtCache.put(identity, jwt);
+        return jwt.token();
+    }
+
+    private CachedJwt createJwt(String identity, Instant now)
+    {
+        Instant expiresAt = now.plusSeconds(JWT_EXPIRATION_SECONDS);
+        JWTCreator.Builder tokenBuilder = JWT.create()
+                .withIssuedAt(Date.from(now))
+                .withNotBefore(Date.from(now))
```
Chaho12: It seems that JWT allows skew of 0 seconds, so if gateway time is ahead of trino cluster, then it would return 401 error. why don't we allow some window for this? maybe 60 secs?

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
```diff
@@ -42,34 +47,63 @@ public class QueryUserInfoParser
...
-        RequestAnalyzerConfig requestAnalyzerConfig = new RequestAnalyzerConfig();
-        requestAnalyzerConfig.setTokenUserField(tokenUserField);
-        requestAnalyzerConfig.setOauthTokenInfoUrl(oauthTokenInfoUrl);
+        if (clientCertificateUserResolver != null && servletRequest != null) {
+            try {
+                clientCertificateUserResolver.resolveMappedUser(servletRequest);
+            }
+            catch (UserMappingException e) {
+                throw unauthorized(e.getMessage());
+            }
```
Chaho12: This feedback from AI seems valid. PTAL

---

QueryUserInfoParser.filter() runs for every path in the PathFilter whitelist and throws 401 whenever resolveMappedUser() fails. That includes /ui, /oauth2, /v1/info and /v1/node, which are not query-protocol traffic.

Since the TLS listener is REQUESTED, a browser may auto-present an unrelated client certificate. No certificate at all is handled fine (Optional.empty(), no throw), but a certificate that doesn't satisfy the mapping rules now blocks access to the proxied Trino Web UI. Restricting the rejection to the query-protocol paths — or letting ClientCertificateJwtRequestAuthenticator be the single place that rejects — would avoid that:
```java
catch (UserMappingException e) {
    if (isQueryProtocolPath(path)) {
        throw unauthorized(e.getMessage());
    }
    log.debug("Ignoring client certificate mapping failure on %s: %s", path, e.getMessage());
}

gateway-ha/src/main/java/io/trino/gateway/ha/config/ClientCertificateJwtAuthenticationConfiguration.java
Chaho12, line 1: Can you check this feedback by AI?

---

X-Trino-User passthrough splits identity between gateway and cluster

The bridge overrides Authorization only, so a client presenting alice's certificate together with X-Trino-User: bob reaches the cluster as principal alice / session user bob. The docs cover the impersonation angle honestly, which I appreciate — but there's a second effect worth calling out: TrinoRequestUser.extractUser() now prefers the certificate identity over the header, so the gateway's routing rules and query history record alice while the cluster records and authorizes bob. The same query is attributed to two different users across the two systems, which makes after-the-fact tracing hard.

Given HeaderOverrideRequestWrapper already supports multiple headers, an opt-in switch is cheap:
```java
// ClientCertificateJwtAuthenticationConfiguration
private boolean overrideTrinoUserHeader;

ImmutableMap.Builder<String, String> headers = ImmutableMap.<String, String>builder()
        .put(AUTHORIZATION, "Bearer " + jwt);
if (overrideTrinoUserHeader) {
    headers.put("X-Trino-User", identity.orElseThrow());

Chaho12 (commented):
The description suggests pointing clusters at https://gateway-host/.well-known/jwks.json, but I don't see a JWKS resource in this change — only jwtKeyId, which sets the kid header. Either add the endpoint or drop that line from the description, otherwise operators following it will get a 404.

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
InfinityHub123: Fixed, The bridge now uses `PathFilter.requiresBackendAuthentication()` instead of a hardcoded path list, so `/v1/statement`  `/v1/query`  `/v1/spooled`. Gateway ui/ general health paths stay exempt

gateway-ha/src/main/java/io/trino/gateway/ha/security/ClientCertificateJwtRequestAuthenticator.java
InfinityHub123: Fixed. JWTs now backdate `nbf` by 60 seconds to tolerate gateway/cluster clock skew

gateway-ha/src/main/java/io/trino/gateway/ha/security/QueryUserInfoParser.java
InfinityHub123: Fixed. `QueryUserInfoParser` now rejects certificate mapping failures only on paths that require backend authentication. For exempt paths like `/ui` `/oauth2` `/v1/info` and `/v1/node`, it ignores the bad cert and falls back to header/cookie identity.

gateway-ha/src/main/java/io/trino/gateway/ha/config/ClientCertificateJwtAuthenticationConfiguration.java
InfinityHub123: Addressed with a config option. By default the bridge preserves client-supplied `X-Trino-User` for existing impersonation workflows, but `overrideTrinoUser: true` now pins both `X-Trino-User` and `X-Trino-Original-User` to the certificate identity so gateway and backend identity stay aligned.

InfinityHub123:
@Chaho12 Thanks for the review. I addressed the remaining feedback, updated docs/tests, and fixed additional flaws I've found during the debugging. Ready for another review/approval.

InfinityHub123:
@Chaho12 gentle ping on this 🙂

Chaho12:
@InfinityHub123 Thx for the PR!

Chaho12:
Can we include this in 21 release note? @mosabua

mosabua:
Done.. still need that review of the release notes so I can actually cut the release

Issue #1018, opened 2026-04-16
API role endpoints can cause inconsistency between DB state and gateway's routing decisions


chiahangchang:
The gateway maintains 2 tiers of state: 

1. Database (source of truth): on every routing request, the list of active backends is queried fresh from the DB 
2. In-memory `RoutingManager.backendToStatus` (health cache): a `ConcurrentHashMap<String, TrinoStatus>` that is only updated by `ActiveClusterMonitor`, which runs on a configurable schedule (default 60 seconds) 

At request time, routing works as a 2-step filter: 
- `getActiveBackends(routingGroup)`: fresh DB query (always current) 
- `removeIf(isBackendNotHealthy())`: checks in-memory health cache (stale by up to 60 seconds)
- route to selected backend 

Only backends with `TrinoStatus.HEALTHY` in the cache are routable. A null entry or anything other than HEALTHY causes the backend to be filtered out. 

There are 5 write endpoints for the API role. 
1. `/gateway/backend/modify/add`
2. `/gateway/backend/modify/update`
3. `/gateway/backend/modify/delete`
4. `/gateway/backend/activate/{name}`
5. `/gateway/backend/deactivate/{name}`

However, the problem is that these API role endpoints only write to the database; they DO NOT update the `RoutingManager`'s in-memory `backendToStatus` health cache. Therefore, this causes several corner cases, where the gateway's routing decisions is inconsistent with the DB state. 

**Corner case 1**

`gateway/backend/modify/update`: stale HEALTHY after URL change 
Suppose a backend's URL is updated (e.g. points to a new cluster), but the in-memory `backendToStatus` still holds the old HEALTHY status from the previous health check against the old URL. THe gateway will route traffic to this new URL as if it were confirmed healthy, when it might actually be unreachable / broken. Requests will fail up to 60 seconds until the new health check occurs. 

**Corner case 2**

`gateway/backend/deactivate/{name}`: stale HEALTHY survives into the next activation
Suppose a backend is deactivated, the DB immediately excludes it from future `getActiveBackends()` queries, so routing is correct in the short term. However, `backendToStatus` retains the last HEALTHY entry, and `ActiveClusterMonitor` only monitors active backends. Therefore, once deactivated, the backend's health entry is never updated again. This stale HEALTHY entry becomes a problem when the backend is later reactivated. It will re-enter `getActiveBackends()` results, and upon finding the stale HEALTHY entry, the gateway may immediately route traffic to it with any new health verification. 

[...]

PR #1029, opened 2026-04-20
Enhance API role /gateway/backend endpoints to update backendToStatus in-memory cache immediately

chiahangchang:
## Description

The gateway maintains 2 tiers of state:

Database (source of truth): on every routing request, the list of active backends is queried fresh from the DB
In-memory `RoutingManager.backendToStatus` (health cache): a `ConcurrentHashMap<String, TrinoStatus>` that is only updated by `ActiveClusterMonitor`, which runs on a configurable schedule (default 60 seconds)
At request time, routing works as a 2-step filter:

- `getActiveBackends(routingGroup)`: fresh DB query (always current)
- `removeIf(isBackendNotHealthy())`: checks in-memory health cache (stale by up to 60 seconds)
- route to selected backend

Only backends with `TrinoStatus.HEALTHY` in the cache are routable. A null entry or anything other than HEALTHY causes the backend to be filtered out.

There are 5 write endpoints for the API role.

1. `/gateway/backend/modify/add`
2. `/gateway/backend/modify/update`
3. `/gateway/backend/modify/delete`
4. `/gateway/backend/activate/{name}`
5. `/gateway/backend/deactivate/{name}`

However, the problem is that these API role endpoints only write to the database; they DO NOT update the `RoutingManager`'s in-memory `backendToStatus` health cache. Therefore, this causes several corner cases, where the gateway's routing decisions is inconsistent with the DB state.

Corner case 1

`gateway/backend/modify/update`: stale HEALTHY after URL change
Suppose a backend's URL is updated (e.g. points to a new cluster), but the in-memory `backendToStatus` still holds the old HEALTHY status from the previous health check against the old URL. THe gateway will route traffic to this new URL as if it were confirmed healthy, when it might actually be unreachable / broken. Requests will fail up to 60 seconds until the new health check occurs.

Corner case 2

`gateway/backend/deactivate/{name}`: stale HEALTHY survives into the next activation
Suppose a backend is deactivated, the DB immediately excludes it from future `getActiveBackends()` queries, so routing is correct in the short term. However, `backendToStatus` retains the last HEALTHY entry, and `ActiveClusterMonitor` only monitors active backends. Therefore, once deactivated, the backend's health entry is never updated again. This stale HEALTHY entry becomes a problem when the backend is later reactivated. It will re-enter `getActiveBackends()` results, and upon finding the stale HEALTHY entry, the gateway may immediately route traffic to it with any new health verification.

The list of corner cases is not comprehensive, but just gives you an idea. 

The ADMIN role endpoint `/entity?entityType=GATEWAY_BACKEND` prevents this inconsistency, by updating both the DB AND the in-memory `backendToStatus` health cache. It sets the backend's status to PENDING, which will prevent the gateway from routing requests to it until the next periodic health check returns HEALTHY.

This PR copies the logic from the ADMIN role endpoint, over to the API role endpoints.

https://github.com/trinodb/trino-gateway/issues/1018

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/docs/operation.md
+++ b/docs/operation.md
@@ -16,6 +16,95 @@ Existing cluster information can also be modified using the edit button.
 
 ![trino.gateway.io/entity](./assets/trinogateway_cluster_page.png)
 
+## Cluster configuration and health status
+
+Trino Gateway tracks cluster state at two separate layers, each serving a
+different purpose.
+
+### Database: source of truth for configuration
+
+All cluster configuration is stored persistently in a database. This includes
+each cluster's name, routing group, proxy URL, external URL, and whether it is
+marked active or inactive.
+
+Any change made through the API or the admin UI — adding, updating, activating,
+deactivating, or deleting a cluster — is written to the database immediately.
+The database represents **what you have configured**: the intended state of the
+system. It persists across restarts.
+
+### Health check cache: source of truth for routing
+
+In addition to the database, Trino Gateway maintains an in-memory record of
+each cluster's current health. This cache is separate from the database and is
+used to make all query routing decisions.
+
+Each cluster in the cache has one of four health statuses:
+
+| Status | Meaning |
+|--------|---------|
+| `HEALTHY` | The cluster is reachable and accepting queries |
+| `UNHEALTHY` | The cluster failed its most recent health check |
+| `PENDING` | The cluster was recently added or changed and has not yet been health-checked |
+| `UNKNOWN` | Health status could not be determined |
+
+Only clusters with a `HEALTHY` status receive query traffic, regardless of
+whether the database marks them as active. A cluster in `PENDING` state does not
+receive traffic — it transitions to `HEALTHY` or `UNHEALTHY` once the next
+background health check evaluates it.
+
+The cache is updated by a background health check that runs at a configurable
+interval (default: every 1 minute, set via `monitor.taskDelay`). The health
+check mechanism itself is also configurable — by default it calls the `/v1/info`
+endpoint on each cluster, but JDBC, JMX, and Prometheus metrics are also
+supported.
+
+### How the two layers interact
+
+The database and health cache serve complementary roles:
+
+- **Database** — "what you configured"
+- **Health cache** — "what is actually reachable right now"
+
+When you make a change via the API or admin UI, both layers are updated
+immediately:
+
+| Action | Database | Health cache |
+|--------|----------|-------------|
+| Add a cluster | Written immediately | Set to `PENDING` if active, `UNKNOWN` if inactive |
+| Update a cluster | Written immediately | Reset to `PENDING` if active, `UNKNOWN` if inactive |
+| Activate a cluster | Written immediately | Set to `PENDING` |
+| Deactivate a cluster | Written immediately | Set to `UNKNOWN`; excluded from routing immediately |
+| Delete a cluster | Removed immediately | Removed from cache and JMX; excluded from routing immediately |
+
+Because the database and health cache are not updated atomically, a failure
+during the in-memory update could leave the two layers out of sync. To prevent
+this, each operation uses a compensation pattern: before applying any change,
+the current state is saved (previous health status and backend configuration).
+If the in-memory update fails after the database write succeeds, the database
+change is reverted using the saved state:
+
+| Action | Compensation on failure |
+|--------|------------------------|
+| Add a cluster | Delete the new entry from DB, remove health |
+| Update a cluster | Restore old config in DB, revert health |
+| Activate a cluster | Deactivate in DB, revert health |
+| Deactivate a cluster | Re-activate in DB, revert health |
+| Delete a cluster | Re-add old config to DB, revert health |
+
+For update and delete, compensation is skipped if the old backend configuration
+was not found before the operation, to avoid reverting to an unknown state. If
+the compensation itself fails, the compensation exception is attached to the
+original exception as a suppressed exception — the original is always the one
+thrown to the caller.
+
+A cluster that crashes or becomes unreachable after the last health check will
+remain active in the database but will be marked `UNHEALTHY` in the cache
+within one polling interval. During that window, in-flight queries already
+routed to that cluster may fail, but no new queries will be sent to it once the
+cache is updated.
+
+The health cache is not persisted — it is rebuilt from the database and
+repopulated by health checks each time Trino Gateway starts.
 
 ## Graceful shutdown
 
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
@@ -15,6 +15,7 @@ package io.trino.gateway.ha.resource;
 
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
+import io.trino.gateway.ha.router.BackendLifecycleManager;
 import io.trino.gateway.ha.router.GatewayBackendManager;
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.ws.rs.GET;
@@ -37,11 +38,15 @@ public class GatewayResource
     private static final Logger log = Logger.get(GatewayResource.class);
 
     private final GatewayBackendManager gatewayBackendManager;
+    private final BackendLifecycleManager backendLifecycleManager;
 
     @Inject
-    public GatewayResource(GatewayBackendManager gatewayBackendManager)
+    public GatewayResource(
+            GatewayBackendManager gatewayBackendManager,
+            BackendLifecycleManager backendLifecycleManager)
     {
         this.gatewayBackendManager = requireNonNull(gatewayBackendManager, "gatewayBackendManager is null");
+        this.backendLifecycleManager = requireNonNull(backendLifecycleManager, "backendLifecycleManager is null");
     }
 
     @GET
@@ -69,7 +74,7 @@ public class GatewayResource
     public Response deactivateBackend(@PathParam("name") String name)
     {
         try {
-            this.gatewayBackendManager.deactivateBackend(name);
+            backendLifecycleManager.deactivateBackend(name);
         }
         catch (Exception e) {
             log.error(e, e.getMessage());
@@ -83,7 +88,7 @@ public class GatewayResource
     public Response activateBackend(@PathParam("name") String name)
     {
         try {
-            this.gatewayBackendManager.activateBackend(name);
+            backendLifecycleManager.activateBackend(name);
         }
         catch (Exception e) {
             log.error(e, e.getMessage());
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
@@ -15,8 +15,7 @@ package io.trino.gateway.ha.resource;
 
 import com.google.inject.Inject;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
-import io.trino.gateway.ha.router.GatewayBackendManager;
-import io.trino.gateway.ha.router.HaGatewayManager;
+import io.trino.gateway.ha.router.BackendLifecycleManager;
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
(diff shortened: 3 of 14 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
```diff
@@ -60,6 +73,14 @@ public Response updateBackend(ProxyBackendConfiguration backend)
...
     {
         ((HaGatewayManager) haGatewayManager).deleteBackend(name);
+        syncBackendHealth(name, false);
         return Response.ok().build();
     }
+
+    private void syncBackendHealth(String name, boolean active)
+    {
+        TrinoStatus trinoStatus = active ? TrinoStatus.PENDING : TrinoStatus.UNHEALTHY;
+        routingManager.updateBackEndHealth(name, trinoStatus);
```
vishalya: This change and new routes mean we need to be thread safe in the class - BackendStateManager.
so could you please change the member type for the BackendStateManager::clusterStats and clusterStatsJMXs   from HashMap to ConcurrentHashMap.

gateway-ha/src/test/java/io/trino/gateway/ha/resource/TestGatewayResource.java
```diff
@@ -0,0 +1,85 @@
...
+import static org.mockito.Mockito.verify;
+
+final class TestGatewayResource
+{
+    private GatewayBackendManager gatewayBackendManager;
+    private RoutingManager routingManager;
+    private BackendStateManager backendStateManager;
+    private GatewayResource resource;
+
+    @BeforeEach
```
vishalya: Trino guidelines discourage the usage of `@BeforeEach`, see if you can have a member instantiated once.

gateway-ha/src/test/java/io/trino/gateway/ha/resource/TestGatewayResource.java
```diff
@@ -0,0 +1,85 @@
...
+    void setUp()
+    {
+        gatewayBackendManager = mock(GatewayBackendManager.class);
```
vishalya: Use the real instances of the classes instead of the mocks.

gateway-ha/src/test/java/io/trino/gateway/ha/resource/TestHaGatewayResource.java
```diff
@@ -0,0 +1,113 @@
...
+{
+    private HaGatewayManager haGatewayManager;
+    private RoutingManager routingManager;
+    private BackendStateManager backendStateManager;
+    private HaGatewayResource resource;
+
+    @BeforeEach
+    void setUp()
+    {
+        haGatewayManager = mock(HaGatewayManager.class);
```
vishalya: Use the real instances of the classes instead of the mocks.

vishalya (commented):
Refer to - `TestGatewayHaMultipleBackend` to write the tests without mocks.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
chiahangchang: Done, I also marked the ClusterStatsJMX fields as volatile, so that the JMX reader thread always sees the latest write

gateway-ha/src/test/resources/test-config-no-monitor-template.yml
```diff
@@ -0,0 +1,20 @@
...
+  jdbcUrl: ${ENV:POSTGRESQL_JDBC_URL}
+  user: ${ENV:POSTGRESQL_USER}
+  password: ${ENV:POSTGRESQL_PASSWORD}
+  driver: org.postgresql.Driver
+
+clusterStatsConfiguration:
+  monitorType: INFO_API
+
+monitor:
+  taskDelay: 1h
```
chiahangchang, line 16: Note to reviewer: I added `test-config-no-monitor-template.yml` with `monitor.taskDelay` set to `1h`, so that the periodic health check does not race with the test assertions

gateway-ha/src/main/java/io/trino/gateway/ha/router/BackendStateManager.java
```diff
@@ -46,29 +46,26 @@ public ClusterStats getBackendState(ProxyBackendConfiguration backend)
 
     public void updateStates(String clusterId, ClusterStats stats)
     {
-        if (!clusterStatsJMXs.containsKey(clusterId)) {
+        clusterStatsJMXs.computeIfAbsent(clusterId, id -> {
```
chiahangchang, line 49: Note to reviewer: I replaced this check-then-act operation with thread-safe `computeIfAbsent`.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
```diff
@@ -60,6 +73,14 @@ public Response updateBackend(ProxyBackendConfiguration backend)
     public Response removeBackend(String name)
     {
         ((HaGatewayManager) haGatewayManager).deleteBackend(name);
+        syncBackendHealth(name, false);
```
Chaho12: There is cache leak on delete.

This writes an UNHEALTHY entry into BackendStateManager.clusterStats and clusterStatsJMXs for a backend that no longer exists. We should remove it from the JMX MBean as well — MBeanExporter.exportWithGeneratedName(...) is never paired with an unexport(...), so repeated add/delete cycles will accumulate orphaned MBeans.

Why not add
1. `BackendStateManager.removeState(clusterId)` that removes from both maps + exporter.unexport(...)) 
2. RoutingManager.removeBackEndHealth(backendId), 
and call those here instead.

gateway-ha/src/test/java/io/trino/gateway/ha/resource/TestHaGatewayResource.java
```diff
@@ -0,0 +1,158 @@
...
+                .url(format("http://localhost:%s/gateway/backend/modify/delete", routerPort))
+                .post(RequestBody.create("remove-test", JSON))
+                .build();
+        assertThat(httpClient.newCall(request).execute().isSuccessful()).isTrue();
+
+        Request stateRequest = new Request.Builder()
+                .url(format("http://localhost:%s/api/public/backends/remove-test/state", routerPort))
+                .get()
+                .build();
+        assertThat(httpClient.newCall(stateRequest).execute().code()).isEqualTo(404);
```
Chaho12: This assertion doesn’t verify what the test name claims, and won’t catch the cache leak

This only checks that /api/public/backends/remove-test/state returns 404. That 404 likely comes from the database lookup, so even if the entry remains in BackendStateManager.clusterStats and clusterStatsJMXs after delete, the test still passes.

Please assert directly against BackendStateManager (or the JMX bean registry) so a future regression that reintroduces the phantom entry is caught.

docs/operation.md
```diff
@@ -17,15 +17,82 @@ Existing cluster information can also be modified using the edit button.
...
+- **Database** — "what you configured"
+- **Health cache** — "what is actually reachable right now"
+
+When you make a change via the API or admin UI, both layers are updated
+immediately:
+
+| Action | Database | Health cache |
+|--------|----------|-------------|
+| Add a cluster | Written immediately | Set to `PENDING`; transitions to `HEALTHY` after the first successful health check |
+| Update a cluster | Written immediately | Reset to `PENDING` |
```
Chaho12: Wouldn't updating a cluster with active=false set to unhealthy?

```java
    private void syncBackendHealth(String name, boolean active)
    {
        TrinoStatus trinoStatus = active ? TrinoStatus.PENDING : TrinoStatus.UNHEALTHY;
```

docs/operation.md
```diff
@@ -17,15 +17,82 @@ Existing cluster information can also be modified using the edit button.
...
+| Activate a cluster | Written immediately | Set to `PENDING` |
+| Deactivate a cluster | Written immediately | Set to `UNHEALTHY`; excluded from routing immediately |
+| Delete a cluster | Removed immediately | Set to `UNHEALTHY`; excluded from routing immediately |
```
Chaho12: Semantically the entry should be removed from the cache entirely so it should be removed from cache.

docs/operation.md
chiahangchang: Good catch, updated the description

docs/operation.md
chiahangchang: Updated the description

gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
chiahangchang: Done!

gateway-ha/src/test/java/io/trino/gateway/ha/resource/TestHaGatewayResource.java
chiahangchang: I renamed the test, and also added a new test file for BackendStateManager.

vishalya (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
```diff
@@ -70,6 +81,8 @@ public Response deactivateBackend(@PathParam("name") String name)
     {
         try {
             this.gatewayBackendManager.deactivateBackend(name);
+            routingManager.updateBackEndHealth(name, TrinoStatus.UNHEALTHY);
```
andythsu: I'm fine with this, but would it be better if we set to `UNKNOWN` upon deactivation? `UNHEALTHY` seems to be a bit assertive. Just IMO

chiahangchang: I agree, let me make this change

mosabua:
@coderabbitai review

gateway-ha/src/main/java/io/trino/gateway/ha/resource/HaGatewayResource.java
```diff
@@ -32,18 +36,26 @@
...
+        this.routingManager = requireNonNull(routingManager, "routingManager is null");
+        this.backendStateManager = requireNonNull(backendStateManager, "backendStateManager is null");
     }
 
     @Path("/add")
     @POST
     public Response addBackend(ProxyBackendConfiguration backend)
     {
         ProxyBackendConfiguration updatedBackend = haGatewayManager.addBackend(backend);
+        syncBackendHealth(updatedBackend.getName(), updatedBackend.isActive());
```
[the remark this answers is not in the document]
chiahangchang: Created new class `BackendLifecycleManager` to perform these lifecycle writes as one operation, with compensation/rollback semantics

gateway-ha/src/main/java/io/trino/gateway/ha/resource/GatewayResource.java
```diff
@@ -70,6 +81,8 @@ public Response deactivateBackend(@PathParam("name") String name)
     {
         try {
             this.gatewayBackendManager.deactivateBackend(name);
+            routingManager.updateBackEndHealth(name, TrinoStatus.UNKNOWN);
+            backendStateManager.updateStates(name, ClusterStats.builder(name).trinoStatus(TrinoStatus.UNKNOWN).build());
```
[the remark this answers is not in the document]
chiahangchang: Created new class `BackendLifecycleManager` to perform these lifecycle writes as one operation, with compensation/rollback semantics

gateway-ha/src/main/java/io/trino/gateway/ha/router/BackendStateManager.java
```diff
@@ -46,29 +46,39 @@ public ClusterStats getBackendState(ProxyBackendConfiguration backend)
...
+        }).updateFrom(stats);
         clusterStats.put(clusterId, stats);
     }
 
+    public void removeStates(String clusterId)
+    {
+        if (clusterStatsJMXs.remove(clusterId) != null) {
+            exporter.unexportWithGeneratedName(ClusterStatsJMX.class, clusterIdProperties(clusterId));
+        }
+        clusterStats.remove(clusterId);
```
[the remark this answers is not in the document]
chiahangchang: Made the methods synchronized

docs/operation.md
```diff
@@ -17,15 +17,105 @@ Existing cluster information can also be modified using the edit button.
...
+repopulated by health checks each time Trino Gateway starts.
+
 ## Graceful shutdown
 
 Trino Gateway supports graceful shutdown of Trino clusters. Even when a cluster
 is deactivated, any submitted query states can still be retrieved based on the
 Query ID.
 
-To graceful shutdown a trino cluster without query losses, the steps are:
+To graceful shutdown a Trino cluster without query losses, the steps are:
```
ebyhr: This change is unrelated to "Enhance API role /gateway/backend endpoints to update backendToStatus in-memory cache immediately". Please separate a commit. Same for other places.

chiahangchang: I reverted the out-of-scope changes, will commit them in a separate PR

mosabua:
Is there anything outstanding .. it was approved by @vishalya previously .. maybe we can just finish off whatever is missing.

vishalya (approved):

PR #779, opened 2025-10-08
Update dependencies to make scanners happy

wendigo:
## Description

This doesn't address any exploitable CVE but makes scanners happy :)

Before:

```
NAME                INSTALLED                FIXED IN                                            TYPE          VULNERABILITY        SEVERITY  EPSS           RISK   
mina-core           2.2.3                    2.2.4                                               java-archive  GHSA-76h9-2vwh-w278  Critical  36.5% (96th)   34.3   
curl-minimal        7.76.1-31.el9_6.1                                                            rpm           CVE-2024-7264        Low       7.1% (91st)    2.9    
libcurl-minimal     7.76.1-31.el9_6.1                                                            rpm           CVE-2024-7264        Low       7.1% (91st)    2.9    
shadow-utils        2:4.9-12.el9                                                                 rpm           CVE-2024-56433       Low       3.6% (87th)    1.2    
tar                 2:1.34-7.el9             (won't fix)                                         rpm           CVE-2005-2541        Medium    1.5% (80th)    0.9    
curl-minimal        7.76.1-31.el9_6.1                                                            rpm           CVE-2024-9681        Low       0.6% (67th)    0.2    
libcurl-minimal     7.76.1-31.el9_6.1                                                            rpm           CVE-2024-9681        Low       0.6% (67th)    0.2    
openssl-libs        1:3.2.2-6.el9_5.1        (won't fix)                                         rpm           CVE-2024-41996       Low       0.4% (62nd)    0.2    
curl-minimal        7.76.1-31.el9_6.1                                                            rpm           CVE-2024-11053       Low       0.3% (55th)    0.1    
libcurl-minimal     7.76.1-31.el9_6.1                                                            rpm           CVE-2024-11053       Low       0.3% (55th)    0.1    
openjdk             24+36                    1.8.0_462, 8.0.462, 11.0.28, 17.0.16, *24.0.2, ...  binary        CVE-2025-30749       High      0.2% (39th)    0.1    
openjdk             24+36                    1.8.0_462, 8.0.462, 11.0.28, 17.0.16, *24.0.2, ...  binary        CVE-2025-50106       High      0.2% (39th)    0.1    
libxml2             2.9.13-12.el9_6                                                              rpm           CVE-2024-34459       Low       0.2% (43rd)    < 0.1  
glib2               2.68.4-16.el9_6.2                                                            rpm           CVE-2023-32636       Low       0.2% (38th)    < 0.1  
[...]

PR #1094, opened 2026-05-26
Add missing libstdc++ dependency in runtime docker image

yardenc2003:
## Description
Since the migration from `ubi-minimal` to `ubi-micro`, the runtime image no longer contains the `libstdc++` runtime library required by the Brotli native library (for Jetty compression).

This causes gateway startup failures with:
```
2026-05-24T09:03:58.935Z        ERROR   main    io.airlift.http.server.HttpServer       Error loading http server compression
java.util.ServiceConfigurationError: org.eclipse.jetty.compression.Compression: Provider org.eclipse.jetty.compression.brotli.BrotliCompression could not be instantiated
...
Caused by: java.lang.UnsatisfiedLinkError: Failed to load Brotli native library...
...
Caused by: java.lang.UnsatisfiedLinkError: Caused by: java.lang.UnsatisfiedLinkError: /tmp/com_aayushatharva_brotli4j_175362864808270/libbrotli.so: libstdc++.so.6: cannot open shared object file: No such file or directorylibbrotli.so: libstdc++.so.6: cannot open shared object file: No such file or directory
```

The fix seems to be adding the `libstdc++` package to the overlay installed in the Dockerfile.
I tested it out by building the new image (for `arm64` only) and making sure the errors no longer appear.

The issue appears to have been introduced after the migration from
```
registry.access.redhat.com/ubi9/ubi-minimal:latest
```
to:
```
registry.access.redhat.com/ubi10/ubi-micro:latest
```
in PR #779 (officially since gateway v17)

## Release notes

() This is not user-visible or is docs only, and no release notes are required.

```markdown
* Fix missing `libstdc++.so.6` runtime error by adding `libstdc++` to Docker image for Brotli native library support.
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,7 +40,8 @@ RUN \
     dnf update -y && \
     dnf install --installroot /tmp/overlay --setopt install_weak_deps=false --nodocs -y \
       less \
-      zlib `#required by java` \
+      libstdc++ `# required by Brotli native library` \
+      zlib `# required by java` \
       curl-minimal grep `# required by health-check` \
       shadow-utils `# required by useradd` \
       tar `# required to support kubectl cp` && \


mosabua (approved):
Looks good. Thank you.

rashworld-max (approved):

PR #1138, opened 2026-06-22
Check Web UI dependency licenses

### LINKED ISSUE #984: Verify license in Web UI dependencies
* Relates to https://github.com/trinodb/trino/pull/27832

hhiamte:
## Description

Add Web UI dependency license verification for Trino Gateway.

This adds an allowlist of accepted Web UI dependency licenses and a `check:licenses` script that validates the `license` fields from installed package `package.json` files. The Web UI `build` script now runs this check before `tsc && vite build`, so the existing frontend build path verifies licenses automatically.

Closes #984.

This follows the same general approach as trinodb/trino#27832. `pnpm-lock.yaml` does not record license metadata, so the checker reads installed package manifests under `webapp/node_modules/.pnpm` after `pnpm install`.

Validation performed:

```text
corepack pnpm@10.27.0 --dir webapp run check:licenses
Verified licenses for 477 Web UI dependencies.
```

I also temporarily removed MIT from `webapp/allowed-licenses.txt` and confirmed the check fails with a list of disallowed packages.

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/pom.xml
+++ b/gateway-ha/pom.xml
@@ -581,6 +581,17 @@
                         </configuration>
                     </execution>
                     <execution>
+                        <id>test licenses script</id>
+                        <goals>
+                            <goal>pnpm</goal>
+                        </goals>
+                        <phase>generate-resources</phase>
+                        <configuration>
+                            <arguments>run test:licenses</arguments>
+                            <skip>${skipTests}</skip>
+                        </configuration>
+                    </execution>
+                    <execution>
                         <id>build</id>
                         <goals>
                             <goal>pnpm</goal>
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -5,7 +5,8 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "tsc && vite build",
+    "build": "node scripts/check-licenses.js && tsc && vite build",
+    "check:licenses": "node scripts/check-licenses.js",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview"
   },
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -5,8 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "node scripts/check-licenses.js && tsc && vite build",
+    "build": "node --test scripts/check-licenses.test.js && node scripts/check-licenses.js && tsc && vite build",
     "check:licenses": "node scripts/check-licenses.js",
+    "test:licenses": "node --test scripts/check-licenses.test.js",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview"
   },
@@ -34,6 +35,7 @@
     "eslint": "^8.53.0",
     "eslint-plugin-react-hooks": "^4.6.2",
     "eslint-plugin-react-refresh": "^0.5.4",
+    "spdx-expression-parse": "5.0.0",
     "typescript": "^5.9.3",
     "vite": "^5.4.21",
     "vite-plugin-svgr": "^4.5.0",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "node --test scripts/check-licenses.test.js && node scripts/check-licenses.js && tsc && vite build",
+    "build": "pnpm run check:licenses && tsc && vite build",
     "check:licenses": "node scripts/check-licenses.js",
     "test:licenses": "node --test scripts/check-licenses.test.js",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -5,8 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "node scripts/check-licenses.js && tsc && vite build",
+    "build": "node --test scripts/check-licenses.test.js && node scripts/check-licenses.js && tsc && vite build",
     "check:licenses": "node scripts/check-licenses.js",
+    "test:licenses": "node --test scripts/check-licenses.test.js",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview"
   },
@@ -34,6 +35,7 @@
     "eslint": "^8.53.0",
     "eslint-plugin-react-hooks": "^4.6.2",
     "eslint-plugin-react-refresh": "^0.4.26",
+    "spdx-expression-parse": "5.0.0",
     "typescript": "^5.9.3",
     "vite": "^5.4.21",
     "vite-plugin-svgr": "^4.5.0",
--- a/webapp/scripts/check-licenses.js
+++ b/webapp/scripts/check-licenses.js
@@ -1,0 +1,105 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+const scriptDirectory = path.dirname(fileURLToPath(import.meta.url));
+const webappDirectory = path.resolve(scriptDirectory, '..');
+const nodeModulesDirectory = path.join(webappDirectory, 'node_modules', '.pnpm');
+const allowedLicensesPath = path.join(webappDirectory, 'allowed-licenses.txt');
+
+function readAllowedLicenses()
+{
+    return new Set(fs.readFileSync(allowedLicensesPath, 'utf8')
+        .split(/\r?\n/)
+        .map((line) => line.trim())
+        .filter((line) => line && !line.startsWith('#')));
+}
+
+function packageLicense(packageJson)
+{
+    if (typeof packageJson.license === 'string' && packageJson.license.trim()) {
+        return packageJson.license.trim();
+    }
+
+    if (Array.isArray(packageJson.licenses)) {
+        const licenses = packageJson.licenses
+            .map((license) => typeof license === 'string' ? license : license?.type)
+            .filter(Boolean)
+            .map((license) => license.trim())
+            .filter(Boolean);
+        if (licenses.length > 0) {
+            return licenses.join(' OR ');
+        }
+    }
+
+    return undefined;
+}
+// Read package manifests because pnpm-lock.yaml does not record licenses.
+function packageJsonFiles()
+{
+    if (!fs.existsSync(nodeModulesDirectory)) {
+        throw new Error('node_modules/.pnpm was not found. Run pnpm install before checking licenses.');
+    }
+
+    return fs.readdirSync(nodeModulesDirectory, {withFileTypes: true})
+        .filter((entry) => entry.isDirectory())
+        .flatMap((entry) => {
+            const nestedNodeModules = path.join(nodeModulesDirectory, entry.name, 'node_modules');
+            if (!fs.existsSync(nestedNodeModules)) {
+                return [];
+            }
+
+            return fs.readdirSync(nestedNodeModules, {withFileTypes: true})
+                .filter((packageEntry) => packageEntry.isDirectory() && !packageEntry.name.startsWith('.'))
+                .flatMap((packageEntry) => {
+                    const packagePath = path.join(nestedNodeModules, packageEntry.name);
+                    if (packageEntry.name.startsWith('@')) {
+                        return fs.readdirSync(packagePath, {withFileTypes: true})
+                            .filter((scopedPackageEntry) => scopedPackageEntry.isDirectory())
+                            .map((scopedPackageEntry) => path.join(packagePath, scopedPackageEntry.name, 'package.json'));
+                    }
+                    return [path.join(packagePath, 'package.json')];
+                });
+        })
+        .filter((packageJsonPath) => fs.existsSync(packageJsonPath));
+}
+
+function checkLicenses()
+{
+    const allowedLicenses = readAllowedLicenses();
+    const packages = new Map();
+
+    for (const packageJsonPath of packageJsonFiles()) {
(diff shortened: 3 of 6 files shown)


mosabua:
Did you submit a CLA already @hhiamte ?

hhiamte:
> Did you submit a CLA already @hhiamte ?

Hi, yes I submitted one a few days ago though I don’t know if it was processed yet.

mosabua:
@cla-bot check

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,105 @@
...
+        .split(/\r?\n/)
+        .map((line) => line.trim())
+        .filter((line) => line && !line.startsWith('#')));
+}
+
+function packageLicense(packageJson)
+{
+    if (typeof packageJson.license === 'string' && packageJson.license.trim()) {
+        return packageJson.license.trim();
+    }
```
Chaho12, line 33: I am not FE expert, but this seems valid feedback. PTAL.

This handles two shapes:
1. license: "MIT" (string) ✓
2. licenses: [{type:"MIT"},...] (array) ✓

But it misses third historic npm shape — license as single object:
{ "license": { "type": "MIT", "url": "http://..." } }

Here typeof packageJson.license === 'string' is false, licenses array absent → returns undefined → package lands in missingLicenses → process.exit(1). Build breaks even though license is valid MIT.

This form deprecated by npm ~2015 but still ships in old transitive deps. 477 deps today pass; risk is a future/updated dep pulling one in. 


```suggestion
    if (packageJson.license && typeof packageJson.license === 'object' && packageJson.license.type) {
        return packageJson.license.type.trim();
    }
```

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,105 @@
...
+            console.error('Packages without license metadata:');
+            console.error(missingLicenses.join('\n'));
+        }
+        if (disallowedLicenses.length) {
+            console.error('Packages with licenses not listed in allowed-licenses.txt:');
+            console.error(disallowedLicenses.join('\n'));
+        }
+        process.exit(1);
+    }
+
```
Chaho12, line 170: PTAL

---

SPDX license match is exact-string, rejects valid compound expressions

`check-licenses.js` compares each package's license against `allowed-licenses.txt` with exact `Set.has()`. SPDX expressions combine with `OR`/`AND`/parens, so a dependency whose every component license is allowed still fails, because the *combined string* isn't a literal allowlist entry.

Failing cases:

| Package declares | Allowlist has | Result |
|---|---|---|
| `MIT OR CC0-1.0` | `(MIT OR CC0-1.0)` | ✗ paren mismatch |
| `Apache-2.0 AND MIT` | `Apache-2.0`, `MIT` | ✗ compound not split |
| `(MIT AND Zlib)` | `MIT`, `BSD-3-Clause` | ✗ parens + `Zlib` missing |
| `MIT OR ISC` (from `licenses` array join) | `MIT`, `ISC` | ✗ join never a literal entry |

**Impact:** false build failures on legitimately-licensed deps. Fails closed (over-strict), so no security risk, but breaks `build` whenever a dep ships a compound SPDX expression.

**Fix:** evaluate the SPDX expression against the allowed set instead of literal matching — `OR` = any operand allowed passes, `AND` = all operands must be allowed. Then reduce `allowed-licenses.txt` to atomic IDs (drop the `(MIT OR CC0-1.0)` compound entry).

hhiamte: Addressed in the latest commit. License values are now parsed as SPDX expressions, with support for OR, AND, and parentheses against an atomic allowlist. I also added tests covering the cases above. All CI checks pass. PTAL.

mosabua:
Wdyt @Chaho12  .. ready to go now?

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,145 @@
...
+    console.log(`Verified licenses for ${packages.size} Web UI dependencies.`);
+}
+
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
```
Chaho12: process.argv[1] is resolved to an absolute path but symlinks are not resolved. 
`import.meta.url` for the ESM entry point is realpath'd by Node. 
If the checkout path contains any symlink (macOS /tmp→/private/tmp, Jenkins workspace links, maven's ${project.basedir}/..), the two differ, checkLicenses() never runs, and the script exits 0. The license gate looks green while checking nothing — a silent failure mode.

```suggestion
const scriptPath = fileURLToPath(import.meta.url);
if (process.argv[1] && fs.realpathSync(process.argv[1]) === scriptPath) {
```

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,145 @@
...
+    const disallowedLicenses = [];
+
+    for (const [packageName, license] of [...packages.entries()].sort()) {
+        if (!license) {
+            missingLicenses.push(packageName);
+        }
+        else if (!isLicenseAllowed(license, allowedLicenses)) {
+            disallowedLicenses.push(`${packageName}: ${license}`);
+        }
+    }
```
Chaho12, line 150: add invalidLicenses
```
    for (const [packageName, license] of [...packages.entries()].sort()) {
        if (!license) {
            missingLicenses.push(packageName);
        }
        else if (!parseLicense(license)) {
            invalidLicenses.push(`${packageName}: ${license}`);
        }
        else if (!isLicenseAllowed(license, allowedLicenses)) {
            disallowedLicenses.push(`${packageName}: ${license}`);
        }
    }
```

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,145 @@
...
+
+    if (missingLicenses.length || disallowedLicenses.length) {
+        if (missingLicenses.length) {
+            console.error('Packages without license metadata:');
+            console.error(missingLicenses.join('\n'));
+        }
+        if (disallowedLicenses.length) {
+            console.error('Packages with licenses not listed in allowed-licenses.txt:');
+            console.error(disallowedLicenses.join('\n'));
+        }
```
Chaho12, line 166: add invalidLicenses error

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,145 @@
...
+    return false;
+}
+
+export function isLicenseAllowed(license, allowedLicenses)
+{
+    try {
+        return isParsedLicenseAllowed(parseSpdxExpression(license), allowedLicenses);
+    }
+    catch {
+        return false;
```
Chaho12: parsefailure will be misreported as "not in allowlist". Why don't we split the buckets

```js
-- add
export function parseLicense(license)
{
    try {
        return parseSpdxExpression(license);
    }
    catch {
        return undefined;
    }
}

-- change
export function isLicenseAllowed(license, allowedLicenses)
{
    const expression = parseLicense(license);
    return expression ? isParsedLicenseAllowed(expression, allowedLicenses) : false;
}

File: webapp/scripts/check-licenses.js [hunk 7]
```diff
@@ -0,0 +1,145 @@
...
+const webappDirectory = path.resolve(scriptDirectory, '..');
+const nodeModulesDirectory = path.join(webappDirectory, 'node_modules', '.pnpm');
+const allowedLicensesPath = path.join(webappDirectory, 'allowed-licenses.txt');
+
+function readAllowedLicenses()
+{
+    return new Set(fs.readFileSync(allowedLicensesPath, 'utf8')
+        .split(/\r?\n/)
+        .map((line) => line.trim())
+        .filter((line) => line && !line.startsWith('#')));
```
Chaho12: A typo such as Apache 2.0 (missing hyphen) silently matches nothing, so hundreds of packages land in the disallowed list with no hint at the real cause. Why don't we validate the allowlist when it is read.

```suggestion
    const licenses = fs.readFileSync(allowedLicensesPath, 'utf8')
        .split(/\r?\n/)
        .map((line) => line.trim())
        .filter((line) => line && !line.startsWith('#'));

    const invalid = licenses.filter((license) => !parseLicense(license));
    if (invalid.length) {
        throw new Error(`Invalid SPDX identifiers in allowed-licenses.txt: ${invalid.join(', ')}`);
    }

    return new Set(licenses);
```

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,145 @@
...
+
+export function isLicenseAllowed(license, allowedLicenses)
+{
+    try {
+        return isParsedLicenseAllowed(parseSpdxExpression(license), allowedLicenses);
+    }
+    catch {
+        return false;
+    }
+}
```
Chaho12, line 84: and new line

webapp/scripts/check-licenses.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+test('allows valid SPDX expressions composed of allowed licenses', () => {
+    assert.equal(isLicenseAllowed('MIT OR ISC', allowedLicenses), true);
+    assert.equal(isLicenseAllowed('Apache-2.0 AND MIT', allowedLicenses), true);
+    assert.equal(isLicenseAllowed('(MIT OR CC0-1.0)', allowedLicenses), true);
+});
+
+test('rejects expressions containing required disallowed licenses', () => {
+    assert.equal(isLicenseAllowed('MIT AND Zlib', allowedLicenses), false);
+    assert.equal(isLicenseAllowed('Zlib OR ISC', allowedLicenses), true);
```
Chaho12: ins this allowed? should be moved to allowed one

webapp/package.json
```diff
@@ -5,7 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "tsc && vite build",
+    "build": "node --test scripts/check-licenses.test.js && node scripts/check-licenses.js && tsc && vite build",
```
Chaho12: This means that even if we do `mvn package -DskipTests`, it doesn't skip all tests (runs FE tests)


```suggestion
    "build": "pnpm run check:licenses && tsc && vite build",
```

Chaho12: Then change gateway-ha/pom.xml
in between install(:566) and build(:577) add the following: 

```xml
                    <execution>
                        <id>test licenses script</id>
                        <goals>
                            <goal>pnpm</goal>
                        </goals>
                        <phase>generate-resources</phase>
                        <configuration>
                            <arguments>run test:licenses</arguments>
                            <skip>${skipTests}</skip>
                        </configuration>
                    </execution>
```

Chaho12:
@hhiamte knock knock

hhiamte:
Sorry for the delay! I've addressed the review comments and rerun the license tests/checks. PTAL when you get a chance.

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    console.log(`Verified licenses for ${packages.size} Web UI dependencies.`);
+}
+
+const scriptPath = fileURLToPath(import.meta.url);
+
+if (process.argv[1] && fs.realpathSync(process.argv[1]) === scriptPath) {
+    checkLicenses();
+}
```
Chaho12, line 178: Add new line

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,178 @@
...
+
+        if (disallowedLicenses.length) {
+            console.error('Packages with licenses not listed in allowed-licenses.txt:');
+            console.error(disallowedLicenses.join('\n'));
+        }
+
+        process.exit(1);
+    }
+
+    console.log(`Verified licenses for ${packages.size} Web UI dependencies.`);
```
Chaho12, line 171: It seems that even if .pnpm is empty, without licenses it would exit 0.
Add an if statement to check if is 0.

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,178 @@
...
+
+
+
+function checkLicenses()
+{
+    const allowedLicenses = readAllowedLicenses();
+    const packages = new Map();
+
+    for (const packageJsonPath of packageJsonFiles()) {
+        const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
```
Chaho12, line 131: Add try/catch so that we know which causes the error, in the event of broken package.json file

webapp/scripts/check-licenses.js
```diff
@@ -0,0 +1,178 @@
...
+        const packageName = `${packageJson.name}@${packageJson.version}`;
```
Chaho12, line 132: if both version and name does not exist, it would return "undefined@undefined". But if there are multiple ones like this, it would all be squashed to a single key.


```suggestion
        const packageName = `${packageJson.name}@${packageJson.version}`;
        const packageName = packageJson.name && packageJson.version
                ? `${packageJson.name}@${packageJson.version}`
                : path.relative(webappDirectory, packageJsonPath);
```

gateway-ha/pom.xml
```diff
@@ -580,6 +580,17 @@
...
+                        <goals>
+                            <goal>pnpm</goal>
+                        </goals>
+                        <phase>generate-resources</phase>
+                        <configuration>
+                            <arguments>run test:licenses</arguments>
+                            <skip>${skipTests}</skip>
+                        </configuration>
+                    </execution>
                     <execution>
```
Chaho12, line 594: Add check licenses. do `run check:licenses` so that releases build with -DSkipTests still verify.

webapp/package.json
```diff
@@ -5,7 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "tsc && vite build",
+    "build": "pnpm run check:licenses && tsc && vite build",
```
Chaho12, line 8: undo this because failure of run build does not show if it is due to license, tsc, or vite. 
We can have the procedure as following:
- install node and pnpm → install → test licenses script → check licenses → build

Chaho12 (changes requested):
There are lost of style issue like adding a new line.
Please take a loot at all. I can't address it all, I see too many.

Also, can you rebase and fix the conflicts?

Issue #926, opened 2026-02-18
Support the Trino spooling protocol


mosabua:
Trino supports a spooling protocol in various configuration.

https://trino.io/docs/current/client/client-protocol.html#protocol-spooling

Trino Gateway should ideally allow use with some or even all of the possible combination. 

The main configuration where results are on object storage separate to Trino and accessed directly should work pretty easily. 

Configurations with access via the coordinator or even directly to the workers will be a bit more difficult. 

In all cases we need to at least test and document necessary configuration. Maybe we even need to document for now that we do NOT yet support it or at least know more.

lozbrown:
@mosabua are you really sure this doesn't work? 

All of our queries pass through a gateway, queries with appropriate drivers show in trino UI as spooled and users have reported significant performance improvement after upgrading to drivers that support spooling.

Maybe wendigo can oppine?

mosabua:
I am sure some scenarios work @lozbrown .. but I am also sure others wont work (such as if access is via workers but the network setup behind Trino Gateway prevents worker access). 

Most importantly I am sure we have no testing or documentation for what does and does not work. This issue is here to improve that and ideally support it all, but also document issues and configurations necessary.

lozbrown:
Cool, I think we are proxying via the coordinator because the need for proxy config on the client's going direct to S3 was causing issues.

koszti:
@mosabua

I documented current behavior and limitations at https://github.com/trinodb/trino-gateway/pull/945

Fixing this I think will require proxying`/v1/spooled/.*` to the correct coordinator with affinity. The main challenge is that spooled URLs do not include a `queryId`, while current sticky routing relies on query ID extraction/mapping. Without a query ID the retrieval requests can be proxied to the wrong cluster which may not have access to the spooled object.

choseungyoon:
## Testing Spooling Protocol with Trino Gateway — STORAGE retrieval mode
 
I've been testing the spooling protocol with Trino Gateway in my environment. Here are my findings.
 
### Environment
 
| Component | Version |
|-----------|---------|
| Trino | 477 |
| Trino Gateway | Milestone 18 |
| trino-python-client | 0.337.0 |
| Object Storage | Ceph S3 compatible |
| Auth | Basic Auth + HTTPS (self-signed certificates) |
| Retrieval Mode | STORAGE |
 
### Setup
 
**Gateway `config.yaml`:**
 
```yaml
serverConfig:
  node.environment: <your-env>
[...]

Issue #670, opened 2025-04-15
[Question] Downloading spooled data


JustinObanor:
My trino cluster has spooling configured in the clusters config file:
```
protocol.spooling.retrieval-mode=COORDINATOR_PROXY
```

I made a query to the gateway using the trino python client with this connection
```
conn = connect(
    http_scheme="https",
    host="gateway.trinoprdt",
    port=8443,
    catalog="hive",
    schema="tpcds_europe_west1_1000",
    verify=False,
    encoding="json+zstd",
    http_session=session,
    http_headers={'X-Trino-Original-User':'xtz'}
)
```

I had to add an extra whitelist in the gateways config file so the gateway will proxy the spooling endpoint:
```
extraWhitelistPaths:
- /v1/spooled/.*
```

This triggers spooling, and eventually fails with
```
DEBUG:urllib3.connectionpool:https://gateway.trinoprdt:8443 "GET /v1/statement/executing/20250415_144309_00019_gntrt/y291e5a9c05bf78e9af33f0bd506421146b3069f6/6 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gateway.trinoprdt:8443
/opt/conda/envs/python3/lib/python3.11/site-packages/urllib3/connectionpool.py:1097: InsecureRequestWarning: Unverified HTTPS request is being made to host 'gateway.trinoprdt'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(
DEBUG:urllib3.connectionpool:https://gateway.trinoprdt:8443 "GET /v1/spooled/download/52bXbfGIdGtpIEt_GfX3KR9yZ6zxt-SyPySruC9HMHI= HTTP/1.1" 200 None
Traceback (most recent call last):
  File "/opt/conda/envs/python3/lib/python3.11/site-packages/trino/client.py", line 1222, in __next__
    return next(self._rows)
           ^^^^^^^^^^^^^^^^
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/dh216100/play2.py", line 33, in <module>
    cur.execute("SELECT * FROM hive.tpcds_europe_west1_1000.store_sales LIMIT 1000000")
  File "/opt/conda/envs/python3/lib/python3.11/site-packages/trino/dbapi.py", line 614, in execute
    self._iterator = iter(self._query.execute())
[...]

PR #1142, opened 2026-06-22
Add /v1/spooled/* to default path whitelist

### LINKED ISSUE #1033: Spooled segments not cleaned up when queries are routed through the Gateway (STORAGE retrieval mode)
### Describe the bug
 
When using the spooling protocol with `STORAGE` retrieval mode through Trino Gateway, spooled segments on S3 are not deleted after the client successfully reads them. The same queries executed directly against the coordinator clean up segments correctly.
 
### Root cause
 
In STORAGE retrieval mode, after a client downloads a segment from S3, it sends an acknowledgment (`ackUri`) request to the coordinator to trigger immediate deletion. When the client connects through the Gateway, these ack requests appear to not reach the coordinator properly, preventing eager segment cleanup.
 
### Impact
 
Without the workaround, spooled segments accumulate indefinitely on S3, eventually consuming significant storage. In a production environment with many concurrent queries returning large result sets, this can lead to storage exhaustion.
 
### Steps to reproduce
 
1. Configure Trino with spooling enabled:
```properties
# Coordinator config.properties
protocol.spooling.enabled=true
protocol.spooling.shared-secret-key=<key>
protocol.spooling.retrieval-mode=STORAGE
```
 
```properties
# spooling-manager.properties (Coordinator + Workers)
spooling-manager.name=filesystem
fs.s3.enabled=true
fs.location=s3://<bucket>/<path>/
fs.segment.explicit-ack=true
```
 
2. Configure Trino Gateway with the Trino cluster as a backend
3. Execute a query that returns enough data to trigger spooling (not inlined):
```sql
-- Through Gateway
SELECT * FROM <large_table> LIMIT 100000;
```
 
4. After the query completes, check S3:
```bash
aws s3 ls s3://<bucket>/<path>/ --recursive
```
 
5. Observe: segment files (`.json+zstd`) remain on S3
6. Repeat the same query directly against the coordinator (bypassing Gateway)
7. Observe: segment files are cleaned up after the query completes
### Workaround
 
Enable TTL-based pruning in `spooling-manager.properties`:
 
```properties
fs.segment.ttl=2h
fs.segment.pruning.enabled=true
fs.segment.pruning.interval=10m
fs.segment.pruning.batch-size=100
```
 
This ensures segments are cleaned up after the TTL expires, regardless of ack delivery. Additionally, an S3 lifecycle policy can serve as a secondary safety net.
 
### Expected behavior
 
Spooled segments should be cleaned up after the client reads them, even when the query is routed through the Gateway. Either:
 
1. The Gateway should transparently proxy ack requests to the coordinator, or
2. The Gateway should be aware of spooling ack endpoints and handle them appropriately
### Environment
 
- Trino Gateway: Milestone 18
- Trino: 477
- trino-python-client: 0.337.0
- Object Storage: Ceph S3 compatible
- Retrieval Mode: STORAGE
### Additional context
 
Related to #926 (Support the Trino spooling protocol).
 
This issue was discovered as part of a broader POC testing spooling protocol compatibility with Trino Gateway. The STORAGE retrieval mode itself works correctly through the Gateway — query results are successfully delivered to clients via S3 pre-signed URIs. Only the cleanup mechanism is affected.
 
Note: Segments from killed queries or Trino shutdown also remain on S3, but that behavior is consistent regardless of Gateway involvement and is handled by TTL pruning as designed.

choseungyoon:
Requests to /v1/spooled/ack/* and /v1/spooled/download/* were not forwarded to backend coordinators because the path was absent from PathFilter's default whitelist. The RouterPreMatchContainerRequestFilter only rewrites the request URI to ROUTE_TO_BACKEND when isPathWhiteListed returns true; unrecognized paths fall through to the Gateway's own JAX-RS resources and receive a 404 without ever reaching a coordinator.

Ack is stateless with respect to coordinator-local state: the opaque token in the URL is AES-encrypted with the shared secret key, and FileSystemSpoolingManager.acknowledge() delegates to a pure fileSystem.deleteFile call whose path is derived entirely from the deserialized handle. Verified empirically on Trino 477:
- same coordinator, restarted between issuing and acking: segment deleted
- ack issued against a different coordinator than the one that produced the segment: segment deleted

Any coordinator with the same protocol.spooling.shared-secret-key and fs.location can process any ack or download request.

Fixes #1033
Related: #926, #670

## Release notes

```markdown
* Fix some things.
```

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
@@ -25,6 +25,7 @@ import static io.trino.gateway.ha.handler.HttpUtils.UI_API_STATS_PATH;
 import static io.trino.gateway.ha.handler.HttpUtils.V1_INFO_PATH;
 import static io.trino.gateway.ha.handler.HttpUtils.V1_NODE_PATH;
 import static io.trino.gateway.ha.handler.HttpUtils.V1_QUERY_PATH;
+import static io.trino.gateway.ha.handler.HttpUtils.V1_SPOOLED_PATH;
 import static io.trino.gateway.ha.handler.HttpUtils.V1_STATEMENT_PATH;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -99,6 +100,15 @@ class TestPathFilter
     }
 
     @Test
+    void testSpooledPaths()
+    {
+        // Spooled segment paths must be forwarded to the backend coordinator.
+        assertThat(pathFilter.isPathWhiteListed(V1_SPOOLED_PATH)).isTrue();
+        assertThat(pathFilter.isPathWhiteListed(V1_SPOOLED_PATH + "/ack/k57HZQqycuX7B_5QQnFtvR9yZ6zxt-SyPySruC9HMHI=")).isTrue();
+        assertThat(pathFilter.isPathWhiteListed(V1_SPOOLED_PATH + "/download/BknNfrjG9rI6GsNoAlSOKB9yZ6zxt-SyPySruC9HMHI=")).isTrue();
+    }
+
+    @Test
     void testNonWhitelistedPaths()
     {
         assertThat(pathFilter.isPathWhiteListed("/v3/statement")).isFalse(); // Not in our statement paths
@@ -145,6 +155,8 @@ class TestPathFilter
 
         // Should still allow hardcoded paths
         assertThat(emptyFilter.isPathWhiteListed(V1_QUERY_PATH)).isTrue();
+        assertThat(emptyFilter.isPathWhiteListed(V1_SPOOLED_PATH + "/ack/sometoken")).isTrue();
+        assertThat(emptyFilter.isPathWhiteListed(V1_SPOOLED_PATH + "/download/sometoken")).isTrue();
         assertThat(emptyFilter.isPathWhiteListed(TRINO_UI_PATH)).isTrue();
         assertThat(emptyFilter.isPathWhiteListed(V1_INFO_PATH)).isTrue();
         assertThat(emptyFilter.isPathWhiteListed(OAUTH_PATH)).isTrue();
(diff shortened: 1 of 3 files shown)


choseungyoon:
Built this branch and deployed it to a Trino Gateway in front of two Trino 477 clusters (STORAGE mode, SSE-C enabled).

A spooling query through the Gateway now cleans up its S3 segments — the ack reaches a coordinator and the segment is deleted instead of waiting for TTL pruning. Before the change the same request 404'd at the Gateway, so the coordinator never received the ack.

No header handling was needed — the SSE-C headers already pass through. The only gap was that `/v1/spooled/*` wasn't in the path whitelist.

mosabua:
This is great. Did you submit a CLA @choseungyoon ?

mosabua (approved):
This is a great find. Nice job @choseungyoon !

I will get this merged as soon as a CLA is submitted and processed. However I dont think we will hold the release for this.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
```diff
@@ -98,6 +99,20 @@ void testDynamicRegexPaths()
...
 
+    @Test
+    void testSpooledPaths()
+    {
+        // /v1/spooled/ack/* and /v1/spooled/download/* must be forwarded to the backend
+        // so that the coordinator can process ack and clean up S3 segments.
+        // Ack is stateless w.r.t. coordinator-local state (verified empirically):
+        // the opaque token is AES-encrypted with the shared secret key, and acknowledge()
+        // delegates to a pure fileSystem.deleteFile call. Any coordinator with the same
+        // shared key and storage config can process any ack.
```
ebyhr: This comment is too verbose. I don't think we need to explain the details of spooled endpoint. Let's simplify.

mosabua: Fair enough. In fact maybe we can just link to the Trino docs at https://trino.io/docs/current/client/client-protocol.html#spooling-protocol and if necessary add more info there.

choseungyoon:
> This is great. Did you submit a CLA @choseungyoon ?

Yes. I submitted a CLA last night.

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestPathFilter.java
choseungyoon: Done, simplified to a single line. Thanks!

ebyhr:

choseungyoon:
> Please squash commits into one.

Done. Thanks!

Chaho12:
Commit body seems to have too much detail. We should shorten it for readability. We don't need implementation and proof detail in the body, PR is enough for that. 

"Commit messages and history are important as well, because they are used by other developers to keep track of the motivation behind changes."

How about sth like
```
Requests to /v1/spooled/{ack,download}/* were not forwarded
to coordinators because these paths were missing from the
default whitelist, resulting in 404 responses from the gateway.

These endpoints are stateless and can be handled by any
coordinator with the same shared secret and filesystem.
```

choseungyoon:
> Commit body seems to have too much detail. We should shorten it for readability. We don't need implementation and proof detail in the body, PR is enough for that.
> 
> "Commit messages and history are important as well, because they are used by other developers to keep track of the motivation behind changes."
> 
> How about sth like
> 
> ```
> Requests to /v1/spooled/{ack,download}/* were not forwarded
> to coordinators because these paths were missing from the
> default whitelist, resulting in 404 responses from the gateway.
> 
> These endpoints are stateless and can be handled by any
> coordinator with the same shared secret and filesystem.
> ```

Done. Thank you!

mosabua:
@cla-bot check

choseungyoon:
I have sended a CLA document again.
Please check it out.

mosabua:
@martint could you process this and other CLAs please

choseungyoon:
My cla is signed. Please check cla-signed verification.

ebyhr:
@cla-bot check

ebyhr (approved):

ebyhr:
Merged, thanks!

PR #1121, opened 2026-06-13
Fix routing rule classification of some statements

jiwen624:
## Description
`StatementUtils` (a copy of `io.trino.util.StatementUtils`) hasn't been synced since it was introduced, so `SHOW BRANCHES`, `CREATE`/`DROP`/`FAST FORWARD BRANCH`, `REFRESH VIEW`, and column `SET`/`DROP DEFAULT` fall through to `"UNKNOWN"`. Routing rules matching on `getResourceGroupQueryType()` (e.g. `"DATA_DEFINITION"`) silently miss these statements.

This PR adds the 7 missing mappings, plus a test that scans `io.trino.sql.tree` and fails if any `Statement` subclass is unmapped — so future Trino dependency bumps catch drift automatically.

## Additional context and related issues
`StatementUtils` is marked `//modified version of io.trino.util.StatementUtils`, and `docs/routing-rules.md` documents `getResourceGroupQueryType()` as matching Trino's values.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
(x) Release notes are required, with the following suggested text:

```markdown
* Fix routing rules classifying `SHOW BRANCHES`, branch DDL, `REFRESH VIEW`, and column `SET`/`DROP DEFAULT` statements as `UNKNOWN` for `getResourceGroupQueryType()`.
```

PR #1146, opened 2026-06-24
Improve StatementUtils drift-guard test failure message

jiwen624:
## Description
Follow-up to #1121. The `testEveryStatementTypeIsClassified` drift guard in `TestStatementUtils` fails when `dep.trino.version` is bumped and Trino adds a `Statement` type that `StatementUtils` doesn't classify. The previous failure message was terse, so this makes it self-explanatory: it names the hand-maintained copy, where it's kept in sync (`dep.trino.version` in `gateway-ha/pom.xml`), why the gap matters (routing rules silently miss the statements), and exactly what to do (copy the query type from upstream `io.trino.util.StatementUtils`, or add it to `UNMAPPED_BY_DESIGN`).

Wording per the discussion in https://github.com/trinodb/trino-gateway/pull/1121#issuecomment-4786127552.

Test-only change. See https://github.com/trinodb/trino-gateway/pull/1121#issuecomment-4792249589.

## Release notes

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStatementUtils.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStatementUtils.java
@@ -25,7 +25,9 @@ import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
 import java.lang.reflect.Modifier;
+import java.util.Comparator;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import static io.trino.gateway.ha.router.StatementUtils.getResourceGroupQueryType;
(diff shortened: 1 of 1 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStatementUtils.java
Chaho12, line 105: Since this is HashSet, %s order is nondeterministic in failure msg added below. 
Why don't we change it to TreeSet for stable, diff-friendly output.

Chaho12 (approved):
LGTM 👍

gateway-ha/src/test/java/io/trino/gateway/ha/router/TestStatementUtils.java
jiwen624: Good point. updated.

mosabua (approved):

mosabua:
Thanks .. this got in with 20 release

PR #1156, opened 2026-07-02
Return 404 instead of 502 when proxying to intentionally scaled-down clusters

prathit06:
## Description

When a backend is intentionally deactivated (e.g. scaled to zero) but health checks still report it, proxy requests to those backends return `502 Bad Gateway` which suggests a transient error that will self-resolve. This is misleading, the cluster is unreachable by design.
This change returns `404 Not Found` instead when the target backend is inactive and not reporting as healthy, making the response correctly signal that the cluster is intentionally unavailable.

## Release notes
```markdown
* Return 404 instead of 502 when proxying to intentionally scaled-down clusters.

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
@@ -20,6 +20,7 @@ import io.trino.gateway.ha.config.HaGatewayConfiguration;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
 import io.trino.gateway.ha.handler.schema.RoutingDestination;
 import io.trino.gateway.ha.handler.schema.RoutingTargetResponse;
+import io.trino.gateway.ha.router.GatewayBackendManager;
 import io.trino.gateway.ha.router.GatewayCookie;
 import io.trino.gateway.ha.router.RoutingGroupSelector;
 import io.trino.gateway.ha.router.RoutingManager;
@@ -46,6 +47,7 @@ public class RoutingTargetHandler
     private static final Logger log = Logger.get(RoutingTargetHandler.class);
     private final RoutingManager routingManager;
     private final RoutingGroupSelector routingGroupSelector;
+    private final GatewayBackendManager gatewayBackendManager;
     private final String defaultRoutingGroup;
     private final List<String> statementPaths;
     private final boolean requestAnalyserClientsUseV2Format;
@@ -56,10 +58,12 @@ public class RoutingTargetHandler
     public RoutingTargetHandler(
             RoutingManager routingManager,
             RoutingGroupSelector routingGroupSelector,
+            GatewayBackendManager gatewayBackendManager,
             HaGatewayConfiguration haGatewayConfiguration)
     {
         this.routingManager = requireNonNull(routingManager);
         this.routingGroupSelector = requireNonNull(routingGroupSelector);
+        this.gatewayBackendManager = requireNonNull(gatewayBackendManager, "gatewayBackendManager is null");
         this.defaultRoutingGroup = haGatewayConfiguration.getRouting().getDefaultRoutingGroup();
         statementPaths = requireNonNull(haGatewayConfiguration.getStatementPaths());
         requestAnalyserClientsUseV2Format = haGatewayConfiguration.getRequestAnalyzerConfig().isClientsUseV2Format();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by prathit06** [comment]
Hi @Chaho12,  can you please review this pull request, thanks !

File: gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java [hunk 1]
```diff
@@ -251,12 +262,34 @@ private FluentFuture<ProxyResponse> executeHttp(Request request)
         return FluentFuture.from(httpClient.executeAsync(request, new ProxyResponseHandler(proxyResponseConfiguration)));
     }
 
-    private static Response handleProxyException(Request request, ProxyException e)
+    private Response handleProxyException(Request request, ProxyException e, RoutingDestination routingDestination)
     {
+        if (isExpectedlyUnavailable(routingDestination.clusterName(), gatewayBackendManager, backendStateManager)) {
+            log.info("Proxy request failed (backend inactive and unreachable): %s %s",
+                    request.getMethod(),
+                    request.getUri());
```
Chaho12, line 270: We should log the exception e here. 
Why don't we change to debug to log the exception

```suggestion
            log.debug(e, "Proxy request failed, backend is inactive and unreachable: %s %s", request.getMethod(), request.getUri());
```

gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java
```diff
@@ -251,12 +262,34 @@ private FluentFuture<ProxyResponse> executeHttp(Request request)
...
+            throw new WebApplicationException(
+                    Response.status(NOT_FOUND)
+                            .type(TEXT_PLAIN_TYPE)
+                            .entity("Backend cluster is unreachable.")
```
Chaho12, line 274:
```suggestion
                            .entity("Backend cluster %s is unreachable.".formatted(getRemoteTarget(request.getUri())))
```

gateway-ha/src/test/java/io/trino/gateway/proxyserver/TestProxyRequestHandler.java
```diff
@@ -185,4 +192,78 @@ void testGetQueryDetailsFromRequest()
         assertThat(queryDetail.getSource()).isEqualTo("trino-cli");
         assertThat(queryDetail.getBackendUrl()).isEqualTo("http://localhost:" + routerPort);
     }
+
+    @Test
+    void testInactiveUnhealthyBackendIsExpectedlyUnavailable()
```
Chaho12, line 197: We don't need to run this inside docker container do we? It is a simple test using mock.

gateway-ha/src/main/java/io/trino/gateway/ha/handler/RoutingTargetHandler.java
```diff
@@ -77,8 +81,13 @@ public RoutingTargetResponse resolveRouting(HttpServletRequest request)
                     .orElse(defaultRoutingGroup);
             String externalUrl = queryId.map(routingManager::findExternalUrlForQueryId)
                     .orElse(cluster);
+            String clusterName = gatewayBackendManager.getAllBackends().stream()
+                    .filter(b -> b.getProxyTo().equals(cluster))
```
Chaho12, line 85: b.getProxyTo() is the config string where cluster is formatted string so if there is a trailing slash at the end, matching will fail.

Chaho12 (changes requested):
Thanks for picking this up — the problem is real. A query pinned to a cluster that was
deliberately scaled down currently fails with a 502, which reads like a transient fault that will
clear on its own, and it never does.

Before going through the implementation, I'd like to settle the direction, because it decides
whether there's any code to write. The description frames this as a semantics problem — the
response misleads the caller — and if that's the goal, 404 doesn't seem to serve it.

- **404** already means "this query is gone" on `/v1/statement/executing/{queryId}` in the Trino
  protocol. Overloading it with "the cluster was scaled down" replaces one ambiguous signal with
  another: a client can no longer tell an expired query from a decommissioned cluster, and neither
  can whoever reads the logs afterwards.
- **503** is defined for exactly this condition — the server is intentionally unavailable right
  now — and can carry `Retry-After`.
- **502, as today**, is also defensible. The gateway genuinely did fail to reach the upstream, so
  it isn't a lie; it's just less specific than it could be.

Related question: how much of the perceived problem is actually the status code? The current body
is `"Backend cluster is unreachable."`, which names neither the cluster nor the fact that it was
taken down on purpose. On a multi-cluster gateway that's what leaves the user guessing, and it's
fixable without touching the status code at all.

So: do we want 503, or do we want to keep 502 and improve the message? Happy to go through the
implementation in detail once that's decided — I have notes, but they're moot if the direction
changes.

Issue #1169, opened 2026-07-10
Trino Gateway should support multiple authenticators


AndyJiang99:
### Description

Currently, the Trino Gateway can only support one authenticator at a time, whether that's password, or OAuth2, etc. However, Trino Coordinator itself can support multiple different authenticators to allow for different authenticators to allow user management in one or more systems. https://trino.io/docs/current/security/authentication-types.html 

It would be extremely helpful to allow users to be able to use OAuth2 to log into the Trino Gateway, but still allow the username/password or certificate based approach for automation to talk to the Trino Gateway API itself

AndyJiang99:
https://github.com/trinodb/trino-gateway/pull/1170

PR #1170, opened 2026-07-10
Trino Gateway can support multiple authentication methods with fallback

AndyJiang99:
Issue: https://github.com/trinodb/trino-gateway/issues/1169
## Summary
At LinkedIn, we need to be able to use multiple different authentication mechanisms, similar to the [Trino Server itself](https://trino.io/docs/current/security/authentication-types.html), as we would like to:
1. Enable our internal users to authenticate via SSO
2. Enable some automated system interactions with Trino Gateway which does not have a human-in-the-loop, and thus, would need another method to performing administrative actions such as activating/deactivating Trino clusters on Trino Gateway.

Configurable to show all or first of the authentication methods: 
Screenshot 2026-08-11 at 11 43 24

Screenshot 2026-08-11 at 11 43 28

## Description

**Authentication — multiple methods with fallback**

- `authentication.defaultType` now accepts **either a single value** (`defaultType: "form"`, as before) **or an ordered list** (`defaultType: ["oauth", "form"]`), so Trino Gateway can be configured with more than one authentication method. Existing single-value configs keep working unchanged, since a scalar is bound as a one-element list via Jackson's `ACCEPT_SINGLE_VALUE_AS_ARRAY`:
```
  authentication:
    defaultType: ["oauth", "form"]
```
- `AuthenticationConfiguration` exposes `getDefaultType()/setDefaultType()` returning `List<String>` — the property keeps its `defaultType` name and binds either a scalar or a list.
- **A configured method is always accepted.** `ChainedAuthFilter` builds its chain from the *configured managers* (whichever `oauth`/`form` blocks exist), not from the `defaultType` list, so a request falls back from one method to the next until one succeeds, and listing a subset in `defaultType` can never drop a configured method. `defaultType` only sets the **order** methods are tried in; a configured-but-unlisted method is appended after the listed ones. This lets humans sign in through the primary browser method (e.g. `oauth`) while automation keeps using form/basic auth — and it keeps existing deployments working on upgrade (a config with both `oauth` and `form` blocks keeps accepting both even if only one is named in `defaultType`).
- `LoginResource`'s `/loginType` endpoint returns the ordered list of methods the login page should show — the `defaultType` entries that have a configured manager — instead of a single type. A configured method that isn't listed in `defaultType` is still accepted by the API but is not shown on the login page.
- The webapp login page now renders a toggle listing every configured authentication method (from `/loginType`) so users can pick which method to log in with, instead of only ever showing the first entry.
- Add `authentication.showFirstTypeOnly` (default `false`). When `true`, the login page shows only the first available method (the first `defaultType` entry with a configured manager). This restricts the login page only — `ChainedAuthFilter` still accepts every configured method, so API clients keep their multi-method fallback (e.g. form/basic for automation even when `oauth` is the method shown on the UI).
- Replace the raw `"oauth"`/`"form"` strings (previously referenced in several places) with an `AuthenticationType` enum, and validate every `defaultType` entry against it at startup so an unknown or misspelled type fails fast instead of being silently skipped.

**Authorization — default privileges**

- `AuthorizationConfiguration` gains an `enableDefaultPrivilege` flag (default `false`) alongside the `defaultPrivilege` value. When `enableDefaultPrivilege` is `true`, any authenticated user who isn't a preset user or resolved via LDAP falls back to `defaultPrivilege` — useful for SSO/OAuth users who aren't individually preset but should still get baseline access. When it is `false` (the default), `defaultPrivilege` is ignored and such users are denied. Even when enabled, a missing/empty/`NONE` (case-insensitive) `defaultPrivilege` denies.
- LDAP is authoritative when configured: an empty LDAP group lookup denies access rather than falling back to `defaultPrivilege`, so a misconfigured search (wrong base DN, member attribute, …) can't silently widen access. Preset users with empty privileges still fall back to `defaultPrivilege` when enabled.
- `AuthorizationManager` resolves the effective default privilege once in its constructor instead of recomputing it on every authorization check; `LoginResource` likewise resolves the `/loginType` list once at startup (in its constructor) rather than on every request, so a misconfiguration fails fast at boot instead of logging on and 500-ing every call.
- Deduplicate `defaultType` (order-preserving) and parse `AuthenticationType` case-insensitively, so `["form","form"]` doesn't double-register filters / duplicate the `/loginType` list and `["OAUTH"]` doesn't fail to boot. `ChainedAuthFilter` logs the resolved chain (accepted methods, in fallback order) at startup, and a warning is logged when a configured method is not listed in `defaultType` (accepted by the API but hidden from the login page).
- `docs/security.md` updated to describe the `defaultType` scalar-or-list property and the two-layer model — configuring an `oauth`/`form` block is what makes a method *accepted*, while `defaultType` only sets the fallback order and what the login page shows, and `showFirstTypeOnly` restricts the page further — plus the `enableDefaultPrivilege` / `defaultPrivilege` authorization options (including the LDAP-authoritative rule).

## Testing

**Java (`gateway-ha`, the project's only Maven module):**
```
mvn test
```

New/updated tests added by this PR:
- `TestAuthorizationManager` (14 tests): `defaultPrivilege` behavior — feature disabled (default) denies, `enableDefaultPrivilege: true` grants the configured default, the value is ignored when the flag is off, `NONE`/empty denies even when enabled (case-insensitive), preset users keep their own privileges, preset users with empty privileges fall back to the default, and LDAP is authoritative (membership wins; an empty LDAP result denies even with a default configured).
- `TestAuthenticationTypeResolver` (12 tests): the login-page resolver returns configured methods in order, filters out listed-but-unconfigured methods, dedups, and fails fast on unknown/empty/no-usable input; the chain resolver additionally keeps every configured method (a configured-but-unlisted method is appended after the listed ones), honors the listed order, dedups, and fails fast on unknown/empty input or when no manager is configured at all.
- `TestAuthenticationType` (3 tests): enum value round-trip, case-insensitive parsing, and unknown-value rejection.
- `TestAuthenticationConfiguration` (4 tests): `defaultType` deserialization as both a scalar (bound to a one-element list) and a list, and `showFirstTypeOnly` binding.
- `TestConfiguredMethodAlwaysAccepted` (3 tests): integration test booting with the single-value `defaultType: "oauth"` alongside a configured `form` block — `/loginType` advertises only `oauth`, a Basic-auth request still authenticates via the configured-but-unlisted form/basic method (the backward-compatibility guarantee), and bad credentials are rejected (4xx).
- `TestAuthenticationFallbacks` (3 tests): integration test booting with `defaultType: ["oauth", "form"]` — `/loginType` reports both methods in order, a Basic-auth request still authenticates via the form/basic fallback even though `oauth` is listed first, and bad credentials are rejected (4xx).
- `TestLoginTypeFiltersUnconfiguredMethods` (2 tests): a listed-but-unconfigured `oauth` is filtered from `/loginType` (no dead login button) and form/basic still authenticates.
- `TestLoginTypeShowFirstTypeOnly` (2 tests): `showFirstTypeOnly` restricts `/loginType` to the first available method while the auth chain still accepts form/basic.

All of the above pass. The only failures observed locally are the pre-existing `TestDatabaseMigrations{PostgreSql,Oracle,MySql}` container-state flakes ("table already exists"), which are unrelated to this change (it touches no persistence/migration code).

**webapp:** no test script exists in this project (`package.json` only defines `build`/`lint`), so those were run instead against the one changed frontend file, `login.tsx`:
- `tsc --noEmit` — passes.
- `eslint src/components/login.tsx` — passes clean.
- Note: `npm run lint` across the whole webapp reports 2 pre-existing errors in `dashboard.tsx` and 1 pre-existing warning in `TimezoneContext.tsx`; verified both are already present on `main` and unrelated to this change.

### Configurations
```
serverConfig:
  [...]

authentication:
  defaultType: ["oauth","form"]
  showFirstTypeOnly: false   # true = login page shows only "oauth"; API still accepts form/basic
  oauth:
    issuer: "[....]"
  form:
    [....]

presetUsers:
  admin:
    password: admin
    privileges: ADMIN_USER_API
  trinoUser:
    password: trinoUserPassword
    privileges: USER

authorization:
  enableDefaultPrivilege: true   # false (default) = ignore defaultPrivilege and deny unmatched users
  defaultPrivilege: "USER"
  admin: (.*)ADMIN(.*)
  user: (.*)USER(.*)
  api: (.*)API(.*)

```
## UI View with defaultPrivilege
1. Login screen with OAuth
489212012-332754a9-7b26-4e40-9b49-df12efdbeeaf

2. As `andjiang` is nto in the presetUsers list, defaults over to `USER` permissions so can only view history
489212036-2af810e1-ff81-445d-913e-a475a2dc5722

## API Calls with OAuth enabled:
1. Query API endpoing for all backends
489212143-9a8720c0-c8b7-4248-bbfa-d1faea27ae19

2. Deactivate one backend with admin account
489212155-6ce2a988-8060-49c4-bbc0-70835b199f84

3. API with deactivate `trino-adhoc1`
489212176-0eca1532-0c10-4498-afda-cb6952dd4012

## Release notes

( x ) Release notes are required, with the following suggested text:

```markdown
* Add support for multiple authentication methods with fallback. `authentication.defaultType` now accepts either a single value (as before) or an ordered list such as `["oauth", "form"]`; existing single-string configurations keep working unchanged. The list sets the order methods are tried in and which methods the login page offers. Any configured `oauth` or `form` method is always accepted by the API even when it is not listed — so setups such as SSO for the browser plus form/basic for automation keep working — and the `/loginType` endpoint now returns the list of available methods instead of a single name. ([#1170](https://github.com/trinodb/trino-gateway/pull/1170))
* Add `authentication.showFirstTypeOnly` (default `false`) to offer only the first available authentication method on the login page while the API keeps accepting every configured method. ([#1170](https://github.com/trinodb/trino-gateway/pull/1170))
* Add `authorization.enableDefaultPrivilege` (default `false`) and `authorization.defaultPrivilege` to grant a baseline privilege to authenticated users who are not preset users and are not resolved via LDAP — useful when admins are preset and everyone else logs in via SSO; when disabled such users are denied. LDAP remains authoritative, so an empty group lookup denies access rather than granting the default, and preset users without configured privileges fall back to the default. ([#1170](https://github.com/trinodb/trino-gateway/pull/1170))
```


gateway-ha/src/main/java/io/trino/gateway/ha/config/AuthenticationConfiguration.java
Chaho12, line 1: Since this is BREAKING change to the user, should we have keep deprecated setter that maps old scalar into list? Jackson binds defaultType: key → this setter, old configs keep booting. 
Or should we just note in release note + add migration explanation. WDYT?

gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
```diff
@@ -172,16 +175,13 @@ else if (oauthManager != null) {
...
     {
-        String loginType;
-        if (formAuthManager != null) {
-            loginType = "form";
-        }
-        else if (oauthManager != null) {
-            loginType = "oauth";
+        List<String> loginTypes;
+        if (haGatewayConfiguration.getAuthentication() != null) {
+            loginTypes = haGatewayConfiguration.getAuthentication().getDefaultTypes();
```
Chaho12: /loginType advertises methods with no manager → dead UI button

LoginResource.java:180 returns getDefaultTypes() verbatim from config. If defaultTypes: ["oauth","form"] but only form: block configured, oauthManager is null. ChainedAuthFilter silently skips oauth (manager null), but /loginType still returns ["oauth",...]. 

Webapp login.tsx:17 renders first entry → OAuth button → click → /sso → LoginResource.java:82 throws "OAuth configuration is not setup". Login unusable.

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
```diff
@@ -55,26 +55,36 @@ public ChainedAuthFilter(
...
-                    new LbUnauthorizedHandler(defaultType)));
+                            authFilters.add(new BasicAuthFilter(
+                                    new ApiAuthenticator(formAuthManager, authorizationManager),
+                                    authorizer,
+                                    new LbUnauthorizedHandler("form")));
+                        }
+                    }
+                    default -> {}
+                }
+            }
```
Chaho12, line 86: If defaultTypes is empty/null, or there are only typos (["oath"]), → authFilters empty → filter() (line 96-105) loops nothing, throwing ForbiddenException on every request.
Also,  default -> {} at line 85 eats typos with no log

Then whole gateway shows 403s, zero clue why.

Fix: log unknown types, fail fast when chain ends up empty.

gateway-ha/src/main/java/io/trino/gateway/ha/config/AuthenticationConfiguration.java
AndyJiang99: Went with the deprecated-setter shim (option 1). Added a `@Deprecated setDefaultType(String)` that maps the old scalar into a single-element `defaultTypes` list, so existing `defaultType: "..."` configs keep booting unchanged — verified with a Jackson round-trip test (`TestAuthenticationConfiguration#testLegacyScalarDefaultTypeDeserializes`).

When both `defaultType` and `defaultTypes` are set, `defaultTypes` wins and we log a warning. Also added a "Migrating from `defaultType`" note to `docs/security.md`. Done in 79e7766.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
AndyJiang99: Fixed in 79e7766. `/loginType` no longer returns `getDefaultTypes()` verbatim — it now delegates to the new shared `AuthenticationTypeResolver`, which filters the configured list down to the types that actually have a manager wired up. So a config like `defaultTypes: ["oauth","form"]` with only a `form:` block resolves to `["form"]`, and the webapp's `data[0]` is always a usable method — no more dead OAuth button.

Both `ChainedAuthFilter` and `LoginResource` consume the same resolver now, so the filter chain and the advertised list can't drift apart. Added integration test `TestLoginTypeFiltersUnconfiguredMethods` asserting `/loginType` returns `["form"]` (and that form/basic auth still works) when oauth is listed but unconfigured.

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
AndyJiang99: Fixed in 79e7766. Extracted `AuthenticationTypeResolver.resolveEffectiveTypes(defaultTypes, oauthConfigured, formConfigured)` which:

- logs a warning for unknown types (the silent `default -> {}` is gone), and
- logs a warning for types listed without a matching manager, and
- **fails fast**: throws when `defaultTypes` is null/empty or when nothing usable remains, so the gateway aborts at startup instead of building an empty chain that 403s every request.

`ChainedAuthFilter` now builds its chain from the resolver's output (and the `default` branch throws rather than swallowing). Covered by `TestAuthenticationTypeResolver` — unknown-only, empty list, and listed-but-unconfigured all assert a thrown exception; order and filtering are asserted too.

vishalya:
Quick questions to understand the usage!
- Is the login form rendering when I use defaultTypes: ["oauth", "form"]. If that's expected, how should users log in using the form? let's say set of uses want to oauth and another set of users want to go with the form (username/password)
- Is there a way to set defaultPrivilege to deny everything by default? Something similar to `defaultPrivilege: None`?

AndyJiang99:
For question #1: The login screen that is rendered is ONLY the first type in the defaultTypes list. How this is designed right now, users cannot toggle between different sets of authentication on the UI. The addition of this switching is to allow for different types of authentication via the API itself.
For question #2: Good call! Addressed it in the latest commit! By default, if that field is empty, users would be denied already. But I added it so it's more explicit now, taking both as options

Chaho12:
Can you squash commit to single one, change commit message (especially the part Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>)

AndyJiang99:
> Can you squash commit to single one, change commit message (especially the part Co-authored-by: Copilot [223556219+Copilot@users.noreply.github.com](mailto:223556219+Copilot@users.noreply.github.com))

Done! Let me know if there's anything else!

Chaho12:
Please follow the guildeline for commit message.
- `Use the [imperative mood, as used in a command or request](https://en.wikipedia.org/wiki/Imperative_mood), in the subject line.`
https://trino.io/development/process.html#pull-request-and-commit-guidelines

AndyJiang99:
> Please follow the guildeline for commit message.
> 
> * `Use the [imperative mood, as used in a command or request](https://en.wikipedia.org/wiki/Imperative_mood), in the subject line.`
>   https://trino.io/development/process.html#pull-request-and-commit-guidelines

@Chaho12 addressed! Please review again! Thanks!

vishalya:
> For question #1: The login screen that is rendered is ONLY the first type in the defaultTypes list. How this is designed right now, users cannot toggle between different sets of authentication on the UI. The addition of this switching is to allow for different types of authentication via the API itself. 

for so, in this config `defaultTypes: ["oauth", "form"]`, how can a user request the login form?

AndyJiang99:
> > For question #1: The login screen that is rendered is ONLY the first type in the defaultTypes list. How this is designed right now, users cannot toggle between different sets of authentication on the UI. The addition of this switching is to allow for different types of authentication via the API itself.
> 
> for so, in this config `defaultTypes: ["oauth", "form"]`, how can a user request the login form?

For this, the user currently cannot request the login form itself, they would be only able to use the oauth method of logging in as that's the first one in the list, similar to how Trino Coordinator does this page rendering as well. If the user wants to do form but enable oauth2 maybe on the API side, they would need to do `defaultTypes: ["form", "oauth"]`. 

If you see this as an issue however, I would be happy to create a way where on the login page, the user can have the option to to use whatever method is available!

vishalya:
> > > For question #1: The login screen that is rendered is ONLY the first type in the defaultTypes list. How this is designed right now, users cannot toggle between different sets of authentication on the UI. The addition of this switching is to allow for different types of authentication via the API itself.
> > 
> > 
> > for so, in this config `defaultTypes: ["oauth", "form"]`, how can a user request the login form?
> 
> For this, the user currently cannot request the login form itself, they would be only able to use the oauth method of logging in as that's the first one in the list, similar to how Trino Coordinator does this page rendering as well. If the user wants to do form but enable oauth2 maybe on the API side, they would need to do `defaultTypes: ["form", "oauth"]`.
> 
> If you see this as an issue however, I would be happy to create a way where on the login page, the user can have the option to to use whatever method is available!

That would be great, please have the links or options available on the landing page. thanks!

gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
```diff
@@ -52,6 +58,27 @@ public Optional<String> getPrivileges(String username)
...
         }
-        return Optional.ofNullable(privs);
+
+        if (privs == null || privs.trim().isEmpty()) {
+            return defaultPrivileges();
+        }
+        return Optional.of(privs);
+    }
+
+    private Optional<String> defaultPrivileges()
```
vishalya: Can this logic be moved to the constructor, so that it's not computed every time.

vishalya (changes requested):
Authentication types are now configured and referenced in multiple places, resulting in repeated string literals and increasing the risk of typos. We should replace these raw strings with an enum and validate all configured values against the supported types at startup so invalid or misspelled entries fail fast.

gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
AndyJiang99: Yep makes sense. Addressed!

AndyJiang99:
@vishalya : Added a configuration to show all or only first of the authentication methods!

vishalya (approved):

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/AuthenticationTypeResolver.java
```diff
@@ -0,0 +1,70 @@
...
+                case OAUTH -> oauthConfigured;
+                case FORM -> formConfigured;
+            };
+            if (configured) {
+                effectiveTypes.add(authType);
+            }
+            else {
+                log.warn("authentication.defaultTypes lists \"%s\" but no matching authentication.%s block is configured; skipping it", authType.value(), authType.value());
+            }
+        }
```
Chaho12: Notice that there is no deduplication for same type.
e.g `defaultTypes: ["form","form"]` would have duplicate key and RadioGroup value would have conflict.

We should use enumset

gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
```diff
@@ -172,16 +181,23 @@ else if (oauthManager != null) {
...
-        String loginType;
-        if (formAuthManager != null) {
-            loginType = "form";
-        }
-        else if (oauthManager != null) {
-            loginType = "oauth";
+        List<String> loginTypes;
+        AuthenticationConfiguration authentication = haGatewayConfiguration.getAuthentication();
+        if (authentication != null) {
+            List<AuthenticationType> resolvedTypes = resolveEffectiveTypes(
```
Chaho12: re-resolves on every request, spamming logs and a 500 risk because it calls resolveEffectiveTypes on every request.

1. If defaultTypes lists a type without a matching configuration block, AuthenticationTypeResolver logs a warn per request rather than once at startup.
2. If the resolver throws (IllegalArgumentException / IllegalStateException), the login page gets a 500 instead of failing fast at boot.

The resolved list is derived purely from static configuration, so it should be computed once in the constructor.

Chaho12 (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
```diff
@@ -52,6 +58,27 @@ public Optional<String> getPrivileges(String username)
         else if (lbLdapClient != null) {
             privs = lbLdapClient.getMemberOf(username);
         }
-        return Optional.ofNullable(privs);
+
+        if (privs == null || privs.trim().isEmpty()) {
+            return defaultPrivilege;
+        }
+        return Optional.of(privs);
```
Chaho12: LbLdapClient.getMemberOf returns an empty string whenever the search yields no entries. It does not distinguish "this user genuinely has no group membership" from "the search returned nothing because ldapUserSearch / ldapUserBaseDn / ldapGroupMemberAttribute is misconfigured". With the new fallback, both cases now resolve to defaultPrivilege.

Why don't we keep the fallback for the preset-user and no-LDAP paths and treat LDAP as authoritative when it is configured.

```
    public Optional<String> getPrivileges(String username)
    {
        UserConfiguration user = presetUsers.get(username);
        if (user != null) {
            String privs = user.privileges();
            return (privs == null || privs.trim().isEmpty()) ? defaultPrivilege : Optional.of(privs);
        }
        if (lbLdapClient != null) {
            // LDAP is authoritative for group membership: an empty search result means
            // "no groups", not "grant the baseline privilege".
            String privs = lbLdapClient.getMemberOf(username);
            return (privs == null || privs.trim().isEmpty()) ? Optional.empty() : Optional.of(privs);
        }
        return defaultPrivilege;
    }
```

If you'd rather keep the current behavior (LDAP misses do fall back to defaultPrivilege), please call it out explicitly in the Default privileges section of docs/security.md, since an operator combining authorization.ldapConfigPath with defaultPrivilege would not expect a failed LDAP search to widen access. Gating it behind a separate flag would also work.

webapp/src/components/login.tsx
```diff
@@ -10,14 +10,29 @@ export function Login() {
...
+  const [loginTypes, setLoginTypes] = useState<string[]>([]);
+  const [loginType, setLoginType] = useState<string>();
 
   useEffect(() => {
     loginTypeApi().then(data => {
-      setLoginType(data);
+      if (data && data.length > 0) {
+        setLoginTypes(data);
+        setLoginType(data[0]);
+      }
```
Chaho12: what happens data is `"form"` a string?

```suggestion
      const types: string[] = Array.isArray(data) ? data : (data ? [data] : []);
      if (types.length > 0) {
        setLoginTypes(types);
        set
```

docs/security.md
```diff
@@ -36,16 +36,38 @@ for more details.
...
+>   defaultTypes: ["oauth", "form"]
+>   showFirstTypeOnly: true  # login page shows only oauth; API still accepts form/basic
+> ```
+
+> **Migrating from `defaultType`:** Earlier releases configured a single string
+> `defaultType: "form"`. That property is deprecated but still honored — it is
+> mapped to a single-element `defaultTypes` list so existing configurations keep
+> booting. If both `defaultType` and `defaultTypes` are set, `defaultTypes` takes
+> precedence. Switch to `defaultTypes`, as `defaultType` may be removed in a
+> future release.
```
Chaho12: Please update the docs because showFirstTypeOnly doesn't match the implementation

Chaho12 (changes requested):

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/AuthenticationTypeResolver.java
AndyJiang99: Good catch. `resolveEffectiveTypes`  now dedups. 

However, I deliberately used a  LinkedHashSet<AuthenticationType>  instead of  EnumSet as order is meaningful here since it drives both the `ChainedAuthFilter` fallback priority and the  `/loginType`  order the login page preselects ( `data[0]` ) and renders.  `EnumSet`  iterates in enum-declaration order rather than configured order, so  `defaultTypes: ["form","oauth"]`  would come back as  `[oauth, form]`  and silently flip the primary method (it'd also break  `testReturnsConfiguredTypesInOrder` ).  

`LinkedHashSet`  gives the dedup after while preserving configured order.

gateway-ha/src/main/java/io/trino/gateway/ha/resource/LoginResource.java
AndyJiang99: Addressed!

gateway-ha/src/main/java/io/trino/gateway/ha/security/AuthorizationManager.java
AndyJiang99: Agreed, and addressed.

I made LDAP authoritative when configured:  `getPrivileges`  now has three explicit branches:
1. preset user (their privileges, else the default)
2. LDAP configured (empty  `getMemberOf`  →  `Optional.empty()` /deny, not the default),
3. Neither (OAuth/SSO → default). So a misconfigured search (wrong base DN, member attribute, …) can't silently widen access; preset users with empty privileges still fall back to the default. 

Added unit tests via a  `@VisibleForTesting` constructor ( `testLdapMembershipTakesPrecedenceOverDefault`, `testLdapEmptyResultDeniesEvenWithDefaultPrivilege` ) and documented the rule in  `docs/security.md`. I also gated the whole fallback behind a new `authorization.enableDefaultPrivilege` flag (default  `false` ), so the baseline is opt-in.

webapp/src/components/login.tsx
AndyJiang99: addressed!

docs/security.md
AndyJiang99: addressed!

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
```diff
@@ -55,28 +56,32 @@ public ChainedAuthFilter(
...
-            authFilters.add(new BasicAuthFilter(
-                    new ApiAuthenticator(formAuthManager, authorizationManager),
-                    authorizer,
-                    new LbUnauthorizedHandler(defaultType)));
+                    authFilters.add(new BasicAuthFilter(
+                            new ApiAuthenticator(requireNonNull(formAuthManager), authorizationManager),
+                            authorizer,
+                            new LbUnauthorizedHandler(AuthenticationType.FORM)));
+                }
+            }
```
Chaho12, line 86: Are you thinking of supporting `backward compatibility` or not? Because if you do want it to support, you need to change code. If not, please mention breaking change in the suggested release notes. 

I see you mentioned that is still honored, but the code doesn't. 
```
> **Migrating from `defaultType`:** Earlier releases configured a single string
> `defaultType: "form"`. That property is deprecated but still honored — it is
```

PTAL at the suggested change from Claude. 

---

Here, `defaultTypes` silently drops configured authentication methods.

Before this change, `ChainedAuthFilter` built its chain purely from *manager presence*.
`defaultType` was never consulted for chain construction — it only picked the
`LbUnauthorizedHandler` redirect path.

After this change the chain is built from defaultTypes only, so this existing
configuration loses two filters on upgrade:
```yaml
authentication:
  defaultType: "oauth"   # deprecated, still honored -> ["oauth"]
  oauth: { ... }
  form: { ... }          # used by automation via Basic auth
```

- before: oauth, form, basic
- after:  oauth only -> every Basic auth request 403s

This is the case that worries me most, because the operator has no signal: the
deprecated property is honored, the gateway boots, only a deprecation warning is
logged — and the automation breaks anyway. Operators who migrate by hand to
defaultTypes: ["oauth"] while keeping a form: block hit the same thing.

It also means the PR removes, for existing users, exactly the capability it is
adding for new ones (humans via SSO, automation via Basic auth).

Chaho12 (changes requested):
@AndyJiang99 PTAL. Other than this, it looks good

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/ChainedAuthFilter.java
AndyJiang99: Thanks for the feedback on this! Took a look and made a couple of changes. Made it so we are leveraging how the authentication methods are constructed previously, while also keeping `defaultType` as well (and allowing it to be a scalar transformed into an array when it is one item), rather than having a new `defaultTypes` config

mosabua:
@AndyJiang99 @Chaho12 @vishalya  - lets get this merged.

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/AuthenticationTypeResolver.java
```diff
@@ -0,0 +1,123 @@
...
+     * @return the ordered, non-empty list of authentication types shown on the login page
+     */
+    public static List<AuthenticationType> resolveEffectiveTypes(List<String> defaultTypes, boolean oauthConfigured, boolean formConfigured)
+    {
+        List<AuthenticationType> resolved = resolveConfiguredListedTypes(defaultTypes, oauthConfigured, formConfigured, true);
+        if (resolved.isEmpty()) {
+            throw new IllegalStateException("No usable authentication methods configured; authentication.defaultType=%s but none had a matching configuration block".formatted(defaultTypes));
+        }
+        return resolved;
+    }
```
Chaho12, line 86: Can you test what happens if defaultType is "oauth" but form is empty block? Because it would have booted fine. Here, it would throw IllegalStateException, which is contradictory of what you said in PR message.

And please add test for this
```java
 assertThat(resolveEffectiveTypes(List.of("oauth"), false, true)).containsExactly(FORM);
```

```suggestion
    public static List<AuthenticationType> resolveEffectiveTypes(List<String> defaultTypes, boolean oauthConfigured, boolean formConfigured)
    {
        List<AuthenticationType> resolved = resolveConfiguredListedTypes(defaultTypes, oauthConfigured, formConfigured, true);
        if (resolved.isEmpty()) {
            return resolved;
        }
        List<AuthenticationType> chainTypes = resolveChainTypes(defaultTypes, oauthConfigured, formConfigured);
        log.warn("None of the authentication.defaultType entries %s has a matching configuration block; the login page falls back to the configured methods %s", defaultTypes, chainTypes);
        return chainTypes;
    }
```

Chaho12 (commented):
Looks good. Just final check please.

Also, here is my suggestion for the release note. 

* Add support for multiple authentication methods with fallback. The
  `authentication.defaultType` property now accepts an ordered list such as
  `["oauth", "form"]` that sets the order methods are tried in and which
  methods the login page offers, a single string keeps working unchanged, any
  configured `oauth` or `form` method stays accepted even when it is not
  listed, and the `/loginType` endpoint now returns the list of available
  methods instead of a single name.
  ([#1170](https://github.com/trinodb/trino-gateway/pull/1170))
* Add `authentication.showFirstTypeOnly` to offer only the first available
  authentication method on the login page while the API keeps accepting every
  configured method.
  ([#1170](https://github.com/trinodb/trino-gateway/pull/1170))
* Add `authorization.enableDefaultPrivilege` and
  `authorization.defaultPrivilege` to grant a baseline privilege to
  authenticated users who are not preset users. LDAP remains authoritative so
  an empty group lookup denies access, and preset users without configured
  privileges fall back to the default.
  ([#1170](https://github.com/trinodb/trino-gateway/pull/1170))

gateway-ha/src/main/java/io/trino/gateway/ha/security/util/AuthenticationTypeResolver.java
AndyJiang99: Thanks for the suggestion! Test added!

Issue #1171, opened 2026-07-10
Extending Trino Gateway to route to Kyuubi (and beyond Trino)


Chaho12:
# Extending Trino Gateway to route to Kyuubi (and beyond Trino)

> Related PR:  https://github.com/trinodb/trino-gateway/pull/1172
> (Adds a lightweight `PING` cluster-stats monitor — a prerequisite for treating Kyuubi as a gateway backend.)

## Motivation

Kyuubi already speaks the **Trino wire (CLI) protocol** through its Trino frontend. So today a client can do:

```
trino CLI  ->  Kyuubi  ->  Trino
```

Kyuubi acts as the entry point and delegates to Trino. But Kyuubi can also delegate to **other engines** — most importantly **Spark on YARN** — not only Trino.

The idea: put **trino-gateway** in front instead, so the gateway owns routing / load balancing / HA, and Kyuubi becomes just one of its backends:

```
trino CLI  ->  trino-gateway  ->  Kyuubi  ->  Spark on YARN
                              \->  Trino (or another Kyuubi)
```

This gives us gateway-level routing/HA in front of engines that trino-gateway could not previously reach — while keeping the same Trino CLI on the client side. Some internal code changes are still needed for full compatibility.

---

## Today (baseline)

```mermaid
flowchart LR
    CLI["Trino CLI<br/>(Trino wire protocol)"]
    K["Kyuubi<br/>(Trino frontend<br/>e.g. port 10999)"]
    T["Trino cluster"]
    SP["Spark engine<br/>on YARN"]

    CLI -->|HTTP| K
    K -->|delegate| T
    K -.->|delegate| SP
```

Kyuubi is the single entry point. No gateway-level routing, load balancing, or HA in front of it.

---

## Proposed

```mermaid
flowchart LR
    CLI["Trino CLI<br/>(Trino wire protocol)"]
    GW["trino-gateway<br/>• routing<br/>• load balancing / HA<br/>• PING health check"]
    K["Kyuubi<br/>(Trino frontend<br/>port 10999)"]
    SP["Spark engine<br/>on YARN"]
    T["Trino cluster<br/>or another Kyuubi"]

    CLI -->|HTTP| GW
    GW -->|proxy| K
    K -->|delegate| SP
    GW -.->|other backend| T

    GW -. "GET /v1/ping" .-> K
```

- **trino-gateway** becomes the single entry point and owns routing / LB / HA.
[...]

PR #1172, opened 2026-07-10
Add PING monitor for backend liveness checks

Chaho12:
## Description

Add PING monitor for backend liveness checks

https://github.com/trinodb/trino-gateway/issues/1171

## Release notes

```markdown
* Add PING monitor for backend liveness checks
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsPingMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsPingMonitor.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.clustermonitor;
+
+import io.airlift.http.client.HttpClient;
+import io.airlift.http.client.Request;
+import io.airlift.http.client.StatusResponseHandler.StatusResponse;
+import io.airlift.log.Logger;
+import io.trino.gateway.ha.config.MonitorConfiguration;
+import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+
+import java.net.URI;
+
+import static io.airlift.http.client.HttpStatus.OK;
+import static io.airlift.http.client.HttpUriBuilder.uriBuilderFrom;
+import static io.airlift.http.client.Request.Builder.prepareGet;
+import static io.airlift.http.client.StatusResponseHandler.createStatusResponseHandler;
+import static io.trino.gateway.ha.clustermonitor.MonitorUtils.shouldRetry;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Lightweight liveness monitor that performs an HTTP ping/pong against a
+ * backend. It issues a plain {@code GET} to a configurable ping endpoint
+ * (default {@code /v1/ping}) and marks the backend {@link TrinoStatus#HEALTHY}
+ * on a {@code 200} response, {@link TrinoStatus#UNHEALTHY} otherwise. Unlike
+ * the info/UI/JDBC monitors it carries no session or query dependency, so it
+ * will not trigger idle-session churn on the backend.
+ */
+public class ClusterStatsPingMonitor
+        implements ClusterStatsMonitor
+{
+    private static final Logger log = Logger.get(ClusterStatsPingMonitor.class);
+    private final HttpClient client;
+    private final String pingPath;
+    private final int retries;
+
+    public ClusterStatsPingMonitor(HttpClient client, MonitorConfiguration monitorConfiguration)
+    {
+        this.client = requireNonNull(client, "client is null");
+        this.pingPath = monitorConfiguration.getPingPath();
+        this.retries = monitorConfiguration.getRetries();
+    }
+
+    @Override
+    public ClusterStats monitor(ProxyBackendConfiguration backend)
+    {
+        return ClusterStats.builder(backend.getName())
+                .trinoStatus(checkStatus(backend.getProxyTo()))
+                .proxyTo(backend.getProxyTo())
+                .externalUrl(backend.getExternalUrl())
+                .routingGroup(backend.getRoutingGroup())
+                .build();
+    }
+
+    private TrinoStatus checkStatus(String baseUrl)
+    {
+        return checkStatus(baseUrl, retries);
+    }
+
+    private TrinoStatus checkStatus(String baseUrl, int retriesRemaining)
+    {
+        try {
+            Request request = prepareGet()
+                    .setUri(uriBuilderFrom(URI.create(baseUrl)).appendPath(pingPath).build())
+                    .build();
+            StatusResponse response = client.execute(request, createStatusResponseHandler());
+            if (response.getStatusCode() == OK.code()) {
+                return TrinoStatus.HEALTHY;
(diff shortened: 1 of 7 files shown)


ebyhr (changes requested):
Please add integration tests with Kyuubi.

```
GET /v1/info	❌	INFO_API monitor	not implemented
```

Why not contribute to Kyuubi?

Chaho12:
> Please add integration tests with Kyuubi.

Isn't it too much for simple ping-pong? I mean Kyuubi is an example that needs ping-pong, but I thought it is too early stage to do integration test with Kyuubi.

> Why not contribute to Kyuubi?

Since healthcheck is meant for client(Trino Gateway), and Kyuubi just supports simple health Ping-Pong health check, it is upto Gateway to check if backend cluster(kyuubi) is alive. So, I thought Gateway should have simple health check instead.

RoeyoOgen:
@coderabbitai

RoeyoOgen:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> gh pr view 1172 --repo trinodb/trino-gateway --json title,body,comments,reviews --jq '.'
> ```
> 
> Length of output: 3073
> 
> `@RoeyoOgen`, I don't have a strong stake in this repo's testing conventions, but here's a pragmatic take on the back-and-forth between Chaho12 and ebyhr:
> 
> **Both points have merit:**
> 
> * ebyhr's ask for integration tests is reasonable in general — testing a liveness monitor against a real (or realistic) backend gives higher confidence than unit tests with mocks alone, especially for network/protocol edge cases (timeouts, malformed responses, connection resets).
> * Chaho12's pushback is also valid — spinning up a full Kyuubi integration test harness is a heavier lift than the feature warrants at this stage, and the health-check contract here is intentionally generic (any backend exposing a simple ping endpoint), not Kyuubi-specific.
> 
> **A middle-ground suggestion:** Rather than a full Kyuubi integration test, consider adding a lightweight test that:
> 
> * Spins up a minimal in-process HTTP/TCP server (e.g., using a test utility server or `MockWebServer`-style helper) that mimics a ping endpoint's success/failure/timeout responses.
> * Verifies the `PingMonitor` correctly interprets healthy vs. unhealthy vs. unreachable states against that fake server.
> 
> This gets you integration-level confidence in the monitor's actual network behavior without requiring a real Kyuubi dependency in CI, and can be extended later if/when a proper Kyuubi integration suite is added. If the maintainers still want a true Kyuubi-backed test eventually, that could be tracked as a follow-up rather than blocking this PR.
> 
> Would it help if I opened a follow-up issue to track "Add Kyuubi-based integration test for PING monitor" so this PR isn't blocked on it? Let me know and I can file it with the appropriate context and backlink.

@Chaho12 what do you think about his suggestions?

@coderabbitai pls review the code as well

Chaho12:
> what do you think about his suggestions?

There is the integration test for spinning a HttpServer serving ping pong test at `TestClusterStatsPingMonitor.java`.
I added some more test on it.

Regarding integration test with real Kyuubi container, what about a separate PR, that introduces a kyuubi container image + service config.

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsPingMonitor.java
```diff
@@ -0,0 +1,92 @@
...
+            Request request = prepareGet()
+                    .setUri(uriBuilderFrom(URI.create(baseUrl)).appendPath(pingPath).build())
+                    .build();
+            StatusResponse response = client.execute(request, createStatusResponseHandler());
+            if (response.getStatusCode() == OK.code()) {
+                return TrinoStatus.HEALTHY;
+            }
+            if (shouldRetry(response.getStatusCode()) && retriesRemaining > 0) {
+                log.warn("Retrying ping health check on status %d for %s", response.getStatusCode(), request.getUri());
+                return checkStatus(baseUrl, retriesRemaining - 1);
```
RoeyoOgen, line 83: just a thought: would we prefer a simple loop instead of recursion?

gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
```diff
@@ -162,6 +162,54 @@ void testMetricsMonitor()
                 new MonitorConfiguration()));
     }
 
+    @Test
```
RoeyoOgen, line 165: i might be missing something but Current tests never verify that retry logic actually works.

perhpse add a test:
First response: 503
Second response: 200
Assert final status is HEALTHY

andythsu:
I'm not familiar with Kyuubi but it looks like it's a gateway just like Trino Gateway

The intention is that we want to put Trino Gateway in front of Kyuubi, which is in front of trino clusters

So it flows like Trino Gateway -> Kyuubi -> Trino cluster?

If I understand correctly, is it meaningful to return a simple 200? Because if Kyuubi is simply a gateway, the Trino clusters behind it it could be dead while Kyuubi is still responding with 200, making this health check virtually useless.

Am I understanding this architecture correctly? Would it be more meaningful to add a health check that actually probes the actual Trino cluster? 

At the end of the day, Trino Gateway was intended to be placed in front of Trino clusters, and the health check is the source of truth that reports the health status of Trino clusters. It's fine to chain Trino Gateway to another gateway, but if it doesn't get the source of truth, it's practically not useful.

Chaho12:
> So it flows like Trino Gateway -> Kyuubi -> Trino cluster?

@andythsu 
I explained it in more detailed in the #1171 , but the flow is as following
So, it practically integrates one (trino) cli to query both Trino SQL and Spark SQL. 
This would then allow setting routing rules for Kyuubi/Spark too!

For example, you can set so that if you do INSERT query (which is very heavy load in Trino) so why not run it in Spark automatically? And If you do SELECT, why wait for Spark to load, but rather let it run in Trino which is much faster? User only wants the results and is not that interested in the how part.

There are much steps to go to handle but this is just a start.

## AS-IS
```
trino CLI  ->  Kyuubi  -> trino
trino CLI -> trino gateway -> trino
```
## TO-BE

```
trino CLI  ->  trino-gateway  ->  Kyuubi  ->  Spark on YARN
                              \->  Trino (or another Kyuubi)
```
OR
```
trino CLI  ->  Kyuubi ->  Spark on YARN
                                  \-> trino-gateway-> Trino (or another Kyuubi)
                                  \-> trino
```

Chaho12:
### FYI: how Kyuubi picks an engine

Kyuubi does not inspect the SQL. The engine comes from one session conf,
`kyuubi.engine.type`, resolved as:

1. Conf the client sends at connect time (JDBC URL)
2. Per-user default in `kyuubi-defaults.conf` (`___<user>___.` prefix)
3. Server default — `SPARK_SQL`

...then overridden by the `SessionConfAdvisor` plugin, which has the
final say. Valid values: `SPARK_SQL`, `FLINK_SQL`, `TRINO`, `HIVE_SQL`,
`JDBC`, `DATA_AGENT`. All share the same Thrift port; only the conf
differs:
```
    beeline -u 'jdbc:hive2://kyuubi:10009/;#kyuubi.engine.type=TRINO'
    beeline -u 'jdbc:hive2://kyuubi:10009/;#kyuubi.engine.type=HIVE_SQL'
```

### Why it matters here

- **`engine.type` is per session, not per statement.** Kyuubi will not
  split one session's statements across engines, so "INSERT on Spark,
  SELECT on Trino" has to be decided *before* the session opens — by the
  gateway or the client, not by Kyuubi.
- **But it is overridable server-side.** A `SessionConfAdvisor` plugin
  can rewrite `kyuubi.engine.type` per user/session and beats whatever
  the client sent. That is the hook if we want engine policy centralized
  in Kyuubi rather than in the gateway.
- **A 200 from Kyuubi does not mean the engine works.** Engine identity
  is `user + engineType + shareLevel + subdomain`, and the engine is only
  launched when a session of that type opens. A Kyuubi frontend can pass
  `PING` while its Trino engine cannot start (missing
  `kyuubi.session.engine.trino.connection.url` / `.connection.catalog`)
  or its Hive engine has no `HIVE_HOME`. Re: @ebyhr's question — `PING`
  proves the frontend is up, not that any engine behind it is usable.

PR #1174, opened 2026-07-11
Add LDAP referral policy support

### LINKED ISSUE #1173: Support LDAP user searches from the Active Directory domain root
### Problem

Trino Gateway uses the Apache Directory LDAP client for form/basic LDAP authentication. When the LDAP user search base is narrow, for example an OU-specific base such as `OU=Users,DC=example,DC=com`, authentication works as expected.

However, in Active Directory deployments it is common to search from a broader base such as `DC=example,DC=com` so users can be found across multiple OUs or naming contexts. In that case, AD can return LDAP referrals. Gateway currently uses the `LdapConnectionTemplate` convenience search/authenticate overloads, which create a `SearchRequest` with the default referral behavior. When referrals are returned, authentication or authorization lookup can fail with errors like:

```text
CursorLdapReferralException
ERR_13100_FAILURE_ON_UNDERLYING_CURSOR
```

### Use case

Gateway should support LDAP deployments where users may exist in different OUs or naming contexts without failing just because the LDAP server returns referrals during user search.

### Proposed solution

Add a Gateway LDAP configuration option for referral handling, for example:

```yaml
ldapReferralPolicy: THROW
```

Supported values:

```text
THROW
FOLLOW
IGNORE
```

`THROW` should remain the default to preserve the existing behavior for users who do not configure the new setting.

Implementation-wise, Gateway can build an explicit Apache Directory `SearchRequest` for LDAP user searches, apply the configured referral policy with `followReferrals()` or `ignoreReferrals()`, and then execute the request through the request-based `LdapConnectionTemplate` overloads for both authentication and group membership lookup.

### Benefits

- Preserves backward compatibility by defaulting to the current behavior.
- Allows Active Directory users to opt into following referrals when using broader search bases.
- Keeps referral behavior centralized in Gateway's LDAP client wrapper instead of spreading it across call sites.
- Provides an explicit configuration surface for deployments that prefer to ignore referrals rather than chase them.

OmerRaifler:
## Description

Adds configurable LDAP referral handling for Gateway LDAP authentication and authorization lookups.

Previously, Gateway relied on Apache Directory `LdapConnectionTemplate` convenience search methods, which internally construct the LDAP `SearchRequest` and always use the default referral behavior. In Active Directory environments with broader search base DNs, LDAP searches may encounter referrals, causing authentication or group lookup to fail with `CursorLdapReferralException`.

This change introduces a new `ldapReferralPolicy` configuration option, allowing administrators to control how LDAP referrals are handled during authentication and authorization searches.

Supported values are:

```text
THROW
FOLLOW
IGNORE
```

The default is `THROW`, preserving the existing behavior unless users explicitly configure a different policy.

This change:
- Adds `LdapReferralPolicy` to `LdapConfiguration`.
- Builds explicit `SearchRequest` objects in `LbLdapClient` so the referral policy can be applied.
- Uses request-based `LdapConnectionTemplate` overloads for user authentication and group lookup.
- Adds unit tests covering the different referral policies.
- Documents the new `ldapReferralPolicy` configuration option.

Closes #1173.

## Release notes

```markdown
* Add support for configuring LDAP referral handling during Gateway LDAP authentication and authorization using the `ldapReferralPolicy` setting.
```

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbLdapClient.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbLdapClient.java
@@ -15,6 +15,8 @@ package io.trino.gateway.ha.security;
 
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.LdapConfiguration;
+import org.apache.directory.api.ldap.model.message.SearchRequest;
+import org.apache.directory.api.ldap.model.message.SearchRequestImpl;
 import org.apache.directory.api.ldap.model.message.SearchScope;
 import org.apache.directory.ldap.client.template.LdapConnectionTemplate;
 import org.apache.directory.ldap.client.template.exception.PasswordException;
@@ -29,7 +31,6 @@ import org.mockito.Spy;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.eq;
 
@@ -69,13 +70,20 @@ final class TestLbLdapClient
         String password = "pass1";
 
         String filter = ldapConfig.getLdapUserSearch().replace("${USER}", user);
+        SearchRequest searchRequest = new SearchRequestImpl();
+
+        Mockito
+                .when(ldapConnectionTemplate.newSearchRequest(
+                        eq(ldapConfig.getLdapUserBaseDn()),
+                        eq(filter),
+                        eq(SearchScope.SUBTREE),
+                        any(String[].class)))
+                .thenReturn(searchRequest);
 
         Mockito
                 .when(ldapConnectionTemplate.authenticate(
-                        ldapConfig.getLdapUserBaseDn(),
-                        filter,
-                        SearchScope.SUBTREE,
-                        password.toCharArray()))
+                        eq(searchRequest),
+                        any(char[].class)))
                 .thenReturn(null);
 
         // Success case
@@ -83,10 +91,8 @@ final class TestLbLdapClient
 
         Mockito
                 .when(ldapConnectionTemplate.authenticate(
-                        ldapConfig.getLdapUserBaseDn(),
-                        filter,
-                        SearchScope.SUBTREE,
-                        password.toCharArray()))
+                        eq(searchRequest),
+                        any(char[].class)))
                 .thenReturn(new TestLbLdapClient.DummyPasswordWarning());
 
         // Warning case
@@ -94,23 +100,12 @@ final class TestLbLdapClient
 
         Mockito
                 .when(ldapConnectionTemplate.authenticate(
-                        ldapConfig.getLdapUserBaseDn(),
-                        filter,
-                        SearchScope.SUBTREE,
-                        password.toCharArray()))
+                        eq(searchRequest),
+                        any(char[].class)))
                 .thenThrow(PasswordException.class);
 
         // failure case
         assertThat(lbLdapClient.authenticate(user, password)).isFalse();
-
-        assertThatThrownBy(() -> Mockito
-                .when(ldapConnectionTemplate.authenticate(
-                        ldapConfig.getLdapUserBaseDn(),
-                        filter,
-                        SearchScope.SUBTREE,
-                        password.toCharArray()))
-                .thenReturn(null))
-                .isInstanceOf(PasswordException.class);
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


RoeyoOgen:
@coderabbitai

RoeyoOgen:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@RoeyoOgen` I see you've mentioned me, but I don't see a specific request in your comment. Could you let me know what you'd like me to help with? For example, I can:
> 
> * Review the changes in this PR
> * Answer questions about the LDAP referral policy implementation
> * Regenerate the PR summary
> * Help with anything else related to this PR
> 
> Just let me know!

@coderabbitai review code please

RoeyoOgen:
@OmerRaifler as discussed can you pls attach relevant code sections from Trino so we could see compatibility?

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbLdapClient.java
```diff
@@ -119,16 +114,22 @@ void testMemberof()
...
+
+        Mockito
+                .when(ldapConnectionTemplate.newSearchRequest(
+                        eq(ldapConfig.getLdapUserBaseDn()),
+                        eq(filter),
+                        eq(SearchScope.SUBTREE),
+                        eq(attributes)))
+                .thenReturn(searchRequest);
 
         java.util.ArrayList users = new java.util.ArrayList();
```
Chaho12, line 127: nit. can you change this raw type? 
```suggestion
        List<LbLdapClient.UserRecord> users = new ArrayList<>();
```

Chaho12 (approved):
LGTM 👍

ebyhr:
Please follow a PR template.

OmerRaifler:
> @OmerRaifler as discussed can you pls attach relevant code sections from Trino so we could see compatibility?

I found the corresponding Trino implementation for reference:

https://github.com/trinodb/trino/commit/8bada1311a4fa80318c64cea5373c079705aa267

One difference between the implementations comes from the LDAP libraries that Trino and Trino Gateway use.

Trino uses JNDI, where referral handling is configured on the LDAP context via the `Context.REFERRAL` property. Before that change, Trino explicitly configured referrals to `follow`, and the commit introduced a boolean configuration to switch between `follow` and `ignore`.

Trino Gateway, on the other hand, uses Apache Directory's `LdapConnectionTemplate`. The convenience `search()` and `authenticate()` overloads construct the `SearchRequest` internally, so there isn't a way to configure referral handling when using those APIs. Because of that, I switched to the request-based overloads, built the `SearchRequest` explicitly, applied the referral policy to the request, and then passed it to the template.

I also chose to expose an enum (`THROW`, `FOLLOW`, `IGNORE`) instead of a boolean. Apache Directory naturally supports all three behaviors on a `SearchRequest`, and Gateway's existing behavior corresponded to `THROW` (i.e., the default behavior before this change). Keeping `THROW` as the default preserves backward compatibility, while allowing users to explicitly opt into following or ignoring referrals.

The same helper is used for both authentication and group lookup searches so that referral handling is applied consistently across both operations.

OmerRaifler:
After testing this implementation against a real Active Directory environment, I discovered that the current approach cannot provide the intended behavior because of a limitation in the Apache Directory LDAP client used by Gateway.

The configured policy is correctly propagated to the `SearchRequest`:

* `FOLLOW` results in `followReferrals=true`
* `IGNORE` results in `ignoreReferrals=true`

However, both policies still fail when Active Directory returns a `SearchResultReference`. The failure occurs inside the Apache Directory client (`LdapConnectionTemplate` / `EntryCursorImpl`), which raises `CursorLdapReferralException` even though the requested referral policy has been applied.

In other words, Apache Directory exposes referral policy settings on the `SearchRequest`, but unlike JNDI (which Trino uses), those settings do not result in automatic referral handling through the current `LdapConnectionTemplate`-based authentication flow.

The referrals started appearing after changing the user base DN from a specific OU to the domain root so that users from all OUs in the domain could authenticate. In a single-domain Active Directory environment, the domain-root subtree search causes AD to return subordinate referrals for additional naming contexts such as `DomainDnsZones` and `ForestDnsZones`.

I verified that adding the Active Directory Domain Scope control (`1.2.840.113556.1.4.1339`) to the search suppresses these subordinate referrals while still allowing users from every OU in the domain to be found.

Replacing the Apache Directory client with JNDI solely to support this behavior would be a much larger architectural change than the original scope of this feature.

Given these findings, I don't think this PR should expose `FOLLOW` and `IGNORE` policies that cannot be implemented reliably through the current Apache Directory client. I'll close this PR and open a new, narrower one proposing optional support for the Active Directory Domain Scope control instead.

Issue #892, opened 2026-02-04
Trino-Gateway Trying to indefinitely perform oauth2 authentication


TessaIO:
# Introduction
We recently integrated Trino-gateway in our infrastructure, and we currently have 2 Clusters (we can name them A and B for simplicity). Both of them support OAuth2 authentication via the same Google OAuth2 client, as recommended by the Trino-gateway security guide.
In the current setup, A is the default cluster and B gets traffic depending on specific headers.

# Issue
What happens is that, when we try to connect via JDBC to cluster A it works smoothly with OAuth2. However, when we set client_tags in the JDBC connection and try to connect to cluster B via OAuth2, the OAuth2 authentication window keeps opening up indefinitely.
In the logs, trino-gateway shows that it's trying to send the request to the correct cluster which is B.

# Configuration

```yaml
config.yaml: |-
  authentication:
    defaultType: oauth
    oauth:
      authorizationEndpoint: https://accounts.google.com/o/oauth2/v2/auth
      clientId: ${ENV:OAUTH_CLIENT_ID}
      clientSecret: ${ENV:OAUTH_CLIENT_SECRET}
      issuer: https://accounts.google.com
      jwkEndpoint: https://www.googleapis.com/oauth2/v3/certs
      privilegesField: email
      redirectUrl: https://URL/oidc/callback
      redirectWebUrl: https://URL/
      scopes:
      - openid
      - email
      - profile
      tokenEndpoint: https://oauth2.googleapis.com/token
      userIdField: email
  authorization:
    admin: ^(tessa)@xxxx\.xxx$
    api: ""
    user: ^(tessa)@xxxx\.xx$
  clusterStatsConfiguration:
    monitorType: INFO_API
  dataStore:
    driver: org.postgresql.Driver
    jdbcUrl: jdbc:postgresql://xxxxxxxxx:5432/prod
    password: ${ENV:PG_PASSWORD}
    queryHistoryHoursRetention: 168
    runMigrationsEnabled: true
    user: gateway
  pagePermissions:
    admin: dashboard_cluster_resource-group_selector_history
    api: null
    user: dashboard_history
  routing:
    defaultRoutingGroup: main
  routingRules:
    rulesConfigPath: /etc/routing-rules/routing_rules.yaml
    rulesEngineEnabled: true
    rulesType: FILE
  serverConfig:
    http-server.http.enabled: true
    http-server.http.port: 8080
    http-server.process-forwarded: true
    node.environment: prod
```

# Version
[...]

Issue #631, opened 2025-02-26
Inconsistent OAuth2 authentication failures when accessing Trino clusters through Trino Gateway


ndrluis:
We have some Trino clusters behind the Trino Gateway. The Trino Clusters login use both PASSWORD and OAUTH authentication. Sometimes when we try to login through the Trino Gateway, we receive the "OAuth2 authentication failed: Authentication response could not be verified" message, and this occurs inconsistently. When we login directly to the Trino Clusters, we never encounter this message.

<img width="451" alt="Image" src="https://github.com/user-attachments/assets/12e3828d-15b0-474c-9da7-948426a2bb83" />

jhorlima:
Hi @ndrluis,

I also experienced the same authentication problem. I ran some tests and based on them, I have come up with a hypothesis that the problem may be related to the way OAuth/OpenID "state" is being handled inside the Trino Gateway. In my view,

-   The "state" is established by a pod (either the Trino Gateway or the Trino AdHoc), and once routed to the external authentication provider (in my case, Google), the callback with the token and state ends up in another pod that does not know the original state established.
-   This unbalanced behavior is what causes the authentication failure with the error message: "OAuth2 authentication failed: Authentication response could not be verified."

I ran the following tests in order to test this hypothesis:

-   **Set the number of Trino Gateway pods to 1**: The problem was still present.
-   **Set the number of Trino AdHoc pods to 1**: The problem disappeared, I tried it 5 times and there were no further failures.

These observations imply that increasing the Trino AdHoc pods might be influencing the authentication process.

To better demonstrate this hypothesis, please see the following sequence diagram:

![Image](https://github.com/user-attachments/assets/fc3076c2-0e17-40e3-a0b6-ac64a347acd4)

As a temporary solution until a permanent one is devised, I suggest reducing the amount of Trino AdHoc pods to 1.

Secondly, I believe a complete solution may be achieved by:

1.  **Sticky sessions**: Configuring the load balancer so that the authentication callback is forwarded to the same pod that initiated the flow.
2.  **Shared state storage**: Utilizing a distributed cache system (e.g., Redis) so that the state generated by one pod is accessible by all pods in the flow.

That said, I would like to know if this hypothesis and these proposals have been considered or if there's another direction planned.

I'm available to help with further tests or to discuss potential implementations.

sdaberdaku:
[...]

PR #1188, opened 2026-07-17
Pin Trino OAuth2 token-exchange handshakes to the minting coordinator

AndyJiang99:
## Description

Trino OAuth2 SSO logins via the CLI/JDBC driver are slow and unreliable behind a multi-coordinator gateway: most logins complete in ~30–90s, but a meaningful minority time out after 2+ minutes and must be retried. The browser Web-UI login is unaffected. This PR adds an opt-in, dark-launch flag `routing.oauth2RoutingEnabled` (default **off**) that pins every leg of an in-flight OAuth2 handshake to the coordinator that minted it, fixing the slowness deterministically.

### Root cause

Trino's OAuth2 token-exchange handshake spans two independent HTTP clients that share only an `authId` (and its `authIdHash`), across these gateway legs:

1. **401 challenge** — the CLI's unauthenticated `POST /v1/statement` hits some coordinator, which mints a fresh `authId` and returns `WWW-Authenticate: Bearer x_redirect_server=".../oauth2/token/initiate/{authIdHash}", x_token_server=".../oauth2/token/{authId}"`.
2. **driver/CLI poll loop** — `GET /oauth2/token/{authId}` (a long-poll that blocks server-side up to ~10s per request).
3. **browser initiate redirect** — `GET /oauth2/token/initiate/{authIdHash}` → IdP.
4. **browser IdP callback** — `GET /oauth2/callback?state=…&code=…`; the `authIdHash` rides back through the IdP inside the signed `state` JWT (the `handler_state` claim).

Each coordinator keeps the token-exchange result in a **per-JVM in-memory cache** with no cross-coordinator propagation (`OAuth2TokenExchange`'s `LoadingCache`). The coordinator that processes the callback calls `setAccessToken(authIdHash, …)` on **its own** cache only (`OAuth2Service`), so only a poll that happens to land on that same coordinator wakes up and receives the token (`OAuth2TokenExchangeResource`). With round-robin load-balancing across multiple coordinators, the callback and the poll loop routinely land on different coordinators, the long-poll times out at ~10s, and the server returns a "keep polling" response — indefinitely.

The reason logins *eventually* succeed rather than hanging forever is accidental: each client re-rolls onto a different backend over time (cookie loss on retries for the OkHttp-based CLI; a cookieless poll path in the OSS Python client, whose `_OAuth2TokenBearer` sends `Request(...).prepare()` with an empty cookie jar, so `HTTPAdapter.send()` never attaches the sticky `TG.OAUTH2` cookie). Without that re-rolling, a poll would stay stuck on the wrong coordinator and fail `(N-1)/N` of the time. The slow tail is just the geometric distribution of how many polls it takes to randomly hit the right coordinator.

### This change

When `routing.oauth2RoutingEnabled` is on, the gateway records the `authId`/`authIdHash` advertised in the `401` `WWW-Authenticate` challenge against the responding backend (`ProxyRequestHandler` → `OAuth2RoutingStore`, a shared, cross-pod `oauth2_routing` table), then pins all three later legs — poll, initiate, and callback — back to that coordinator (`RoutingTargetHandler`):

- The callback is pinned by decoding the `state` JWT's `handler_state` claim (base64url, **without** verifying the signature — the gateway only selects a backend; the coordinator still validates the token, so a forged `state` can at worst mis-route, never bypass auth), so it's cookie-independent and works cross-pod.
- The pin is honored only while the coordinator is **active and healthy** — deactivation (an operator drain signal), an unhealthy verdict, or removal from the fleet drops the pin and forces re-auth, since only the minting coordinator holds the exchange state. The browser legs get a plain `401`; the driver poll loop gets Trino's token-poll failure contract (HTTP 200 + JSON `error`) so `HttpTokenPoller` ends cleanly.

Pins are pruned automatically; retention is configurable via `dataStore.oauth2RoutingRetention` (a duration, default `1h`). No new database is required — the `oauth2_routing` table ships in the existing Flyway migrations (MySQL/PostgreSQL/Oracle).

This mirrors the gateway's existing query-id sticky routing (`RoutingManager.findBackendForQueryId`) and reuses the same shared-datastore infrastructure, eliminating the routing mismatch deterministically while keeping the auth path load-balanced across coordinators — with no coordinator-side changes.

### Prior art and why it doesn't fully solve this

- **[#165](https://github.com/trinodb/trino-gateway/issues/165) (closed) + [#188](https://github.com/trinodb/trino-gateway/pull/188) (merged):** #165 reported that the **browser** OAuth2 handshake to the Web UI fails with multiple backends. #188 fixed it by adding cookie-based sticky routing (the `TG.OAUTH2` cookie / `OAuth2GatewayCookie`) for configurable `cookiePaths` such as `/oauth2/callback`, at lower precedence than query-id routing. That solves the **browser-only** flow, where a single browser session carries the cookie across the initiate → callback legs. It does **not** solve the CLI/JDBC token-exchange flow this PR targets: the CLI poll loop and the browser are separate processes with separate cookie jars, so the poll requests never carry the `TG.OAUTH2` cookie and still land on a different coordinator than the callback. This PR builds on #188, adding the cookie-independent, id-based pinning the CLI flow needs — and pinning the callback by the `state` JWT's `handler_state` instead of by cookie.
- **[#631](https://github.com/trinodb/trino-gateway/issues/631) (open):** intermittent `OAuth2 authentication failed: Authentication response could not be verified` through the gateway, never seen when connecting to a coordinator directly. This is the same multi-coordinator state mismatch — the callback/verification landing on a coordinator that doesn't hold the initiating handshake state — and cookie routing alone doesn't cover it consistently. Deterministically pinning every leg by `authId`/`authIdHash` (this PR) removes the mismatch that produces those inconsistent failures.
- **[#892](https://github.com/trinodb/trino-gateway/issues/892) (open):** JDBC OAuth2 to a non-default cluster (selected via client tags) loops the auth window indefinitely, with the gateway logs showing it routing to the correct cluster. This is exactly the failure this PR fixes when the target routing group has more than one coordinator: the cookieless JDBC poll loop never lands on the coordinator that minted and stores the token, so it polls forever. Pinning the poll and callback to the minting coordinator ends the loop.

### Other notes

- **Alternatives considered:** (a) routing all `/oauth2/*` to a single dedicated coordinator or small group — trivial, but makes that group a SPOF for SSO; (b) terminating OAuth2 at the gateway entirely so coordinators only validate the resulting bearer token — architecturally cleanest long-term, but a much larger change spanning both gateway and coordinator; (c) backing the coordinator's token cache with a shared store — adds an infra dependency in a security-critical path with awkward long-poll semantics. This PR takes the smallest deterministic fix that lives entirely in the gateway.
- Trino wire-format coupling (paths, challenge params, `state`/`handler_state`) is isolated in `OAuth2RoutingUtils`, verified against Trino 446 (`OAuth2Service`, `OAuth2CallbackResource`, `OAuth2TokenExchange.hashAuthId`).
- Fully behind `routing.oauth2RoutingEnabled`; when off, routing is unchanged. On any DB error the store falls back to normal (non-pinned) routing rather than failing the request.
- OSS code references: `OAuth2TokenExchange` (per-coordinator in-memory cache), `OAuth2TokenExchangeResource` (poll long-poll), `OAuth2Service` (`setAccessToken` on callback), `OAuth2GatewayCookie`, `RoutingTargetHandler`, `ProxyRequestHandler`.
- Docs: `docs/installation.md` (how to enable + DDL + retention) and `docs/routing-logic.md` (routing behavior for all three legs).
- Tests: `TestOAuth2RoutingUtils`, `TestProxyRequestHandlerOAuth2`, `TestRoutingTargetHandlerOAuth2`, `TestHaOAuth2RoutingStore`, `TestStochasticRoutingManager`, and the DB-migration tests all pass.

## Release notes

* Add opt-in `routing.oauth2RoutingEnabled` configuration that pins Trino OAuth2 token-exchange requests (CLI/JDBC poll, browser initiate, and IdP callback) to the coordinator that minted the handshake, fixing slow and unreliable OAuth2 logins behind a multi-coordinator deployment. Pin retention is configurable via `dataStore.oauth2RoutingRetention` (a duration, default `1h`). ({issue}`631`, {issue}`892`)

Also credits to : @xkrogen for helping ideate this solution!

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
@@ -22,6 +22,8 @@ public class DataStoreConfiguration
     private boolean queryHistoryEnabled = true;
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
+    // OAuth2 pins only need to outlive an in-flight handshake (minutes); keep them an hour.
+    private Integer oauth2RoutingHoursRetention = 1;
 
     public DataStoreConfiguration(String jdbcUrl, String user, String password, String driver, boolean queryHistoryEnabled, Integer queryHistoryHoursRetention, boolean runMigrationsEnabled)
     {
@@ -104,5 +106,15 @@ public class DataStoreConfiguration
     public void setRunMigrationsEnabled(boolean runMigrationsEnabled)
     {
         this.runMigrationsEnabled = runMigrationsEnabled;
+    }
+
+    public Integer getOauth2RoutingHoursRetention()
+    {
+        return this.oauth2RoutingHoursRetention;
+    }
+
+    public void setOauth2RoutingHoursRetention(Integer oauth2RoutingHoursRetention)
+    {
+        this.oauth2RoutingHoursRetention = oauth2RoutingHoursRetention;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
@@ -17,6 +17,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.DataStoreConfiguration;
+import io.trino.gateway.ha.persistence.dao.OAuth2RoutingDao;
 import io.trino.gateway.ha.persistence.dao.QueryHistoryDao;
 import jakarta.annotation.Nullable;
 import org.jdbi.v3.core.Jdbi;
@@ -32,6 +33,7 @@ import java.util.concurrent.TimeUnit;
 import static java.util.Objects.requireNonNull;
 
 public class JdbcConnectionManager
+        implements AutoCloseable
 {
     private static final Logger log = Logger.get(JdbcConnectionManager.class);
 
@@ -106,12 +108,36 @@ public class JdbcConnectionManager
     {
         executorService.scheduleWithFixedDelay(
                 () -> {
-                    log.info("Performing query history cleanup task");
-                    long created = System.currentTimeMillis() - TimeUnit.HOURS.toMillis(this.configuration.getQueryHistoryHoursRetention());
-                    jdbi.onDemand(QueryHistoryDao.class).deleteOldHistory(created);
+                    // Each cleanup is isolated in its own try-catch: scheduleWithFixedDelay silently
+                    // suppresses all future runs once a task throws, so a failure in one cleanup (e.g. a
+                    // table not yet created during a rolling deploy) must neither skip the other cleanup
+                    // nor propagate out of the scheduled task.
+                    try {
+                        log.info("Performing query history cleanup task");
+                        long created = System.currentTimeMillis() - TimeUnit.HOURS.toMillis(this.configuration.getQueryHistoryHoursRetention());
+                        jdbi.onDemand(QueryHistoryDao.class).deleteOldHistory(created);
+                    }
+                    catch (RuntimeException e) {
+                        log.warn(e, "Query history cleanup failed; will retry on next run");
+                    }
+
+                    try {
+                        log.info("Performing OAuth2 routing cleanup task");
+                        long oauthCutoff = System.currentTimeMillis() - TimeUnit.HOURS.toMillis(this.configuration.getOauth2RoutingHoursRetention());
+                        jdbi.onDemand(OAuth2RoutingDao.class).deleteOldOAuth2Pins(oauthCutoff);
+                    }
+                    catch (RuntimeException e) {
+                        log.warn(e, "OAuth2 routing cleanup failed; will retry on next run");
+                    }
                 },
                 1,
                 120,
                 TimeUnit.MINUTES);
+    }
+
+    @Override
+    public void close()
+    {
+        executorService.shutdownNow();
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/OAuth2RoutingDao.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/OAuth2RoutingDao.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.persistence.dao;
+
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+public interface OAuth2RoutingDao
+{
+    @SqlQuery(
+            """
+            SELECT backend_url FROM oauth2_routing
+            WHERE oauth_id = :oauthId
+            """)
+    String findBackendByOAuthId(String oauthId);
+
+    @SqlUpdate(
+            """
+            INSERT INTO oauth2_routing (oauth_id, backend_url, created)
+            VALUES (:oauthId, :backendUrl, :created)
+            """)
+    void insert(String oauthId, String backendUrl, long created);
+
+    @SqlUpdate(
+            """
+            DELETE FROM oauth2_routing
+            WHERE oauth_id = :oauthId
+            """)
+    void delete(String oauthId);
+
+    @SqlUpdate(
+            """
+            DELETE FROM oauth2_routing
+            WHERE created < :created
+            """)
+    void deleteOldOAuth2Pins(long created);
+}
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaOAuth2RoutingStore.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/HaOAuth2RoutingStore.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import com.google.inject.Inject;
+import io.airlift.log.Logger;
+import io.trino.gateway.ha.persistence.dao.OAuth2RoutingDao;
+import org.jdbi.v3.core.Jdbi;
+
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Shared, cross-pod {@link OAuth2RoutingStore} backed by the {@code oauth2_routing} table.
+ * <p>
+ * The DB is the single source of truth — there is no cache. Pins are written once per handshake and
+ * read only a handful of times, so the DB load is negligible and a cache would only add staleness.
+ * Every DB call is wrapped so that if the DB is unreadable for any reason, {@link #findBackend}
+ * returns empty and the request falls back to normal (non-pinned) routing.
+ */
+public class HaOAuth2RoutingStore
+        implements OAuth2RoutingStore
+{
+    private static final Logger log = Logger.get(HaOAuth2RoutingStore.class);
+
+    private final Jdbi jdbi;
+    private final OAuth2RoutingDao dao;
+
+    @Inject
+    public HaOAuth2RoutingStore(Jdbi jdbi)
+    {
+        this.jdbi = requireNonNull(jdbi, "jdbi is null");
+        this.dao = jdbi.onDemand(OAuth2RoutingDao.class);
+    }
+
+    @Override
+    public void setBackend(String oauthId, String backend)
+    {
+        try {
+            // Idempotent upsert without dialect-specific syntax: a repeated challenge for the same
+            // id (or a re-pin) replaces the row rather than failing on the primary key. Run as one
+            // transaction on a single connection so the delete+insert is atomic and concurrent pins
+            // of the same id serialize instead of racing into a primary-key violation.
+            jdbi.useTransaction(handle -> {
+                OAuth2RoutingDao txDao = handle.attach(OAuth2RoutingDao.class);
+                txDao.delete(oauthId);
+                txDao.insert(oauthId, backend, System.currentTimeMillis());
+            });
+        }
+        catch (RuntimeException e) {
+            log.warn(e, "Failed to persist OAuth2 pin [%s]", oauthId);
+        }
+    }
+
+    @Override
+    public Optional<String> findBackend(String oauthId)
+    {
+        try {
+            return Optional.ofNullable(dao.findBackendByOAuthId(oauthId));
+        }
+        catch (RuntimeException e) {
+            // DB unreadable: fall back to normal (non-pinned) routing rather than failing the request.
+            log.warn(e, "Failed to load OAuth2 pin [%s]; falling back to normal routing", oauthId);
+            return Optional.empty();
+        }
+    }
+
+    @Override
+    public void removeBackend(String oauthId)
+    {
+        try {
+            dao.delete(oauthId);
+        }
+        catch (RuntimeException e) {
+            log.warn(e, "Failed to remove OAuth2 pin [%s]", oauthId);
+        }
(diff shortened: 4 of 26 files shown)


vishalya:
Thanks for the clean solution! The other OAuth2 approaches were practical workarounds, but this feels like the right fix. I will take a closer look as time permits.

vishalya:
@coderabbitai full review

vishalya:
Let's add some tests -
- The write path has no coverage. `recordOAuth2Challenge` has no test: nothing asserts that a `401` with a challenge writes both ids, that a `non-401` writes nothing, or that a `401` without `x_token_server` writes nothing. There's also no test that the feature is fully inert when `oauth2RoutingEnabled=false` - the key property for a dark launch.
- `TestStochasticRoutingManager.testIsBackendActiveAndHealthy` only exercises exact `scheme://host:port` values, add a path-prefixed and a *mixed-case* proxyTo case.

gateway-ha/src/main/java/io/trino/gateway/ha/router/OAuth2RoutingUtils.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+        try {
+            byte[] payload = Base64.getUrlDecoder().decode(state.substring(firstDot + 1, secondDot));
+            JsonNode claim = OBJECT_MAPPER.readTree(payload).get(HANDLER_STATE_CLAIM);
+            if (claim == null || claim.isNull() || isNullOrEmpty(claim.asText())) {
+                return Optional.empty();
+            }
+            return Optional.of(claim.asText());
+        }
+        catch (Exception e) {
```
vishalya, line 234: Narrow to `IllegalArgumentException | IOException;`

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -22,6 +22,8 @@ public class DataStoreConfiguration
     private boolean queryHistoryEnabled = true;
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
+    // OAuth2 pins only need to outlive an in-flight handshake (minutes); keep them an hour.
+    private Integer oauth2RoutingHoursRetention = 1;
```
vishalya, line 26: Let's get the unit along with the number like below -
```
dataStore:
  oauth2RoutingRetention: "10m"    # or "90s", "2h"

```

gateway-ha/src/main/java/io/trino/gateway/proxyserver/ProxyRequestHandler.java
```diff
@@ -300,6 +314,29 @@ private ProxyResponse recordBackendForQueryId(
...
+     * {@code WWW-Authenticate} challenge of its {@code 401}; recording both here lets every later
+     * request of the handshake be routed back here. See {@link OAuth2RoutingUtils}.
+     */
+    private ProxyResponse recordOAuth2Challenge(URI remoteUri, ProxyResponse response)
+    {
+        if (response.statusCode() != UNAUTHORIZED.getStatusCode()) {
+            return response;
+        }
+        String backend = getRemoteTarget(remoteUri);
+        response.headers().entries().stream()
```
vishalya, line 329: `response.headers().get(HeaderName.of(WWW_AUTHENTICATE))` would be quick and easy.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
```diff
@@ -22,6 +22,8 @@ public class DataStoreConfiguration
     private boolean queryHistoryEnabled = true;
     private Integer queryHistoryHoursRetention = 4;
     private boolean runMigrationsEnabled = true;
+    // OAuth2 pins only need to outlive an in-flight handshake (minutes); keep them an hour.
```
vishalya, line 25: Is this really 1hr or 120mins? the jdbcConnectionManager cleanup runs on a schedule of 120mins.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/JdbcConnectionManager.java
```diff
@@ -106,12 +108,36 @@ private void startCleanUps()
...
+                        log.warn(e, "OAuth2 routing cleanup failed; will retry on next run");
+                    }
                 },
                 1,
                 120,
                 TimeUnit.MINUTES);
     }
+
+    @Override
+    public void close()
```
vishalya, line 139: Add annotation `@PreDestroy` to cleanup.

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaOAuth2RoutingStore.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Inject
+    public HaOAuth2RoutingStore(Jdbi jdbi)
+    {
+        this.jdbi = requireNonNull(jdbi, "jdbi is null");
+        this.dao = jdbi.onDemand(OAuth2RoutingDao.class);
+    }
+
+    @Override
+    public void setBackend(String oauthId, String backend)
```
vishalya, line 49: Let's try to insert both records for oauthId here in a single transaction. Apart from the performance, it would also keep the state for oauth2 consistent as we will need both records or none.
so the new signature of the method would look like -  
`public void setBackends(Set<String> oauthIds, String backend)`

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/OAuth2RoutingDao.java
```diff
@@ -0,0 +1,48 @@
...
+    @SqlUpdate(
+            """
+            INSERT INTO oauth2_routing (oauth_id, backend_url, created)
+            VALUES (:oauthId, :backendUrl, :created)
+            """)
+    void insert(String oauthId, String backendUrl, long created);
+
+    @SqlUpdate(
+            """
+            DELETE FROM oauth2_routing
```
vishalya, line 37: How many records will be deleted here? do we need any batching support?

gateway-ha/src/main/java/io/trino/gateway/ha/router/HaOAuth2RoutingStore.java
AndyJiang99: Done — replaced `setBackend` with `setBackends(Set<String> oauthIds, String backend)`; the ids are written (delete+insert) in one `jdbi.useTransaction`, so a handshake's pins are all-or-nothing. Interface, impl, caller, and tests updated.

gateway-ha/src/main/java/io/trino/gateway/ha/persistence/dao/OAuth2RoutingDao.java
AndyJiang99: Kept it a single statement — the delete set is small and bounded. Each handshake writes exactly 2 rows, retention is short (default 1h), and the `created` column is indexed, so each sweep is a bounded index-range delete of just the rows from OAuth2 logins in the window (interactive SSO — thousands at most), not a full scan. I avoided batching because a bounded/`LIMIT` delete has no portable form across MySQL/Postgres/Oracle (Postgres has no `DELETE … LIMIT`; needs a ctid/subquery, Oracle needs `ROWNUM`), which would reintroduce the dialect-specific SQL this DAO deliberately avoids. If we ever expect large backlogs (e.g. a much longer retention), happy to add a batched variant then.

gateway-ha/src/main/java/io/trino/gateway/ha/config/DataStoreConfiguration.java
AndyJiang99: They're independent: `oauth2RoutingRetention` (default `1h`) is the minimum age before a pin is eligible for pruning; the 120-min figure is the fixed sweep cadence it shares with the existing query-history cleanup. So a pin lives at least its retention and is removed on the first sweep after that (worst case ~retention + 120m). A pin only needs to outlive an in-flight handshake (minutes), and an expired-but-not-yet-swept pin is harmless since authIds are unique per handshake and never re-matched. I documented this lower-bound behavior in the field comment. If you'd prefer sub-120m retention values to take effect promptly, I can make the sweep interval configurable / derive it from the retention — it just seemed like over-engineering for a minutes-scale need.

developer-Bushido:
Thanks for this PR - it solves a real problem for us. We run several gateway instances behind an L3 ECMP VIP (no stickiness) in front of Trino 483 clusters with OAuth2 + JDBC `externalAuthentication`, and pinning the handshake by id is exactly what we need. We cherry-picked 6489c489eb onto `21`, built it, ran the new tests (all green) and applied the migrations against TiDB 8.5 (MySQL protocol) without problems.

Before running it in production we reviewed it with a security focus, and a few things came up that we think should be addressed here. Sharing them with the evidence, plus how we are addressing them in our internal build.

### 1. Raw `authId` is persisted, and it is a bearer credential (security)

In Trino (checked against 483, `io.trino.server.security.oauth2`):
- `GET /oauth2/token/{authId}` needs no authentication (`OAuth2TokenExchangeResource`), and it returns the token for `sha256(authId)` once the callback has completed.
- With refresh tokens off, that is the IdP access token itself.
- `DELETE /oauth2/token/{authId}` after delivery does not remove the result: `OAuth2TokenExchange.dropToken` calls `set` on a future that is already set, which is a no-op. The entry lives until it expires: `challenge-timeout` (15m by default) + 100s.

The PR stores both the raw `authId` and the `authIdHash` in `oauth2_routing`, as two rows (`recordOAuth2Challenge` → `setBackends`). Anyone with read access to the gateway database (including DB-wide admin/monitoring accounts, statement logs or processlists) can take a fresh UUID from the table, poll `/oauth2/token/<uuid>` through the gateway (which even routes the poll to the right coordinator) and receive the user's token alongside the user. The window is about 16m40s per login.

The ids also reach the logs:
- `HaOAuth2RoutingStore`'s WARN messages include them when a DB call fails.
- The WARN in `RoutingTargetHandler` ("OAuth2 pinned backend ... is unavailable for [%s]") prints `getRequestURI()`, i.e. `/oauth2/token/<authId>`.
- Unrelated to this PR, but worth noting, several places on `21` already print the full URI:
  - `RoutingTargetHandler.logRewrite` logs every rerouted request at INFO, on both sides of the arrow (the target is built by `buildUriWithNewCluster`, so it contains the path and the query string);
  - the WARN `Proxy request failed: %s %s` in `ProxyRequestHandler`;
  - DEBUG lines in `ProxyUtils`, `QueryMetadataParser` and `FileBasedRoutingGroupSelector` (the last one logs the whole request object, query string included).

  All of these can contain `/oauth2/token/<authId>`, `/oauth2/token/initiate/<hash>` or the callback's `code`/`state`. The airlift HTTP request log, on by default, records the same URIs.

Suggested fix: persist one row per handshake, keyed by a one-way value that all three legs can derive. For example `pinKey = sha256("<prefix>:" + authIdHash)`:
- poll leg: `authIdHash = sha256hex(UUID.fromString(pathSegment).toString())`, and no lookup if the segment is not a UUID;
- initiate leg: `authIdHash` = the path segment (validate 64 hex chars);
- callback leg: `authIdHash` = the `handler_state` claim (validate 64 hex chars);
- 401 challenge: derive the key from `x_token_server` only.

Nothing stored can then be turned back into a usable `authId` or `authIdHash`, the table gets half the rows, and routing stays correct. We would also drop the ids from the new log lines, and mask `/oauth2/token/*` and the callback query string wherever a request or backend URI is logged.

### 2. Every OAuth2 401 now costs a DB transaction (availability)

Trino generates a fresh `authId` for every failed authentication - anonymous requests, expired tokens, a wrong Basic password (`OAuth2Authenticator`). With the flag on, each such 401 leads to a transaction (delete + insert per id), and the response waits for it. As far as we can see, `Jdbi.create(url, user, password)` has no connection pool, so every operation opens a new connection.

At around 1,000 anonymous requests/s, which one laptop can easily produce, that is millions of rows per hour. Rows live for `oauth2RoutingRetention` (1h) plus up to the 120-minute sweep interval, and the sweep is a single large `DELETE`. Before this PR, an anonymous 401 did no DB work at all.

Suggestions:
- a rate limit on pin writes: when it is exceeded, skip the pin and route as today. It should be per client address, or generous enough, that an anonymous flood cannot use up the budget and leave real logins unpinned;
- a shorter default retention, which only needs to exceed `challenge-timeout` + 100s (e.g. 20m);
- a dedicated, more frequent sweep that deletes in batches.

### 3. One instance's health view deletes the shared pin

When the pinned backend is not `isBackendActiveAndHealthy` in *this* instance's in-memory view, `RoutingTargetHandler` calls `oauth2RoutingStore.removeBackend(oauthId)` and returns a terminal error. A freshly started instance sees every backend as UNKNOWN until its first monitor round. With the poller's requests spread across instances (any LB without stickiness), a rolling restart of the gateway fleet, or a flapping health check on one instance, deletes pins that the other instances would have honoured, and the login fails fleet-wide.

Suggestion: in that case, skip the pin (route normally, or return the retryable response) instead of deleting shared state. Let retention clean up stale rows.

### 4. `V5` collides with #947, and `repair()` hides it

#947 adds `V5__add_tags_to_gateway_backend.sql` for all three databases. `FlywayMigration.migrate` runs `flyway.repair()` before `migrate()`, so whichever `V5` lands second would have its checksum/description realigned to the recorded one and would never run on databases that already applied the other `V5`. Worth coordinating the version number with #947 before either one merges.

### Smaller notes

- The comment says the wire format was checked against Trino 446. We re-checked it against 483 (`OAuth2Authenticator`, `OAuth2TokenExchangeResource`, `OAuth2Service`): it still matches.
- Since the pin write runs on an unbounded cached thread pool and Connector/J has no connect/socket timeout by default, a slow or unreachable DB under a 401 flood keeps adding blocked threads. Documenting `connectTimeout`/`socketTimeout` for the `jdbcUrl` would help.

We have implemented 1-3 (and a non-colliding migration number) on top of your commit in our internal build, with tests. We would be glad to contribute that code here once our CLA is in place. In the meantime, feel free to take any of the above.

PR #1197, opened 2026-07-22
Add support for Active Directory LDAP Domain Scope control

### LINKED ISSUE #1173: Support LDAP user searches from the Active Directory domain root

OmerRaifler:
## Description

Adds optional support for the Microsoft Active Directory Domain Scope LDAP control.

When `ldapUserBaseDn` is configured to the Active Directory domain root (for example, `DC=example,DC=com`), subtree searches can discover users from all organizational units. However, Active Directory may also return subordinate referrals for naming contexts such as `DomainDnsZones` and `ForestDnsZones`, which can cause authentication to fail when using the Apache Directory LDAP client.

This change introduces an optional configuration that enables the Microsoft Active Directory Domain Scope control (OID `1.2.840.113556.1.4.1339`) for LDAP user searches. When enabled, the search is restricted to the current domain naming context, allowing users across all OUs within the domain to be found without returning subordinate referrals.

The feature is specific to Microsoft Active Directory and is disabled by default.

This PR supersedes the previous referral-handling approach #1174 .

During the investigation, it became clear that the underlying requirement was not to follow LDAP referrals, but to prevent Active Directory from generating subordinate referrals when searching from the domain root. The Microsoft Active Directory Domain Scope control provides the desired behavior without requiring generic LDAP referral handling.

Closes #1173 

## Release notes

```markdown
* Add optional support for the Microsoft Active Directory Domain Scope LDAP control, allowing LDAP user searches from the Active Directory domain root without returning subordinate referrals.
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
@@ -39,6 +39,7 @@ public class LdapConfiguration
     private Integer poolMaxTotal;
     private Integer poolMinIdle;
     private boolean poolTestOnBorrow;
+    private boolean ldapAdDomainScopeControl;
 
     public LdapConfiguration(
             String ldapHost,
@@ -237,5 +238,15 @@ public class LdapConfiguration
     public void setPoolTestOnBorrow(boolean poolTestOnBorrow)
     {
         this.poolTestOnBorrow = poolTestOnBorrow;
+    }
+
+    public boolean isLdapAdDomainScopeControl()
+    {
+        return ldapAdDomainScopeControl;
+    }
+
+    public void setLdapAdDomainScopeControl(boolean ldapAdDomainScopeControl)
+    {
+        this.ldapAdDomainScopeControl = ldapAdDomainScopeControl;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
@@ -48,6 +48,7 @@ public class LdapConfiguration
     private Integer poolMaxTotal;
     private Integer poolMinIdle;
     private boolean poolTestOnBorrow;
+    private boolean ldapAdDomainScopeControl;
 
     public LdapConfiguration(
             String ldapHost,
@@ -291,5 +292,15 @@ public class LdapConfiguration
     public void setPoolTestOnBorrow(boolean poolTestOnBorrow)
     {
         this.poolTestOnBorrow = poolTestOnBorrow;
+    }
+
+    public boolean isLdapAdDomainScopeControl()
+    {
+        return ldapAdDomainScopeControl;
+    }
+
+    public void setLdapAdDomainScopeControl(boolean ldapAdDomainScopeControl)
+    {
+        this.ldapAdDomainScopeControl = ldapAdDomainScopeControl;
     }
 }
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
@@ -18,6 +18,7 @@ import io.trino.gateway.ha.config.LdapConfiguration;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.exception.LdapException;
+import org.apache.directory.api.ldap.model.message.SearchRequest;
 import org.apache.directory.api.ldap.model.message.SearchScope;
 import org.apache.directory.ldap.client.api.DefaultLdapConnectionFactory;
 import org.apache.directory.ldap.client.api.LdapClientTrustStoreManager;
@@ -34,13 +35,24 @@ import java.util.List;
 public class LbLdapClient
 {
     private static final Logger log = Logger.get(LbLdapClient.class);
-    private LdapConnectionTemplate ldapConnectionTemplate;
-    private LdapConfiguration config;
-    private UserEntryMapper userRecordEntryMapper;
+    private final LdapConnectionTemplate ldapConnectionTemplate;
+    private final LdapConfiguration config;
+    private final UserEntryMapper userRecordEntryMapper;
 
     public LbLdapClient(LdapConfiguration ldapConfig)
     {
+        this(ldapConfig, createLdapConnectionTemplate(ldapConfig));
+    }
+
+    LbLdapClient(LdapConfiguration ldapConfig, LdapConnectionTemplate ldapConnectionTemplate)
+    {
         config = ldapConfig;
+        this.ldapConnectionTemplate = ldapConnectionTemplate;
+        userRecordEntryMapper = new UserEntryMapper(config.getLdapGroupMemberAttribute());
+    }
+
+    private static LdapConnectionTemplate createLdapConnectionTemplate(LdapConfiguration ldapConfig)
+    {
         LdapConnectionConfig connectionConfig = new LdapConnectionConfig();
         connectionConfig.setLdapHost(ldapConfig.getLdapHost());
         connectionConfig.setLdapPort(ldapConfig.getLdapPort());
@@ -61,7 +73,7 @@ public class LbLdapClient
         DefaultLdapConnectionFactory defaultFactory =
                 new DefaultLdapConnectionFactory(connectionConfig);
 
-        // A single connection and keep it alive
+        // Configure the LDAP connection pool.
         GenericObjectPoolConfig poolConfig = new GenericObjectPoolConfig();
         poolConfig.setMaxIdle(ldapConfig.getPoolMaxIdle());
         poolConfig.setMaxTotal(ldapConfig.getPoolMaxTotal());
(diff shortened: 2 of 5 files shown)


ebyhr:
@coderabbitai full review

gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
```diff
@@ -238,4 +239,14 @@ public void setPoolTestOnBorrow(boolean poolTestOnBorrow)
     {
         this.poolTestOnBorrow = poolTestOnBorrow;
     }
+
+    public boolean isLdapAdDomainScopeControl()
+    {
+        return this.ldapAdDomainScopeControl;
```
ebyhr: Remove redundant `this.`

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -34,13 +35,24 @@
...
 
     public LbLdapClient(LdapConfiguration ldapConfig)
+    {
+        this(ldapConfig, createLdapConnectionTemplate(ldapConfig));
+    }
+
+    LbLdapClient(LdapConfiguration ldapConfig, LdapConnectionTemplate ldapConnectionTemplate)
     {
         config = ldapConfig;
+        this.ldapConnectionTemplate = ldapConnectionTemplate;
```
ebyhr: `requireNonNull` is missing.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -137,18 +152,18 @@ String getMemberOf()
...
             implements EntryMapper<UserRecord>
     {
-        String memberOf;
+        private final String memberOfAttribute;
 
-        public UserEntryMapper(String memberOfAttr)
+        public UserEntryMapper(String memberOfAttribute)
         {
-            memberOf = memberOfAttr;
+            this.memberOfAttribute = memberOfAttribute;
```
ebyhr: `requireNonNull` is missing.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -104,24 +112,31 @@ public String getMemberOf(String user)
...
 
+    private SearchRequest newUserSearchRequest(String filter, String... attributes)
+    {
+        SearchRequest searchRequest = ldapConnectionTemplate.newSearchRequest(
+                config.getLdapUserBaseDn(),
+                filter,
+                SearchScope.SUBTREE,
+                attributes);
+
+        return searchRequest;
```
ebyhr, line 194: Remove a redundant local variable.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -131,9 +140,35 @@ private SearchRequest newUserSearchRequest(String filter, String... attributes)
...
+
+    static LdapApiService createLdapApiService()
+    {
+        try {
+            LdapApiService ldapApiService = new StandaloneLdapApiService();
+            ldapApiService.registerRequestControl(
+                    new OpaqueControlFactory(ldapApiService, AD_DOMAIN_SCOPE_CONTROL_OID));
+            return ldapApiService;
+        }
+        catch (Exception exception) {
```
ebyhr: Exception looks too broad.

ebyhr (commented):
Just skimmed.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -71,20 +92,16 @@ public LbLdapClient(LdapConfiguration ldapConfig)
...
         try {
             String filter = config.getLdapUserSearch().replace("${USER}", user);
+            SearchRequest searchRequest = newUserSearchRequest(filter);
             PasswordWarning passwordWarning =
-                    ldapConnectionTemplate.authenticate(
-                            config.getLdapUserBaseDn(),
-                            filter,
-                            SearchScope.SUBTREE,
-                            password.toCharArray());
+                    ldapConnectionTemplate.authenticate(searchRequest, password.toCharArray());
```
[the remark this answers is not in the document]
OmerRaifler: Fixed in the current branch: username values are escaped with FilterEncoder.encodeFilterValue() before ${USER} substitution in both authenticate() and getMemberOf(). I also added regression tests for *, (, ), and \ in the username.

gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
OmerRaifler: Done.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
OmerRaifler: Added requireNonNull checks for ldapConfig and ldapConnectionTemplate.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
OmerRaifler: Added requireNonNull for memberOfAttribute in UserEntryMapper.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
OmerRaifler: Kept intentionally. The SearchRequest is now conditionally mutated to attach the optional Domain Scope control before it is returned, so the local variable is no longer redundant.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
OmerRaifler: Narrowed the initialization failure boundary around StandaloneLdapApiService() and wrapped it in IllegalStateException.

vishalya (commented):
For the tests, we need to get rid of the mocks and start using the actual containers. (may be in a separate PR!)

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -101,27 +131,63 @@ public boolean authenticate(String user, String password)
...
+    {
+        OpaqueControl control = new OpaqueControl(AD_DOMAIN_SCOPE_CONTROL_OID, false);
+        // Apache Directory's opaque control encoder requires an encoded value,
+        // while the Active Directory Domain Scope control has no payload.
+        control.setEncodedValue(Strings.EMPTY_BYTES);
+        return control;
+    }
+
+    static LdapApiService createLdapApiService()
+            throws Exception
```
vishalya, line 207: `Exception` is too generic, can we narrow it down?

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbLdapClient.java
```diff
@@ -13,52 +13,53 @@
...
 import static org.mockito.Mockito.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
 final class TestLbLdapClient
 {
-    private static final Logger log = Logger.get(TestLbLdapClient.class);
+    private static final String AD_DOMAIN_SCOPE_CONTROL_OID = "1.2.840.113556.1.4.1339";
```
vishalya: Can we use the same definition for `AD_DOMAIN_SCOPE_CONTROL_OID` from the LbLdapClient?

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
OmerRaifler: `StandaloneLdapApiService()` declares `throws Exception`, so the catch cannot be narrowed at that boundary.

gateway-ha/src/test/java/io/trino/gateway/ha/security/TestLbLdapClient.java
OmerRaifler: agreed to make the production constant package-private and reuse it in the test.

gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -61,10 +87,18 @@ public LbLdapClient(LdapConfiguration ldapConfig)
...
+        try {
+            ldapApiService = createLdapApiService();
+        }
+        catch (Exception exception) {
+            throw new IllegalStateException("Failed to initialize LDAP client", exception);
+        }
+        connectionConfig.setLdapApiService(ldapApiService);
         DefaultLdapConnectionFactory defaultFactory =
                 new DefaultLdapConnectionFactory(connectionConfig);
+        defaultFactory.setLdapApiService(ldapApiService);
```
vishalya, line 100: We are also setting the `ldapApiServer` on line 97 and here, do we need both?

OmerRaifler: Removed the redundant assignment on `connectionConfig`.

PR #1205, opened 2026-07-27
Support direct LDAP bind without a user search

Chaho12:
Add an optional ldapUserDnPattern to derive the user DN from the login name and bind as it directly, instead of binding as the admin account to search for the entry first. The search based flow stays the default.

## Description

Support direct LDAP bind without a user search
- Add an optional ldapUserDnPattern to derive the user DN from the login name and bind as it directly, instead of binding as the admin account to search for the entry first. The search based flow stays the default.

## Release notes

```markdown
* Support direct LDAP bind without a user search
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
@@ -16,14 +16,22 @@ package io.trino.gateway.ha.config;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.airlift.log.Logger;
+import org.apache.directory.api.ldap.model.exception.LdapInvalidDnException;
+import org.apache.directory.api.ldap.model.name.Dn;
+import org.apache.directory.api.ldap.model.name.Rdn;
 
 import java.io.IOException;
 import java.nio.file.Path;
+
+import static com.google.common.base.Preconditions.checkArgument;
 
 public class LdapConfiguration
 {
     private static final Logger log = Logger.get(LdapConfiguration.class);
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
+    private static final String USER_PLACEHOLDER = "${USER}";
+    // Stand-in login name used to check that the pattern resolves to a parseable DN
+    private static final String VALIDATION_USER = "validation-user";
     private String ldapHost;
     private Integer ldapPort;
     private boolean useTls;
@@ -31,6 +39,7 @@ public class LdapConfiguration
     private String ldapAdminBindDn;
     private String ldapUserBaseDn;
     private String ldapUserSearch;
+    private String ldapUserDnPattern;
     private String ldapGroupMemberAttribute;
     private String ldapAdminPassword;
     private String ldapTrustStorePath;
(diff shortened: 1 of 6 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/security/LbLdapClient.java
```diff
@@ -77,14 +80,34 @@ public LbLdapClient(LdapConfiguration ldapConfig)
 
     public boolean authenticate(String user, String password)
     {
+        // A bind with an empty password is an unauthenticated bind, which some directory
+        // servers answer with success. Reject it before it reaches the server.
+        if (password == null || password.isEmpty()) {
```
vishalya, line 85: Good check!

vishalya (dismissed):
Thanks for filling gap in ldap authn!
Only suggestion, I have is to validate the pattern which is expected in `ldapUserDnPattern` as a part of the config.

Chaho12:
@vishalya changed to validating in config and not runtime !

mosabua:
> Thanks for filling gap in ldap authn! Only suggestion, I have is to validate the pattern which is expected in `ldapUserDnPattern` as a part of the config.

Can you actually do that though .. different LDAP servers support different pattern use right?

gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
```diff
@@ -86,9 +95,36 @@ public static LdapConfiguration load(String path)
...
+    public void validate()
+    {
+        if (ldapUserDnPattern == null || ldapUserDnPattern.isEmpty()) {
+            return;
+        }
+        if (!ldapUserDnPattern.contains(USER_PLACEHOLDER)) {
+            throw new IllegalArgumentException(
+                    "ldapUserDnPattern must contain the %s placeholder, otherwise every login binds as the same DN: %s"
+                            .formatted(USER_PLACEHOLDER, ldapUserDnPattern));
+        }
```
ebyhr: We could use `checkArgument` here.

gateway-ha/src/main/java/io/trino/gateway/ha/config/LdapConfiguration.java
```diff
@@ -159,6 +195,22 @@ public void setLdapUserSearch(String ldapUserSearch)
...
 
+    /**
+     * Template used to derive a user DN directly from the login name, for example
+     * {@code uid=${USER},ou=people,dc=example,dc=com}. When set, authentication binds
+     * as that DN without first searching for the user entry. When unset, the
+     * {@link #getLdapUserSearch() search} based flow is used instead.
+     */
+    public String getLdapUserDnPattern()
+    {
+        return this.ldapUserDnPattern;
```
ebyhr:
```suggestion
        return ldapUserDnPattern;
```

vishalya:
> > Thanks for filling gap in ldap authn! Only suggestion, I have is to validate the pattern which is expected in `ldapUserDnPattern` as a part of the config.
> 
> Can you actually do that though .. different LDAP servers support different pattern use right?

We are not validating any specific LDAP query construct, the ${USER} is just a placeholder and replaces just that part of the string. The rest is allowed as long as the final string still reads as a valid DN.

vishalya (approved):

andythsu:
I literally ran into this just now and found your PR fixed this lol

PR #1215, opened 2026-08-02
Validate pull request commit messages

mosabua:
Add a `check-commit-messages` CI job that runs the airlift `check-commit-messages` action on pull requests, enforcing the project commit message conventions before merge.

Ported from the equivalent change in the core Trino repository: trinodb/trino@aee9e39624e4148f9193a96f7e249c8635bb123c

The Gateway CI workflow has no aggregate `pull-request-check` job, so only the standalone job was ported; the `needs`/status-gate hunks from the upstream diff do not apply here. The `if: github.event_name == 'pull_request'` guard is kept because this workflow also runs on push to `main`.

See https://github.com/trinodb/trino/pull/30489

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,20 @@ env:
   MAVEN_FAST_INSTALL: "-B -V -T 1C -DskipTests -Dmaven.source.skip=true -Dair.check.skip-all"
 
 jobs:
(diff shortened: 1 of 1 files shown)


ebyhr (approved):

mosabua:
Thank you @ebyhr !

PR #1218, opened 2026-08-04
Reword Dependabot commits to pass message check

mosabua:
## Description

Dependabot writes its own commit messages and offers no setting for subject
length, so a share of its updates fail the `check-commit-messages` job added in
#1215 and block the pull request.

Measured against the last 269 Dependabot commits on `main`, using the checker
itself, 41 break the rules:

* 38 exceed the 60 character subject limit, mostly npm packages with a
  `in /webapp` suffix and long Maven coordinates.
* 3 carry a description line longer than 79 characters. These are the
  `Bumps <artifact> from A to B.` lines that Dependabot writes without a
  Markdown link when it cannot resolve a repository URL for the package.

This adds a script that amends the commit message, and a workflow that runs it
on Dependabot pull requests and force pushes the result.

Subjects are shortened one step at a time, keeping the first form that fits, so
no more detail is dropped than necessary:

```
Bump com.github.eirslett:frontend-maven-plugin from 2.0.0 to 2.0.1  (66)
Bump com.github.eirslett:frontend-maven-plugin to 2.0.1             (54)

Bump @douyinfe/semi-illustrations from 2.101.0 to 2.101.1 in /webapp  (68)
Bump @douyinfe/semi-illustrations to 2.101.1 in /webapp               (55)
```

Nothing is lost, because Dependabot repeats the full name and both versions in
the description and in its `updated-dependencies` metadata, which the script
leaves untouched so the YAML stays valid.

## Verification

Replaying the transform over all 269 Dependabot commits and rechecking each
result with the airlift checker:

```
commits=269  violating_before=41  violating_after=0  changed=41  non_idempotent=0
```

Only the 41 failing messages are rewritten. The other 228 are left byte for
byte identical, and running the script twice is a no-op.

## Configuration required before this takes effect

The workflow needs a token with `Contents: read and write` on this repository,
stored as a **Dependabot** secret named `REWORD_PAT`, either a fine-grained
personal access token or a GitHub App installation token. Actions secrets do
not work, because workflows triggered by Dependabot can only read Dependabot
secrets.

A separate token is needed even though the workflow grants `contents: write`,
because [pushes made with the automatic `GITHUB_TOKEN` do not start new
workflow runs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/trigger-a-workflow),
so CI would never run against the amended commit and the pull request would
wait for status checks that never arrive.

Until that secret exists the job logs a notice and stops, so merging this
changes nothing and blocks nothing.

## Notes for reviewers

* The script is runnable and inspectable locally with
  `.github/scripts/reword-dependabot-commit.py --dry-run`.
* It only amends `HEAD`. Dependabot creates single commit pull requests,
  including for grouped updates.
* If Dependabot recreates the branch, for example after `@dependabot rebase`,
  it regenerates the default message, the `synchronize` event runs the workflow
  again, and the message is reworded again.

### DIFF
--- a/.github/scripts/reword-dependabot-commit.py
+++ b/.github/scripts/reword-dependabot-commit.py
@@ -1,0 +1,271 @@
+#!/usr/bin/env python3
+
+"""Rewrite the HEAD commit message to satisfy the project commit message policy.
+
+Dependabot generates its own commit messages and offers no configuration for
+subject length. Most of them already comply, but two cases do not:
+
+  * Long subjects, such as
+    "Bump @douyinfe/semi-illustrations from 2.99.3 to 2.100.0 in /webapp" (68).
+  * Long description lines, such as the "Bumps <artifact> from A to B." line
+    that Dependabot writes without a Markdown link when it cannot resolve a
+    repository URL for the package.
+
+Both fail the check-commit-messages job in ci.yml, which blocks the pull
+request. This script amends HEAD so those commits pass.
+
+The rules mirror airlift/github-actions/check-commit-messages: subjects are at
+most 60 characters and should be at most 50, and ordinary description lines are
+at most 79 characters and should wrap at 72.
+
+Subjects are shortened one step at a time, keeping the first result that fits,
+so no more detail is dropped than necessary:
+
+  Bump com.github.eirslett:frontend-maven-plugin from 2.0.0 to 2.0.1  (66)
+  Bump com.github.eirslett:frontend-maven-plugin to 2.0.1             (54)
+
+Nothing is lost in the process, because Dependabot repeats the full name and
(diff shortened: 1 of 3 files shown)


ebyhr:
```
git commit -am xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx --allow-empty
```

```
.github/scripts/reword-dependabot-commit.py --dry-run
...
```

What is the output `...`?

Chaho12:
@mosabua knock knock~

Issue #1219, opened 2026-08-05
ClusterStatsHttpMonitor is broken with Trino 483+


mosabua:
### Trino Gateway version

21-SNAPSHOT, current `main`

### Trino version(s)

483

### Description

`ClusterStatsHttpMonitor` does not work against Trino 483. The monitor reports
`TrinoStatus.UNKNOWN` for a healthy cluster, so a backend fronted by Trino Gateway
looks unavailable when `monitorType: "UI_API"` is configured.

**Cause**

The monitor collects statistics through the legacy Web UI endpoints defined in
`HttpUtils`:

- `/ui/login` for form authentication
- `/ui/api/stats`
- `/ui/api/query?state=QUEUED`

Trino 483 made the preview Web UI the default at `/ui` and moved the previous UI
to `/ui/legacy`, and the previous UI is deactivated by default. See the
[Trino 483 release notes](https://trino.io/docs/current/release/release-483.html):

> The previous Web UI is deactivated by default and can be restored by setting
> `web-ui.legacy.enabled` to `true`.

The form login endpoint the monitor posts to therefore no longer accepts a POST,
so no authentication cookie is obtained and every subsequent request comes back
empty.

**Observed behavior**

```
INFO   login request received response code 405
INFO   Unauthorized to fetch cluster stats
SEVERE Received null/empty response for /ui/api/stats
```

The monitor then returns `TrinoStatus.UNKNOWN`.

**Steps to reproduce**

`TestClusterStatsMonitor` covers this, but the class is pinned to
`trinodb/trino:476` so the failure is not visible in CI. Point the container at
`trinodb/trino:483` and run `testHttpMonitor`:

```
expected: HEALTHY
 but was: UNKNOWN
	at io.trino.gateway.ha.clustermonitor.TestClusterStatsMonitor.testClusterStatsMonitor
	at io.trino.gateway.ha.clustermonitor.TestClusterStatsMonitor.testHttpMonitor
```

**Expected behavior**

The monitor reports `HEALTHY` for a healthy Trino 483 cluster.

**Workarounds**

- Set `web-ui.legacy.enabled=true` on the backend Trino clusters, which restores
  the endpoints the monitor depends on.
- Use a different `ClusterStatsMonitor` implementation. Note that the JMX and
  metrics monitors are broken on Trino 477 and later for an unrelated reason,
  tracked in #773 and #1155.

**Possible resolution**

The monitor needs a source of cluster statistics that is stable across Trino
[...]

Issue #1155, opened 2026-07-01
Trino gateway expects incorrect metric name that is not configurable


lozbrown:
### Trino Gateway version

19

### Trino version(s)

481

### Description

when useing:  
```
clusterStatsConfiguration:
  monitorType: "METRICS"
```

Trino gateway is exspecting trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount however this seems to no longer exists in 481 (likely replaced by trino_node_name_CoordinatorNodeManager_ActiveNodeCount)

The names of the other two main metrics are configurable as follows 

```
monitor:
    runningQueriesMetricName: io_starburst_galaxy_name_GalaxyMetrics_RunningQueries
    queuedQueriesMetricName: io_starburst_galaxy_name_GalaxyMetrics_QueuedQueries
```
	
however this one does not seem to be configurable, best resolution is likely to make it configurable so it can continue to be compatible with older versions of trino and document a note that it may need to be configured

the error we experience is as follows

> UnexpectedResponseException{message=Request is missing required keys:
> 
> trino_execution_name_QueryManager_RunningQueries
> 
> trino_execution_name_QueryManager_QueuedQueries
> 
> trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount
> 
> in response: '# TYPE trino_execution_name_QueryManager_RunningQueries gauge
> 
> trino_execution_name_QueryManager_RunningQueries 0.0
> 
> # TYPE trino_execution_name_QueryManager_QueuedQueries gauge
> 
> trino_execution_name_QueryManager_QueuedQueries 0.0
> 
> # EOF

lozbrown:
likely this is related to https://github.com/trinodb/trino-gateway/issues/773

Chaho12:
Have you tried setting `discovery.type=airlift-discovery` ?

Issue #773, opened 2025-10-01
ClusterStatsJmxMonitor is broken with trino version 477+


vishalya:
The routing would fails if you are using `ClusterStatsJmxMonitor` with trino version 477 onwards.

Workaround
- Do not ugrade to trino 477 if you are using this monitor
- Use any other `ClusterStatsMonitor`

Cause:
The `ClusterStatsJmxMonitor` relies on the `MBean trino.metadata:name=DiscoveryNodeManager` to collect cluster statistics. In Trino release 477, this MBean was removed. The change is documented in Trino [PR 26083](https://github.com/trinodb/trino/pull/26083).

vishalya:
- The fix would involve adding a new endpoint to get cluster stats information and use it inside - `ClusterStatsJmxMonitor`
- The PR checks are also failing due this issue and to move forward with the development process we will pin the trino image to the earlier version.

PR #1222, opened 2026-08-05
Create test containers from shared factories

mosabua:
## Description

The Trino and MySQL container images were hardcoded at every use in the tests,
and had drifted apart:

| Image | Uses |
|---|---|
| `trinodb/trino` unpinned | 6 |
| `trinodb/trino:476` | 2 |
| `trinodb/trino:466` | 2 |
| `mysql:8.0.36` | 3 |
| `mysql:5.7` | 1 |

The six unpinned uses resolved to whatever was last published as `latest`, so
those tests were not reproducible and silently changed behavior whenever a new
Trino release appeared. Meanwhile the project depends on Trino 483 through
`dep.trino.version`, so most tests exercised Trino Gateway against a Trino release
it does not build against.

Container creation now goes through factory methods in `TestcontainersUtils`,
alongside the existing `createPostgreSqlContainer`. The Trino version is read
from `dep.trino.version` through a Maven-filtered properties file, so a
Dependabot update of the Trino dependency also moves the Trino release under
test, with no second place to remember to update.

Each factory takes its image from a single named source, so all three are
consistent and there is one place to change per image:

| Factory | Image source |
|---|---|
| `createPostgreSqlContainer()` | `POSTGRESQL_IMAGE` constant |
| `createMySqlContainer()` | `MYSQL_IMAGE` constant |
| `createTrinoContainer()` | `dep.trino.version` through `test-versions.properties` |

The PostgreSQL and MySQL versions stay as constants rather than moving into the
filtered properties file, because neither has a corresponding Maven property to
track. Only the Trino version does.

Maven test resource filtering is deliberately scoped to
`test-versions.properties` alone. The other test resources are configuration
templates containing `${...}` placeholders that the tests substitute themselves
at runtime, and filtering those would let Maven consume the placeholders first.

`TestClusterStatsMonitor` keeps its existing pin to `trinodb/trino:476`, because
the cluster stats monitors do not work against Trino 477 and later. The reason
and the tracking issues are now recorded next to the pin, rather than the pin
sitting there unexplained.

Moving `TestClusterStatsMonitor` to 483 fails four tests, which is how the pin
turned out to be load-bearing:

```
testHttpMonitor     expected: HEALTHY  but was: UNKNOWN
testJmxMonitor      expected: HEALTHY  but was: UNHEALTHY
testMetricsMonitor
testMetricsRanges
```

Those failures are three separate defects, all pre-existing and none introduced
here:

- #773 — Trino 477 removed the `trino.metadata:name=DiscoveryNodeManager`
  MBean, which `ClusterStatsJmxMonitor` depends on
- #1155 — the metrics monitor looks for
  `trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount`, gone for the same
  reason
- #1219 — Trino 483 made the preview Web UI the default and deactivated the
  legacy UI, so the UI API monitor can no longer authenticate

Keeping the pin in this pull request is intentional, so that the change stays a
test infrastructure cleanup. Fixing the monitors is separate work.

## Release notes

### DIFF
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
@@ -42,13 +42,21 @@ import static org.testcontainers.utility.MountableFile.forClasspathResource;
 @TestInstance(PER_CLASS)
 final class TestClusterStatsMonitor
 {
+    // Deliberately pinned instead of using TestcontainersUtils.createTrinoContainer, because the
+    // cluster stats monitors do not work against Trino 477 and later. The JMX monitor breaks
+    // because Trino 477 removed the DiscoveryNodeManager MBean, see
+    // https://github.com/trinodb/trino-gateway/issues/773, the metrics monitor looks for a metric
+    // of the same name, see https://github.com/trinodb/trino-gateway/issues/1155, and the HTTP
+    // monitor fails as well. Switch to createTrinoContainer once the monitors are fixed.
+    // TODO https://github.com/trinodb/trino-gateway/issues/773 Update Trino version
+    private static final String PINNED_TRINO_IMAGE = "trinodb/trino:476";
+
     private TrinoContainer trino;
 
     @BeforeAll
     void setUp()
     {
-        // TODO https://github.com/trinodb/trino-gateway/issues/773 Update Trino version
-        trino = new TrinoContainer("trinodb/trino:476");
+        trino = new TrinoContainer(PINNED_TRINO_IMAGE);
         trino.withCopyFileToContainer(forClasspathResource("trino-config-with-rmi.properties"), "/etc/trino/config.properties");
         trino.withCopyFileToContainer(forClasspathResource("jvm-with-rmi.config"), "/etc/trino/jvm.config");
         trino.start();
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/util/TestcontainersUtils.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/util/TestcontainersUtils.java
@@ -16,23 +16,66 @@ package io.trino.gateway.ha.util;
 import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
+import org.testcontainers.mysql.MySQLContainer;
 import org.testcontainers.postgresql.PostgreSQLContainer;
+import org.testcontainers.trino.TrinoContainer;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.time.Duration;
+import java.util.Properties;
+
+import static com.google.common.base.Preconditions.checkState;
 
 public final class TestcontainersUtils
 {
+    private static final String VERSIONS_RESOURCE = "/test-versions.properties";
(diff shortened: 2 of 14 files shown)


gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
```diff
@@ -42,13 +42,21 @@
 @TestInstance(PER_CLASS)
 final class TestClusterStatsMonitor
 {
+    // Deliberately pinned instead of using TestcontainersUtils.createTrinoContainer, because the
```
mosabua, line 45: Follow up PR for that is in #1223

gateway-ha/src/test/java/io/trino/gateway/ha/util/TestcontainersUtils.java
```diff
@@ -16,23 +16,66 @@
...
+        Properties properties = new Properties();
+        try (InputStream input = TestcontainersUtils.class.getResourceAsStream(VERSIONS_RESOURCE)) {
+            checkState(input != null, "Resource %s not found. It is filtered by Maven and only exists in a Maven build.", VERSIONS_RESOURCE);
+            properties.load(input);
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        String version = properties.getProperty("trino.version");
+        checkState(version != null && !version.isBlank(), "Property trino.version not set in %s", VERSIONS_RESOURCE);
```
Chaho12, line 78:
```suggestion
        checkState(version != null && !version.isBlank(), "Property trino.version not set in %s", VERSIONS_RESOURCE);
        checkState(!version.contains("${"), "Property trino.version in %s was not filtered by Maven: %s", VERSIONS_RESOURCE, version);
```
How about adding another check for an unfiltered copy? It yields the literal "${dep.trino.version}", which is neither null nor blank, so the check passes and the failure surfaces as a docker pull error.

Chaho12 (approved):
LGTM 👍

gateway-ha/src/test/java/io/trino/gateway/ha/util/TestcontainersUtils.java
```diff
@@ -16,23 +16,66 @@
...
                         .withStrategy(new LogMessageWaitStrategy()
                                 .withRegEx(".*database system is ready to accept connections.*\\s")
                                 .withTimes(2)
                                 .withStartupTimeout(Duration.ofMinutes(1))));
     }
+
+    public static MySQLContainer createMySqlContainer()
+    {
+        //noinspection resource
+        return new MySQLContainer(MYSQL_IMAGE);
```
ebyhr, line 54: Image

gateway-ha/src/test/java/io/trino/gateway/ha/util/TestcontainersUtils.java
```diff
@@ -16,23 +16,66 @@
...
+        return new PostgreSQLContainer(POSTGRESQL_IMAGE)
                 .waitingFor(new WaitAllStrategy()
                         .withStrategy(Wait.forListeningPort())
                         .withStrategy(new LogMessageWaitStrategy()
                                 .withRegEx(".*database system is ready to accept connections.*\\s")
                                 .withTimes(2)
                                 .withStartupTimeout(Duration.ofMinutes(1))));
     }
+
+    public static MySQLContainer createMySqlContainer()
```
ebyhr, line 51: Extracting a helper method for a single line is overkill. Just providing a public constant MYSQL_IMAGE seems sufficient.

wendigo (approved):
LGTM % existing comments

mosabua:
I will get around to working on this soon I hope

PR #1223, opened 2026-08-05
Support current Trino in cluster stats monitors

mosabua:
## Description

Three of the six cluster stats monitor types do not work against a current
Trino release. This fixes two of them and documents the third, which cannot be
fixed from the Trino Gateway side.

**`METRICS`** — Trino 477 renamed the active node count metric from
`trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount` to
`trino_node_name_CoordinatorNodeManager_ActiveNodeCount`. That metric is the
default entry of `metricMinimumValues`, so the monitor reports every cluster on
Trino 477 or later as unhealthy with a missing required key. The default now
uses the current name. Reported in #1155.

**`UI_API`** — Trino 483 replaced the previous Web UI with the preview Web UI.
The form login the monitor posts to is no longer served at `/ui/login`, which
now answers `GET` but rejects `POST` with `405`. Without a session cookie every
subsequent request returns `401` and the monitor reports `UNKNOWN`. Reported in
#1219.

The preview Web UI logs in through `/ui/auth/login` with a JSON body, so this is
not only a path change, and a new `uiLoginType` setting on `backendState`
selects between the two:

| `uiLoginType` | Login resource | Body | Backend Trino version |
|---|---|---|---|
| `AUTH_API`, the default | `/ui/auth/login` | JSON | 483 and later |
| `FORM` | `/ui/login` | form encoded | 482 and earlier |

**`JMX`** — not fixed, and not fixable here. See the following section.

### Verified behavior

Measured directly against stock Trino containers rather than inferred from
release notes:

| Trino | Active node count metric | `v1/jmx/mbean` | Web UI login that authenticates |
|---|---|---|---|
| 476 | `trino_metadata_name_DiscoveryNodeManager_ActiveNodeCount` | 200 | `/ui/login`, form |
| 477 | `trino_node_name_CoordinatorNodeManager_ActiveNodeCount` | 200 | `/ui/login`, form |
| 478, 479 | `trino_node_name_...` | 200 | `/ui/login`, form |
| 480, 482 | `trino_node_name_...` | **404** | `/ui/login`, form |
| 483 | `trino_node_name_...` | 404 | **`/ui/auth/login`, JSON** |

None of these changes appear in the Trino release notes for 477 through 483,
which contain no mention of JMX at all.

### Why not `/ui/legacy/login`

On Trino 483 the previous Web UI moved to `/ui/legacy`, and `/ui/legacy/login`
still accepts a form post and issues a working cookie. Using it would be a
smaller change, and it is deliberately avoided. The previous Web UI is scheduled
for removal, and that resource goes with it, so the monitor would break a second
time. `/ui/auth/login` and the `/ui/api` endpoints the monitor reads are the
preview Web UI's own, confirmed from the endpoints its front end calls, so they
survive that removal.

### The `JMX` monitor

`ClusterStatsJmxMonitor` reads `v1/jmx/mbean`, which is not served from Trino
480 onwards. It is also not part of the
[documented Trino JMX interface](https://trino.io/docs/current/admin/jmx.html),
which covers RMI access only, so there is no supported replacement to point the
monitor at. This pull request documents the limitation and leaves #773 open.
Removing or reimplementing that monitor is a separate discussion.

### Configuration granularity

`uiLoginType` and `metricMinimumValues` are both gateway wide, so a Trino
Gateway fronting clusters on both sides of one of these version boundaries
cannot satisfy all of them with one setting. That is a known limitation of this
approach, called out in the documentation, with `INFO_API` and `JDBC` as the
recommended alternatives for mixed deployments.

Two better solutions, both larger and deliberately out of scope here:

1. **Per backend configuration.** Backends are configured in the database rather
   than the configuration file, so this needs a schema change, but it models a
   fleet mid-upgrade correctly.
2. **Detect per backend.** Probe the cluster once and cache which interface it
   offers, so a mixed fleet needs no configuration at all. `/ui/auth/info` looks
   like a usable signal. The cost is detection logic and a less obvious failure
   mode when detection is wrong.

The underlying problem is that both monitors read interfaces that are not
covered by any compatibility guarantee: the Web UI is explicitly not an API, and
JMX MBean names are internals. `INFO_API` and `JDBC` keep working across all
versions tested precisely because they use stable, documented interfaces.

### Test coverage gap

`TestClusterStatsMonitor` is pinned to `trinodb/trino:476` because the JMX
monitor cannot work on anything newer, so the tests exercise `uiLoginType: FORM`
and the previous metric name. **The new defaults are therefore not covered by
CI.** Both were verified by hand against Trino 483:

```
$ curl -s -c c.txt -o /dev/null -w "%{http_code}\n" -X POST \
    -H "Content-Type: application/json" -d '{"username":"probe","password":""}' \
    http://localhost:8080/ui/auth/login
204
$ curl -s -b c.txt -o /dev/null -w "%{http_code}\n" http://localhost:8080/ui/api/stats
200
```

Closing that gap means running the non-JMX monitor tests against a current Trino
release, which needs the test container split by monitor type.

## Release notes

```markdown
* {Breaking} Fix the `METRICS` cluster stats monitor for Trino 477 and later, where the
  active node count metric was renamed. Backends running Trino 476 or earlier
  require the previous metric name in `metricMinimumValues`. ({issue}`1155`)
* {Breaking} Fix the `UI_API` cluster stats monitor for Trino 483 and later, where the Web
  UI login resource changed. Backends running Trino 482 or earlier now require
  `uiLoginType` set to `FORM`. ({issue}`1219`)
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
@@ -19,9 +19,11 @@ import io.airlift.http.client.HttpStatus;
 import io.airlift.log.Logger;
 import io.trino.gateway.ha.config.BackendStateConfiguration;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import io.trino.gateway.ha.config.UiLoginType;
 import okhttp3.Call;
 import okhttp3.FormBody;
 import okhttp3.HttpUrl;
+import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
@@ -37,7 +39,6 @@ import static com.google.common.net.HttpHeaders.X_FORWARDED_PROTO;
 import static io.airlift.http.client.HttpStatus.fromStatusCode;
 import static io.trino.gateway.ha.handler.HttpUtils.UI_API_QUEUED_LIST_PATH;
 import static io.trino.gateway.ha.handler.HttpUtils.UI_API_STATS_PATH;
-import static io.trino.gateway.ha.handler.HttpUtils.UI_LOGIN_PATH;
 import static java.util.Objects.requireNonNull;
 
 public class ClusterStatsHttpMonitor
@@ -46,16 +47,21 @@ public class ClusterStatsHttpMonitor
     private static final Logger log = Logger.get(ClusterStatsHttpMonitor.class);
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private static final String SESSION_USER = "sessionUser";
+    private static final String AUTH_API_LOGIN_PATH = "/ui/auth/login";
+    private static final String FORM_LOGIN_PATH = "/ui/login";
+    private static final MediaType JSON_MEDIA_TYPE = MediaType.parse("application/json");
 
     private final String username;
     private final String password;
     private final boolean xForwardedProtoHeader;
+    private final UiLoginType uiLoginType;
 
     public ClusterStatsHttpMonitor(BackendStateConfiguration backendStateConfiguration)
     {
         username = backendStateConfiguration.getUsername();
         password = backendStateConfiguration.getPassword();
         xForwardedProtoHeader = backendStateConfiguration.getXForwardedProtoHeader();
+        uiLoginType = backendStateConfiguration.getUiLoginType();
     }
 
     @Override
@@ -105,17 +111,38 @@ public class ClusterStatsHttpMonitor
         return clusterStats.userQueuedCount(clusterUserStats).build();
     }
 
+    private String loginPath()
+    {
+        return switch (uiLoginType) {
+            case AUTH_API -> AUTH_API_LOGIN_PATH;
+            case FORM -> FORM_LOGIN_PATH;
+        };
+    }
+
+    private RequestBody loginBody()
+    {
+        return switch (uiLoginType) {
+            // The login resource of the Web UI in Trino 483 and later rejects a form encoded body
+            case AUTH_API -> RequestBody.create(
+                    OBJECT_MAPPER.createObjectNode()
+                            .put("username", username)
+                            .put("password", password)
+                            .toString(),
+                    JSON_MEDIA_TYPE);
+            case FORM -> new FormBody.Builder()
+                    .add("username", username)
+                    .add("password", password)
+                    .build();
+        };
+    }
+
     private OkHttpClient acquireClientWithCookie(String loginUrl)
     {
         UiApiCookieJar cookieJar = new UiApiCookieJar();
         OkHttpClient client = new OkHttpClient.Builder().cookieJar(cookieJar).build();
-        RequestBody formBody = new FormBody.Builder()
-                .add("username", username)
-                .add("password", password)
-                .build();
         Request loginRequest = new Request.Builder()
                 .url(HttpUrl.parse(loginUrl))
-                .post(formBody)
+                .post(loginBody())
                 .build();
 
         Call call = client.newCall(loginRequest);
@@ -132,7 +159,7 @@ public class ClusterStatsHttpMonitor
 
     private String queryCluster(ProxyBackendConfiguration backend, String path)
     {
-        String loginUrl = backend.getProxyTo() + UI_LOGIN_PATH;
+        String loginUrl = backend.getProxyTo() + loginPath();
         OkHttpClient client = acquireClientWithCookie(loginUrl);
         if (client == null) {
             log.error("Client received is null");
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
@@ -25,6 +25,7 @@ import io.airlift.units.Duration;
 import io.trino.gateway.ha.config.BackendStateConfiguration;
 import io.trino.gateway.ha.config.MonitorConfiguration;
 import io.trino.gateway.ha.config.ProxyBackendConfiguration;
+import io.trino.gateway.ha.config.UiLoginType;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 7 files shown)


mosabua:
Filed [trinodb/trino#30591](https://github.com/trinodb/trino/issues/30591) upstream, asking Trino to keep accepting a form encoded `POST` on `/ui/login` as an alias for `/ui/auth/login`.

If that is accepted, the `uiLoginType` setting added here becomes unnecessary for new Trino releases, though it is still needed for the versions already released.

I am not convinced this is worth doing in Trino.

lozbrown:
Hey Manfred

What do we need here to move this along, this is an important fix

lozbrown:
Hey @mosabua 

I've spent some time thinking about this, it's an important fix but the more I think about this the more I worry about the way this blocks Canary upgrades (upgrading one prod cluster at a time to b prevent me impact) 

I think it would be better to do something to handle this now granularly

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
```diff
@@ -105,17 +111,38 @@ public ClusterStats monitor(ProxyBackendConfiguration backend)
...
         UiApiCookieJar cookieJar = new UiApiCookieJar();
         OkHttpClient client = new OkHttpClient.Builder().cookieJar(cookieJar).build();
-        RequestBody formBody = new FormBody.Builder()
-                .add("username", username)
-                .add("password", password)
-                .build();
         Request loginRequest = new Request.Builder()
                 .url(HttpUrl.parse(loginUrl))
-                .post(formBody)
+                .post(loginBody())
```
Chaho12, line 145: Can you log more details on failure in line 150? Because this change would set default as AUTH_API and it is hard to tell simply with status code. 

```java
        try (Response res = call.execute()) {
            log.info("login request received response code %d", res.code());
            if (cookieJar.loadForRequest(url).isEmpty()) {
                log.error("Login to %s returned %d without a session cookie. Verify that backendState.uiLoginType "
                                + "matches the Trino version of this cluster: AUTH_API for Trino 483 and later, "
                                + "FORM for Trino 482 and earlier. Currently configured: %s",
                        loginUrl, res.code(), uiLoginType);
            }
            return client;
        }
```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
```diff
@@ -46,16 +47,21 @@ public class ClusterStatsHttpMonitor
...
     private final String password;
     private final boolean xForwardedProtoHeader;
+    private final UiLoginType uiLoginType;
 
     public ClusterStatsHttpMonitor(BackendStateConfiguration backendStateConfiguration)
     {
         username = backendStateConfiguration.getUsername();
         password = backendStateConfiguration.getPassword();
         xForwardedProtoHeader = backendStateConfiguration.getXForwardedProtoHeader();
+        uiLoginType = backendStateConfiguration.getUiLoginType();
```
Chaho12, line 64: Why don't we require non null? Because it would cause NPE

```suggestion
        username = requireNonNull(backendStateConfiguration.getUsername(), "backendState username is null");
        password = backendStateConfiguration.getPassword();
        xForwardedProtoHeader = backendStateConfiguration.getXForwardedProtoHeader();
        uiLoginType = requireNonNull(backendStateConfiguration.getUiLoginType(), "backendState uiLoginType is null");
```

gateway-ha/src/test/java/io/trino/gateway/ha/clustermonitor/TestClusterStatsMonitor.java
Chaho12, line 1: I ran through my claude for this, but since we have MockWebServer (used at TestProxyRequestHandler), we can test AuthAPI's path/content-type/body without container, no?
```java
    @Test
    void testHttpMonitorAuthApiLogin()
            throws Exception
    {
        try (MockWebServer server = new MockWebServer()) {
            server.enqueue(new MockResponse()
                    .setResponseCode(204)
                    .addHeader("Set-Cookie", "__Secure-Trino-UI-Token=token; Path=/ui"));
            server.enqueue(new MockResponse()
                    .setResponseCode(200)
                    .setBody("{\"activeWorkers\":1,\"queuedQueries\":0,\"runningQueries\":0}"));
            server.enqueue(new MockResponse().setResponseCode(204).addHeader("Set-Cookie", "__Secure-Trino-UI-Token=token; Path=/ui"));
            server.enqueue(new MockResponse().setResponseCode(200).setBody("[]"));
            server.start();

            BackendStateConfiguration configuration = new BackendStateConfiguration();
            configuration.setUsername("probe");

            ProxyBackendConfiguration backend = new ProxyBackendConfiguration();
            backend.setName("mock");
            backend.setProxyTo(server.url("").toString().replaceAll("/$", ""));

            ClusterStats stats = new ClusterStatsHttpMonitor(configuration).monitor(backend);
            assertThat(stats.trinoStatus()).isEqualTo(TrinoStatus.HEALTHY);

            RecordedRequest login = server.takeRequest();
            assertThat(login.getPath()).isEqualTo("/ui/auth/login");
            assertThat(login.getHeader("Content-Type")).startsWith("application/json");
            assertThat(login.getBody().readUtf8()).isEqualTo("{\"username\":\"probe\",\"password\":\"\"}");
        }
    }

```

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
```diff
@@ -46,16 +47,21 @@ public class ClusterStatsHttpMonitor
     private static final Logger log = Logger.get(ClusterStatsHttpMonitor.class);
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private static final String SESSION_USER = "sessionUser";
+    private static final String AUTH_API_LOGIN_PATH = "/ui/auth/login";
+    private static final String FORM_LOGIN_PATH = "/ui/login";
```
Chaho12, line 51: Why is this here and removed from HttpUtils?

gateway-ha/src/main/java/io/trino/gateway/ha/clustermonitor/ClusterStatsHttpMonitor.java
```diff
@@ -46,16 +47,21 @@ public class ClusterStatsHttpMonitor
...
+    private static final MediaType JSON_MEDIA_TYPE = MediaType.parse("application/json");
```
Chaho12, line 52: What about MediaType.get?

PR #1258, opened 2026-09-05
Wait for backend probes when searching for an unknown query id

### LINKED ISSUE #1241: searchAllBackendForQuery fails to locate running queries due to non-blocking isDone() check
### Trino Gateway version

20

### Trino version(s)

416

### Description

Description:

In the BaseRoutingManager class, the searchAllBackendForQuery method contains a critical logic flaw that prevents it from reliably finding a query that is actively running on a specific backend.

When the Gateway needs to locate the correct backend for an unknown queryId (e.g., when the query ID is not found in the cache or history), this method is invoked. It sends concurrent HTTP HEAD requests to all known backends. However, when checking the results, it uses a non-blocking, one-time status check with if (entry.getValue().isDone()).

Since these requests are submitted asynchronously, the vast majority of Future tasks are not yet complete by the time the loop iterates over the responseCodes map. As a result, isDone() almost always returns false, and the loop exits without ever entering the inner logic to process a successful response. Consequently, the method always falls back to returning the first healthy backend, which is likely incorrect.

This leads to several issues:

Incorrect Routing: Requests that should be routed to the specific backend that owns the query (e.g., to view query details or terminate a query) may be routed to a different backend, causing request failures or unexpected behavior.

Performance Overhead: Although tasks are executed concurrently, the main thread does not wait for them. These background tasks may be wasted or delayed in cancellation.

Steps to Reproduce:

Submit a query to the Trino Gateway that gets routed to Backend A.

While the query is executing, attempt to fetch its status or plan through the Gateway's interface (e.g., by accessing /ui/query.html?queryId={queryId}), ensuring this queryId is not in the Gateway's internal cache.

Observe the Gateway's logs or behavior. The request will likely be routed to Backend B instead of Backend A, where the query is actually running.

Alternatively, set a breakpoint inside the searchAllBackendForQuery method. When multiple backends exist, you will observe that isDone() returns false for almost all Future objects during the first iteration over the map.

Expected Behavior:

The method should block and wait until at least one backend request succeeds (returns a 200 status code), or until all requests have completed or timed out, before making a routing decision. It must accurately identify which backend holds the queried queryId.

Suggested Fix:

Use get() with a timeout: Replace the isDone() check with a get() call on each Future, using a reasonable timeout to block and wait for each task to complete.

Refactor with CompletableFuture: Reimplement the logic using CompletableFuture.anyOf() to wait for any task to complete, process its result, and gracefully cancel other pending tasks.

Here is a sample fix using CompletableFuture:

java
private String searchAllBackendForQuery(String queryId) {
    List<ProxyBackendConfiguration> backends = gatewayBackendManager.getAllBackends();
    List<CompletableFuture<Map.Entry<String, Integer>>> futures = backends.stream()
        .map(backend -> CompletableFuture.supplyAsync(() -> {
            try {
                String target = backend.getProxyTo() + "/v1/query/" + queryId;
                URL url = URI.create(target).toURL();
                HttpURLConnection conn = (HttpURLConnection) url.openConnection();
                conn.setConnectTimeout(5000);
                conn.setReadTimeout(5000);
                conn.setRequestMethod("HEAD");
                int code = conn.getResponseCode();
                return Map.entry(backend.getProxyTo(), code);
            } catch (Exception e) {
                return Map.entry(backend.getProxyTo(), -1);
            }
        }, executorService))
        .collect(Collectors.toList());

    // Wait for any one task to complete
    CompletableFuture<Map.Entry<String, Integer>> anyCompleted = 
        CompletableFuture.anyOf(futures.toArray(new CompletableFuture[0]))
            .thenApply(result -> (Map.Entry<String, Integer>) result);

    try {
        Map.Entry<String, Integer> result = anyCompleted.get(10, TimeUnit.SECONDS);
        if (result.getValue() == 200) {
            log.info("Found query [%s] on backend [%s]", queryId, result.getKey());
            setBackendForQueryId(queryId, result.getKey());
            futures.forEach(f -> f.cancel(true)); // Cancel other pending requests
            return result.getKey();
        }
        // If the first completed response is not 200, the logic should continue waiting for others.
        // This is a simplified example; further refinement is needed for a complete solution.
    } catch (Exception e) {
        log.warn("Search for query [%s] failed: %s", queryId, e.getMessage());
    }
    // Fallback ...
}
Environment:

Trino Gateway Version: main branch (or any version containing this logic)

Java Version: All versions

Additional Context:

This issue arises from a fundamental misunderstanding of the Future.isDone() method. The method is non-blocking and intended for status checks, but the required behavior in this scenario is to block and wait for a result. A refactoring using CompletableFuture is recommended for a more robust and readable asynchronous implementation.

adityaanikam:
## Description

`findBackendForUnknownQueryId` submits a probe to every backend and then inspects the futures with `isDone()`:

```java
for (Map.Entry<String, Future<Integer>> entry : responseCodes.entrySet()) {
    if (entry.getValue().isDone()) {
        int responseCode = entry.getValue().get();
```

The loop runs immediately after submission, so in practice no probe has completed and `isDone()` is false for all of them. No response code is ever read, and the method falls through to the fallback at the end, returning the first active backend in the default routing group. The backend that actually owns the query is found only by luck.

This affects requests that arrive without a routing cookie or cached mapping. The gateway forwards them to an arbitrary cluster rather than the one running the query.

Fixes #1241.

The change waits for the probes rather than polling them, but against **one deadline shared across all backends** rather than a timeout per future. `executorService` is `Executors.newFixedThreadPool(5)`, so with more backends than threads the later probes do not start until earlier ones finish; a per future timeout would compound to five seconds per backend in the worst case, while a shared budget bounds the whole search at five seconds regardless of backend count.

`ExecutionException` and `TimeoutException` are caught per backend so that one unreachable cluster does not abort the search. Previously any throw from `get()` was caught by the outer `catch (Exception)` and logged as "not found", which ended the search entirely.

Outstanding probes are cancelled in a `finally` block. Nothing reads them once the search returns, and the executor is shared with the rest of the routing manager, so leaving them to run would occupy pool threads that other lookups need.

## Testing

`TestRoutingManagerSearchAllBackends` starts two stub backends: one that answers `200` after a short delay (the query owner) and one that answers `404`. The owner is placed in a routing group other than the default, and the second backend in the default group, so the fallback path cannot return the owner by accident. The assertion passes only if the search loop actually located it.

Verified with a negative control: reverting only `BaseRoutingManager.java` and keeping the test makes it fail with the fallback backend's URL rather than the owner's, matching the reported symptom.

## Release notes

```markdown
* Fix routing of requests for a query whose backend is not cached, which
  previously went to an arbitrary active backend instead of the one running
  the query. ({issue}`1241`)
```

### DIFF
--- a/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
+++ b/gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
@@ -35,10 +35,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /**
  * This class performs health check, stats counts for each backend and provides a backend given
@@ -48,6 +50,7 @@ public abstract class BaseRoutingManager
         implements RoutingManager
 {
     private static final Logger log = Logger.get(BaseRoutingManager.class);
+    private static final long BACKEND_SEARCH_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(5);
     private final ExecutorService executorService = Executors.newFixedThreadPool(5);
     private final GatewayBackendManager gatewayBackendManager;
     private final ConcurrentHashMap<String, TrinoStatus> backendToStatus;
@@ -230,19 +233,36 @@ public abstract class BaseRoutingManager
                                 });
                 responseCodes.put(backend.getProxyTo(), call);
             }
+            // The probes run concurrently, so wait for them against a single shared budget
+            // rather than per future: the executor is a fixed pool, so with more backends
+            // than threads a per-future timeout would compound.
+            long deadline = System.nanoTime() + BACKEND_SEARCH_TIMEOUT_NANOS;
             for (Map.Entry<String, Future<Integer>> entry : responseCodes.entrySet()) {
-                if (entry.getValue().isDone()) {
-                    int responseCode = entry.getValue().get();
+                long remaining = deadline - System.nanoTime();
+                if (remaining <= 0) {
+                    log.warn("Timed out searching backends for query [%s]", queryId);
+                    break;
+                }
+                try {
+                    int responseCode = entry.getValue().get(remaining, TimeUnit.NANOSECONDS);
                     if (responseCode == 200) {
                         log.info("Found query [%s] on backend [%s]", queryId, entry.getKey());
                         setBackendForQueryId(queryId, entry.getKey());
                         return entry.getKey();
                     }
                 }
+                catch (ExecutionException | TimeoutException e) {
+                    log.debug(e, "Could not check backend [%s] for query [%s]", entry.getKey(), queryId);
+                }
             }
         }
         catch (Exception e) {
             log.warn("Query id [%s] not found", queryId);
+        }
+        finally {
+            // Nothing reads the remaining probes, and the executor is shared, so do not
+            // leave them occupying its threads.
+            responseCodes.values().forEach(future -> future.cancel(true));
         }
         // Fallback on first active backend if queryId mapping not found.
         return gatewayBackendManager.getActiveBackends(defaultRoutingGroup).stream()
--- a/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingManagerSearchAllBackends.java
+++ b/gateway-ha/src/test/java/io/trino/gateway/ha/router/TestRoutingManagerSearchAllBackends.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.gateway.ha.router;
+
+import com.google.common.net.HostAndPort;
+import com.sun.net.httpserver.HttpServer;
+import io.trino.gateway.ha.config.DataStoreConfiguration;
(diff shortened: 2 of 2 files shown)


gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -230,20 +233,37 @@ private String searchAllBackendForQuery(String queryId)
                                 });
                 responseCodes.put(backend.getProxyTo(), call);
             }
+            // The probes run concurrently, so wait for them against a single shared budget
+            // rather than per future: the executor is a fixed pool, so with more backends
+            // than threads a per-future timeout would compound.
+            long deadline = System.nanoTime() + BACKEND_SEARCH_TIMEOUT_NANOS;
             for (Map.Entry<String, Future<Integer>> entry : responseCodes.entrySet()) {
```
RoeyoOgen: isnt the loop is still sequential over a HashMap? so a slow or hung backend that appears earlier in iteration order can consume the rest of the shared deadline before later backends are checked?

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
```diff
@@ -230,20 +233,37 @@ private String searchAllBackendForQuery(String queryId)
...
+                }
+                try {
+                    int responseCode = entry.getValue().get(remaining, TimeUnit.NANOSECONDS);
                     if (responseCode == 200) {
                         log.info("Found query [%s] on backend [%s]", queryId, entry.getKey());
                         setBackendForQueryId(queryId, entry.getKey());
                         return entry.getKey();
                     }
                 }
+                catch (ExecutionException | TimeoutException e) {
```
RoeyoOgen: do we need to catch 'InterruptedException' as well?

RoeyoOgen:
other than the two comments. looks good :) 
@adityaanikam

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
adityaanikam: Good catch, you're right. The probes run concurrently, but I was reading the results in HashMap order, so a hung backend that came up first used the whole five second budget and a probe that had already answered 200 was never looked at. I switched to an ExecutorCompletionService so the results are read in the order the probes finish, still against the one shared deadline. The test now adds a few hung backends next to the owner, and it failed for me with the old read order.

gateway-ha/src/main/java/io/trino/gateway/ha/router/BaseRoutingManager.java
adityaanikam: Yes, thanks. Future.get throws it and the outer catch was swallowing it, so the interrupt flag was lost. I now catch it separately, restore the flag, and fall back to the default group like the other failure paths.

PR #1280, opened 2026-09-21
Confirm cluster deletion in a modal naming the cluster

maswin:
## Description
Deleting a cluster from the Cluster tab used a Popconfirm bubble with a generic "Are you sure you want to delete?" message that never named the target. The bubble is small, anchored to the row, and easy to dismiss or misread, so an accidental click on the wrong row deleted the wrong cluster with no way to notice before confirming.

Replace it with a centered modal that spells out the cluster name in the prompt, so the operator can see exactly which cluster is about to go away. The confirm button is styled as a danger action and the existing "cannot be recovered" warning is kept.

While the delete request is in flight the modal stays open with a loading spinner on the confirm button, and every dismiss path is blocked (cancel button, mask click, close icon, and Esc) so the dialog cannot be closed before the result is known. On failure it stays open and shows the error toast instead of silently disappearing.

Screen Shot-
Screenshot 2026-09-20 at 8 30 59 PM

## Release notes

```markdown
* Fix some things.
```


webapp/src/components/cluster.tsx
```diff
@@ -28,6 +31,18 @@ export function Cluster() {
       }).catch(() => { });
   }
 
+  const remove = () => {
+    if (deleteTarget === undefined) return;
```
koszti, line 35: Can we guard against repeated execution while deletion is pending to prevent duplicate delete requests?

```suggestion
    if (deleteTarget === undefined || deleting) return;
```

webapp/src/components/cluster.tsx
```diff
@@ -210,6 +215,26 @@ export function Cluster() {
...
         </Form>
       </Modal>
+      <Modal
+        title={Locale.Cluster.DeleteTitle}
+        visible={visibleDelete}
+        onOk={remove}
+        onCancel={() => { setVisibleDelete(false) }}
+        okText={Locale.UI.Delete}
+        cancelText={Locale.UI.Cancel}
+        okButtonProps={{ type: 'danger', theme: 'solid', loading: deleting }}
```
koszti, line 225: Similar to the above one, can we also add `disabled: deleting`? The loading state blocks pointer clicks but still allows keyboard activation while the request is pending.

```suggestion
        okButtonProps={{ type: 'danger', theme: 'solid', loading: deleting, disabled: deleting }}
```

koszti (changes requested):
Can we fill the release note section in the PR description with something like:

  * Confirm cluster deletion in a modal showing the cluster name.

webapp/src/components/cluster.tsx
maswin: done

webapp/src/components/cluster.tsx
maswin: done

